PR #4, opened 2017-02-24
Start a sqlite package and add cmd/rebuilddb app

chappjc:
Well, the basic store/query functions in package `dcrsqlite` seem to work, as does the app `rebuilddb`.  The app is a prototype at this point as it doesn't reconstruct the database as needed by the web API, just the summaries for all blocks.

Also note that the sqlite drivers use cgo.  This is unfortunate for several reasons so I'm not eager to merge just yet.  Perhaps go build flags should be used to opt-in (e.g. `+sqlite`).

At some point in the near future, there will also need to be a bridge between the full scan function like this and the existing API.  These things are independent test projects for the time being.  But it can wait a bit. 
  Next up should be tests, which should make development easier anyway.


chappjc:
kinda sucks that the go sqlite drivers use cgo.  Makes it harder on Windows users and sucks in general, but it's working.

chappjc:
good enough for now.  The sqlite package using cgo doesn't need to build to build the main app so that's ok

PR #7, opened 2017-02-26
scanblocks README.md

chappjc:
This adds a README.md for scanblocks, along with a helper .sh script, and some related code cleanup.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,17 @@
 debug
 *.cert
 *.key
+dcrdata
 dcrdata.conf
 vendor/
 *.orig
+*.db
+*.db-journal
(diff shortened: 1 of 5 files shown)


raedah:
Data catchup and repopulation of new blocks is going to need to happen every time dcrdata is launched. Instead of it being a separate tool, I suggest its just part of the startup function. If the db is empty, it would start from the genesis 0 block, else it would query the db and pickup from the latest stored block after verifying a delta of say 6 blocks.

chappjc:
Agreed.  This functionality needs to be absorbed by the startup routines.

This also means it needs to ability to start from block N rather than 0, depending on the state of the DB.

PR #11, opened 2017-02-27
extended stake info for sqlite and rebuilddb

chappjc:
This addresses https://github.com/dcrdata/dcrdata/issues/8

Seems to be working.

### DIFF
--- a/apimiddleware.go
+++ b/apimiddleware.go
@@ -29,11 +29,11 @@ func (c *appContext) StatusCtx(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		status := &apitypes.Status{}
 		// When no data yet, BlockData.Height = -1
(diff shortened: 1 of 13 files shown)


chappjc:
The SQLite DB seems to be working (see screenshot below), and the branch has expanded in scope quite a bit, so it's time to merge.

In addition to the SQLite stuff, the stake difficulty estimates (from estimatestakediff and the next element of getstakedifficulty) have been removed from the /block/X/pos endpoints because it makes no sense to have this for anything but the current height.  This there are now /stake endpoints for this information.  Only the current sdiff remains in /block/...

Some silly bugs were fixed, namely the broken store/retrieve methods in dcrdatasql/sqlite.go.

A bunch of utility functions were created too.

After this, the rebuild of the db needs to be ported from the rebuilddb tool into the main dcrdata app.  Also, some basic caching of recent responses should be implemented to reduce db access, starting with the current best block!

sqlite-browser

PR #14, opened 2017-03-03
Webui start

chappjc:
Create basic WebUI type and plug it in.  Just a starting point.  No html/tmpl exists yet!

Begins to address https://github.com/dcrdata/dcrdata/issues/6


chappjc:
@karamble and @fai04d There's a framework for a web ui now.  See web.go, and the webMux in main.go, and the /public and /view folders.

PR #18, opened 2017-03-09
Resync rewrite

chappjc:
Addresses https://github.com/dcrdata/dcrdata/issues/17

This addresses the insane memory usage of rebuilddb.  Maybe also resync from arbitrary height instead of 0.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ CURRENT
 LOCK
 LOG
 MANIFEST-*
+ffldb_stake/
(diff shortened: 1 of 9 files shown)


chappjc:
Seems to work as intended.  However, it's complicated and a big change.

Now requires an on-disk ticket database (leveldb) to be able to get live tickets at the last recorded height before dcrdata exited.

chappjc:
Also note that glide install is required before build as depends changed.

PR #22, opened 2017-03-13
Update README.md.

### LINKED ISSUE #19: README missing some commands

chappjc:
Closes #19


fai04d:
Forgot to add; need to add that you need to disable tls in dcrd to get it to work. I assume this is expected as notls is specified in the conf.

chappjc:
@fai04d I've been using it w/o TLS, but it should work with it.  If it doesn't, that should be fixed.

PR #29, opened 2017-03-19
Add time stamp to mempool data.

chappjc:
Addresses https://github.com/dcrdata/dcrdata/issues/28


chappjc:
Closes https://github.com/dcrdata/dcrdata/issues/28

PR #47, opened 2017-06-21
Switch to new logger (btclog+jrick's logrotate)

chappjc:
This also removes seelog, and updates several other dependencies.


chappjc:
Version is bumped to 0.4.0, but it won't be tagged that way until some other changes are done.

PR #49, opened 2017-07-02
enable CORS for /api

gozart1:
I've started working on a separate front end and have enabled CORS so that it can make ajax requests. Saw an issue related to this https://github.com/dcrdata/dcrdata/issues/33

New to go so feedback welcome.

### DIFF
--- a/apirouter.go
+++ b/apirouter.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/go-chi/chi"
 	"github.com/go-chi/chi/middleware"
+	"github.com/rs/cors"
 )
 
 type apiMux struct {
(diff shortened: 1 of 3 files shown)


chappjc:
Good find with `rs/cors`.  Looks like this will work, but there are a few changes needed first:

1. Add github.com/rs/cors to glide, the dependency manager: `glide get github.com/rs/cors`, then say "Y" to semantic versioning and "P" to patch releases.  This should update glide.yaml and glide.lock, so you will then need to push those changes too.
2. Middleware, like cors, are currently added in apirouter.go.  I'd suggest putting it in right before `mux.Get("/", app.root)` with something like:

```go
corsMW := cors.Default();
mux.Use(corsMW.Handler)
```

Then finish up with `go fmt $(glide nv)` and it should be good to go.

Thanks for the PR!

chappjc:
Hmm, looks like `github.com/pressly/chi` has moved to `github.com/go-chi/chi`, so it will probably not build after updating glide.  But I can follow up this PR with fixes for that.

gozart1:
Made the suggested changes, squashed and rebased. Thanks for the actionable feedback.

chappjc:
Nice, thanks!

PR #51, opened 2017-07-03
WebSocket support

chappjc:
This adds websocket support to the web interface:

- /ws is the websocket path
- JS in the main page html automatically connects a websocket on this path, and updates a elements on the main page (height, hash, size, diff, sdiff) when a new block is connected.
- Added the new`WebsocketHub` to manage active connections
- `(*WebUI).WSBlockUpdater` is the new http.HandlerFunc for the `/ws` path.  It manages a single connection, from register, updates, and disconnect.
- A bunch more (will edit this PR description before merging)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 .vscode/
 *.exe
 *.log
+*.log.*.gz
 *.bak
 debug
 *.cert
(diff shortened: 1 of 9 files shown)


chappjc:
Now running on https://dcrdata.ssgen.io/

Please report any problems or errors in brower's developer console.

PR #58, opened 2017-07-20
Server-sent event / event stream support

chappjc:
Like PR https://github.com/dcrdata/dcrdata/pull/51 but using an HTML5 event stream instead of a websocket.

### DIFF
--- a/apimiddleware.go
+++ b/apimiddleware.go
@@ -10,7 +10,6 @@ import (
 
 	"github.com/go-chi/chi"
 	"github.com/go-chi/chi/docgen"
-	//"github.com/decred/dcrrpcclient"
 )
 
 // Middlewares
--- a/blockdata/blockdata.go
+++ b/blockdata/blockdata.go
@@ -47,7 +47,10 @@ func (b *BlockData) ToStakeInfoExtendedEstimates() apitypes.StakeInfoExtendedEst
 		StakeDiff: apitypes.StakeDiff{
 			GetStakeDifficultyResult: b.CurrentStakeDiff,
 			Estimates:                b.EstStakeDiff,
+			IdxBlockInWindow:         b.IdxBlockInWindow,
+			PriceWindowNum:           b.PriceWindowNum,
 		},
+		// PriceWindowNum and Idx... are repeated here since this is a kludge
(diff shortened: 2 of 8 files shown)


chappjc:
This seems pretty good now.  Server-sent events (SSE) a.k.a. event source or event streaming automatically connects when the web ui is loaded in a supported browser (not IE or Edge), and displays an indicator so you know if the event source connection is still connected:

image


When using nginx as a reverse proxy for the dcrdata web app, some settings are need to permit HTML streaming such as SSE to work.

In **nginx config**:

    proxy_http_version 1.1;

In the **HTML response header** (this PR does it in web.go):

```
Content-Type: text/event-stream;
Cache-Control: no-cache;
X-Accel-Buffering: no;
```

chappjc:
Refs for SSE/EventSource in general:

- https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events
- https://www.html5rocks.com/en/tutorials/eventsource/basics/
- https://www.w3.org/TR/2011/WD-eventsource-20111020/

For the connected indicator widget thingy:  https://github.com/html5rocks/www.html5rocks.com/tree/master/content/tutorials/eventsource/basics/static/demo

For the SSE server component details: https://github.com/kljensen/golang-html5-sse-example/blob/master/server.go

This was all based on the `WebsocketHub` from that PR, but the above resources helped for SSE differences.

chappjc:
Would like to fix the drop down menu on mobile.  @gozart do you know why it drops in front of the page like this?

menu

chappjc:
Would like to fix the drop down menu on mobile.  @gozart do you know why it drops in front of the page like this?

menu

chappjc:
This is live on https://dcrdata.ssgen.io/

After adding mempool updates, fixing the menu, reviewing by the web experts, and stress testing, then this can merge.

chappjc:
Closing this in favor of the WebSocket PR https://github.com/dcrdata/dcrdata/pull/51

@gozart1 I'd still love to make the dropdown menu work on a mobile phone.  Do you think you know why it looks like it does as shown in this comment: https://github.com/dcrdata/dcrdata/pull/58#issuecomment-317200730

PR #63, opened 2017-07-31
Speed up block range API requests with a caching system

chappjc:
The small change here is changing how the response to a block range API request is written to the `http.ResponseWriter`.  Previously, the entire marshalled JSON response had to be generated, which involved all the DB retrievals, prior to sending to the client.  Now the response is streamed as the blocks are retrieved and marshalled.  Newly-enabled zip compression complements the streaming response very well as the added latency is hidden, and bandwidth is drastically reduced allowing the full response to reach the client quickly.

The larger change in this PR is the addition of a cache for accessed blocks, `APICache`.  Access of a given block includes API requests for the block, or the addition of a new block to the cache via one of the chain monitors.

The capacity of the cache is specified in the number of blocks it should be able to hold before evicting low priority (I'll get to priority in a sec) blocks.  This is currently hard-coded to 50000, but it may be converted to a config parameter in the future.  `CachedBlock` represents a cached block, and it comprises a `BlockDataBasic` and fields for tracking access count, access time, and heap position index.  Within the `APICache` is a map APICacheof block hashes to `CachedBlock`s, and a priority queue `BlockPriorityQueue` for the `CachedBlock`s stored by `APICache`, which facilitates evicting certain blocks as needed.

This can be improved a lot, but this is enough to start testing on the interwebs.

Version bumped to 0.6.2.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 *.log.*.gz
 *.bak
 debug
+debug.test
 *.cert
 *.key
 dcrdata
--- a/dcrdataapi/apicache.go
+++ b/dcrdataapi/apicache.go
@@ -110,8 +110,17 @@ func (apic *APICache) StoreBlockSummary(blockSummary *BlockDataBasic) error {
 	// insert into the cache and queue
 	cachedBlock := newCachedBlock(blockSummary)
 	cachedBlock.Access()
-	apic.blockCache[*hash] = cachedBlock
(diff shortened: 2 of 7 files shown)


chappjc:
Closing in favor of two separate PRs:
- https://github.com/dcrdata/dcrdata/pull/176
- https://github.com/dcrdata/dcrdata/pull/177

PR #64, opened 2017-07-31
Fixed nodaemontls value

RogueElement:
Simply changed the default nodaemontls value


raedah:
This will make new developer on boarding easier with a working default value.

chappjc:
Ah, yes, I made the same mistake with hardfork demo web site.  You can tell I've only ever used it without TLS.

Thanks for the PR!

PR #72, opened 2017-08-07
Add block size endpoints, like /block/.../size.

chappjc:
This includes range, where an array of integers is returned.

A new SQLite query is added to make this fast.  Only the size column is returned.

TODO: add transaction count somewhere, somehow.


chappjc:
@gozart1 Perhaps for https://dcr-data.netlify.com/#/transactions
BTW, I need to make an issue to get a range of items with a step>1

Issue #62, opened 2017-07-27
app panics if dcrdata.conf does not exist


raedah:
should exit cleanly

chappjc:
closed by PR 76

PR #76, opened 2017-08-08
Removed log warning

RogueElement:
In reference to issue #62, log.warnf seems to always have give a run time error but removing it should exit the app cleanly and still show the error

### DIFF
--- a/config.go
+++ b/config.go
@@ -274,8 +274,7 @@ func loadConfig() (*config, error) {
 	// succeeds.  This prevents the warning on help messages and invalid
 	// options.
 	if configFileError != nil {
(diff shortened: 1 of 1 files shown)


chappjc:
Looks like other spots use `fmt.Fprintln(os.Stderr, err)`.  Does that work?

Please delete instead of comment unless there is a reason.

RogueElement:
No reason, although I feel like even the print is unnecessary since the  fourth line of maincore will print the same error

chappjc:
Yeah, you do get a double message, but that's ok.  I'd like config to give it's own errors as main or someone else might forget to do it in the future.

PR #77, opened 2017-08-10
Explorer WebUI

RogueElement:
I added a basic block explorer page with a main explorer page showing recent blocks, a block page and a transactions page


chappjc:
The commit list here is really weird.  Instead of merging in branches to this branch as you did, periodically rebasing with master is preferable.

I'd have to look at your repo to give specific remedies, but I'd suggest starting with a git rebase master, then squash (`git rebase -i` then squash all but one commit, or any other method to squash that works for you.

RogueElement:
Yeah, sorry about that, not that used to git, I'll close this now sort everything out then open another one later when I've added an address page

PR #79, opened 2017-08-15
Block Explorer

chappjc:
This is the block explorer branch (rebased and squashed) by RogueElement.

Looks like a nice start!  However, we need to get some fancy web development going.

### DIFF
--- a/explorer.go
+++ b/explorer.go
@@ -1,0 +1,134 @@
+package main
+
+import (
+	"html/template"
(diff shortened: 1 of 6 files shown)


RogueElement:
On it, plan on adding websocket support to the main explorer page and adding an address page, but I'm still terrible at design

chappjc:
@RogueElement will open a new one when ready

PR #84, opened 2017-08-18
collectionQueue to run certain collection functions synchronously and in order

chappjc:
This still allows the `OnBlockConnected` handler to return quickly and not block the RPC client, but it band-aids some issues with execution order.

Create the `collectionQueue` in `make`/`getNodeNtfnHandlers` and return it to main so that the synchronous collection functions can be set once they are available.

No longer send directly to the block connected channels for  stakedb, wiredDB, or blockdata.  Instead call the new synchronous function `BlockConnectedSync` of each package's chain monitor.  The individual block connected channels are still used in `BlockConnectedSync`, but signals are used to wait until handling of the signaled block has completed in the old handler loops.

Create `blockdata.(*Collector).CollectHash(*chainhash.Hash)` to be used in the event that the chain node best block is higher than the block requested for data collection.  Pool value may be wrong if the stake DB is not on the same height however.  But the new synchronous execution changes are intended to prevent the stake DB from getting ahead.  **EDIT**: This now adds `PoolInfoCache` in stakedb so that ticket pool info for recent (since program start) previous blocks are available.

Update `blockdata.(*chainMonitor).BlockConnectedHandler` so that it will call `Collect()` only if chain height equals the height of the requested block data, and call `CollectHash(hash)` if collection is behind.  A caveat with `CollectHash`, in addition to the pool value bit described above, is that it will not get the next block window estimates, however this is not needed if a higher block is waiting to be processed subsequently.

<s>Change reorg handling in dcrsqlite so that instead of making a slice of block hashes for the new main chain and then trying to collect data for each, which is bad because the stake DB will be at tip and pool value for intermediate blocks will not be available, instead collect with `CollectAPITypes()` and make a slice of  the the data.  The new `blockAPISourceData` type holds this data.  Then in `switchToSideChain()`, just unpack the data.</s>  Reorg handling in dcrsqlite still just stores the block hashes of the side chain, deferring actual data collection until `switchToSidechain`, but the `PoolInfoCache` in stakedb will have pool info for the entire side chain since it's block-connected handler runs _before_ dcrsqlite's.  So as dcrsqlite's chainmonitor runs `CollectAPITypes(hash)` for each side chain block, the correct pool info will be obtained from stakedb.

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,8 +190,8 @@ by other dcrdata packages, but they may be of general value in the future.
 * The `chainMonitor` type and its `BlockConnectedHandler()` method that handles
   block-connected notifications and triggers data collection and storage.
 * The `BlockData` type and methods for converting to API types.
-* The `blockDataCollector` type its `Collect()` method that is called by
-  the chain monitor when a new block is detected.
+* The `blockDataCollector` type and its `Collect()` and `CollectHash()` methods
+  that are called by the chain monitor when a new block is detected.
 * The `BlockDataSaver` interface required by `chainMonitor` for storage of
   collected data.
 
--- a/blockdata/blockdata.go
+++ b/blockdata/blockdata.go
@@ -130,7 +130,8 @@ func (t *Collector) CollectBlockInfo(hash *chainhash.Hash) (*apitypes.BlockDataB
 	// Ticket pool info (value, size, avg)
 	ticketPoolInfo, sdbHeight := t.stakeDB.PoolInfo()
 	if sdbHeight != height {
-		log.Warnf("Collected block height %d != stake db height %d. Pool info will not match.", height, sdbHeight)
+		log.Warnf("Collected block height %d != stake db height %d. Pool info "+
+			"will not match the rest of this block's data.", height, sdbHeight)
 	}
 
 	// Fee info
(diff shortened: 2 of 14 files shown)


chappjc:
Running this with a `-race` build on dcrdata.ssgen.io

chappjc:
The need for the `PoolInfoCache` is this.

A reorg is handled in `stakedb` by first switching into reorg mode, in which subsequent connect block ntfns pertain to side chain blocks, building a slice of block hashes for the side chain, and on the final/best block of the side chain actually switching to the side chain.  The switch is done by first disconnecting main chain blocks back to the common ancestor of the two chains, then connecting blocks from the side chain slide to form the new main chain.  During this final connection process is where ticket pool info at each block gets computed and stored in `PoolInfoCache`.

The `dcrsqlite` chain monitor flips into reorg mode at roughly the same time as stakedb.  But the difference is that a given block connected ntfn gets handled after the stakedb's chain monitor handles it.  So the switch to side chain in dcrsqlite will occur once the switch is complete in stakedb.  And unless there is a single switch-to-sidechain method that synchronizes each packages reorg better, it is easiest to acquire pool info as blocks are connected in stakedb's switch-to-sidechain function and save the data for when dcrsqlite (or whoever asks) requests pool info for a given block hash.

PR #87, opened 2017-08-20
Step param in block range requests

chappjc:
`/api/block/range/{idx0}/{idx}/{step}`
`/api/block/range/{idx0}/{idx}/{step}/size`


chappjc:
Darn I meant to squash this.

PR #88, opened 2017-08-20
Add sample nginx.conf and rate_limiting.html

chappjc:
The sample nginx.conf demonstrates several important settings when using nginx as a reverse proxy:

- websocket upgrade
- RealIP module
- rate limiting
- auto-HTTPS redirect
- SSL cert setup
- gzip settings

The sample rate_limiting.html demonstrates a possible response to the client when requests are being refused due to rate limiting.


chappjc:
Partially addresses https://github.com/dcrdata/dcrdata/issues/30, but need golang side of that

Issue #67, opened 2017-08-03
Starting another instance of dcrdata clears database


RogueElement:
if another instance of  dcrdata attempted while one is running, it sees that dcrdata.sqlt.db is in use and overwrites it, effectively clearing a fully loaded db and the original process is not aware of this. dcrdata should refuse to start if a dcrdata process is confirmed to be running

PR #90, opened 2017-08-21
Do not continue startup if stake DB is already opened (unavailable)

chappjc:
Fixes Issue #67.


chappjc:
If anyone wants to test this:

```
git fetch origin pull/75/head:no-overwrite-stakedb
git checkout no-overwrite-stakedb
```

PR #98, opened 2017-08-25
Make elements on web ui pulse when stuff happens on the websocket handlers

chappjc:
I suck at HTML and CSS, so I don't really care about doing this right at the moment.  Maybe later.

Fix some other stuff like links not opening in new tabs.


chappjc:
No longer relevant with redesign.  Besides, I was just playing around with this.

PR #103, opened 2017-08-28
Block Explorer UI

RogueElement:
Added a simple recent blocks page `/explorer`, a block page `/explorer/block/{blockhash}`, a transaction page `/explorer/tx/{txid}` and an address page `/explorer/address`.


views/extras.tmpl
```diff
@@ -0,0 +1,87 @@
...
+                </a>
+            </div>
+            <div id="navbar" class="navbar-collapse collapse">
+                <ul class="nav navbar-nav">
+                </ul>
+                <ul class="nav navbar-nav navbar-right">
+                    <li><a href="https://github.com/dcrdata/dcrdata" title="dcrdata on GitHub" title="_blank">GitHub</a></li>
+                    <li><a href="https://github.com/dcrdata/dcrdata#json-rest-api" title="API Endpoints" title="_blank">API</a></li>
+                    <li><a href="/api/status" title="API Status" title="_blank">API Status</a></li>
+                    <li><a href="/explorer" title="Block Explorer" title="_blank">Explore</a></li>
```
chappjc: I discovered in some work on another branch that this as supposed to be `target="_blank"` instead of `title="_blank"`.  Might as well change them here.

explorer.go
```diff
@@ -0,0 +1,229 @@
...
+		"getTime": getTime,
+	}
+	explorerTemplate, _ := template.New("explorer").Funcs(helpers).ParseFiles("views/explorer.tmpl", "views/extras.tmpl")
+
+	idx := c.BlockData.GetHeight()
+	N := 25
+	start, errS := strconv.Atoi(r.URL.Query().Get("startat"))
+	if errS != nil || start == 0 {
+		start = idx - N + 1
+	}
```
gozart1: Noticed browser crashes when I ask for large range, probably simply can't render that much html.
Maybe add a pagesize param that defaults to N?

views/block.tmpl
```diff
@@ -0,0 +1,67 @@
...
+<body>
+    <!-- Fixed navbar -->
+    {{template "navbar"}}
+
+    <div class="wrapper" style="margin-top:60px;">
+        <div id="boxagenda">
+            <div style="clear:both;">&nbsp;</div>
+            <h4><span>Block #{{.Height}}</span></h4>
+
+            <div class="col-md-15">
```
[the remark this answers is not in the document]
chappjc: Or this could be in two columns depending on the window size

explorer.go
chappjc: yes, me too.  there needs to be a limit.  The API request for 10k blocks takes less than 10 seconds (still a alot), but the browser just can't handle it.

explorer.go
```diff
@@ -0,0 +1,229 @@
...
+
+func (c *appContext) explorerUI(w http.ResponseWriter, r *http.Request) {
+	helpers := template.FuncMap{
+		"getTime": getTime,
+	}
+	explorerTemplate, _ := template.New("explorer").Funcs(helpers).ParseFiles("views/explorer.tmpl", "views/extras.tmpl")
+
+	idx := c.BlockData.GetHeight()
+	N := 25
+	start, errS := strconv.Atoi(r.URL.Query().Get("startat"))
```
chappjc: just `start`

explorer.go
chappjc: Some way to control parameters `start=A&end=B` would perhaps be better than specifying one end and having the other end be best block height.

views/block.tmpl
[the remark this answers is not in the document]
RogueElement: I tried the two column thing but I was either overlapping or leaning to a side, I just don't know how to do it

explorer.go
chappjc: hmm, I was thinking start<end, but I suppose if you consider start is at the top of the page, I can see why start would be the larger number.  But IMO it does seem counter-intuitive to do `start=100&end=0` and not the other way around.

explorer.go
```diff
@@ -29,30 +29,36 @@ func (c *appContext) explorerUI(w http.ResponseWriter, r *http.Request) {
...
+	if errE != nil || end == 0 || end > idx {
+		end = idx
 	}
-	end, errS := strconv.Atoi(r.URL.Query().Get("end"))
-	if errS != nil || end == 0 || start-end < 10 {
-		end = start - 26
-	} else if start-end > 200 {
-		end = start - 200
+	start, errS := strconv.Atoi(r.URL.Query().Get("start"))
+	if errS != nil || start == 0 || end-start < 10 {
```
chappjc: we should have data for block 0 (genesis block), although I can't recall how this is handled.  If there's nothing breaking about the data in genesis, we should not prevent it from being queried

views/explorer.tmpl
```diff
@@ -41,11 +41,9 @@
                 </tr>
             {{end}}
             </table>
-            <form action="/explorer#{{.Last}}"  class="form">
-                from: <input type="text" id="start" name="start" style="color:blue;">
-                to: <input type="text" id="end" name="end" style="color:blue;">
-                <input type="submit" value="go" style="color:blue;">
-            </form>
+            <p id="best"class="hidden">{{.BestBlock}}</p>
```
chappjc: need space here

explorer.go
```diff
@@ -169,6 +161,34 @@ func (c *appContext) addressPage(w http.ResponseWriter, r *http.Request) {
 	io.WriteString(w, str)
 }
 
+func (c *appContext) search(w http.ResponseWriter, r *http.Request) {
+	searchStr := r.URL.Query().Get("search")
```
chappjc: I think we want a chi path here with a middleware like the usual pattern.

explorer.go
```diff
@@ -169,6 +161,34 @@ func (c *appContext) addressPage(w http.ResponseWriter, r *http.Request) {
...
+	_, err := c.BlockData.GetBlockHeight(searchStr)
```
chappjc: As this is refined, let's please put plenty of comments within the function in addition to the docs above the function.  Specifically, talk through the logic of how you decide what the string represents.

views/block.tmpl
```diff
@@ -26,7 +26,7 @@
                     {{if .NextHash}}
                     Next Block Hash: <span style="float: right;font: 12px monospace;" >{{template "copyButton" .NextHash}}<a href="/explorer/block/{{.NextHash}}">{{.NextHash}}</a></span><br>
                     {{end}}
-                    Transactions: <span style="float: right;" >{{.TxCount}}</span><br>
+                    Transactions: <span style="float: right;" >{{.RawTx | len}}</span><br>
```
chappjc: do we need to pipeline this here?  Maybe I'm forgetting how this works, but I though you could just do `{{len .RawTx}}`

explorer.go
```diff
@@ -0,0 +1,257 @@
...
+
+type explorerMux struct {
+	*chi.Mux
+}
+
+func (c *appContext) explorerUI(w http.ResponseWriter, r *http.Request) {
+	helpers := template.FuncMap{
+		"getTime": getTime,
+	}
+	explorerTemplate, _ := template.New("explorer").Funcs(helpers).ParseFiles("views/explorer.tmpl", "views/extras.tmpl")
```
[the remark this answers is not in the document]
raedah: There was a suggestion from gozart to make this template parsing configurable with an environment variable. I can see how designers wouldnt want to have to restart the daemon while doing edits.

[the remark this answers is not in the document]
chappjc: Agreed that it's absolutely necessary to avoid restarting just to update templates.  The ability to signal a reload if the templates with SIGUSR1 was added for this purpose: https://github.com/dcrdata/dcrdata/pull/37/commits/2460e455fda8f9cac8905f217af7151432fbe04b
@gozart1 let me knew if this addresses the need

explorer.go
```diff
@@ -29,12 +29,12 @@ func (c *appContext) explorerUI(w http.ResponseWriter, r *http.Request) {
...
-	end, errE := strconv.Atoi(r.URL.Query().Get("end"))
-	if errE != nil || end == 0 || end > idx {
+	end, err := strconv.Atoi(r.URL.Query().Get("end"))
+	if err != nil || end == 0 || end > idx {
 		end = idx
 	}
-	start, errS := strconv.Atoi(r.URL.Query().Get("start"))
-	if errS != nil || start == 0 || end-start < 10 {
+	start, err := strconv.Atoi(r.URL.Query().Get("start"))
+	if err != nil || start < 0 || end-start < 10 {
```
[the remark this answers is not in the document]
RogueElement: This is so that the explorer list always shows a minimum of ten blocks

views/block.tmpl
RogueElement: Forgot I could do that, so used to working with the mapped functions

explorer.go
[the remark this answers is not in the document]
RogueElement: Added this to the explorer as well

explorer.go
```diff
@@ -39,219 +49,333 @@ func (c *appContext) explorerUI(w http.ResponseWriter, r *http.Request) {
...
-	}
-	txTemplate, _ := template.New("address").Funcs(helpers).ParseFiles("views/address.tmpl", "views/extras.tmpl")
-
 	if !ok {
 		apiLog.Trace("address not set")
 		http.Error(w, "address not set", http.StatusInternalServerError)
 		return
 	}
-	data := c.BlockData.GetAddressTransactions(address, 10)
+	data := exp.app.BlockData.GetAddressTransactions(address, 2000)
```
chappjc: I guess eventually this will be some kind of reactive display thing where you click for more, but it's probably OK like this for now.

explorer.go
```diff
@@ -39,219 +49,333 @@ func (c *appContext) explorerUI(w http.ResponseWriter, r *http.Request) {
...
 }
 
-func (c *appContext) search(w http.ResponseWriter, r *http.Request) {
-	searchStr := r.URL.Query().Get("search")
-	_, err := c.BlockData.GetBlockHeight(searchStr)
-	if err == nil {
-		http.Redirect(w, r, "/explorer/block/"+searchStr, http.StatusPermanentRedirect)
+// search implements a primitive search algorithm by checking if the value in
+// question is a block index, block hash, address hash or transaction hash and
+// redirects to the appropiate page or displays an error
```
chappjc: appropriate

chappjc:
In the future it would be better to do one feature at a time with PRs.  The layout overhaul is great, but ideally it would be a different PR.

raedah:
@chappjc Maybe time soon to merge this in soon so we can start getting smaller PRs going?

chappjc:
@raedah yeah, I'm itching to see this go in too.  Will be exciting to deploy.  But there are a couple things to iron out.  If that's it with added commits, the changes should be relatively small.

@RogueElement On the Stake Transactions part of the /explorer/block/... page the types are incorrect.  At least it appears that "Vote" and "Ticket" are switched:

image

Also, looking at the tables on these pages, there are some other issues:

- The "In" side of regular transactions is not right, the values are always something like "-0.0000001 DCR"  See /explorer/tx/1e31018e0f56911fd41cca57be298b258d275028328b66ad51032c73a46f009c
image

- The "Total Value" column on the Address page has "DCR" repeated: 
image

- The search form still uses URL query instead of path (http://127.0.0.1:7777/explorer?search=12 vs http://127.0.0.1:7777/explorer/block/12).   Which is "right" is up for debate, but the routing framework favors the latter.

- I think the headers should dispense with the ":" suffix.
- The "Transactions ID" column header -- would be clear as just "Transaction ID"


@gozart1 The redesign looks awesome BTW!

chappjc:
Needs rebase.  Thanks!

RogueElement:
Cannot reproduce the error for the In values on `/explorer/tx/1e31018e0f56911fd41cca57be298b258d275028328b66ad51032c73a46f009c`

chappjc:
It's not showing those values for me either.  Not sure what happened there on there.

chappjc:
Looks like root.tmpl needs `{{template "html-head"}}` too.
Also, Ticket fees (mempool and next block) do not seem to be live updating.  I think the id tags might be missing from them.
Nothing else is a show stopper though.

RogueElement:
Did a messy rebase, looking into root

views/extras.tmpl
```diff
@@ -21,7 +21,8 @@
     <link rel="icon" href="/favicon.ico">
     <title>dcrdata web</title>
     <link href="/css/main.css" rel="stylesheet">
-    <link rel="stylesheet/less" type="text/css" href="/css/dynamic.less">
+    <!-- <link rel="stylesheet/less" type="text/css" href="/css/dynamic.less"> -->
+    <script src="js/messagesocket.js"></script>
```
chappjc: leading forward slash:
<script src="/js/messagesocket.js"></script>

Issue #106, opened 2017-08-29
Integrate data from various time periods


chappjc:
Summaries of certain counts such as number of blocks, transactions, etc. could be generated on running time windows (e.g. hourly, daily, weekly, monthly).

the time of each block is in sqlite, so this can be done.

xaur:
I would love it so we could have some montly network health report. Is it possible to also include average ticket price and average hashrate?

raedah:
Related to https://github.com/decred/dcrdata/issues/135

PR #107, opened 2017-09-01
Added querying for blocks from time frame

RogueElement:
for issue #106, added a way to get blocks from a timeframe and added an example endpoint `/api/block/time/day/{MM-DD-YYY}` to get blocks on an specific day,  `/api/block/time/day/` returns blocks from the current day. Wanted a review before proceeding

### DIFF
--- a/apiroutes.go
+++ b/apiroutes.go
@@ -4,9 +4,11 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"math"
 	"net/http"
 	"strconv"
 	"sync"
+	"time"
 
 	apitypes "github.com/dcrdata/dcrdata/dcrdataapi"
 	"github.com/decred/dcrd/dcrjson"
@@ -39,6 +41,7 @@ type APIDataSource interface {
 	GetBestBlockSummary() *apitypes.BlockDataBasic
 	GetBlockSize(idx int) (int32, error)
 	GetBlockSizeRange(idx0, idx1 int) ([]int32, error)
+	GetBlockSummaryByTime(start, end int) ([]*apitypes.BlockDataBasic, error)
 	GetPoolInfo(idx int) *apitypes.TicketPoolInfo
 	GetPoolInfoByHash(hash string) *apitypes.TicketPoolInfo
 	GetPoolInfoRange(idx0, idx1 int) []apitypes.TicketPoolInfo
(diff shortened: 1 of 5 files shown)


apiroutes.go
```diff
@@ -242,6 +245,21 @@ func getAddressCtx(r *http.Request) string {
...
+		today := time.Now()
+		return int(math.Floor(float64(today.Unix() / 86400)))
+	}
+	dateTime, err := time.Parse("01-02-2006", dateString)
+	if err != nil {
+		apiLog.Trace("cannot parse date %s: %s", dateString, err.Error())
+		today := time.Now()
+		return int(math.Floor(float64(today.Unix() / 86400)))
+	}
+	return int(math.Floor(float64(dateTime.Unix() / 86400)))
```
chappjc: I think the math.Floor is unnecessary since the division is two integers.

apiroutes.go
```diff
@@ -803,6 +821,36 @@ func (c *appContext) getBlockRangeSteppedSummary(w http.ResponseWriter, r *http.
 	fmt.Fprintf(w, "]")
 }
 
+func (c *appContext) getBlockSummaryByDate(w http.ResponseWriter, r *http.Request) {
+	date := getDateCtx(r)
+	blocks, er := c.BlockData.GetBlockSummaryByTime(date*86400, (date+1)*86400)
```
chappjc: Might as well keep the int64 all over and keep the units in seconds.

Also, and alt solution to day rounding:  https://play.golang.org/p/9jVLG6vQci

dcrsqlite/sqlite.go
```diff
@@ -596,6 +597,45 @@ func (db *DB) RetrieveBlockSizeRange(ind0, ind1 int64) ([]int32, error) {
 	return blockSizes, nil
 }
 
+// RetrieveBlockSummaryRangeTime returns an array of apitypes.BlockDataBasic for
+// blocks in given time frame
+func (db *DB) RetrieveBlockSummaryRangeTime(start, end int64) ([]*apitypes.BlockDataBasic, error) {
+	N := start - end
+	if N == 0 {
```
chappjc: Also check for end<start

dcrsqlite/sqlite.go
```diff
@@ -596,6 +597,45 @@ func (db *DB) RetrieveBlockSizeRange(ind0, ind1 int64) ([]int32, error) {
...
+		return []*apitypes.BlockDataBasic{}, nil
+	}
+
+	blocks := []*apitypes.BlockDataBasic{}
```
chappjc: Will also need a limit on max blocks

chappjc: So we should estimate as days*288

dcrsqlite/sqlite.go
```diff
@@ -596,6 +597,45 @@ func (db *DB) RetrieveBlockSizeRange(ind0, ind1 int64) ([]int32, error) {
...
+		return nil, err
+	}
+	defer rows.Close()
+
+	for rows.Next() {
+		bd := new(apitypes.BlockDataBasic)
+		if err = rows.Scan(&bd.Height, &bd.Size, &bd.Hash, &bd.Difficulty,
+			&bd.StakeDiff, &bd.Time, &bd.PoolInfo.Size, &bd.PoolInfo.Value,
+			&bd.PoolInfo.ValAvg); err != nil {
+			log.Errorf("Unable to scan for sdiff fields: %v", err)
```
chappjc: if scan errors, I think we should `continue` instead of appending an empty bd

raedah:
@RogueElement Closing this since what we will summing the totals of all blocks in the range, rather then displaying a list of blocks for the range.

PR #113, opened 2017-09-06
dock footer nav to bottom of page

gozart1:
This change makes sure the page body expands the full height of the page and the footer nav is absolutely position to the bottom. 

This means the footer won't be visible in mobile unless you scroll all the way down, but the upside is it maximizes the amount of vertical screen real estate for viewing data.

Sample mobile view:
screen shot 2017-09-05 at 6 52 44 pm
 
But we can also do fixed position, which means it will always be visible, eg:
screen shot 2017-09-05 at 6 53 03 pm

Let me know which approach you prefer to address
https://github.com/dcrdata/dcrdata/issues/109

### DIFF
--- a/public/css/main.css
+++ b/public/css/main.css
@@ -9,19 +9,28 @@
 /* Quick and Dirty CSS, just layer on top of boostrap */
 
 html {
-    color: #5f5f5f;
+  color: #5f5f5f;
+  min-height: 100vh;
+}
+body {
+  position: relative;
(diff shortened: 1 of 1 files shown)


chappjc:
I prefer fixed so you can always tell if the page is stale.

gozart1:
closed in favor of https://github.com/dcrdata/dcrdata/pull/114

Issue #126, opened 2017-09-06
Give clear message instead of 422 InternalServiceError when tx not found


chappjc:
For example, the first transaction in the genesis block: e7dfbceac9fccd6025c70a1dfa9302b3e7b5aa22fa51c98a69164ad403d60a2c

It's listed as a transaction on block 0, but it can't display it up because `getrawtransaction` fails (that's another issue as this is a special tx).  The web browser just shows:

    Internal Server Error

And the dcrdata logs show:

```
[ERR] DSQL: GetRawTransactionVerbose failed for: e7dfbceac9fccd6025c70a1dfa9302b3e7b5aa22fa51c98a69164ad403d60a2c
[ERR] JAPI: Unable to get transaction e7dfbceac9fccd6025c70a1dfa9302b3e7b5aa22fa51c98a69164ad403d60a2c
"GET http://dcrdata.ssgen.io/explorer/tx/e7dfbceac9fccd6025c70a1dfa9302b3e7b5aa22fa51c98a69164ad403d60a2c HTTP/1.1" from x.x.x.x - 500 22B in 1.316908ms
```

There should be a nice error message in the browser.  For instance, Insight via the search box has a transient text box that says "No matching records found!", while if you type in the URL for the missing transaction it redirects to the main explorer page showing latest blocks and a flash message that says "Transaction Not Found".

raedah:
Completed in https://github.com/dcrdata/dcrdata/pull/128/
though template needs a restyling, tracking in https://github.com/dcrdata/dcrdata/issues/118

Issue #124, opened 2017-09-06
Show more granularity on Age


raedah:
The blocks view has an Age column. Currently it shows 1 minutes, 1 hour, 1 day, 1 month, 1 year.

I suggest these
1s
1m 30s
1h 30m
1d 12h
1mo 15d
1y 6mo

raedah:
Completed in https://github.com/dcrdata/dcrdata/pull/128/

Issue #123, opened 2017-09-06
Pagination next button does not work


raedah:
Pagination Next button is not working.
Also, Instead of 'start' and 'end' parameters. Should be 'height' and 'rows', with a sane max row return. Probably 2000 since that is what we use elsewhere. Make default 12 ( 1 hours worth ).

raedah:
Completed in https://github.com/dcrdata/dcrdata/pull/128/

Issue #122, opened 2017-09-06
Show Total Sent per block


raedah:
Sum of all transactions in the block on the blocks view.

~~Sum of all transactions in the block on the block view.~~ DONE in https://github.com/dcrdata/dcrdata/pull/128

raedah:
This is resource intensive and needs a caching layer before being implemented on the blocks view.

chappjc:
perhaps also mining fees collected in block (spent - sent)

raedah:
ok, ive added the mining fee display to https://github.com/dcrdata/dcrdata/issues/138

chappjc:
@gozart1 is progressing on this

chappjc:
Blocks view now shows total sent in the "DCR" column.

PR #128, opened 2017-09-07
Bug Fixes and Enhancements

RogueElement:
This should address issues:
#116, #117, #118, #119, #120, #122, #123, #124 and #126

### DIFF
--- a/views/root.tmpl
+++ b/views/root.tmpl
@@ -2,7 +2,7 @@
 <!DOCTYPE html>
 <html lang="en">
 
-{{ template "html-head"}}
+{{ template "html-head" "Dcrdata Web"}}
 
 <body>
         <style>
--- a/views/root.tmpl
+++ b/views/root.tmpl
@@ -334,7 +334,7 @@
         <h4><span>Ticket Fees</span></h4>
         <div class="row">
             <div class="col-md-6">
-                <div>Number of Tickets in Mempool (queued for mining): <strong id="mempoolfeeinfo_number">{{.MempoolFeeInfo.Number}}</strong></div>
+                <div>Tickets in Mempool: <strong id="mempoolfeeinfo_number">{{.MempoolFeeInfo.Number}}</strong></div>
                 <strong>Fees:</strong>
                 <table class="table table-sm">
                     <thead>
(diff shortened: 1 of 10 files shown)


raedah:
ACK I have tested this and updated the related issues. Can be merged following code review.

views/root.tmpl
```diff
@@ -356,7 +356,7 @@
                 </table>
             </div>
             <div class="col-md-6 text-left">
-                <div>Number of Tickets Mined in Last Block: <strong id="feeinfo_number">{{.StakeSummary.Feeinfo.Number}}</strong></div>
+                <div>Tickets in Lastest Block: <strong id="feeinfo_number">{{.StakeSummary.Feeinfo.Number}}</strong></div>
```
gozart1: Should be Latest, not Lastest

views/tx.tmpl
```diff
@@ -50,13 +70,24 @@
...
+                            N/A
+                        {{else}}
+                            <a href="/explorer/block/{{$.Block.BlockHash}}">{{.BlockHeight}}</a></td>
+                        {{end}}
                         </td>
-                        <td>{{if ne .BlockHeight 0}}{{.BlockHeight}}{{end}}</td>
                         <td>{{.Sequence}}</td>
                         <td>{{.Tree}}</td>
-                        <td>{{.AmountIn | getAmount}}</td>
+                        <td>{{if eq .AmountIn -1e-8}} N/A {{else}} {{.AmountIn}} {{end}}</td>
```
chappjc: Unless we understand clearly where -1e-8 is coming from (is it hard-coded somewhere?) I'd prefer if we change this to <0, but it's OK for now.

web.go
```diff
@@ -177,6 +187,31 @@ func (td *WebUI) RootPage(w http.ResponseWriter, r *http.Request) {
...
+	str, err := TemplateExecToString(td.errorTempl, "error", struct {
+		ErrorCode   string
+		ErrorString string
+	}{
+		code,
+		msg,
+	})
+	if err != nil {
+		http.Error(w, "template execute failure", http.StatusInternalServerError)
+		return
```
chappjc: need to log `err`

chappjc:
The "dcrdata.org" link in the footer goes to the wrong place.  I'm not sure where it should go tbh as we don't have an general organization page, and only a basic GH page for the dcrdata repository at http://www.dcrdata.org/

PR #146, opened 2017-09-08
simplify cookie logic, fixes issue of multiple cookies created based on path

gozart1:
@chappjc I noticed that multiple cookies where being created based on path and was causing issue persisting the selected theming option while navigating though the explorer.
This simplifies the logic and solves that issue.


chappjc:
Seems to work alright, but the light theme always seems to be shown briefly (sometimes noticeable such as with a new address with lots of transactions).

gozart1:
yea, we need a way to set the `darkBG` class on body on the server side when that is the preferred state.

@raedah is looking into using cookie to accomplish that.

gozart1:
For now this should be merged in though because the current master has a bug where the theme state does not persist reliably between routes to multiple cookies being created for various paths.

PR #148, opened 2017-09-08
cleanup on templates

raedah:
Simple cleanups on some templates.
Removed the ticket fee info on the front page since its no longer useful with the current sdiff algo.
Removed block info to prepare for addition of https://github.com/dcrdata/dcrdata/issues/147

### DIFF
--- a/views/error.tmpl
+++ b/views/error.tmpl
@@ -1,19 +1,15 @@
 {{define "error"}}
 <!DOCTYPE html>
 <html lang="en">
-{{template "html-head" printf "Error %s" .ErrorCode}}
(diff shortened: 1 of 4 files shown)


chappjc:
I mentioned in Issue 147 that I think removing all the root explorer page stuff should go with those addition suggested in that issue.  Just that this takes away a lot of core info.  Should this stuff be relocated to a /explorer/status page?

raedah:
All the same block info is going to be in the new explorer rows that are being added, so the only thing that is removed is the ticket fee info, because only the min fee is necessary now.

PR #151, opened 2017-09-09
Middleware for Cache-Control: max-age=X on FileServer routes

chappjc:
Add config option cachecontrol-maxage, defaulting to 86400 seconds.

This hints to the browser that the css and js and stuff in files served by FileServer may be cached.


chappjc:
We should carefully review https://developers.google.com/speed/pagespeed/insights/

chappjc:
We can get to the other tweaks later.  Gonna merge this now.

PR #152, opened 2017-09-09
Fix light theme flash

RogueElement:
Moving the cookie logic script to the top of the body should fix the temporary light theme flash


chappjc:
I'll defer review to @gozart1 , but if it works, it works.

gozart1:
Nice solution.

PR #155, opened 2017-09-10
Previous OutPoint Addresses in Tx view

chappjc:
Modify the tx.tmpl for the Transaction page so that it shows the Previous Outpoints, their first Address (could be multisig), Block, and Value. Tree is dropped.
Add txhelpers.OutPointAddresses, which gets the addresses payed to by a transaction output.
The outpoint CSS class is added to be a bit wider than the hash type.

This probably needs tweaking, but here it is:
image


chappjc:
Addresses Issue https://github.com/dcrdata/dcrdata/issues/140

PR #160, opened 2017-09-12
Reverse address tx list

chappjc:
Addresses Issue https://github.com/dcrdata/dcrdata/issues/143

Dependencies are updated to get current revision of dcrrpcclient, which contains the fix for the `SearchRawTransactions` bug described in https://github.com/decred/dcrrpcclient/pull/89

Please verify that this is the intended operation.


chappjc:
Remember to wipe `vendor/` and then `glide i` or this won't work.

raedah:
Tested, verified newest address now show up first.

PR #162, opened 2017-09-13
format explorer numbers with commas (WIP)

gozart1:
This change imports the go-humanize lib to:

- format various kinds of numbers with commas in the explorer.
- formats bytes so that it shows in "kB" or "B" depending on size.
- Instead of printing out 'DCR' in every cell, use the column header to label the unit

What is the preferred way to share these functions so they can also be used in root.tmpl?

### DIFF
--- a/explorer.go
+++ b/explorer.go
@@ -10,15 +10,15 @@ import (
 	"strconv"
 	"time"
 
-	"github.com/decred/dcrutil"
-
 	apitypes "github.com/dcrdata/dcrdata/dcrdataapi"
 	"github.com/decred/dcrd/chaincfg/chainhash"
 	"github.com/decred/dcrd/dcrjson"
-
+	"github.com/decred/dcrutil"
+	
 	"github.com/go-chi/chi"
 	"github.com/go-chi/chi/middleware"
 	"github.com/rs/cors"
+	"github.com/dustin/go-humanize"
 )
 
 const (
@@ -290,6 +290,25 @@ func newExplorerMux(app *appContext, userRealIP bool) *explorerUI {
 	exp.templateFiles["extras"] = filepath.Join("views", "extras.tmpl")
 
 	exp.templateHelpers = template.FuncMap{
+		"uint32Comma": func(v uint32) string {
+			i64 := int64(v)
+			return humanize.Comma(i64)
+		},
+		"int64Comma": func(v int64) string {
+			t := humanize.Comma(v)
+			return t
+		},
+		"float64Commaf": func(v float64) string {
+			return humanize.Commaf(v)
+		},
+		"formatBytes": func(v int32) string {
+			i64 := uint64(v)
+			return humanize.Bytes(i64)
+		},
+		"formatBytesInt": func(v int) string {
+			i64 := uint64(v)
+			return humanize.Bytes(i64)
+		},
 		"timezone": func() string {
 			t, _ := time.Now().Zone()
 			return t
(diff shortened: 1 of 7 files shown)


gozart1:
looks like CI had some issue building. Is there something else that needs to be done besides updating glide.yaml with the go-humanize lib?

chappjc:
right, you'll need to do a `glide get github.com/dustin/go-humanize`, which will update glide.yaml and glide.lock

explorer.go
```diff
@@ -1,6 +1,8 @@
 package main
 
 import (
+	"github.com/decred/dcrutil"
+	"github.com/dustin/go-humanize"
```
chappjc: please group imports by site (all github.com imports together)

glide.yaml
```diff
@@ -31,3 +31,6 @@ import:
   version: ~3.0.0
 - package: github.com/rs/cors
   version: ~1.1.0
+- package: github.com/rs/cors
```
chappjc: oops, looks like you've got cors repeated here

glide.yaml
```diff
@@ -31,3 +31,6 @@ import:
...
+  version: ~1.1.0
+- package: github.com/dustin/go-humanize
```
chappjc: to add this, first remove the manually added line, then use `glide get github.com/dustin/go-humanize`.  Then you'll have an updated glide.yaml and glide.lock

views/address.tmpl
```diff
@@ -17,7 +17,7 @@
                 <table class="table table-sm striped">
                     <thead>
                         <th>Transactions ID</th>
-                        <th>Total Value</th>
+                        <th>DCR Value</th>
```
chappjc: I kind of liked Total, because it implied it was a sum over the transactions paying to the address.  Perhaps Total DCR Value or Total Value (DCR) if that's not to much space?

views/address.tmpl
```diff
@@ -26,10 +26,10 @@
                         {{range .Transactions}}
                         <tr>
                             <td><a href="../tx/{{.TxID}}" class="hash collapse">{{.TxID}}</a></td>
-                            <td>{{.Value | getAmount}}</td>
+                            <td>{{.Value | float64Commaf}}</td>
```
chappjc: or you can put "DCR" at the end of this <td> instead of in the column header, but I like not having it on every single amount, as you made it.

chappjc:
I'm cool with the way you've made the functions accessible in the template.  Maybe there's better/more elegant ways, but that's not my forte.

views/block.tmpl
```diff
@@ -127,7 +127,7 @@
                 <table class="table table-sm striped">
                     <thead>
                         <th>Transactions ID</th>
-                        <th>Total Value</th>
+                        <th>Total DCR</th>
```
gozart1: @chappjc how about "Total DCR" ?

chappjc: Yeah, that works

glide.lock
```diff
@@ -61,7 +61,7 @@ imports:
 - name: github.com/decred/dcrrpcclient
   version: c25ea196a4f6bd9c37a7d74224f6e6027c3ee3d7
 - name: github.com/decred/dcrutil
-  version: ddbde93f65ab0692e54ed8a5ad325fa2e8af4daa
+  version: a5fab53cab39b793142c8453caa4c6f83bc152d4
   subpackages:
```
chappjc: This actually is going backwards.  Glide is kind of messed up as it loves to update everything even when but requested, but also it seems incapable of hosting is cache.  Before a glide update, or apparent glide get, it's necessary to glide cc.  But usually I just pick pieces of glide.lock edits to keep with git add -p glide.lock

chappjc:
Needs a rebase to resolve some conflicts too.  Thanks

chappjc:
All good now except it looks like dcrutil is still being moved to an older commit:

```
-  version: ddbde93f65ab0692e54ed8a5ad325fa2e8af4daa
+  version: a5fab53cab39b793142c8453caa4c6f83bc152d4
```

The tip was ddbde9, so that can stay.  Manually editing glide.lock back to that revision should work.

gozart1:
I've rebased again and corrected the dcrutil version hash

PR #163, opened 2017-09-13
Added latest blocks, coin supply and block subsidy to home page. Added age to all pages

RogueElement:


web.go
```diff
@@ -176,9 +194,24 @@ func (td *WebUI) RootPage(w http.ResponseWriter, r *http.Request) {
...
+	str, err := TemplateExecToString(td.templ, "home", struct {
+		InitialData []*dcrjson.GetBlockVerboseResult
+		Data        WebTemplateData
+	}{
+		initialBlocks,
+		td.TemplateData,
+	})
 	td.templateDataMtx.RUnlock()
 	if err != nil {
+		fmt.Print(err)
```
chappjc: `log.Errorf`

blockdata/blockdata.go
```diff
@@ -119,14 +136,22 @@ func (t *Collector) CollectAPITypes(hash *chainhash.Hash) (*apitypes.BlockDataBa
...
 	msgBlock, err := t.dcrdChainSvr.GetBlock(hash)
 	if err != nil {
-		return nil, nil, nil, err
+		return nil, nil, nil, nil, err
 	}
 	height := msgBlock.Header.Height
 	block := dcrutil.NewBlock(msgBlock)
-
+	txLen := len(block.Transactions())
+	coinSupply, _ := t.dcrdChainSvr.GetCoinSupply()
```
chappjc: why can we ignore the errors?

blockdata/blockdata.go
```diff
@@ -119,14 +136,22 @@ func (t *Collector) CollectAPITypes(hash *chainhash.Hash) (*apitypes.BlockDataBa
...
+	nbSubsidy, _ := t.dcrdChainSvr.GetBlockSubsidy(int64(msgBlock.Header.Height)+1, 1)
```
chappjc: please comment about the 1 voter thing, assuming the idea is to get a per/voter PoS subsidy, rather than a block total

chappjc:
Refreshing the age every second is too fast I think  maybe 5 or 10 seconds

chappjc:
With the current ticket price and price window, I think the estimated net price would go nicely after the window progress text as a third column

blockdata/blockdata.go
RogueElement: Since I can't really tell how many votes would be in the next block(it's usually 5 but sometimes 4), just decided to use 1 but I can use 5 if you want

blockdata/blockdata.go
```diff
@@ -144,8 +144,14 @@ func (t *Collector) CollectBlockInfo(hash *chainhash.Hash) (*apitypes.BlockDataB
 	height := msgBlock.Header.Height
 	block := dcrutil.NewBlock(msgBlock)
 	txLen := len(block.Transactions())
-	coinSupply, _ := t.dcrdChainSvr.GetCoinSupply()
-	nbSubsidy, _ := t.dcrdChainSvr.GetBlockSubsidy(int64(msgBlock.Header.Height)+1, 1)
+	coinSupply, err := t.dcrdChainSvr.GetCoinSupply()
+	if err != nil {
+		log.Error("GetCoinSupply failed")
```
chappjc: With `err` please

blockdata/blockdata.go
```diff
@@ -144,8 +144,14 @@ func (t *Collector) CollectBlockInfo(hash *chainhash.Hash) (*apitypes.BlockDataB
...
+	}
+	nbSubsidy, err := t.dcrdChainSvr.GetBlockSubsidy(int64(msgBlock.Header.Height)+1, 1)
+	if err != nil {
+		log.Errorf("GetBlockSubsidy for %d failed", msgBlock.Header.Height)
```
chappjc: also with err

blockdata/blockdata.go
chappjc: Right, so I think it only makes sense to show the reward for a single voter (and indicating somewhere that it is per vote) or for all 5 which would imply the max possible subsidy.  I like 1 if we label it clearly.

views/root.tmpl
```diff
@@ -313,70 +306,36 @@
...
             <div class="col-md-6">
-                <strong>Next Price Window</strong>
+                <strong>Block Subsidy</strong>
                 <table class="table table-sm">
                     <thead>
-                        <th>Min</th>
-                        <th>Max</th>
-                        <th>Guess</th>
+                        <th>Developer</th>
+                        <th>POS</th>
```
chappjc: PoS (per vote)

views/root.tmpl
```diff
@@ -313,70 +306,36 @@
...
+                        <th>POW</th>
```
chappjc: PoW

views/root.tmpl
```diff
@@ -313,70 +306,36 @@
...
-                        <tr>
-                            <td><span id="mempoolfeeinfo_lowestmineable" >{{printf "%.8f" .MempoolFeeInfo.LowestMineable}}</span></td>
-                            <td><span id="mempoolfeeinfo_min" >{{printf "%.8f" .MempoolFeeInfo.Min}}</span></td>
-                            <td><span id="mempoolfeeinfo_max" >{{printf "%.8f" .MempoolFeeInfo.Max}}</span></td>
-                            <td><span id="mempoolfeeinfo_median" >{{printf "%.8f" .MempoolFeeInfo.Median}}</span></td>
-                            <td><span id="mempoolfeeinfo_mean" >{{printf "%.8f" .MempoolFeeInfo.Mean}}</span></td>
-                        </tr>
-                    </tbody>
-                </table>
+                <div><strong>Coin Supply:</strong> <span id="coin_supply" >{{.BlockSummary.CoinSupply}}</span></div>
```
chappjc: Swap "Coin Supply ..." and "Tickets in mempool..." so it better matches what's above them

PR #164, opened 2017-09-13
Transaction view should say what type of transaction it is

McEdward:


dcrsqlite/apisource.go
```diff
@@ -289,20 +289,41 @@ func (db *wiredDB) GetAllTxOut(txid string) []*apitypes.TxOut {
...
 // each previous outpoint consumed by the transaction.
-func (db *wiredDB) GetRawTransactionWithPrevOutAddresses(txid string) (*apitypes.Tx, [][]string) {
+func (db *wiredDB) GetRawTransactionWithPrevOutAddresses(txid string) (*apitypes.Tx, [][]string, string) {
 	tx := db.GetRawTransaction(txid)
 	prevOutAddresses := make([][]string, len(tx.Vin))
 	if tx == nil {
-		return tx, prevOutAddresses
+		return nil, prevOutAddresses, ""
+	}
+	txhash, err := chainhash.NewHashFromStr(txid)
```
chappjc: I think we already have the txid in tx.TxID

dcrsqlite/apisource.go
```diff
@@ -289,20 +289,41 @@ func (db *wiredDB) GetAllTxOut(txid string) []*apitypes.TxOut {
...
+	if err != nil {
+		log.Errorf("Unknown transaction %s", txid)
+		return nil, prevOutAddresses, ""
+	}
 	for i := range tx.Vin {
 		vin := &tx.Vin[i]
 		prevOutAddresses[i] = txhelpers.OutPointAddressesFromString(
 			vin.Txid, vin.Vout, vin.Tree, db.client, db.params)
 	}
+	var txType string
```
chappjc: since we keep reusing this switch, I think this should also be a function in txhelpers

dcrsqlite/apisource.go
```diff
@@ -289,20 +289,41 @@ func (db *wiredDB) GetAllTxOut(txid string) []*apitypes.TxOut {
...
-		return tx, prevOutAddresses
+		return nil, prevOutAddresses, ""
+	}
+	txhash, err := chainhash.NewHashFromStr(txid)
+	if err != nil {
+		log.Errorf("Invalid transaction hash %s", txid)
+		return nil, prevOutAddresses, ""
 	}
 
+	txR, err := db.client.GetRawTransaction(txhash)
```
chappjc: Would be best to avoid a second call to dcrd.  We can refactor a little to pass the transaction bytes back:

```go
func (db *wiredDB) GetRawTransaction(txid string) *apitypes.Tx {
	tx, _ := db.getRawTransaction(txid)
	return tx
}

func (db *wiredDB) getRawTransaction(txid string) (*apitypes.Tx, string) {
	tx := new(apitypes.Tx)
...
```
then
```go
func (db *wiredDB) GetRawTransactionWithPrevOutAddresses(txid string) (*apitypes.Tx, [][]string, string) {
	tx, txhex := db.getRawTransaction(txid)
...
```

And then we can get the `MsgTx` by decoding the hex string representation of the tx:

```go
	txBytes, err := hex.DecodeString(txhex)
	if err != nil {
		log.Errorf("Unable to read tx %v: %v", txid, err)
		return nil, nil, ""
	}
	msgTx := wire.NewMsgTx()
	if err = msgTx.FromBytes(txBytes); err != nil {
		log.Errorf("Failed to deserialize MsgTx: %v", err)
		return nil, nil, ""
	}
```

This last snippet could be a function in txhelpers.

views/tx.tmpl
```diff
@@ -47,7 +48,7 @@
                         {{else}}
                             <a href="/explorer/block/{{$.Block.BlockHash}}">{{$.Block.BlockHeight}}</a></td>
                         {{end}}
-
+                        <td>{{$.Type}}</td>
```
chappjc: A new column in the summary table is fine for now, but I think we need the tx type to be more conspicuous.  Something like this:

image

chappjc:
I went ahead and merged this since @RogueElement and I implemented a follow-up branch.  Now in PR https://github.com/dcrdata/dcrdata/pull/165

views/tx.tmpl
raedah: yes, will be addressed later

PR #166, opened 2017-09-14
MsgTx from hex and reuse code in txhelpers

chappjc:


chappjc:
@RogueElement I meant that I rebased _this_ branch so if you want to add to it, you'll need to do git pull --rebase

PR #167, opened 2017-09-14
Made transaction type a tag

McEdward:
Added last change requested from #164

### DIFF
--- a/views/tx.tmpl
+++ b/views/tx.tmpl
@@ -9,7 +9,7 @@
     <div class="row">
         <div class="col-md-12">
             <h4>Transaction</h4>
-            <p>{{.TxID}}   <span style="font-size: 12px;"><a href="/api/tx/{{.TxID}}?indent=true">view raw</a></span></p>
+            <p>{{.TxID}}   <span style="font-size: 12px;"><a href="/api/tx/{{.TxID}}?indent=true">view raw</a></span> <span style="background-color: #7FFFD4; border: 1px solid black"><strong>{{$.Type}}</strong></span> </p>
             <table class="table">
                 <thead>
                     <th>Transactions ID</th>
@@ -18,7 +18,6 @@
                     <th>Age</th>
                     <th>Confirmations</th>
                     <th>Included in block</th>
-                    <th>Type</th>
                     <th>Size</th>
                 </thead>
                 <tbody>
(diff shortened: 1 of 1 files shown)


chappjc:
This gives us:

image

@raedah and @gozart1 may want to give feedback on this

RogueElement:
I feel it may be a nice visual cue if each type had a different color and regular transactions had no tag.

views/tx.tmpl
```diff
@@ -138,7 +135,24 @@
...
+            document.getElementById("type-tag").style.backgroundColor = "#a1cfd9";
+        }
+        else if (txtype == "Ticket") {
+            document.getElementById("type-tag").style.backgroundColor = "#7edb71";
+        }
+        else if (txtype == "Revocation") {
+            document.getElementById("type-tag").style.backgroundColor = "#f02431";
+        }
+        else{
+            document.getElementById("type-tag").style.backgroundColor = "#e6a1f2";
```
chappjc: I think we can hide the tag for regular tx.

views/tx.tmpl
```diff
@@ -130,6 +126,7 @@
                                 <div class="hash"><a href="/explorer/address/{{.}}">{{.}}</a></div>
                             {{end}}
                         </td>
+                        <th>{{.ScriptPubKeyDecoded.Type}}</th>
```
chappjc: nice, placement looks good
Although I think non-bold and maybe italics would be good for this data

chappjc: While here, please remove the `{{ if ne .Value 0.0}}` condition since we will want to show all outputs regardless of value

views/tx.tmpl
```diff
@@ -9,7 +9,7 @@
     <div class="row">
         <div class="col-md-12">
             <h4>Transaction</h4>
-            <p>{{.TxID}}   <span style="font-size: 12px;"><a href="/api/tx/{{.TxID}}?indent=true">view raw</a></span> <span style="background-color: #7FFFD4; border: 1px solid black"><strong>{{$.Type}}</strong></span> </p>
+            <p>{{.TxID}}   <span style="font-size: 12px;"><a href="/api/tx/{{.TxID}}?indent=true">view raw</a></span> <span id ="type-tag" style="background-color: #7FFFD4; border: 1px solid black"><strong>{{.Type}}</strong></span> </p>
```
chappjc: we still need to make this box a bit prettier.  There needs more padding or border or whatever on the inside between the text and the border.  Maybe add ` padding: 2px 4px; border-radius: 3px; box-shadow: 1px 1px 1px grey;` to style.

raedah:
There is still more tx details being worked out such as https://github.com/dcrdata/dcrdata/issues/149 , and more information we will be displaying about votes and tickets as well. Deferring this design issue so it can be properly ordered into the queue, and likely dealt with by @gozart1.

PR #172, opened 2017-09-15
2 column root layout

gozart1:
Changes:
- Split home page into 2 columns for large screens.
- Add comma formatting to block height values

before:
before

after ( on big screens ):
screen shot 2017-09-15 at 10 31 12 am

after ( on smaller screens ):
screen shot 2017-09-15 at 10 31 26 am


chappjc:
This is really quite nice, but the footer doesn't stretch all the way now, and it's a little shifted.  It's up on beta.dcrdata.org now.

BTW, I think for really small views, like regular size phones, we should abbreviate Transactions and Revocations as they make the columns too wide and the table doesn't fit on my screen, making a horizontal scroll.  yuk.

gozart1:
Good catch, the .container div was missing a closing tag. Fixed and rebased now.
Agreed we should do something to make headings fit small screens.
Maybe "Txns" for transactions. Not sure how to abbreviate about Revocations though.

gozart1:
just FYI, rebased again to clean up line spacing in the html markup a bit.

chappjc:
Cool.  It's also good to just push lots of commits and fixes, and squash them later.

chappjc:
I should mention that @raedah is somewhat opposed to commas in block height.  Personally I am OK either way.

gozart1:
Ok, sure I'll stop rebasing until ready to merge.
Hmm, I just added another commit to format block heights w/ commas when js updates it so that its consistent.
But looking at the other block explorers out there, none of them format height w/ commas, so I guess there must be a reason. But I'm curious what it is.

gozart1:
I can make those changes now ( shortening based on view size and moving Transactions to 2nd column )

raedah:
Explorer button is not needed anymore. Redundant with 'More Blocks...'

raedah:
I am okay with the commas on the block height. Please add the comma to the heading on the block template too for consistency.

gozart1:
alright, made last round of changes. lmk and ill squash commits

gozart1:
I think for now it makes sense to keep the Explore nav item so that there is an easy/predictable way to get to explorer when viewing a transaction, block, address.

chappjc:
Pretty good.  There's still a minor horizontal scroll on mobile though.  Just barely:

image

Other than dropping the Time column or something creative, I don't know how to fix that.

gozart1:
we could remove the "s"s so its just 
Txn, Ticket, Vote, Revoke

gozart1:
try with the latest commit, reads fine to me w/ out the s

gozart1:
hah, removing commas format in block height would give us another pixel or too :)

gozart1:
Another option to deal w/ mobile is to force the viewport size to a large width.

gozart1:
yet another option would be to move to a fixed top nav, similiar to the footer, you would never see that odd clipping due to table overflowing the container

gozart1:
for now I've added 'table-responsive' to the table, this way the table has its own scroll when it overflows instead of the page itself. I think this is a reasonable solution of the interim while we figure out a more holistic approach to handling small screens.

gozart1:
since the headings reflow on small screen anyways, they could be stacked, eg
```
Vote   Ticket  Revoke
Txn    Txn     Txn
```

chappjc:
current solution is good, I think.  scrolling the table rather than the view is way better

Another think I noticed is that the size column does not get human formatting on new blocks via websocket update

gozart1:
yea, for formatting bytes in JS I think we'll need to pull in https://github.com/HubSpot/humanize or something similar

chappjc:
OK, that can wait.  I'm good to merge.  @raedah this is up at beta.dcrdata.org if you wanna preview quickly.

chappjc:
BTW, no need to squash or rebase manually.  Github can do it:
image

gozart1:
Ok, sg. Merge away!

chappjc:
We should follow up with similar width H resizing stuff on Blocks table at /explorer

Issue #173, opened 2017-09-15
Block Subsidy is wrong on home page


gozart1:
In the block subsidy table introduced in https://github.com/dcrdata/dcrdata/commit/64159cfb90215ddaa3bfd35981259afbf0b6c805
It looks like all the values are getting divided by 5, instead of just the PoS

chappjc:
Ah that's right, setting voters to 1 affects the other components too. That RPC line will have to be for 5 voters, then divide the PoS by 5 (or just leave it depending on the semantics we are after).

gozart1:
As a rule of thumb, could we keep the data being piped to the templates as raw as possible and leave the formatting to the template helper functions?

For example, CoinSupply is a string at the moment, it would be better if it could be some sort of numeric so we can run it through a humanize comma function.

gozart1:
I think it would be fine to simply divide by 5 in the view layer for PoS per Ticket, since in a sense its just another way to look at the PoS subsidy.

chappjc:
Closed by PR 175

PR #175, opened 2017-09-15
Corrected block subsidy on home page

RogueElement:
Should address #173

### DIFF
--- a/views/root.tmpl
+++ b/views/root.tmpl
@@ -85,6 +85,9 @@
             };
         }
 
+        var subsidyToString = function(x, y = 1) {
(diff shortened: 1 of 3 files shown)


views/root.tmpl
```diff
@@ -85,6 +85,9 @@
             };
         }
 
+        var subsidyToString = function(x, y = 1) {
```
chappjc: note that back in golang, you can use `nbSubsidy.Developer.ToCoin()` to get the usual floating point amount in DCR.  But this is fine.  No need to change it.

PR #176, opened 2017-09-15
Add a block data cache, APICache.

chappjc:
This is broken out of PR https://github.com/dcrdata/dcrdata/pull/63.

`APICache` implements `BlockSummarySaver`, which takes a `*BlockDataBasic`
and caches it. A block priority queue, `BlockPriorityQueue`, is maintained
by `APICache` to decide which blocks to evict if needed. The "LessFn" can
be set to an arbitrary comparator of two `CachedBlock` values. There are
methods to remove and retrieve blocks from the cache.

### DIFF
--- a/dcrdataapi/apicache.go
+++ b/dcrdataapi/apicache.go
@@ -59,6 +59,16 @@ type APICache struct {
 }
 
 // NewAPICache creates an APICache with the specified capacity.
+//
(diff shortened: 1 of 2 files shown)


chappjc:
Note that the added code here is completely unused presently.  Merging so is can be used for multiple purposes.

PR #177, opened 2017-09-16
Plug APICache into dcrsqlite.wiredDB (TESTING)

chappjc:
Plug the new APICache into dcrsqlite.wireDB to have a block data cache.

Add wiredDB.GetHash, which uses dcrsqlite.RetrieveBlockHash.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 *.log.*.gz
 *.bak
 debug
+debug.test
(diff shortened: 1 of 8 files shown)


chappjc:
this will probably be obsolete and overly complicated once db v2 is done.

chappjc:
yup, looking a bit obsolete now

PR #179, opened 2017-09-17
add custom humanize.js, humanize time & bytes when block is added via js

gozart1:
Ended up hand-rolling the humanize.bytes in javascript since https://github.com/HubSpot/humanize makes different choices than https://github.com/dustin/go-humanize with regard to bytes formatting

@RogueElement do I have your blessings to move `timeSince` into humanize.js as well?

### DIFF
--- a/public/js/humanize.js
+++ b/public/js/humanize.js
@@ -1,4 +1,4 @@
-// this is a handrolled humanize lib designed to mirror the go-humanize package used by backend
+// For all your client side value formatting needs...
 var humanize = (function() {
   function logn(n, b) {
     return Math.log(n) / Math.log(b)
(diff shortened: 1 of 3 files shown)


RogueElement:
Sure. Could you also move float64Commaf to the js as well? Or at least fix it because the total values tend to have variable decimal places.

gozart1:
Do you mean hiding trailing 0's?
That seems like a feature to me. 
decimals

chappjc:
I think GDAX figured out the best compromise, although I expect implementing it would be challenging:

image

gozart1:
Yea, that is a nice touch!
I've been wanting to play around with styling split on the decimal anyways, might just give this a shot.

chappjc:
So should this go in now, or did it need to wait for PR https://github.com/dcrdata/dcrdata/pull/185 ?  Would be good to get this in if it is OK on its own.

gozart1:
Lets merge this one, and I'll rebase 185 after this is in.

PR #181, opened 2017-09-18
Layout tweaks

gozart1:
- Explorer
Align styling and column order of explorer table with that of 6 blocks table in root.tmpl.
Fix prev/next links ( current logic is thrown off by commas in block heights )

- Transaction
Add Links to "In" addresses 
Splits address onto 2 lines on small screens

- Sticky Top Nav
Made the top nav sticky. The main reason for this to allow for a more flexible layout on small screens where the tables often overflow. Instead of making all the tables scroll within themselves when they don't fit. We can just let them expand the width of the page as needed this way.

### DIFF
--- a/views/explorer.tmpl
+++ b/views/explorer.tmpl
@@ -22,14 +22,20 @@
 
         <div class="row">
             <div class="col-md-12">
-                <table class="table striped">
+                <table class="table striped table-responsive">
                     <thead>
                         <tr>
                             <th>Height</th>
-                            <th>Votes</th>
-                            <th>Tickets</th>
-                            <th>Transactions</th>
-                            <th>Revocations</th>
+                            <th>
+                                <span class="d-none d-md-inline">Transactions</span>
+                                <span class="d-md-none">Txn</span>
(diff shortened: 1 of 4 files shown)


views/explorer.tmpl
```diff
@@ -22,14 +22,20 @@
...
-                            <th>Revocations</th>
+                            <th>
+                                <span class="d-none d-md-inline">Transactions</span>
+                                <span class="d-md-none">Txn</span>
+                            </th>
+                            <th>Vote<span class="d-none d-md-inline">s</span></th>
+                            <th>Ticket<span class="d-none d-md-inline">s</span></th>
+                            <th>
+                                <span class="d-none d-md-inline">Revocations</span>
+                                <span class="d-md-none">Revoke</span>
```
chappjc: I tried just "Rev" and it gets the whole table on the phone without scrolling.  It's not a bad abbreviation considering we have Txns for Transactions.  :)

gozart1:
agreed, on the In/Out I'll rework that.

PR #182, opened 2017-09-18
Address cache

Henrychuks25:
why not use a specific range for the comparison blocks rather than this dangling nature?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 *.log.*.gz
 *.bak
 debug
+debug.test
 *.cert
 *.key
 dcrdata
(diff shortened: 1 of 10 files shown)


RogueElement:
What exactly are you referring to here?

Henrychuks25:
When the codes were created, the range was set to > 2000 for instance so my suggestion is that we could use >2000 || <3000

chappjc:
For discussion, you can comment on specific lines in existing pull requests.  There is already a PR for the block cache that should include the lines of code under discussion here.  I've not created a PR for this branch yet because it's still a WIP (work in progress).

PR #185, opened 2017-09-19
fade out trailing zeroes & use DroidSansMono for various monospace values

gozart1:
This is a prototype of the faded trailing zeroes effect suggested by @chappjc in https://github.com/dcrdata/dcrdata/pull/179

screen shot 2017-09-19 at 12 11 37 am

While working on this I noticed some dcr totals have 15 decimal places. Eg. https://beta.dcrdata.org/explorer/block/00000000000000937548da9c7b809b200ff7782c731ee33a7bea372e3d37dd9e

I assume this is a bug? Maybe a type mismatch or flouting point math issue? 

As a temporary fix, the `styleDCRVal` formatter function simply discards values after 8 decimals. This way, via monospace font and right aligned text, the decimals points line up vertically between the table rows as suggested by @raedah  in https://github.com/dcrdata/dcrdata/issues/118 

Things that could be improved
- Implement formatting server side
- Figure out 15 decimals points issue
- Get a nice monospace font


chappjc:
Damn that looks good (the decimal fading).  This will be the shit when a monospace font is put in place.

Regarding the 15 decimal floating point math issue, I haven't looked at what's going on in JavaScript or templates for that, but if we have to use float math, it should be double precision.  But when possible in golang, the `dcrutil.Amount` type is `int64` so these issues are largely avoided.

chappjc:
I bet the split formatting would be good for distinguishing the hash from the index portion of the outpoint (hash:index)

gozart1:
Updated this to use DroidSansMono.
I think Inconsolata may be a brand font for Decred (needs confirmation). 
But I don't like its 0's with a strikethrough. Of the open source fonts monospace fonts available https://fonts.google.com/?category=Monospace DroidSansMono is pretty nice.

I also changed the default font to use Source Sans Pro which is what decred.org uses

chappjc:
But without a slash or dot in the zero, it looks too much like a capital O to me.  Inconsolata and Roboto are not bad with their slash, and Source Code Pro has a dot.

chappjc:
It still looks good with Droid though.  We can try this out.

gozart1:
Alright, switched to inconsolata for monospace. Slashes are growing on me :) 
inconsolata is a really compact font so this last commit also has further spacing and font size tweaks.

gozart1:
I think for the home page especially we need to figure out something else for the layout, but for now I've been tweaking the headings to look decent in the rest of the explorer. I think it can be improved further but lets make it a separate PR.

chappjc:
Yeah, this looks great.  Address pages load slower now on account of the browser working harder, but server-side formatting can perhaps help with this.

BTW, I think we'll need to work some javascript magic to get "More..." buttons on the explorer page's block list and other lists, but most of that will require v2 of the database so we can query instead of using inefficient RPCs.

PR #186, opened 2017-09-19
Show vote info and choices on Tx explorer page

chappjc:
- Create Vote Info table that displays on `/explorer/tx/...` for a ssgen.

  image

- Call `APIDataSource.GetVoteInfo` from `explorerUI.txPage` and execute the template with the vote info included in the existing template data struct.
- Add vote bits to `dcrdataapi.VoteInfo`.
- `txhelpers.SSGenVoteChoices` now returns vote bits too.
- `GetRawTransactionWithPrevOutAddresses` returns `*wire.MsgTx` as new 4th output argument.

Addresses Issue https://github.com/dcrdata/dcrdata/issues/141, but we should consider improved styling.

### DIFF
--- a/apiroutes.go
+++ b/apiroutes.go
@@ -10,6 +10,7 @@ import (
 
 	apitypes "github.com/dcrdata/dcrdata/dcrdataapi"
 	"github.com/decred/dcrd/dcrjson"
+	"github.com/decred/dcrd/wire"
 	"github.com/decred/dcrrpcclient"
 )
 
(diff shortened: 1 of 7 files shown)


gozart1:
nice! I'll see what I can do to make to match a bit more with https://voting.decred.org/ style

chappjc:
I like how that groups the Issue and Choice elements.

image

chappjc:
@RogueElement Good question.  It might get cramped on the block pages's vote table, but we could find a way to compress the information or at least show the vote bits, or maybe just the block validation part of the vote bits.

I think this is good for now though.  Would you mind raising an issue for that suggestion though?

chappjc:
normally I would use Squash and merge for more than 3 commits, but since there are multiple commiters, I'm going to rebase so everyone gets credit where credit's due.

Issue #138, opened 2017-09-08
Show transaction fees


raedah:
Transaction view should show Fee and Fee/size
Lets try and fit this on the transaction rows partial as well (and stake tx since they also have fees.)

For blocks, calculate mining fees collected in block (spent - sent).

raedah:
I am seeing bad math on the fee calc, example 1280fabbc22201862d0d17ce41ba2429ca849a1851703da6d871d62e8e26d1b6

raedah:
Looks like transaction fee is showing N/A on unconfirmed transactions, but we already know the fee before it gets confirmed.

chappjc:
The unconfirmed issue goes back to this: https://github.com/dcrdata/dcrdata/issues/142

Regarding votes and fees, I'm pretty sure votes don't pay fees.

chappjc:
closed by https://github.com/dcrdata/dcrdata/pull/202

PR #188, opened 2017-09-19
Tx fees

RogueElement:
Added tx fees to tx page. Should partly address #138 but I add more commits to finish up

### DIFF
--- a/dcrsqlite/apisource.go
+++ b/dcrsqlite/apisource.go
@@ -324,19 +324,16 @@ func (db *wiredDB) GetExplorerTxData(txid string) *apitypes.ExplorerTxData {
 	}
 
 	txType := txhelpers.DetermineTxTypeString(msgTx)
-	var txFee, txFeeRate float64
-	if txType == "Ticket" || (txType == "Regular" && len(tx.Vin[0].Coinbase) < 5) {
-		txFee = txhelpers.TxFee(msgTx)
-		txFeeRate = txhelpers.TxFeeRate(msgTx)
-	} else {
-		txFee = 0.0
-		txFeeRate = 0.0
+	var txFee, txFeeRate dcrutil.Amount
+	if !(txType == "Vote" || (txType == "Regular" && len(tx.Vin[0].Coinbase) > 4)) {
+		txFee, txFeeRate = txhelpers.TxFeeRate(msgTx)
 	}
 
-	vinfo := new(apitypes.VoteInfo)
+	var vinfo *apitypes.VoteInfo
 	if stake.DetermineTxType(msgTx) == stake.TxTypeSSGen {
 		validation, version, bits, choices, err := txhelpers.SSGenVoteChoices(msgTx, db.params)
 		if err != nil {
+			log.Debugf("Cannot get vote choices for %s", txid)
 			return &apitypes.ExplorerTxData{
 				Tx:       tx,
 				VinAddrs: prevOutAddresses,
@@ -355,8 +352,6 @@ func (db *wiredDB) GetExplorerTxData(txid string) *apitypes.ExplorerTxData {
 			Bits:    bits,
 			Choices: choices,
 		}
-	} else {
-		vinfo = nil
 	}
 
 	return &apitypes.ExplorerTxData{
--- a/dcrsqlite/apisource.go
+++ b/dcrsqlite/apisource.go
@@ -19,7 +19,6 @@ import (
 	"github.com/decred/dcrd/chaincfg/chainhash"
 	"github.com/decred/dcrd/dcrjson"
 	"github.com/decred/dcrd/txscript"
-	"github.com/decred/dcrd/wire"
 	"github.com/decred/dcrrpcclient"
 	"github.com/decred/dcrutil"
 )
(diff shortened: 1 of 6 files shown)


dcrsqlite/apisource.go
```diff
@@ -320,12 +317,83 @@ func (db *wiredDB) GetRawTransactionWithPrevOutAddresses(txid string) (*apitypes
...
-		return tx, prevOutAddresses, "", nil
+		return &apitypes.ExplorerTxData{
+			Tx:       tx,
+			VinAddrs: prevOutAddresses,
+		}
 	}
 
 	txType := txhelpers.DetermineTxTypeString(msgTx)
+	var txFee, txFeeRate float64
+	if txType == "Ticket" || (txType == "Regular" && len(tx.Vin[0].Coinbase) < 5) {
```
chappjc: Revocations too have a fee.  You might be better off reversing this and checking for votes and regular coinbase.

But also i don't follow the `len(tx.Vin[0].Coinbase) < 5` part.  What is 5?

txhelpers/txhelpers.go
```diff
@@ -584,3 +584,29 @@ func DetermineTxTypeString(msgTx *wire.MsgTx) string {
...
+	}
+	var amtOut int64
+	for iv := range msgTx.TxOut {
+		amtOut += msgTx.TxOut[iv].Value
+	}
+	return dcrutil.Amount(amtIn - amtOut).ToCoin()
+}
+
+// TxFeeRates computes and returns the fee rate in DCR/KB for a given tx
+func TxFeeRate(msgTx *wire.MsgTx) float64 {
```
chappjc: This could have a second output to give absolute fee too.  This would avoid the need to call TxFee after this.

dcrsqlite/apisource.go
```diff
@@ -320,12 +317,83 @@ func (db *wiredDB) GetRawTransactionWithPrevOutAddresses(txid string) (*apitypes
...
+		txFee = txhelpers.TxFee(msgTx)
+		txFeeRate = txhelpers.TxFeeRate(msgTx)
+	} else {
+		txFee = 0.0
+		txFeeRate = 0.0
+	}
+
+	vinfo := new(apitypes.VoteInfo)
```
chappjc: no need to do an actual alloc here since it's done below.  can just `var vinfo *apitypes.VoteInfo`

dcrsqlite/apisource.go
```diff
@@ -320,12 +317,83 @@ func (db *wiredDB) GetRawTransactionWithPrevOutAddresses(txid string) (*apitypes
...
+	if stake.DetermineTxType(msgTx) == stake.TxTypeSSGen {
+		validation, version, bits, choices, err := txhelpers.SSGenVoteChoices(msgTx, db.params)
+		if err != nil {
+			return &apitypes.ExplorerTxData{
```
chappjc: Since this is unexpected (vote with invalid vote bits, perhaps we should log it as Debug)

dcrsqlite/apisource.go
```diff
@@ -320,12 +317,83 @@ func (db *wiredDB) GetRawTransactionWithPrevOutAddresses(txid string) (*apitypes
...
+			Validation: apitypes.BlockValidation{
+				Hash:     validation.Hash.String(),
+				Height:   validation.Height,
+				Validity: validation.Validity,
+			},
+			Version: version,
+			Bits:    bits,
+			Choices: choices,
+		}
+	} else {
```
chappjc: Can get rid of this if `vinfo` defn. is changed to avoid alloc

dcrsqlite/apisource.go
```diff
@@ -320,12 +317,83 @@ func (db *wiredDB) GetRawTransactionWithPrevOutAddresses(txid string) (*apitypes
...
+	tx, _ := db.getRawTransaction(txid)
+	if tx == nil {
+		return nil, nil
+	}
+
+	prevOutAddresses := make([][]string, len(tx.Vin))
+
+	for i := range tx.Vin {
+		vin := &tx.Vin[i]
+		if vin.IsCoinBase() /* || vin.IsStakeBase() */ {
```
chappjc: FYI, I left this comment because there is a PR for dcrd to introduce changes to `Vin` that will include a Stakebase field and similar method to check it.

views/tx.tmpl
```diff
@@ -18,7 +18,11 @@
                     <th>Confirmations</th>
                     <th>Included in block</th>
                     <th>Type</th>
-                    <th>Size</th>
+                    {{if .FeeRate}}
+                    <th>Fee</th>
+                    <th>Fee Rate</th>
+                    {{end}}
+                    <th>Size {{(index .Vin 0).Coinbase}}</th>
```
chappjc: I don't follow this.  Can you explain what this should show?

dcrsqlite/apisource.go
RogueElement: Used it because when there is no coinbase the coinbase string consists of 4 spaces

dcrsqlite/apisource.go
RogueElement: Noted

views/tx.tmpl
RogueElement: That should not be there

dcrsqlite/apisource.go
chappjc: oh, that's interesting.  Is that set somewhere explicitly in dcrd?  we should check that out

RogueElement: I don't know if it's set somewhere in dcrd, just an observation I made while debugging

views/tx.tmpl
```diff
@@ -52,6 +56,10 @@
                             <a href="/explorer/block/{{$.Block.BlockHash}}">{{$.Block.BlockHeight | int64Comma}}</a></td>
                         {{end}}
                         <td>{{$.Type}}</td>
+                        {{if $.FeeRate}}
+                        <td>{{$.Fee}}</td>
+                        <td>{{$.FeeRate}}</td>
```
chappjc: just noticed that this prints as x DCR, meaning we need the "/kB" here too

Issue #171, opened 2017-09-14
Transaction sort order in Block View


raedah:
Transaction in the block view have the same date, date of block mined.
So the best way I see to do it would be sort by Total Value (largest value first).

PR #191, opened 2017-09-20
Sorting on the block page

McEdward:
Resolving issue #171

### DIFF
--- a/explorer.go
+++ b/explorer.go
@@ -32,23 +32,11 @@ const (
 	addressRows     = 2000
 )
 
-type ByValue []dcrjson.TxRawResult
-
-func (s ByValue) Len() int {
-	return len(s)
-}
-func (s ByValue) Swap(i, j int) {
-	s[i], s[j] = s[j], s[i]
-}
-func (s ByValue) Less(i, j int) bool {
-	srt := func(vout []dcrjson.Vout) float64 {
-		total := 0.0
-		for _, v := range vout {
-			total = total + v.Value
-		}
-		return total
+func voutTotal(vouts []dcrjson.Vout) (total float64) {
+	for i := range vouts {
+		total += vouts[i].Value
 	}
-	return srt(s[i].Vout) < srt(s[j].Vout)
+	return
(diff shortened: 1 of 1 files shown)


explorer.go
```diff
@@ -31,6 +32,25 @@ const (
 	addressRows     = 2000
 )
 
+type ByValue []dcrjson.TxRawResult
```
chappjc: In the event that we want other things in this package sorted by value, this naming will turn out to be too general.  Just an unfortunately consequence of go's lack of generics.  Anyway, if implementing `sort.Interface` is the chosen solution, I would go with something like `TxRawByValue`.

However, since `Len` and `Swap` are completely boilerplate here, this is a prime case to utilize the new [`sort.Slice`](https://golang.org/pkg/sort/#Slice) function.  For example:

```go
func voutTotal(vouts []dcrjson.Vout) (total float64) {
	for i := range vouts {
		total += vouts[i].Value
	}
	return
}

sort.Slice(data.RawTx, func(i, j int) bool {
	return voutTotal(data.RawTx[i].Vout) > voutTotal(data.RawTx[j].Vout)
})
```

But I leave that up to you.

explorer.go
```diff
@@ -109,10 +97,18 @@ func (exp *explorerUI) blockPage(w http.ResponseWriter, r *http.Request) {
...
+	})
+	sort.Slice(data.Tickets, func(i, j int) bool {
+		return voutTotal(data.Tickets[i].Vout) > voutTotal(data.Tickets[j].Vout)
+	})
+	sort.Slice(data.Votes, func(i, j int) bool {
+		return voutTotal(data.Votes[i].Vout) > voutTotal(data.Votes[j].Vout)
+	})
+	sort.Slice(data.Revs, func(i, j int) bool {
+		return voutTotal(data.Revs[i].Vout) > voutTotal(data.Revs[j].Vout)
+	})
```
chappjc: hah, it ended up bloating this section, but I still like this better.

Issue #118, opened 2017-09-06
Template enhancement roundup


raedah:
**Below items completed by @McEdward in https://github.com/dcrdata/dcrdata/pull/194**

Address template is showing invalid date Jan 1 00:00:00 1970 on unconfirmed transactions.

Transaction view with a Vote Reward should say Created for the block instead of N/A.

In Address on transaction template should be a link. **previously done**

On block and transaction templates, move Age field onto the same line as date.
Time (UTC) | 9/14/17 21:09:19 (4m 13s ago)

**Completed by @gozart1 in https://github.com/dcrdata/dcrdata/pull/185**

On the Block page, magic so that Total Value column so that the decimal points are always stacked ... and the 'DCR'. Same thing for the Size column with the B. Being right justified with other items, but left justified within the column.

**Below items are completed by @gozart1 in https://github.com/dcrdata/dcrdata/pull/150**

Need restyling on new search not found page.

Reduce the amount of surrounding white space on the tables so that more rows are shown. Depending on how much more data can be reasonably shown, default row count may be adjusted.

Pagination on blocks page, white space surrounding Older and Newer should be fixed, and font size needs to be worked with. Also, the Newer button is initially hidden, and the rest of the page shifts down when it appears. The page should not shift, so we should find a way to leave enough space at the top or can include the text 'Head of Chain' or 'Top' which is then replaced by the Newer link.

**Below items are completed by @gozart1 in https://github.com/dcrdata/dcrdata/pull/128**

Change default number of blocks listed on Blocks page to 12. (its one hour)

Reorder header
From:  Home   GitHub  API   Status  Explore
To:  Home  Explore  Status  API  Github
(or move items to footer)

On the Blocks and Address pages, from the Time column remove the repetitive +0000 UTC from each item, and change the column title to Time (UTC)

On the blocks page
Instead of Voters | Fresh Stake | Transactions
Use Votes, Tickets, Revocations, Transactions

Change pagination buttons from Previous Blocks Next Blocks, to  Older  Newer

Currently all html titles for pages are 'dcrdata web'
I suggest the title should say 
'Decred Blocks'
'Decred Transaction cc50e968f87476342e9...'   (truncated to 20 chars)
'Decred Address DsarUwfqxgrTryZbTDegCQ6ffm5CybmtSBv'
'Decred Block 166274'

Front page text can be simplified.
"Number of Tickets in Mempool (queued for mining):" -> "Tickets in Mempool:"
[...]

PR #194, opened 2017-09-21
template enhancement

McEdward:
Resolving issue #118.
I feel like some of the changes requested have already be done. Please explain in more detail if i overlooked anything. Thank you

### DIFF
--- a/explorer.go
+++ b/explorer.go
@@ -29,7 +29,7 @@ const (
 	addressTemplateIndex
 	maxExplorerRows = 2000
 	minExplorerRows = 20
-	addressRows     = 700
+	addressRows     = 500
 )
 
 func voutTotal(vouts []dcrjson.Vout) (total float64) {
--- a/explorer.go
+++ b/explorer.go
@@ -29,7 +29,7 @@ const (
 	addressTemplateIndex
 	maxExplorerRows = 2000
 	minExplorerRows = 20
-	addressRows     = 200
+	addressRows     = 700
 )
 
 func voutTotal(vouts []dcrjson.Vout) (total float64) {
--- a/explorer.go
+++ b/explorer.go
@@ -29,7 +29,7 @@ const (
 	addressTemplateIndex
 	maxExplorerRows = 2000
 	minExplorerRows = 20
-	addressRows     = 2000
+	addressRows     = 200
 )
 
 func voutTotal(vouts []dcrjson.Vout) (total float64) {
--- a/views/tx.tmpl
+++ b/views/tx.tmpl
@@ -91,7 +91,7 @@
                         {{else if eq $vin.Txid "0000000000000000000000000000000000000000000000000000000000000000"}}
                             Created
                         {{else if eq $vin.BlockHeight 0}}
-                            Created
+                            Pending
                         {{else}}
                             <a class="mono fs13" href="/explorer/block/{{$.Block.BlockHash}}">{{$vin.BlockHeight | uint32Comma}}</a></td>
                         {{end}}
--- a/views/tx.tmpl
+++ b/views/tx.tmpl
@@ -93,7 +93,7 @@
                         <!-- <td>{{$vin.Tree}}</td> -->
                         <td>
                         {{if eq $vin.BlockHeight 0}}
-                            N/A
+                            Created for the block
                         {{else}}
                             <a class="mono fs13" href="/explorer/block/{{$.Block.BlockHash}}">{{$vin.BlockHeight | uint32Comma}}</a></td>
                         {{end}}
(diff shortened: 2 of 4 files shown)


views/tx.tmpl
```diff
@@ -93,7 +93,7 @@
                         <!-- <td>{{$vin.Tree}}</td> -->
                         <td>
                         {{if eq $vin.BlockHeight 0}}
-                            N/A
```
chappjc: This is generally referred to as either "unconfirmed" or "in mempool".

@raedah and @gozart1 Any preferences here?

ruemic: unconfirmed sounds good

raedah: unconfirmed

chappjc: unconfirmed sounds good to me

raedah: Actually the 0 in this case is referring to a Vote Reward where the coins are created for the first time and have no input block height. So I had suggested 'Created'. Could also be mined or coinbase.

chappjc: oh, I see this is the "In" table.  My bad.  Yes, since the coins are generate this is input does not actually have a previous outpoint at all.  However, the "Previous Outpoint" column says something like "Coinbase: 00002f646372642f".  For example:  https://beta.dcrdata.org/explorer/tx/941505dd59ff29bebfa986c5192fcf93e86a10732df26e32643809345eae3ddd

So showing N/A does seem appropriate for the "Block" column, although blank would be good too.  Not sure created makes sense to indicate the origin block for the coinbase input.

raedah: Generated coins do not have a origin block. There is no source height because the coins did not previously exist. N/A suggests that its not available for some reason, like not yet being confirmed, but its actually not even logical. Leaving it blank just looks like a bug where data is missing. Both of those a liable to confuse and mislead the user, and neither of those are an actual block height either. If Created or Generated is not clear enough, more verbose would be (Newly Generated Coins). In the same way that Vote Reward is not a previous outpoint, Created is also not a block. Instead we are accurately informing the user of what is occurring.

raedah: N/A suggests that its not available for some reason, like not yet being confirmed, but its actually not even logical. Some users may not know what a Coinbase or Vote Reward is and could use additional description. In the same way that a Coinbase or Vote Reward is not a previous outpoint, Created is also not a block. The Block column is just the right size to put the word Created which informs the user of what is occurring, and can be used for both Coinbase and Vote Reward special cases.

McEdward: 2000 for addressRows was a bit much, so I finally decided to cut it down to 700 to save a bit of load time.

chappjc: just to reduce review iteration time, a screenshot is sometimes helpful  :)

chappjc: So this is still not right because for regular (non-stake) unconfirmed transactions, it will show "Created" in the Block column for the inputs.  There is an issue with having no input values for unconfirmed transactions, which we do need to solve, but in the meantime this check should not be `{{if eq $vin.BlockHeight 0}}` but the same logic that was used for the Previous Outpoint column:

```
{{if $vin.Coinbase}}
    Coinbase: {{ $vin.Coinbase }}
{{else if eq $vin.Txid "0000000000000000000000000000000000000000000000000000000000000000"}}
    Vote Reward
{{else}}
...
```

except we'll need to OR the first two conditions

views/tx.tmpl
```diff
@@ -92,7 +86,11 @@
                         </div></td>
                         <!-- <td>{{$vin.Tree}}</td> -->
                         <td>
-                        {{if eq $vin.BlockHeight 0}}
+                        {{if $vin.Coinbase}}
+                            Created
+                        {{else if eq $vin.Txid "0000000000000000000000000000000000000000000000000000000000000000"}}
+                            Created
+                        {{else if eq $vin.BlockHeight 0}}
```
raedah: since the above two are the coinbase and votereward cases, what is this third case for?

explorer.go
```diff
@@ -29,7 +29,7 @@ const (
 	addressTemplateIndex
 	maxExplorerRows = 2000
 	minExplorerRows = 20
-	addressRows     = 2000
+	addressRows     = 700
```
raedah: 500 so its a round number

views/tx.tmpl
chappjc: unconfirmed regular transaction

chappjc: oh, but it shouldn't say created for the third case.  For that it should say N/A since it's just not there, at least not yet since the miner fills this in.  We should do it though in the near future, just not possible here.

chappjc: or maybe "pending" until we extract it manually

McEdward: Alright, I'm on it

chappjc:
Merged, but @McEdward note that your branch forked from master quite a ways back.  To be sure you are testing the same thing that will be merged into master, you should rebase on to master:

```
git checkout master
git pull
git checkout template-enhancement
git rebase master
```

PR #203, opened 2017-09-27
layout rework wip

gozart1:
This PR is a work in progress for https://github.com/dcrdata/dcrdata/issues/200.

I'll keep refining this, and also planning to rebase and work on top of https://github.com/dcrdata/dcrdata/pull/202 once that merges to master.

### DIFF
--- a/public/css/main.css
+++ b/public/css/main.css
@@ -321,4 +321,7 @@ body.darkBG .navbar-fixed-bottom a {
   .xs-w91 {
     width: 91px;
   }
+  .xs-w117 {
+    width: 117px;
+  }
 }
--- a/public/css/main.css
+++ b/public/css/main.css
@@ -251,9 +251,6 @@ body.darkBG .navbar-fixed-bottom a {
   font-family: 'inconsolata-v15-latin-regular';
   font-size: 14px;
 }
-.table .height {
-  font-size: 18px;
-}
 
 /*content*/
 .dcr-total .dcr {
--- a/public/css/main.css
+++ b/public/css/main.css
@@ -272,23 +272,6 @@ body.darkBG .navbar-fixed-bottom a {
   word-wrap: break-word;
   transition: .133s transform;
 }
-.hash.collapse {
-  font-size: 12px;
-  display: inline-block;
-  width: 197px;
-}
-
-.outpoint {
-  font-family: 'inconsolata-v15-latin-regular', monospace;
-  display: inline-block;
-  word-wrap: break-word;
-  transition: .133s transform;
-}
-.outpoint.collapse {
-  font-size: 12px;
-  display: inline-block;
-  width: 206px;
-}
 
 /*resemble voting.decred.org*/
 .highlight-text {
@@ -325,15 +308,7 @@ body.darkBG .navbar-fixed-bottom a {
     max-width: 100% !important;
   }
 }
-@media only screen and (max-width: 940px)  {
-  .hash.collapse {
-    width: 142px;
-  }
-}
 @media only screen and (max-width: 768px)  {
-  .hash.collapse {
-    width: 197px;
-  }
   .table {
     margin-bottom: 0;
     width: 100%;
@@ -342,13 +317,6 @@ body.darkBG .navbar-fixed-bottom a {
     width: 151px;
   }
 }
-@media only screen and (max-width: 630px)  {
-  .table .hash {
-    font-size: 12px;
-    width: 197px;
-  }
-}
-
 @media only screen and (max-width: 576px)  {
   .xs-w91 {
     width: 91px;
--- a/public/css/main.css
+++ b/public/css/main.css
@@ -128,6 +128,9 @@ body.darkBG .navbar-fixed-bottom a {
 .fs18 {
   font-size: 18px !important;
 }
+.fs24 {
+  font-size: 24px !important;
+}
 .fs28 {
   font-size: 28px !important;
 }
@@ -152,6 +155,9 @@ body.darkBG .navbar-fixed-bottom a {
 }
 .w80 {
   width: 80px;
+}
+.w142 {
+  width: 142px;
 }
 .vam {
   vertical-align: middle;
--- a/public/css/main.css
+++ b/public/css/main.css
@@ -99,6 +99,10 @@ body.darkBG .navbar-fixed-bottom a {
 }
 
 /*utils*/
+.anchor {
+  position: relative;
+  top: -33px;
+}
 .hidden {
   display: none;
 }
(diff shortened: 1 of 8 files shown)


gozart1:
I've rebased this and ready for some review.
This still needs work to look good in Firefox ( tables will overlap in various screen width scenarios ) Planning to take care of that next.

chappjc:
Nice.  I'll have a look today!

gozart1:
Firefox layout issue now fixed.

chappjc:
Just for discussion:
image

My first thought was that things run together a bit, in that the lines of text perhaps lacked some visual cues to make each row more clear.

The dev/pos/pow reward are components of the block reward, so it would be nice to somehow show this relationship, visually.

If I were to order the information:

1. Ticket price
2. ticket window progress
3. tickets in mempool
4. rewards
5. total suply
6. pow diff

clearly this is coming from a staking point of view, but regardless of the viewer, the last two items are hard to use in a meaningful way.  Well, they are quite interesting on a chart, but as a single number, not so much.

chappjc:
The new block view layout:

image

This is a big improvement I think.  Before everything was small and it was hard to know where to look.

I'm glad the block and stake version are prominently featured as they are important for the upgrade progress.

Perhaps final state and bits can be omitted?  Bits is another representation of pow difficulty, so it's actually redundant with the decimal pow difficulty we are showing. And final state, well, I don't know what to make of it.  :)  I think it's good to nave nonce and the roots though.

This might be beyond the scope of this PR, but it would be cool if when you clicked on "transactions" or "votes", etc. in the block summary it would pulse the section header that it jumps you to.  The scroll is rarely long enough that you get the section at the top of the page, so this would be pretty useful I think.

"TOTAL SENT" should have DCR units, I think.

Some reordering of the right side:

1. Time
2. size
3. transactions
4. votes
5. revocations
6. tickets
7. ticket price
8. pool size
9. pow diff

When showing the best block, instead of hiding NEXT BLOCK, perhaps it would be cool to show "TIP" where the hash would be if the block were not at the tip of the chain.

I like the transaction page, except that confirmations is gone and I really think we need to show that.  maybe just next to INCLUDED IN BLOCK.

gozart1:
I've incorporated your feedback, except for these items 
- spicing up the reward info in root.tmpl
- anchor link animation ( to orient user when page is too short to scroll )
- removing final state ( not sure what to make of it either lol )

gozart1:
The main goal of this PR was to explore a way to break out of using the same table layout style for all the data.

I think for the home page especially, we should think of what content we actually want there, implement that and focus on a good layout for it in a separate PR.

raedah:
One thing I would add is to not have transactions, votes, revocations, tickets at the top at all. Just incorporate the count into each section heading below.

gozart1:
@raedah Firefox fix is in. Do you see any remaining issues with Firefox?

gozart1:
I like having the counts because it provides a nice summary and on mobile it doubles as a navigational aid.

But maybe that's not worth the business it adds. I'd suggest we keep it around for now and review canning or moving the counts when we inevitably will find something more useful to take up that screen real estate.

raedah:
ok

gozart1:
@chappjc for the rewards in root.tmpl, maybe we can create a component similar to this https://v4-alpha.getbootstrap.com/components/progress/#multiple-bars

The one issue I ran into when playing around w/ that style in https://dcr-data.netlify.com/#/ was that the labeling is tricky to do in a space efficient and readable at the same time.

chappjc:
Yeah, I love those multiple bars in a single progress bar.

gozart1:
Here's a sketch of how that could work.
Never mind the colors and spacing, see the label design challenge?

Hmm now that I think about it, maybe we could do "hamburger style" for mobile. ( making a vertical stack would allow for dev fund to have room for label ). 

But in any case, it's tough to find a configuration that isn't wasting space in some way.

subsidy_bar
screen shot 2017-10-01 at 12 24 14 am

raedah:
I agree, it doesnt use space effectively, becomes illegible, and I also dont find it particularly useful since its a static ratio that does not change over time. To inform the users, would be easier just to put (70/30/10) next to it. We also have a whole lot of data that is not yet available on the front page (Other data to be added at https://github.com/dcrdata/dcrdata/issues/208 ), so I wouldnt do anything too tricky with the design till we are displaying more since it will likely need to be rearranged as more data is added. Ticket window progress bar looks useful. but with the long size the text becomes not very readable. If used, id keep the bar to a small static length so that the proportions remain constant.

chappjc:
@raedah makes a good point about the ratio of rewards being static.  Also it would be hard to fit the text inside nicely, although dev/PoS/PoW could be abbreviations.
The ticketwindow progress bar is useful for sure, but it's a bit out of proportion.

gozart1:
Another interesting progress bar would be to the current block reward reduction interval.

chappjc:
That's a good point. Unlike Bitcoin where it happens very rarely in decred you can actually see it happening slowly though

gozart1:
reworked root a bit to add ticket window progress
screen shot 2017-10-03 at 5 00 50 pm
screen shot 2017-10-03 at 5 01 03 pm

gozart1:
@chappjc what do you think of bunching the reward breakdown like this for now?

screen shot 2017-10-03 at 5 08 16 pm

gozart1:
Added another small change to compact those latest blocks headings earlier so we can keep it 2 columns for wider range of screens
screen shot 2017-10-03 at 5 12 30 pm

chappjc:
Yeah, that looks really good.  I'm happy.  @raedah?

gozart1:
Here's how progress bar looks when its emtpy

screen shot 2017-10-03 at 5 18 32 pm

gozart1:
don't merge this yet, there are some bugs right now when the js updates the html

chappjc:
@gozart1 this one? https://github.com/dcrdata/dcrdata/issues/207

gozart1:
Ah yes, lets fix the time column in separate PR then. I just fixed the issue of the font size of block height not being set to 18px on newly added blocks.

chappjc:
Just found one peculiarity in this width:

image

It's not terrible, but the bottom table should be wider and/or the price window bar should be shorter.

gozart1:
The approach to html/css that I've been using is to leverage the latest version of Bootstrap 4 as much as possible.

And when it lacks a certain utility class that I need to tweak something like, I add my own reusable utility classes. 

A lot of the classes you see in my markup like `pl-1` are Bootstrap classes, this one in particular means to add bit of left padding.

gozart1:
I haven't figured out a way to make the table overflow _and_ fill the width of container in a smart way. But I've added a bit of css just now to limit the progress bar so it doesn't stick out further than the table in that screen size.

gozart1:
I'd really love to figure out a cool way to show the voting outcome of each block in these block indexes.

chappjc:
That's cool about the table overflow thing.  no biggie.
Regarding "voting outcome of each block in these block indexes", I'm not sure what you mean?  A summary on the list of blocks?

gozart1:
Something along the lines of this...
screen shot 2017-10-03 at 10 06 11 pm

But there are obvious issues that we'd need to address, like how do deal with multiple things being voted on.

gozart1:
Edited above image for clarity. The vote bubbles could also be links to the vote txn.
Think something like this would be neat to have for when a vote is contentious.

gozart1:
lemmi rebase this and I think its in good shape to merge

gozart1:
alright, rebased. One last thing I noticed after rebasing is that total and size where not on right side of table like the other tables on block page, so I moved those 2 cols to right side as well.

raedah:
Looks great. I am happy to get things merged sooner and can patch any bugs later rather then deal with lots of merge conflicts.

PR #205, opened 2017-09-28
block view: show vote version in table

chappjc:
resolves issue https://github.com/dcrdata/dcrdata/issues/204


chappjc:
@RogueElement Cool, thanks.  Give `git cherry-pick [commit-hash]` a shot to save some effort and avoid a merge commit.

Issue #187, opened 2017-09-19
Show minor vote info on Block Page


RogueElement:
Maybe just a column to show the vote bits or columns to show the vote choice on LN or SdiffAlgo? Just something along those lines.

raedah:
Can also be noted that we should be showing the basic vote info, which is the vote on if the previous block is accepted as valid.  Soon we will also have to be time boxing the additional votes, because their op_return identifiers will only be valid during their voting period and will be replaced with new voting info. Will be conditional upon the block height for a specific range that the voting takes place.

chappjc:
Here's what we have now:

![image](https://user-images.githubusercontent.com/9373513/30725423-938c1fc6-9ef9-11e7-9b0a-1216aaa88ee5.png)

So we have a clear summary of the vote choices.  We do however lack an block validity indication, which is easily added.

The vote bits are there, as well as the vote version.  The vote version is used to fetch the data shown in the table out of the network's `chaincfg.Params`.  It is true that when dcrdata is rebuilt with a new version of `package chaincfg` that includes the next agenda and removes the previous, then this particular vote will no longer have any information displayed beyond version and bits.

raedah:
Because of the new agendas, we will need to save the old the old agenda descriptions locally and display based on date. Unfortunately, dcrd wont have this historic data.

chappjc:
Correct about storing agendas.  We had this issue for the official voting/hardfork demo site, and I implemented a database using a BoldDB ORM:  https://github.com/decred/hardforkdemo/pull/78
We can do something similar, or use the primary DB that's in the works.

chappjc:
Note that it's not an issue yet as the v4 and v5 agendas are both still in the code: https://github.com/decred/dcrd/blob/master/chaincfg/params.go#L523-L606

PR #206, opened 2017-09-29
Adding vote details to block page

McEdward:
resolving issue #187

### DIFF
--- a/views/block.tmpl
+++ b/views/block.tmpl
@@ -131,18 +131,27 @@
                         <th class="text-right">Total DCR</th>
                         <th>Version</th>
                         <th>Size</th>
+                        <th>Issue ID</th>
+                        <th>Choice ID</th>
                     </thead>
                     <tbody>
-                        {{ range .Votes}}
+                        {{range $i, $v := .Votes}}
                         <tr>
                             <td>
                                 <span>
-                                    <a class="hash" href="/explorer/tx/{{.Txid}}">{{.Txid}}</a>
+                                    <a class="hash" href="/explorer/tx/{{.Txid}}">{{$v.Txid}}</a>
                                 </span>
                             </td>
+<<<<<<< HEAD
                             <td class="dcr mono fs15 text-right">{{.Vout | getTotalFromBlock | float64Commaf}}</td>
                             <td class="mono fs15">{{.VoteInfo.Version}}</td>
                             <td class="mono fs15">{{.Hex | size | formatBytesInt}}</td>
+=======
+                            <td class="dcr mono fs15 text-right">{{$v.Vout | getTotalFromBlock | float64Commaf}}</td>
+                            <td class="mono fs15">{{$v.Hex | size | formatBytesInt}}</td>
+                            <td>{{(index (index $.VoteInfo $i).Choices 0).ID}}</td>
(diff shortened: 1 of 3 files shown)


chappjc:
So this all looks like good work, but we may need to make some changes.

PR https://github.com/dcrdata/dcrdata/pull/205 adds the vote version to the table, while this one adds the vote's issue and agenda ID.  We should show both.  So we can put both PRs in, but one will have to be rebased (and conflicts resolved).

The problem with the way the vote info is added in this PR is that it does not account for the sort of the Votes slices going on here:  https://github.com/dcrdata/dcrdata/blob/0f0aae7a123312e716e49c0fce3386872e60059e/explorer.go#L105 in the block page handler.  The way i addressed that is to make slice of a new type that embeds the raw tx with the vote info.  This way the vote info goes along for the ride during the sort and it stays with the correct vote.

If the above makes sense, what I suggest is this:

1. PR 205 goes in as is
2. you pull the updated master
3. rebase this PR on master, simultaneously resolving conflicts and refactoring to use the `[]apitypes.TxRawWithVoteInfo` as in PR 206 so the vote info stays aligned with the sorted votes.

When doing 3., block.tmpl will hopefully become simpler as you should no longer have to index with `i`.

@McEdward let me know if this makes sense.  Comments from anyone? @RogueElement and @raedah ?

chappjc:
A problem this has is when there is a vote on a version that does not have an agenda in the hard-coded map of deployments.  Basically right now the code tries to index a nil Choices slice in the block template and the page errors out.

What we need is this:

image

I suggest something like this:

```go
                            <td class="mono fs15">{{.VoteInfo.Version}}</td>
                            <td class="mono fs15">{{.Hex | size | formatBytesInt}}</td>
                            <td class="mono fs15">{{if (len .VoteInfo.Choices) gt 0}}
                                {{(index .VoteInfo.Choices 0).ID}}
                            {{else}}
                                -
                            {{end}}</td>
                            <td class="mono fs15">{{if (len .VoteInfo.Choices) gt 0}}
                                {{(index .VoteInfo.Choices 0).Choice.Id}}
                            {{else}}
                                -
                            {{end}}</td>
```

chappjc:
Actually, I suggest putting the vote version next to the other vote info columns in that table.

views/block.tmpl
```diff
@@ -143,6 +145,8 @@
                             <td class="dcr mono fs15 text-right">{{.Vout | getTotalFromBlock | float64Commaf}}</td>
                             <td class="mono fs15">{{.VoteInfo.Version}}</td>
                             <td class="mono fs15">{{.Hex | size | formatBytesInt}}</td>
+                            <td class="mono fs15">{{(index .VoteInfo.Choices 0).ID}}</td>
+                            <td class="mono fs15">{{(index .VoteInfo.Choices 0).Choice.Id}}</td>
```
chappjc: See my comment https://github.com/dcrdata/dcrdata/pull/206#issuecomment-333959553 about a nil Choices slice.

chappjc:
nice.  Going in.

Although, we'll have to make this work for deployments/agendas with multiple choices, such as vote version 4, where there was the `sdiffalgorithm` as the first choice and then `lnsupport`, which we just aren't showing with the code in this PR.  With vote version 4, there is just the one choice on `lnfeatures`, so this is fine.

PR #216, opened 2017-10-05
split layout 50/50 for tx.tmpl input/output tables

gozart1:
this simply splits the layout so its 50/50 which solves the one character overflow in output address when screen size is large.

before 
screen shot 2017-10-05 at 9 42 32 am

after
screen shot 2017-10-05 at 9 42 45 am

### DIFF
--- a/views/tx.tmpl
+++ b/views/tx.tmpl
@@ -84,7 +84,7 @@
     </div>
 
     <div class="row">
-        <div class="col-md-7 mb-3">
+        <div class="col-md-6 mb-3">
             <h4>Input</h4>
             <table class="table table-sm striped">
                 <thead>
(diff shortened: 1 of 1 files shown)


chappjc:
Looks good on chrome, but something is wrong on firefox:

image

chappjc:
It seems that for `word-break` to work in Firefox, it needs to be in a block or something like that.  Anyway, this seems to work, although it seems really wrong to my untrained CSS eye:

```css
td > * > .break-word {
  display: table-cell;
}
```

chappjc:
I'm testing without the 50/50 and just the `display: table-cell` change and it seems like we don't need the 50/50 change anymore.

gozart1:
closing in favor of #222

gozart1:
generally you should never use `*` in css, because its really inefficient to compute due to the way css works. https://css-tricks.com/efficiently-rendering-css/ 
This particular issue can be solved by simply moving the class to the wrapping block level element in the DOM. ( the `<td>` in this case )

PR #219, opened 2017-10-05
Show block validity for ssgen, reintroduce BlockValidation type

chappjc:
Addresses issue https://github.com/dcrdata/dcrdata/issues/217

As usual, this is just terrible get-it-done web programming.  Sorry.


chappjc:
Looks like this:

image

PR #223, opened 2017-10-06
rm commas in block height, rm mono font for explorer table

gozart1:


chappjc:
Still gotta try this out, but I thought we wanted commas in the height after all? Maybe just at top of block page and not in table?

Also, I always felt that monospace was the nicest way to show hex strings, but like I said I haven't actually looked yet.  :)

raedah:
See comma conversation in issue https://github.com/dcrdata/dcrdata/issues/200

PR #225, opened 2017-10-06
add turbolinks

gozart1:
This is a spike to explore if we want to add turbolinks as discussed in https://github.com/dcrdata/dcrdata/issues/220#issuecomment-334666580

Give it a whirl and let me know if you notice any issues.

If we end up going this route, it just takes a bit of awareness to structure our js in a way that doesn't create memory leaks, since the javascript context sticks around as you navigate.

Luckily turbolinks has a great api for controlling js execution, so it shouldn't be much of an issue.


chappjc:
Seems fine, although when run locally I don't see much difference.  :)  Maybe I need to try on the server.

gozart1:
A good way to test it is to simulate a slow connection via Chrome dev tools

screen shot 2017-10-08 at 11 29 02 am

You'll notice when navigating the html gets fetched via XHR.
Since we already do a great job setting cache headers adding turbolinks doesn't really optimize amount transferred over the network, but it does reduce the work the browser needs to do since it doesn't need to reload the whole page. 

BTW, don't merge this yet. The js in root.tmpl that handles socket updates still needs to be scoped properly.

chappjc:
Cool, I'll give that a shot.  In any case, this seems really nice.  Are there any downsides or caveats to this?  Crawlers or search engines OK with this?

gozart1:
As I was looking at the js for root.tmpl to make sure it plays nice w/ turbolinks, I ended up refactoring it a bit. I added zepto.js to give us a bit more of a toolkit for writing terse client side html/css manipulations.

 Sorry to lump the refactor into a single PR, but the changes are minimal functionality wise:

- Ticket window progress bar now updates when new block is added
- "Time" column now gets set when new block is added
- The dark/light theme is now non-blocking after the initial load and gets applied to the turbolinks body before it hits the DOM.

One thing to note is that currently the socket connection stays open as you navigate away from root.tmpl, at first I was thinking we should close it. But another way to approach it would be to make it universal and use it to display best block in top right an keep that updated. Thoughts?

chappjc:
I do like the idea of reusing the same socket connection on each page for some bit of information like block height or maybe a status thing like nodes connected and sync height like Insight.  Perhaps number of transactions in mempool would be cool too since that would change more frequently than other values.

But that can wait for another PR of course.  I'm cool with the PR as is, just need to review one more time and we can merge in a few hours.

PR #226, opened 2017-10-07
format trailing DCR zeroes on server instead of client

gozart1:
This moves formatting of trailing zeroes in DCR values to the server.
There should be no difference in terms of presentation, other than removed the flash of un-styled content on page load.

This simply uses Sprintf to round to 8 decimals, so this doesn't fully address rounding issues in mentioned in https://github.com/dcrdata/dcrdata/issues/195

I haven't tried optimizing the formatting function, still getting familiar with Go so if you see any obvious optimizations or issues feel free to push a commit/comment.

### DIFF
--- a/explorer/explorer.go
+++ b/explorer/explorer.go
@@ -4,6 +4,7 @@
 package explorer
 
 import (
+	"fmt"
 	"html/template"
 	"io"
 	"net/http"
@@ -292,6 +293,20 @@ func New(dataSource explorerDataSource, userRealIP bool) *explorerUI {
 		"timezone": func() string {
 			t, _ := time.Now().Zone()
 			return t
+		},
+		"fancyDCR": func(v float64) []string {
+			roundedV := fmt.Sprintf("%.8f", v)
+			trailingZeroes := ""
+			for i := range roundedV {
+				if roundedV[len(roundedV)-1-i] != 48 {
+					break
+				}
+				trailingZeroes += "0"
+			}
+			sliceIndex := len(roundedV) - len(trailingZeroes)
+			roundedV = roundedV[0:sliceIndex]
+			result := append(make([]string, 0, 2), roundedV, trailingZeroes)
+			return result
 		},
 	}
 
--- a/explorer/explorer.go
+++ b/explorer/explorer.go
@@ -12,8 +12,8 @@ import (
 	"os/signal"
 	"path/filepath"
 	"strconv"
-	"time"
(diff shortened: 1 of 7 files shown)


explorer/explorer.go
```diff
@@ -293,6 +294,20 @@ func New(dataSource explorerDataSource, userRealIP bool) *explorerUI {
...
 			return t
 		},
+		"fancyDCR": func(v float64) []string {
+			roundedV := fmt.Sprintf("%.8f", v)
+			trailingZeroes := ""
+			for i := range roundedV {
+				if roundedV[len(roundedV)-1-i] != 48 {
+					break
+				}
+				trailingZeroes += "0"
```
chappjc: This seems OK, but in general I try to use the standard library where possible.  In this case, `roundedV = strings.TrimRight(roundedV, "0")` would get you most of the way, and then you just need `trailingZeros := string.Repeat("0", oldLength - newLength)`

gozart1: Well aren't those handy! That's much better.
Pushed another commit using strings lib.

raedah:
The best way to deal with this really is to remove this rounding code completely so that we can expose where the bad var types are and get them fixed at the root so that we do not return inaccurate values.

chappjc:
@raedah Agreeed.  But I this PR, the inputs to `FancyDCR` are the already mangled numbers from the go code.  The issue AFAICT is that there's too much math in floating point when all amounts are fundamentally int64 values.  They should only be converted and shown as float at the very last step.  

Anyway, the question of how to styling numbers that do have trailing zeros exists in any case.

raedah:
Yes, I want the mangled numbers exposed so we know what needs to be fixed instead of hiding them, which causes them to not be known and get fixed. Im not aware of anywhere that we are doing math and would end up with a floating point. All information I can think of should be dcrvalues and should not be leaving 8 decimal points to begin with.

chappjc:
One way to keep it pretty while we fix this, but have some visibility into the problem is to have the values have a hover/tooltip showing the value prior to sprintfing

raedah:
That would be some extra work to make it pretty, and I rather it not be pretty so that we stay focused on fixing the problem. It will just be in master and should not be being pushed out to the live explorer site.

chappjc:
Precision issue fix will be in follow-up PR based on https://github.com/dcrdata/dcrdata/pull/229 (I'm gonna split that one up)

PR #229, opened 2017-10-10
Float amount precision and coin supply formatting

chappjc:
This fixes floating point precision issues with amount sums.

- Add `txhelpers.TotalVout `to sum each Value for a slice of `Vout`s in `dcrutil.Amount `instead of `float64`.
- Fix `Total` sent precision by summing in `dcrutil.Amount`.

<s>Related, fix coin coin supply display, which was sent to template exec as a string.

- `BlockExplorerExtraInfo.CoinSupply` is now a `int64 `instead of a string.  Previously the string had the `" DCR"` prefix baked in, which messed up the trailing zero styling.
- Add js function `styleAmountAsDCRVal `to handle trailing zero styling similar to `styleDCRVal `but taking an `int64 `with decred atom semantics instead of a `float64 `(coin).  This applies automatically to elements of class `"dcr-amount"`.</s> coin supply fix later

### DIFF
--- a/dcrsqlite/apisource.go
+++ b/dcrsqlite/apisource.go
@@ -577,14 +577,10 @@ func (db *wiredDB) GetAddressTransactions(addr string, count int) *apitypes.Addr
 	}
 	tx := make([]*apitypes.AddressTxShort, 0, len(txs))
 	for i := range txs {
-		var value float64
-		for j := range txs[i].Vout {
-			value += txs[i].Vout[j].Value
-		}
 		tx = append(tx, &apitypes.AddressTxShort{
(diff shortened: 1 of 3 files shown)


chappjc:
Aw crap, I should have done this on top of PR https://github.com/dcrdata/dcrdata/pull/226.  Don't worry @gozart1 we'll merge your PR as is and I'll rebase and rework my solution to fit.

gozart1:
Ok, feel free to just push to that branch as well, whatever works.

chappjc:
coin supply fix will be in follow-up PR now

Issue #110, opened 2017-09-06
Transaction outputs should indicate if unspent or spending tx


chappjc:

raedah:
I imagine this could be done with a Spent column that indicates Yes or No.

chappjc:
On Insight, the outputs have either a green (U) indicating unspent, or a ">" that is a link to the transaction that spends it.

raedah:
Looks like what insight does is link to the address, and from the address page it is shown if it is spent or not. So lets show on the TX page if it is spent, and show on the Address page if its incoming or outgoing. These can just be put in a column for now, and we can change visual indicators once we are settled on the layout and move to design.

chappjc:
For reference, here's Insight's display of spend outputs with a ">" that links to the spending tx:

![image](https://user-images.githubusercontent.com/9373513/30457067-e537f0ec-995a-11e7-870e-204a442c3042.png)

And unspent, with a green "U" that takes you nowhere, of course:

![image](https://user-images.githubusercontent.com/9373513/30457128-1fc2679c-995b-11e7-8920-b5bd01e5fc96.png)

raedah:
Ah, never noticed that before. Insight has a few 'hidden' features. Will work on UI to make these things accessible.

RogueElement:
I feel like there should be tool tips to give a bit of info on a lot of things.

chappjc:
For now, just showing unspent if unspent is enough.  I'm going to make a new issue to create v2 of our database, with the ability to query for spent status, etc.

chappjc:
@RogueElement I know you looked checking spent status some.  Was there a branch you were working on for this?

RogueElement:
No, didn't actually have anything to code just theory, seemed like looking for the spending scriptsig that matches the spent scriptpubkey would work but that seem super inefficient so I just kept on looking.

chappjc:
Yeah, we're gonna have to skip the link to spending tx for now.  Just showing spend/unspent status is enough for now.

chappjc:
Addressed by @RogueElement in PR https://github.com/dcrdata/dcrdata/pull/231 (merged) and @chappjc in PR https://github.com/dcrdata/dcrdata/pull/209 (ready for review and merge)

PR #231, opened 2017-10-10
Show if tx ouput is spent or unspent

RogueElement:
Should address #110. For an unconfirmed tx, all outputs should show false in the spent column. @gozart you probably have a better idea on how to display this.

### DIFF
--- a/dcrsqlite/apisource.go
+++ b/dcrsqlite/apisource.go
@@ -889,12 +889,17 @@ func (db *wiredDB) GetExplorerTx(txid string) *explorer.TxInfo {
 	tx.Vin = inputs
 
 	outputs := make([]explorer.Vout, 0, len(txraw.Vout))
-	for _, vout := range txraw.Vout {
+	for i, vout := range txraw.Vout {
+		txout, err := db.client.GetTxOut(txhash, uint32(i), true)
+		if err != nil {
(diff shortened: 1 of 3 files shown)


dcrsqlite/apisource.go
```diff
@@ -889,12 +889,17 @@ func (db *wiredDB) GetExplorerTx(txid string) *explorer.TxInfo {
 	tx.Vin = inputs
 
 	outputs := make([]explorer.Vout, 0, len(txraw.Vout))
-	for _, vout := range txraw.Vout {
+	for i, vout := range txraw.Vout {
+		txout, err := db.client.GetTxOut(txhash, uint32(i), true)
+		if err != nil {
+			log.Warnf("Failed to determine if tx out is spent for ouput %d of tx %s", i, txid)
+		}
```
chappjc: for `if err == nil && txout == nil`, I'd say that we want to check with `getrawtransaction` if it is really just in mempool

I'm not really sure what the mempool arg for GetTxOut is supposed to help with

RogueElement:
Shows false for all outputs for a tx in the mempool

dcrsqlite/apisource.go
chappjc: nvm.  If it presently shows unconfirmed transaction outputs as unspent, then we're all good.

PR #232, opened 2017-10-10
Change coin supply to int64 and format nicely.

chappjc:
NOTE:  Coin supply is not updated by websocket yet, so this number never changes unless the page reloads

This PR removes `" DCR"` suffix from value passed from go to browser.  Now a JS function expecting just a number that operates on it will not fail.
Add `amountAsFloatParts` function to web and explorer template maps.  It takes an amount as `int64` and returns a `[]string` of: part left of decimal place, right with trailing zeros removed, and trailing zeros.
Add `fancyFloatParts` html template definition that takes the `[]string` from `amountAsFloatParts` and creates a stylized floating point coin value.
Adjust title of root page.

### DIFF
--- a/explorer/explorer.go
+++ b/explorer/explorer.go
@@ -296,12 +296,25 @@ func New(dataSource explorerDataSource, userRealIP bool) *explorerUI {
 			return t
 		},
 		"fancyDCR": func(v float64) []string {
-			roundedV := fmt.Sprintf("%.8f", v)
-			oldLength := len(roundedV)
-			roundedV = strings.TrimRight(roundedV, "0")
-			trailingZeros := strings.Repeat("0", oldLength-len(roundedV))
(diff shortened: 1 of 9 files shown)


explorer/explorer.go
```diff
@@ -296,13 +296,26 @@ func New(dataSource explorerDataSource, userRealIP bool) *explorerUI {
 			return t
 		},
 		"fancyDCR": func(v float64) []string {
```
gozart1: I like the nomenclature change to `clipped` here.

How about we rename this to `float64asFloatParts` and make its return value a string slice with three values just like `amountAsFloatParts`?  That way they can share the same render partial.

chappjc:
@gozart1 OK.  I did some renaming and rebased.  Hopefully it is still correct.

chappjc:
hmm, github seems to have gotten confused.  I'll make a new PR if it doesn't see the new commit soon.

Issue #121, opened 2017-09-06
Top of an address page shows summary


raedah:
Transactions:
Total Received:
Current Balance:  or  'sum of unspent outputs (UTXOs)'
Unconfirmed:

We have a fixed limit of tx being displayed currently. If the # of transactions is less then the max, then display the sums. If we are hitting it we should not show the sums which would be incomplete, but do show a warned that all transactions are not being displayed below.

raedah:
dup https://github.com/dcrdata/dcrdata/issues/131

chappjc:
Needs aux/auxiliary DB (DB v2) to be built out first.  Namely, it needs address queries.

raedah:
We were able to get all of this functioning without problem, just limited to addresses that return less then 500 transactions. https://github.com/dcrdata/dcrdata/compare/master...McEdward:summay_at_the_top_of_the_address_page

Just working out the dcrutil.Amount issue for the templates is left.  Looks like .ToUnit(-8) does the trick to return satoshis.

PR #233, opened 2017-10-11
Adding a summary at the top of the Address Page

McEdward:
Trying to resolve issue #121

### DIFF
--- a/views/address.tmpl
+++ b/views/address.tmpl
@@ -24,8 +24,8 @@
                 <p>Transactions:   {{$Transactions}}</p>
                 <p>Current Balance: {{$CurrentBalance}}</p>
                 <p>Unconfirmed: {{$Unconfirmed}}</p>
-                {{if eq $Transactions 2000}}
-                    Currently only the first 2000 transactions are being used.
+                {{if eq $Transactions 500}}
+                    Currently only the first 500 transactions are being used.
                 {{end}}
                 <h5>Transactions</h5>
                 <table class="table table-mono-cells table-sm striped">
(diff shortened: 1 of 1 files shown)


McEdward:
I seem to be having issues adding the Current balance, Please feel free to give your opinion.

chappjc:
I don't like the "balance" nomenclature, but that's how people seem to like it.  Really the right name is sum of unspent outputs (UTXOs) with this address.  Given that more accurate name, it's more clear how to get it.

However unless we get *all* transaction for an address, the notion of a balance is 👎 .  So for this number we should wait until we have a DB to query.

views/address.tmpl
```diff
@@ -13,6 +13,20 @@
...
+                {{range .Transactions}}
+                    {{if eq .Time 0}}
+                        {{$Unconfirmed +1}}
+                    {{end}}
+                {{end}}
+                <p>Transactions:   {{$Transactions}}</p>
+                <p>Current Balance: {{$CurrentBalance}}</p>
+                <p>Unconfirmed: {{$Unconfirmed}}</p>
+                {{if eq $Transactions 2000}}
+                    Currently only the first 2000 transactions are being used.
```
chappjc: I think the number is more like 700 now, but I don't love putting a magic number here.

McEdward: I also wasn't sure it was still 2000, because i kept getting a max high of 500. I could change this now and wait for the new DB before anything else here. Just confirming if it's 700 first?

chappjc: yup, looks to be 500 actually, like you said: https://github.com/dcrdata/dcrdata/blob/b98c0b9bbbf7d0e3a7288e80f33f55a842d1fed1/explorer/explorer.go#L31

raedah:
I like 'sum of unspent outputs (UTXOs)'
We can do this because most addresses dont hit the limit, and we can easily detect if we are hitting the limit.

Issue #154, opened 2017-09-09
Ticket commitment state


raedah:
Tickets have an OP_RETURN script for commitments ... locked funds. Funds should show as locked and not sent to the receiving address until the state has changed.

Stake Commitments: https://github.com/decred/dcrd/blob/master/blockchain/chaingen/generator.go#L439-L482

chappjc:
Along the same line, a vote should show as locked funds until 256 confirms

chappjc:
@raedah did @RogueElement hit everything in the PR just merge?

PR #234, opened 2017-10-12
Added ticket maturity to tx page

RogueElement:
Addresses #154

### DIFF
--- a/views/tx.tmpl
+++ b/views/tx.tmpl
@@ -156,12 +156,15 @@
                     <th class="text-center">Spent</th>
                 </thead>
                 <tbody>
-                    {{range .Vout}}
-                    {{ if ne .Amount 0.0}}
+                    {{range .Vout}} 
                     <tr>
                         <td class="break-word">
-                            {{range .Addresses}}
-                                <a class="mono address" href="/explorer/address/{{.}}">{{.}}</a><br>
+                            {{if ne .Amount 0.0}}
+                                {{range .Addresses}}
+                                    <a class="mono address" href="/explorer/address/{{.}}">{{.}}</a><br>
+                                {{end}}
+                            {{else}}
+                                    <a class="mono address">{{.OP_RETURN}}</a>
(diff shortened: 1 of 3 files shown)


raedah:
Works well, I like. Saw a few things...
Stakebase output on the vote page is missing.
In addition to tickets, mature applies to votes and coinbase txs as well so should be added to those too.

views/tx.tmpl
```diff
@@ -38,6 +38,22 @@
                         {{.Type}}
                     </td>
                 </tr>
+                {{if eq .Type "Ticket"}}
+                <tr>
+                    <td class="text-right pr-2 h1rem p03rem0">Mature</td>
```
raedah: Capitalize to match rest of template.

views/tx.tmpl
```diff
@@ -38,6 +38,22 @@
...
+                    <td>
+                        {{.TicketMature}}
+                    </td>
+                </tr>
+                {{end}}
+                {{if eq .Type "Vote"}}
+                <tr>
+                    <td class="text-right pr-2 h1rem p03rem0">Funds locked?</td>
```
raedah: LOCKED

views/tx.tmpl
```diff
@@ -38,6 +38,22 @@
                         {{.Type}}
                     </td>
                 </tr>
+                {{if .Mature}}
+                <tr>
+                    <td class="text-right pr-2 h1rem p03rem0">Mature</td>
```
raedah: caps like the rest of the labels

chappjc:
This still has the WIP label.  Is it ready?

raedah:
Yeah its ready.

chappjc:
LGTM, runs fine, but it would be nice if we could limit the length of `OP_RETURN` data with ellipses because this is data with no limit on size.

chappjc:
You may also want to rebase and dep ensure.

PR #236, opened 2017-10-13
handle age formatting via data attributes

gozart1:
- Corrects a bug were the wrong type was being passed to timeSince ( requires unixTime, was getting native Date object )
- Simplifies handling of timestamp values by using data attribute, instead of coordinating index of 2 separate arrays of elements.


chappjc:
could you remove the link for the "dcrdata.org" in the footer while you are here?  Eventually we'll have a groovy organization page, but for now it just redirects to the explorer.

gozart1:
done

chappjc:
thanks!

PR #239, opened 2017-10-17
Remove lots of pointless RPC calls.

chappjc:
`FeeInfoBlock` and `FeeRateInfoBlock` are much faster now.  Hence sync is way faster now (40% or more).

Add TicketTxnsInBlock.


raedah:
Tested all /api/mempool/sstx end points. Data looks good.

PR #242, opened 2017-10-20
standardize decimal formatting on server & client

gozart1:
- Standardize output of float64 and int64 formatting helpers so that they can share the same template. 
- Adds decimalParts formatter to humanize.js so we can keep decimal styling consistent when new data comes in via websocket 
- Add useCommas option to formatter functions
- Fixes various formatting inconsistencies when a new block updates values root.tmpl ( the js now formats values in the same way server does )

root.tmpl
root

tx.tmpl ( block.tmpl has similiar treatment ) 
total sent


chappjc:
This looks great!  Looks like Travis wants a `go fmt ./...` though.

gozart1:
thx, fmt'ed

raedah:
Looks really good.

explorer/explorer.go
```diff
@@ -295,15 +296,21 @@ func New(dataSource explorerDataSource, userRealIP bool) *explorerUI {
...
 			trailingZeros := strings.Repeat("0", oldLength-len(clipped))
-			result := append(make([]string, 0, 2), clipped, trailingZeros)
-			return result
+			valueChunks := strings.Split(clipped, ".")
+			integer := valueChunks[0]
+			if useCommas {
+				integerAsInt64, _ := strconv.ParseInt(integer, 10, 64)
+				integer = humanize.Comma(integerAsInt64)
+			}
+			dec := valueChunks[1]
```
chappjc: this could panic if valueChunks length <2.  I don't see how that could happen easily, but it would be best to guard for this.

gozart1:
@chappjc Ok, added some defense for the following scenarios:

1. If NaN ( or some other val that could cause there to be no decimal ) is input we log the occurrence, but let it pass through:
nan

2. If `strconv.ParseInt` fails, we log the error and display an error value in the UI
error

public/js/humanize.js
```diff
@@ -8,6 +8,31 @@ var humanize = (function() {
...
+      while (i--) {
+        if (decimal[i] == "0") {
+          numTrailingZeros++
+        } else {
+          break
+        }
+      }
+      var decimalVals = decimal.slice(0,decimal.length - numTrailingZeros)
+      var trailingZeros = (numTrailingZeros == 0) ? "" : decimal.slice(-(numTrailingZeros))
+      return "<span class='int'>"+int+"</span>"+
```
gozart1: When we build chunks of html strings that will get evaluated by the client, we should really be scrub out scripts as a security best practice. We could use [jQuery.parseHTML()](https://api.jquery.com/jquery.parsehtml/) for example.

Zepto doesn't come with that so we should either move to a lightweight build of jQuery ( since we're trying to keep page weight as low as possible ) or add some other template system in that guards against this efficiently.

gozart1: @chappjc if I strip out everything we don't need from the latest jQuery, its down to 24.72 Kb when minified and gzipped. You OK w/ switching to that? Zepto is currently 10.5Kb over the wire.

chappjc:
Yeah I'm cool with that from a size standpoint, mainly because of caching.  The only concern I have is speed.  I'm guessing it's trivial.

gozart1:
Yea, it def adds some computational overhead
https://measurethat.net/Benchmarks/Show/857/0
But the workload is light so not a problem.
Gonna go ahead and make that change.

gozart1:
Pushed a few more minor updates, looking ready to ship

Issue #244, opened 2017-10-22
Fix width ratio between columns on transaction page


raedah:
Previous Outpoint, Addresses
The prior has too much space, and the later does not have enough. It should be adjusted so that the address can be displayed on 2 rows instead of 3, the same as on the Output side.

This is a good example, https://explorer.dcrdata.org/explorer/tx/45197998ac8b430f7b8a5d0053ef0a28c51c0a7c43f64765a701899e227e6492

Also check the display in firefox.

chappjc:
@raedah have you looked at this on Firefox? I still think there's work to be done here

raedah:
Its improved on firefox, will leave open for further consideration though.

chappjc:
I see the addresses wrapping to a third line, just barely.

ZeroASIC:
I see two options here.  One would be widening the entire container to at least 1173px and the div to 232px, but that requires overriding the bootstrap container.  The other option, although maybe more complicated, is shrinking the Type column and the Ouput div and then widening the Input div.  I'm not sure which option I like more, but I thought I saw you guys talking about getting rid of the bootstrap overrides.

A third option could be shrinking the font size to 12 instead of 13, but that makes it less readable.  I think the real problem here is that the browsers don't render the same and that means one of the two has a bug.  I think all the webkit browsers render it the same way.

gozart1:
Changes to bootstrap base frameworks can now be done using variables
https://getbootstrap.com/docs/4.0/getting-started/theming/#variable-defaults

Right now, we are not making use of this yet, but I'm planning to do some cleanup of our stylesheets for things that can be solved with variables at the framework layer instead.

If you want to make changes like modifying the layout framework params, you would define variables after line 3 here https://github.com/decred/dcrdata/blob/master/public/scss/application.scss#L3 ( in a separate import since application.scss is intended to be just a manifest). See defaults here https://github.com/twbs/bootstrap/blob/v4-dev/scss/_variables.scss

ZeroASIC:
This issue doesn't occur on alpha anymore.

PR #246, opened 2017-10-24
balance cell dimensions of Input/Output tables in tx.tmpl

gozart1:
This addresses #244 by tweaking the layout of table cells in tx.tmpl so that the input and output rows are the same height as much as possible over a wide range of screen sizes. Text balancing is tricky business and it's not really possibly to have maintainable/efficient code AND have it look perfect for ALL screen widths. This is an effort to improve over the current state while keeping the CSS simple.

As the old woodcutters saying goes:
"Hew to the line and let the chips fall where they may"

Here's a bunch of screen grabs at various screen widths:

screen shot 2017-10-24 at 12 11 18 am
screen shot 2017-10-24 at 12 11 42 am
screen shot 2017-10-24 at 12 11 55 am
screen shot 2017-10-24 at 12 12 11 am
screen shot 2017-10-24 at 12 12 21 am
screen shot 2017-10-24 at 12 28 53 am

### DIFF
--- a/public/css/main.css
+++ b/public/css/main.css
@@ -258,8 +258,13 @@ body.darkBG .navbar-fixed-bottom a {
   font-size: 14px;
 }
 .table .address {
-  max-width: 127px;
-  display: inline-block;
+  max-width: 123px;
+}
+.table .outpoint {
+  max-width: 222px;
+}
+.table-fixed {
+  table-layout: fixed;
 }
 
 /*content*/
@@ -337,6 +342,10 @@ body.darkBG .navbar-fixed-bottom a {
   .table {
     margin-bottom: 0;
     width: 100%;
+  }
+  .sm-fullwidth {
+    width: 100%;
+    max-width: 100% !important;
   }
   .sm-w151 {
     width: 151px;
--- a/views/tx.tmpl
+++ b/views/tx.tmpl
@@ -101,17 +101,18 @@
     <div class="row">
         <div class="col-md-7 mb-3">
             <h4>Input</h4>
-            <table class="table table-sm striped">
+            <table class="table table-sm table-fixed striped">
                 <thead>
                     <th>Previous Outpoint</th>
                     <th>Addresses</th>
-                    <th class="text-center">Block</th>
-                    <th class="text-center">DCR</th>
+                    <th class="text-center" width="60">Block</th>
+                    <th class="text-center" width="100">DCR</th>
                 </thead>
                 <tbody>
                     {{range .Vin}}
                     <tr>
-                        <td class="break-word">
+                        <td class="break-word mono fs13">
+                            <div class="outpoint">
                             {{if .Coinbase}}
                                 Coinbase: {{ .Coinbase }}
                             {{else if .Stakebase}}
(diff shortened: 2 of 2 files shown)


raedah:
Looking at just the first screenshot, I see that the Input/Address has a consist overlap of 3 chars of the top row over the bottom row. This could be narrowed so that it becomes only a 1 char overlap. The Output/Address has a consistent overlap of 7 chars, and it could also be narrowed to 1.

chappjc:
@raedah It looks like the Address columns do not have a defined width, leaving that to Block and DCR columns, so the address ends up wrapping like a sentence .To work with any screen size, I don't know if a robust solution.

gozart1:
Alright, the address and previous outpoint text is now balanced via a max-width set on them.

While testing this out a found a transaction with some other kinds of values for in the address column, those will not be constrained by the max-width.

screen shot 2017-10-24 at 11 08 11 pm

chappjc:
Firefox 57 (linux) must be doing something different on my end.  On 03bd84c, it's doing this for a wide range of inputs:

image

I see 8f7b2bc now but I haven't tested that yet.

gozart1:
did you do hard refresh? ( your css may be cached )

chappjc:
Yeah I think so.  CTRL+SHIFT+R usually does it.

On Chrome I see what you see.  But this reveals one issue with the Previous Outpoint width:

image

The vout index can up to 3 digits long in practice.

chappjc:
Confirmed cache isn't the culprit with Firefox.  Even checked Disable Cache in the dev tools console, similar to Chrome's.

gozart1:
It's a bit of a bop a mole situation. 

Tightening up the styling to optimize for one type of value at one type of screen width tends to mess things up in a different configuration ( screen-size or a different kind of value in the row that has different space requirements )

Sure in theory we could write a very large amount of css to hard code styling for each foreseable scenario, but the would be brittle and slow us down as I'm sure the content in this view will evolve and we'd have to play the bop a mole game all over again.

I think its worth it to give up pixel perfection for cross device functionality and think we should keep the css as light as possible and only noodle with it to fix egregious things like overlapping text or anything else that makes things truly hard to read.

screen shot 2017-10-24 at 11 29 01 pm

chappjc:
Yeah I see what you mean.  It just seems like the `max-width` of the outpoint table style needs to be wider so it can get on 2 rows on a wide enough screen.  The current `max-width` seems to be picked for cases with only 1 out index digit width.

chappjc:
Aaaand you're one step ahead of me again.  :)

gozart1:
There's definitely some nuanced differences in font rendering for various os. I made a little tweak to prev outpoint, lmk if that fixes it for you.

gozart1:
ah ok, good spot, whats the max on that index?

gozart1:
squashed and rebased this

chappjc:
Hmm Firefox 56 and 57 beta are showing the same issue at max dynamic width:

image

chappjc:
Let's see if we get feedback from firefox users.

PR #249, opened 2017-10-24
Add /api/stake/vote/info?version=X route.

chappjc:
Add RPC calls to getvote info.
Add middleware to get latest vote version.

@karamble


chappjc:
Running on https://beta.dcrdata.org/api/stake/vote/info.

Without the `version` query, it figures out the latest stake version info.

Issue #247, opened 2017-10-24
Tx page shows immature status bar


raedah:
For immature tickets and votes, show a progress bar indicating the number of blocks remaining before maturity. 

This could apply to coinbase inputs as well since they have a maturity period, but we would be challenged to find a good place to the progress bar because its just a single input rather then applicable to the transaction as a whole.

PR #252, opened 2017-10-25
add maturity progress bar to tx.tmpl

gozart1:
As described in #247

Some sample scenarios:

screen shot 2017-10-25 at 2 01 45 am

screen shot 2017-10-25 at 1 51 09 am
screen shot 2017-10-25 at 1 48 19 am

### DIFF
--- a/dcrsqlite/apisource.go
+++ b/dcrsqlite/apisource.go
@@ -926,6 +926,7 @@ func (db *wiredDB) GetExplorerTx(txid string) *explorer.TxInfo {
 			tx.Mature = "True"
 		} else {
 			tx.Mature = "False"
+			tx.TicketMaturity = int64(db.params.TicketMaturity)
 		}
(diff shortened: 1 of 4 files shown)


gozart1:
At 256 confirmations,`.Maturity` is still considered "False", so I've added +1 to the progress logic to make the progress bar represent when it is actually spendable. Does that make sense?

gozart1:
this is still a wip, this should use b.params.TicketMaturity so it works on testnet

chappjc:
I love the concept and the look.  Haven't tested yet though.

gozart1:
alright, updated the progress bar to use the network params and put some better text for when its about ticket progress
screen shot 2017-10-25 at 11 13 36 pm

Thinking this ones ready to go

raedah:
I gave this a test run, looks good.

Issue #132, opened 2017-09-08
Utilized websockets for real time updates on explorer pages


raedah:
This is needed on most all of the views, blocks, block, address, transaction.

PR #253, opened 2017-10-25
Explorer websocket

RogueElement:
For #132. Adds the websocket system to explorer, the explore page now works like the root page (~~still have to make it not update when the first block on the page isn't the best block~~) and the other pages update the amount of confirmations for now.


gozart1:
Travis wants you to run `go fmt ./...` to make gometalinter happy.

views/explorer.tmpl
```diff
@@ -5,10 +5,39 @@
...
+        var updateBlockData = function (event) {
+            console.log("Received newblock message", event);
+            var newblock = JSON.parse(event);
+
+            // block summary data
+            var b = newblock;
+            var expTableRows = $('#explorertable tbody tr');
+            if (!expTableRows) return
+            //expTableRows.last().remove()
+            var newRow = '<tr id="' + b.height + '">' +
```
gozart1: As a best practice, we should always guard against script injection when server-side data is used to build an html string that gets evaluated in the dom.

Otherwise we may create openings for remote code execution exploits.

For example, someone could sign a transaction with some text containing javascript code and we for whatever reason it comes in via web socket to the explorer, including that value in an html string would leave to the script being executed.

Or maybe we start pulling in data from a 3rd party to get exchange rates and their server gets compromised. Just some hypotheticals, I'm not a security expert.

But now that we're loading jquery we have [$.parseHTML](https://api.jquery.com/jquery.parsehtml/) avail to turn strings into sanitized html.

chappjc:
This starts accumulating websocket connections.  If you go to the dev console (F12 -> console) you'll see a debug message about users online.  Two observations about that: it grows, and there are two ping messages going.

Also, it seems like the root page does not have a websocket because the green light is gone from the root page.  I'm actually seeing issues with this on explorer.dcrdata.org right now.  When you browse away from the root page, then click the browser's back button, it doesn't even show the connected/disconnected text or a light of any color.  Dunno what's going on here.

gozart1:
Both of those things are related to turbolinks.
Since the page does not reload, it means the socket connection stays open as you navigate between pages.

The handling of the web socket connection should be made global by moving logic that instantiates `ws` into the html-head of extras tmpl.

chappjc:
Ah, I thought that might be related.  This is great though... that the way we change pages lets the socket persist.

gozart1:
Yea, it's an interesting property of using turbolinks for sure.
To do this right we'll a way make ensure event that only the event handlers related to specific view the user is one run when a message comes in.

- One way would be to registers all events on every view change
- Another approach would be to let the handlers persist, but put filters in the view specific ones to do an early return if its not related to the current path the user is on. Eg
```
  if (!window.location.pathname.startsWith("/explorer/tx") { return }
  ... user is viewing tx, handle event accordingly here
}
```

chappjc:
@RogueElement Are you clear on what needs to be done to fix the issues that we are discussing with @gozart1?

RogueElement:
I think so, all that needs to be done is detect when turbolinks changes the view, close the current websocket connection then open the new connection meant for that view, right?

chappjc:
Reusing the websocket connection would probably be more efficient.
I haven't reviewed the architecture recently though.

gozart1:
2 suggestions:
- add  `data-turbolinks-permanent` as an attribute to the #connection span
https://github.com/dcrdata/dcrdata/blob/master/views/extras.tmpl#L130
this will prevent it from getting overwritten when new html loads via turbolinks.

- Remove ```var ws = createwebsocket()```
from the views and instead run it immediately after defining the function in extras.tmpl
https://github.com/dcrdata/dcrdata/pull/253/files#diff-cbc831f79556b345269731c1132a8412R167

Sidenote: Please use camelCase for naming all functions and data defined in JS. Reason being it's simply a convention, you won't find any popular javascript open source project _not_ using camelCase.

gozart1:
Since all the view are registering handlers on the same key "newblock" there's no need to guard for multiple handlers, since they get overwritten anyways as you change views.

If we do end up managing multiple keys, or have views that shouldn't handle any events, we can simply add a `clearHandlers()` method to MessageSocket that we can hook into routing events. That way we can keep a single connection open, and not have to worry about managing handlers.

But right now it's not really necessary since, each each view now has its own handler that overwrites any existing on since they all share the same key in the handlers hash.

views/tx.tmpl
```diff
@@ -4,7 +4,16 @@
...
 {{template "navbar"}}
-
+<script type="text/javascript">
+    ws.registerEvtHandler("newblock", function (event) {
+        console.log("tx Received newblock message", event);
+        var confirms = $(".confirmations")
+        if (isNaN(confirms.text())) {
+            confirms.text("0")
+        }
+        confirms.text(parseInt(confirms.text())+1)
```
chappjc: what's going on here?  I'm also confused why tx.tmpl has a `registerEvtHandler` running, but not address and block

views/extras.tmpl
```diff
@@ -165,6 +165,59 @@
             console.log("ping. users online: ", evt)
             ws.send("pong", "copy")
         });
+        var updateBlockData = function (event) {
+            console.log("Received newblock message", event);
+            var newBlock = JSON.parse(event);
+            var confirms = $(".confirmations")
+            $.each(confirms, function(i, v) {
```
gozart1: There's an edge case here where if a socket update is missed, the number of confirmations will get out of sync with the backend.

Maybe we could set the block height of when something was mined at as a data attribute eg `$('[data-confirmation-block-height]')` on the span that holds the confirmation value. 
Then we can derive confirmations by subtracting b.height from that value whenever new block comes in.

views/address.tmpl
```diff
@@ -115,8 +116,10 @@
                                 <span data-age="{{.Time}}"></span>
                             {{end}}
                             </td>
-                            <td>{{.Confirmations}}</td>
+                            <span data-confirmation-block-height="{{index $heights $i}}"></span>
```
gozart1: Instead of using a separate span element `data-confirmation-block-height="{{index $heights $i}}` should be set _directly_ on the element that wraps the value to be updated. This saves a bit of page weight and eliminates bug surface area.

views/extras.tmpl
```diff
@@ -117,6 +117,131 @@
...
+        });
+
+        ws.registerEvtHandler("ping", function(evt) {
+            console.log("ping. users online: ", evt)
+            ws.send("pong", "copy")
+        });
+        var updateBlockData = function (event) {
+            console.log("Received newblock message", event);
+            var newBlock = JSON.parse(event);
+            var confirms = $(".confirmations");
```
gozart1: If the data attributes with the confirmation heights are set on the elements holding the confirmations values themselves, the confirmation values can be updated in a single loop.

Check out the ages are updated  https://github.com/dcrdata/dcrdata/blob/master/views/extras.tmpl#L138

Should be possible to eliminate `initialBlocks` and `heights`

gozart1: Funny, I just noticed we could be passing unixTime in directly to humanize.timeSince in the ages updater, instead of calling `.data('age')` and second time. https://github.com/dcrdata/dcrdata/blob/master/views/extras.tmpl#L142

explorer/websocket.go
```diff
@@ -0,0 +1,148 @@
+package explorer
+
+import (
+	"sync"
+	"time"
+)
+
+// WebSocketMessage represents the JSON object used to send and received typed
```
chappjc: I think a follow-up PR should make a package for the reusable websocket stuff

views/extras.tmpl
```diff
@@ -126,7 +246,7 @@
         <a class="nav-item" href="https://github.com/dcrdata/dcrdata" title="dcrdata on GitHub" target="_blank">GitHub</a>
         <a class="nav-item" href="https://github.com/dcrdata/dcrdata#json-rest-api" title="API Endpoints" target="_blank">API</a>
         <a class="nav-item" href="/api/status" title="API Status" target="_blank" data-turbolinks="false">Status</a>
-        <span class="nav-item hidden" id="connection">Connecting to WebSocket...<div></div></span>
+        <span data-turbolinks-permanent class="nav-item hidden" id="connection">Connecting to WebSocket...<div></div></span>
```
chappjc: This no longer shows on any page.

chappjc: scratch that.  User error

chappjc:
@gozart1 If the changes you requested of @RogueElement are cool, I'm good to merge this.

gozart1:
Looks good to me! I tried opening a bunch of browser tabs and did various navigation scenarios and updates worked well.

Issue #262, opened 2017-11-03
Address pagination


chappjc:
The new address page that comes with the DB v2 PR (https://github.com/dcrdata/dcrdata/pull/209) can be paginated with URL query params, but there are no user controls to do this presently.  For example,

https://beta.dcrdata.org/explorer/address/Dcur2mcGjmENx4DhNqDctW5wJCVyT3Qeqkx<b>?n=8&start=666</b>

We just need the frontend elements.  Note that this is not necessarily as snappy as the block pagination since this can potentially be a tough SQL query.

This is a must for v1.0.0.

chappjc:
closed by https://github.com/dcrdata/dcrdata/pull/264

PR #264, opened 2017-11-05
add simple transactions pagination to address.tmpl

gozart1:
This adds some basic pagination UI to address.tmpl 
issue #262

screen shot 2017-11-05 at 1 11 47 pm
screen shot 2017-11-05 at 1 11 41 pm
screen shot 2017-11-05 at 12 48 32 pm
screen shot 2017-11-05 at 12 48 09 pm
screen shot 2017-11-05 at 1 08 13 pm

### DIFF
--- a/explorer/explorer.go
+++ b/explorer/explorer.go
@@ -230,6 +230,7 @@ func (exp *explorerUI) addressPage(w http.ResponseWriter, r *http.Request) {
 		addrData.Limit, addrData.Offset = limitN, offsetAddrOuts
 		addrData.KnownFundingTxns = balance.NumSpent + balance.NumUnspent
 		addrData.Balance = balance
+		addrData.Path = r.URL.Path
 		// still need []*AddressTx filled out and NumUnconfirmed
 
 		// Query database for transaction details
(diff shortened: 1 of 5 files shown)


gozart1:
Thanks for the feedback.
What do you think about the wording/styling?
Here's some ideas for styling
https://codepen.io/RRoberts/pen/jVOWEQ?q=pagination&limit=all&type=type-pens

I was planning to add page size controls in a separate PR, but could probably manage to add them into this one this evening if thats easier.

chappjc:
I'm not sure which option is best.  It depends on how we show "first" and "last", also how we deal with skipping way ahead if we have say 400 pages.

raedah:
Im noticing that the pagination is different then the blocks page. It may be useful for the user experience to unify and reuse these. Blocks page could probably use page size control as well. For skipping way ahead, could be totally bad ass to have a slider control.

gozart1:
added some page size controls and set default to 20 rows
screen shot 2017-11-05 at 10 45 40 pm
screen shot 2017-11-05 at 10 45 50 pm

chappjc:
The page size looks awesome!

Just two things:

1. The displayed range might need to be tweaked:

> 0 — 20 of 184,107 funding transactions.

I read this as "0 to 20 of ... displayed in the table below."   But with a page size of 2 and 0-based indexing, this is "0 — 19".

And when you click next you get:

> 20 — 40 of 184,108 funding transactions

So tx number 20 would be in both tables if these ranges are accurate.

2. When selecting a new page size from the drop down box, it doesn't load the new page.  You have to click next or previous.

gozart1:
Ok, good catch.
Solved this by adding +1 to the lower bound,
Eg. 1-20 => 21-40 => 41-60

Also, selecting a page size now reloads the current page.

chappjc:
Good stuff!  Going in.

Issue #211, opened 2017-10-04
Feature the coinbase transaction on the block page


chappjc:
Presently to find the coinbase transaction of a block, you have to look down at the regular transactions table and find the one that is 200 bytes.  It should be featured prominently above all the transaction tables, just below the block summary.

PR #269, opened 2017-11-07
Featuring the coinbase on the block page

McEdward:
resolving issue #211


McEdward:
screenshot 11

McEdward:
screenshot 13 I think this is a better representation @chappjc , or what do you think?

db/dcrsqlite/apisource.go
```diff
@@ -720,6 +720,13 @@ func makeExplorerTxBasic(data dcrjson.TxRawResult, msgTx *wire.MsgTx, params *ch
 	tx.FormattedSize = humanize.Bytes(uint64(len(data.Hex) / 2))
 	tx.Total = txhelpers.TotalVout(data.Vout).ToCoin()
 	tx.Fee, tx.FeeRate = txhelpers.TxFeeRate(msgTx)
+	for _, i := range data.Vin {
+		if i.IsCoinBase() == true {
+			tx.Coinbase = true
```
chappjc: I much prefer this approach.  Thanks.

db/dcrsqlite/apisource.go
```diff
@@ -720,6 +720,13 @@ func makeExplorerTxBasic(data dcrjson.TxRawResult, msgTx *wire.MsgTx, params *ch
...
+		} else {
+			tx.Coinbase = false
```
chappjc: Can omit the else part since false is default zero value for a boolean field

explorer/explorertypes.go
```diff
@@ -26,12 +26,13 @@ type BlockBasic struct {
...
-	Fee           dcrutil.Amount
-	FeeRate       dcrutil.Amount
-	VoteInfo      *VoteInfo
+	TxID          	string
+	FormattedSize 	string
+	Total         	float64
+	Fee           	dcrutil.Amount
+	FeeRate       	dcrutil.Amount
+	VoteInfo      	*VoteInfo
+	Coinbase		bool
```
chappjc: `go fmt ./...`

views/block.tmpl
```diff
@@ -155,6 +147,38 @@
...
+            <span class="anchor" id="coinbase"></span>
+            <div class="col-sm-12">
+                <h4><span>Coinbase</span></h4>
+                {{range .Tx}}
+                    {{if eq .Coinbase true}}
+                        <table class="table table-sm striped">
+                            <thead>
+                                <th>Transaction ID</th>
+                                <th class="text-right">Total DCR</th>
+                                <th class="text-right">Fee</th>
```
chappjc: Fee is always 0 DCR, so this column should be removed.

chappjc:
May want to enable "format on save" if your IDE supports it.

raedah:
The coinbase is still showing in the transaction list, which is a dup. It should be removed there.

db/dcrsqlite/apisource.go
chappjc: BTW, this entire if-else could also just be `tx.Coinbase = i.IsCoinBase()`

McEdward:
screenshot 15

raedah:
We should get the Size column aligned across all the headers, probably for @gozart1 .

McEdward:
screenshot 16

chappjc:
OK to merge this without fixing the Size column?

PR #273, opened 2017-11-13
tx page: Always show any decoded addresses, toggle OP_RETURN display.

chappjc:
If an output has an address, always display it, and if there is also OP_RETURN data, show a "toggle OP_RETURN" txt to display the data on click.
If an output has no decoded address, but OP_RETURN, show it like before.
Unspendable 0 value outputs marked as n/a in Spent column.

Addresses issues https://github.com/dcrdata/dcrdata/issues/272 and https://github.com/dcrdata/dcrdata/issues/248

A vote's outputs:

image

A ticket's outputs:

image

Then, after clicking the "toggle OP_RETURN" text:

image

### DIFF
--- a/public/css/main.css
+++ b/public/css/main.css
@@ -282,6 +282,21 @@ body.darkBG .page-item.disabled .page-link {
   table-layout: fixed;
 }
 
+.addressAndScriptData {
+  overflow: hidden;
+}
+.scriptDataStar {
+  z-index: 1;
+  text-align: center;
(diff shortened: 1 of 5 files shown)


chappjc:
@gozart1 your stripped jquery.min.js was missing the `slideToggle` function, so I added the minified version from jquery as is.  If you want to somehow fixup your version or strip this one, that might be good, but not needed immediately AFAIK.

Issue #228, opened 2017-10-09
Block view vote rows vote choice


raedah:
Currently data is being shown on one row, so only one vote item is being showed. Though, depending on what agendas are being voted on there may be multiple votes. Also not being shown is the default vote choice, Last Block Valid.

First stage: For now I suggest we just show just the Last Block Valid status on the vote row. All the agenda choices can still be seen by selecting the individual vote to view. 

Later stage: Display multiple hierarchical rows which show all the vote choices for a each of the votes. We are currently missing data needed to display this completely, such as the date range for a given agenda voting period which tells us if the agenda is still up for vote. Dependent on https://github.com/dcrdata/dcrdata/issues/210

raedah:
Wont do. This information will just be on the tx details page to see additional vote agenda information. This was because would have been duplicate information, it was more complex to do, and there was limited screen space to display it.

PR #274, opened 2017-11-16
Adding the last block valid to the vote row

McEdward:
Resolving the first stage of issue #228

### DIFF
--- a/views/block.tmpl
+++ b/views/block.tmpl
@@ -195,6 +195,7 @@
                         <thead>
                             <th>Transactions ID</th>
                             <th>Version</th>
+                            <th>Last Block Valid</th>
                             <th>Issue ID</th>
                             <th>Choice ID</th>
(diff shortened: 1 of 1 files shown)


views/block.tmpl
```diff
@@ -209,6 +210,9 @@
                                     </span>
                                 </td>
                                 <td class="mono fs15">{{.VoteInfo.Version}}</td>
+                                {{with .VoteInfo}}
+                                    <td>{{.Validation.Validity}}</td>
```
chappjc: Why not `.VoteInfo.Validation.Validity`?

McEdward: It was just a choice i made, but i think .VoteInfo.Validation.Validity would be a much cleaner way to write it. I'll change it.

chappjc: OK, it's just that in general I find it's good to use fewer lines if there's no reason not to

Issue #190, opened 2017-09-19
Find other formats available for raw transaction display


raedah:
Currently 'view raw' is actually linking to json output. We will want a way to display actual raw tx data (pre json conversion). This applies to blocks as well which also use 'view raw'.

getrawtransaction returns hex-encoded string, how is the hex converted into readable script?
Is there something similiar to getrawtransaction for blocks?

gozart1:
There's the Hexadecimal representation of the binary that gets sent over the wire 
Might be nice to include this, but not sure as to the best way to fit it in.

One option to not clutter things would be have a button to click to reveal it on the page, or link to it as a separate endpoint similar to how "view raw" currently links to the parsed transaction.

One neat feature the the BlockCypher's explorer has is the ability to paste in a transaction hex string and decode it. See https://live.blockcypher.com/bct/decodetx/

Here I just put it at the bottom:

<img width="724" alt="hex" src="https://user-images.githubusercontent.com/25571523/32763677-ea1aa638-c8b5-11e7-881e-7d2cefb0f407.png">

Note the hex strings can be quite big, here's the famous pizza transaction for 10k btc

```
[...]

Issue #136, opened 2017-09-08
Submit raw transaction page


raedah:
A public page where someone can submit raw transaction data for it to be broadcast out through the node.

PR #276, opened 2017-11-17
Added decode tx page and view raw link

RogueElement:
For issues #190 and #136 . The rpc api has no "getrawblock" function so that can't be added, but the view raw link works as intended now, showing the txid and hex while the view decoded link shows the json output. The decode tx pages works as intended as well but doesn't look so good for now.

### DIFF
--- a/explorer/explorer.go
+++ b/explorer/explorer.go
@@ -58,7 +58,7 @@ type explorerDataSourceLite interface {
 	GetBlockHash(idx int64) (string, error)
 	GetExplorerTx(txid string) *TxInfo
 	GetExplorerAddress(address string, count, offset int64) *AddressInfo
-	DecodeRawTransaction(hex string) *dcrjson.TxRawResult
+	DecodeRawTransaction(txhex string) *dcrjson.TxRawResult
 	GetHeight() int
 }
 
@@ -146,18 +146,21 @@ func (exp *explorerUI) rootWebsocket(w http.ResponseWriter, r *http.Request) {
 				if msg.EventId == "decodetx" {
 					log.Debug("Recieved decodetx signal for hex", msg.Messsage)
 					tx := exp.blockData.DecodeRawTransaction(msg.Messsage)
-					if tx == nil {
-						log.Debugf("Could not decode raw tx")
-						continue
-					}
-					buff := new(bytes.Buffer)
-					enc := json.NewEncoder(buff)
-					enc.SetIndent("", "    ")
-					enc.Encode(tx)
 					webData := WebSocketMessage{
-						EventId:  "decodedtx",
-						Messsage: buff.String(),
+						EventId: "decodedtx",
 					}
+
+					if tx != nil {
+						buff := new(bytes.Buffer)
+						enc := json.NewEncoder(buff)
+						enc.SetIndent("", "    ")
+						enc.Encode(tx)
+						webData.Messsage = buff.String()
+					} else {
+						log.Debugf("Could not decode raw tx")
+						webData.Messsage = fmt.Sprintf("Error: Could not decode hex %s", msg.Messsage)
+					}
+
 					err := websocket.JSON.Send(ws, webData)
 					if err != nil {
 						log.Debugf("Failed to encode WebSocketMessage decodedtx: %v", err)
@@ -165,7 +168,6 @@ func (exp *explorerUI) rootWebsocket(w http.ResponseWriter, r *http.Request) {
 						// the connection and quit.
 						return
 					}
-					log.Debugf("Sent decoded tx")
 				}
 			}
 		}()
--- a/explorer/explorer.go
+++ b/explorer/explorer.go
@@ -59,6 +59,7 @@ type explorerDataSourceLite interface {
 	GetExplorerTx(txid string) *TxInfo
 	GetExplorerAddress(address string, count, offset int64) *AddressInfo
 	DecodeRawTransaction(txhex string) *dcrjson.TxRawResult
+	SendRawTransaction(txhex string) string
 	GetHeight() int
 }
 
@@ -143,13 +144,12 @@ func (exp *explorerUI) rootWebsocket(w http.ResponseWriter, r *http.Request) {
 			for {
 				msg := &WebSocketMessage{}
 				websocket.JSON.Receive(ws, &msg)
-				if msg.EventId == "decodetx" {
-					log.Debug("Recieved decodetx signal for hex", msg.Messsage)
+				var webData WebSocketMessage
+				switch msg.EventId {
+				case "decodetx":
+					webData.EventId = "decodedtx"
+					log.Debug("Recieved decodetx signal for hex: ", msg.Messsage)
 					tx := exp.blockData.DecodeRawTransaction(msg.Messsage)
-					webData := WebSocketMessage{
-						EventId: "decodedtx",
-					}
-
 					if tx != nil {
 						buff := new(bytes.Buffer)
 						enc := json.NewEncoder(buff)
@@ -160,7 +160,17 @@ func (exp *explorerUI) rootWebsocket(w http.ResponseWriter, r *http.Request) {
 						log.Debugf("Could not decode raw tx")
 						webData.Messsage = fmt.Sprintf("Error: Could not decode hex %s", msg.Messsage)
 					}
-
+				case "sendtx":
+					webData.EventId = "senttx"
+					log.Debug("Recieved sendtx signal for hex: ", msg.Messsage)
+					txid := exp.blockData.SendRawTransaction(msg.Messsage)
+					if txid == "" {
+						webData.Messsage = fmt.Sprintf("Error: Could not send hex %s\nTry decoding it first to see if it's valid", msg.Messsage)
+					} else {
+						webData.Messsage = fmt.Sprintf("Transaction sent: %s", txid)
+					}
+				}
+				if webData.EventId != "" {
 					err := websocket.JSON.Send(ws, webData)
 					if err != nil {
 						log.Debugf("Failed to encode WebSocketMessage decodedtx: %v", err)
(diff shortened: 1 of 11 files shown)


gozart1:
Just pushed up a bit more styling for this https://github.com/dcrdata/dcrdata/tree/raw-tx
buttons

RogueElement:
@gozart1 thanks looks pretty good right now. @chappjc this is ready for merge.

gozart1:
Might be nice to pass the rejection reason out to the UI.
In the server log I see:
```
Rejected transaction 005969df2bc92ac33b97255521762ec4d69a27d6188e958c3246f01856d78259: transaction already exists
```
But the UI currently just shows 
```
Error: Could not send hex
Try decoding it first to see if it's valid
```

RogueElement:
Added that

chappjc:
We need to be cautious with accepting and processing external data.  Some thoughts:

- broadcast needs to be rate limited (per IP and server-wide)
- websocket input needs to somehow stop receiving data after X bytes (otherwise an attacker can send an infinite data stream, eating bandwidth and eventually filling up the server's memory)
- the html form should screen the data before sending, but an attacker would just ignore the form and send directly to the websocket
- the decode handler needs to have a final check on the size of the data before decoding it

explorer/explorer.go
```diff
@@ -21,6 +21,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/decred/dcrd/dcrjson"
+
```
chappjc: extra line

explorer/explorer.go
```diff
@@ -135,9 +140,50 @@ func (exp *explorerUI) rootWebsocket(w http.ResponseWriter, r *http.Request) {
 				exp.wsHub.HubRelay <- sigPingAndUserCount
 			}
 		}()
-
+		go func() {
+			for {
+				msg := &WebSocketMessage{}
+				websocket.JSON.Receive(ws, &msg)
```
chappjc: Here's where I'm concerned about someone sending a massive message.  I'm not sure how to limit this though.

explorer/explorer.go
```diff
@@ -135,9 +140,50 @@ func (exp *explorerUI) rootWebsocket(w http.ResponseWriter, r *http.Request) {
...
+				var webData WebSocketMessage
+				switch msg.EventId {
+				case "decodetx":
+					webData.EventId = "decodedtx"
+					log.Debug("Recieved decodetx signal for hex: ", msg.Messsage)
```
chappjc: *Received

explorer/explorer.go
```diff
@@ -135,9 +140,50 @@ func (exp *explorerUI) rootWebsocket(w http.ResponseWriter, r *http.Request) {
...
+					tx := exp.blockData.DecodeRawTransaction(msg.Messsage)
+					if tx != nil {
+						buff := new(bytes.Buffer)
+						enc := json.NewEncoder(buff)
+						enc.SetIndent("", "    ")
+						enc.Encode(tx)
```
chappjc: This is basically the internals of `MarshalIndent`.  Might as well do this, and check for encode error:

```go
				message, err := json.MarshalIndent(webData.Messsage, "", "    ")
				if err != nil {
					log.Warn("Invalid JSON message: ", err)
					continue
				}
				webData.Messsage = string(message)```

File: explorer/explorer.go [hunk 5]
```diff
@@ -135,9 +140,50 @@ func (exp *explorerUI) rootWebsocket(w http.ResponseWriter, r *http.Request) {
...
+						webData.Messsage = buff.String()
+					} else {
+						log.Debugf("Could not decode raw tx")
```
chappjc: decode error from RPC, as @gozart1 suggested

explorer/explorer.go
```diff
@@ -135,9 +140,50 @@ func (exp *explorerUI) rootWebsocket(w http.ResponseWriter, r *http.Request) {
...
+						webData.Messsage = fmt.Sprintf("Error: Could not decode hex %s", msg.Messsage)
+					}
+				case "sendtx":
+					webData.EventId = "senttx"
+					log.Debug("Recieved sendtx signal for hex: ", msg.Messsage)
+					txid, err := exp.blockData.SendRawTransaction(msg.Messsage)
```
chappjc: a final check on message length here is warranted IMO

explorer/explorer.go
```diff
@@ -135,9 +140,50 @@ func (exp *explorerUI) rootWebsocket(w http.ResponseWriter, r *http.Request) {
...
+					if err != nil {
+						webData.Messsage = fmt.Sprintf("Error: %v", err)
+					} else {
+						webData.Messsage = fmt.Sprintf("Transaction sent: %s", txid)
+					}
+				}
+				if webData.EventId != "" {
```
chappjc: It's a matter of style, but I prefer to minimize indent.  That is,

```go
if webData.EventID == "" {
    continue
}
```

However, this seems like something that should be in the `default:` section of the `switch`

explorer/explorer.go
```diff
@@ -135,9 +140,50 @@ func (exp *explorerUI) rootWebsocket(w http.ResponseWriter, r *http.Request) {
...
+						// If the send failed, the client is probably gone, so close
+						// the connection and quit.
+						return
+					}
+				}
+			}
+		}()
 	loop:
 		for {
+
```
chappjc: delete

explorer/explorer.go
RogueElement: How long can the message be?

RogueElement: Looking at https://godoc.org/golang.org/x/net/websocket#Codec.Receive, it seems the default max it can accept is 32MB, is that too much?

chappjc: Good find.  Since the block size is 1 MB, I think 32 MB for a tx is too big.  For whatever reason, we can just set this at 2 MB.

Also, we'll need to check the error code and deal with the next read containing the remaining data:

> The whole frame payload is read to an in-memory buffer; max size of payload is defined by `ws.MaxPayloadBytes`. If frame payload size exceeds limit, `ErrFrameTooLarge` is returned; in this case frame is not read off wire completely. The next call to Receive would read and discard leftover data of previous oversized frame before processing next frame. ```

RogueElement: I think what it says here ```The next call to Receive would read and discard leftover data of previous oversized frame before processing next frame. ``` means that we don't need to deal with the extra data ourselves as Receive would discard the leftover data before processing the next frame, but I could be wrong.

chappjc: Yeah, I think you're right.  The key is `before processing next frame`.

chappjc: Oh, I guess we can't change the 32MB value.  But if someone submits a tx larger than 1MB, it's definitely not OK to broadcast.

chappjc:
Rate limiting may not be needed as the node won't broadcast junk or duplicates, and txns cost $$.

chappjc:
Took a shot at abusing this.  My initial attempt was passing 460,000 valid hex characters to decode transaction.  It completely crippled dcrdata.  Local network request, just major CPU usage, forever.

I assume it's at least as costly to do this to send transaction.

chappjc:
I started with some valid transaction, like this one:

> 01000000020000000000000000000000000000000000000000000000000000000000000000ffffffff00ffffffffc0c343c6fb9ed34f26dd6a2bc7e233f22f82b4da4f363a7c5d437f7a555d30c60000000001ffffffff0400000000000000000000266a2463769d4b15965779ec9e542f888848426ec49dd80c0d175f3500000000000000b8df020000000000000000000000086a06010005000000a6e567000000000000001abb76a914a92c9ac541dd5ac40c630e6659ecce25e9fdd70e88aca53471d90100000000001abb76a91454ee8f3f4ceb3dfbdf38b3f55ec34c318c55b65c88ac0000000000000000028ffa46080000000000000000ffffffff020000bd1f92d10100000082d702000600000091483045022100f317885cfed85eb7355ea25a6fee0125b5fcd81afdcad91071a8e8ecc991f3ee02201683600dedd352d88a2ae0e11a5550ddfde68f87d53f9e072963f984f3bb8b1d014751210306559376b41006b6e16341b003c8d957cd3974ec665cea4e932826d9e7f1c7c72102f536ba2b34501eb3d3522c4397272d99803223087034a04f687ac6424c5aa8b652ae

Then I pasted that in the text box, holding down CTRL+V for a few seconds, then select all, and paste for another couple of seconds.  😆

chappjc:
We also need to limit the length of the hex logged: `Received decodetx signal for hex:  `.  
But what I was seeing is the process chugging away with multiple CPU cores _after_ printing the offending hex.

chappjc:
It seems easy to clog up the websocket.

```
2017-11-21 17:25:27.351 [ERR] SQLT: SendRawTransaction failed: -1: Rejected transaction 24a719667306c2284af9cb564be3178417bd4906da09746d3f73ed34df94b2be: transaction already exists
2017-11-21 17:25:27.351 [DBG] EXPR: Failed to encode WebSocketMessage decodedtx: write tcp 127.0.0.1:7777->127.0.0.1:44582: i/o timeout
2017-11-21 17:25:27.921 [DBG] EXPR: Failed to encode WebSocketMessage 3: write tcp 127.0.0.1:7777->127.0.0.1:44582: i/o timeout
2017/11/21 17:25:27 "GET http://127.0.0.1:7777/explorer/decodetx/ws HTTP/1.1" from 127.0.0.1:44582 - 000 0B in 1m48.000770061s
```

I'm just clicking both buttons like a maniac.

chappjc:
Going in the right direction, but it still chokes.  Try this:  https://pastebin.com/raw/FRAfMiqb

chappjc:
@RogueElement nice job with this.  I'm just being picky on this because of the attack surface it exposes.

PR #286, opened 2017-11-24
explorer websocket: check error on websocket.JSON.Receive, fix client ping

chappjc:
Addresses issue https://github.com/dcrdata/dcrdata/issues/285.

- Set server's response event IDs to the incoming message's event ID + "Resp".
- Client side ping (in JS) was not sending a JSON response, which the server's websocket handler did not appreciate.  Fix this by using the same send function (i.e. `function(eventID, message)`) as the other messages.
- Recognize the "ping" event ID in server's websocket read loop.
- Change client side ping from every 1 sec to 7 seconds.
- Create websocket read timeout, set to 12 seconds. Write is 10 seconds.
- Set both read and write deadlines where appropriate.
- Docs.
- Fix reloading of explorer templates with SIGUSR1.
- Add pulse effect to rawtx result box to ack click/keypress.

### DIFF
--- a/explorer/explorer.go
+++ b/explorer/explorer.go
@@ -145,7 +145,7 @@ func (exp *explorerUI) rootWebsocket(w http.ResponseWriter, r *http.Request) {
 			}
 		}()
 
-		// Start listening for websocket messages from clinet with raw
+		// Start listening for websocket messages from client with raw
 		// transaction bytes (hex encoded) to decode or broadcast.
 		go func() {
 			for {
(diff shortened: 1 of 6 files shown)


explorer/explorer.go
```diff
@@ -138,19 +138,30 @@ func (exp *explorerUI) rootWebsocket(w http.ResponseWriter, r *http.Request) {
...
 		defer ticker.Stop()
 
+		// Periodically ping clients over websocket connection
 		go func() {
 			for range ticker.C {
 				exp.wsHub.HubRelay <- sigPingAndUserCount
 			}
 		}()
+
+		// Start listening for websocket messages from clinet with raw
```
RogueElement: *client

RogueElement:
Look good. Great job.

PR #288, opened 2017-11-24
explorer: make int64Comma generic

chappjc:
Change the template function `int64Comma(v int64)` into `intComma(v interface{})` using a type switch to accept all integer types.

Use `intComma` to give the block size text a hover text for the comma-formatted bytes, as show below:

image

(the mouse pointer isn't shown, but it's hovering over the size text)

### DIFF
--- a/explorer/explorer.go
+++ b/explorer/explorer.go
@@ -629,8 +629,27 @@ func New(dataSource explorerDataSourceLite, primaryDataSource explorerDataSource
 			p := (float64(a) / float64(b)) * 100
 			return p
 		},
-		"int64Comma": func(v int64) string {
-			return humanize.Comma(v)
+		"intComma": func(v interface{}) string {
+			var vi64 int64
+			switch vt := v.(type) {
+			case int64:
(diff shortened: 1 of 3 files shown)


gozart1:
Maybe more explicit & reusable to add type coercion as separate helpers?
Eg move the switch into a helper function called`int64` and use it to normalize inputs to our  current formatter functions like so:
```{{int64Comma (int64 .Size)}} ```

**Comment by chappjc** [comment]
@gozart1 That's a good idea.  Although I think we can get reusability and the simplicity at the template function call sites.  Here's what I'm trying now:

```go
		toInt64 := func(v interface{}) int64 {
			switch vt := v.(type) {
			...
			}
		}
...
		"int64": toInt64,
		"intComma": func(v interface{}) string {
			return humanize.Comma(toInt64(v))
		},
		"int64Comma": func(v int64) string {
			return humanize.Comma(v)
		},
```

So you can use `int64` and `int64Comma` if you want, but also just `intComma`.

Issue #261, opened 2017-11-03
Search error has fixed url and does not allow for refresh


raedah:
Upon doing a failed search, the user is returned to a url that looks like this
https://explorer.dcrdata.org/error/.......
When the user does a page refresh to check if the address or hash is now valid, they will always get the same error page regardless of if new data is available.
The expected behavior is for a page refresh to trigger the search again.

ugwuchuks:
/home/chucks/Downloads/dcrdata-address-cache-correct error/error.go

raedah:
Bug in this, search is often not working on the first attempt, returning back to the same page. Usually works on the second attempt. For the current red highlighting, it does no work for user with javascript disabled, and it also does not clearly state to the user what is happening, "A transaction with id xxx could not be found." For the page that is displaying that error, it should use web sockets to refresh as soon as the transaction does become available. This should be applied to transactions and all other searchable types.

RogueElement:
The first attempt bug is probably due to the search occurring before the websocket connection is fully established. I dont know where to put the error message it should show which is why I just went with the red highlight. I can probably try something for the no javascript case like falling back to the url `/search`.

Issue #282, opened 2017-11-19
removing /explorer/ from the url


raedah:
Having the extra word in the url is redundant and makes the url longer.
https://explorer.dcrdata.org/explorer/block/187492
https://explorer.dcrdata.org/block/187492   would be better.
https://dcrdata.org/block/187492   would be even shorter.

xaur:
Also please https://explorer.dcrdata.org/explorer -> https://explorer.dcrdata.org/blocks

RogueElement:
Would need to merge explorer and the root page for this to work.

raedah:
I like @xaur's idea of making /explorer into /blocks , more descriptive of the element that is actually being shown. Id like dcrdata.org to be identical to explorer.dcrdata.org instead of forwarding to it so that sharing urls would stay compact.

chappjc:
- I agree that /explorer, which currently shows a list of blocks, should be at /blocks.
- Root page is good as the root page.
- A specific block's page should go from /explorer/block/190365 to /blocks/190365.
- The other pages are good, just without the /explorer prefix.

xaur:
If you plan to deploy other apps on dcrdata.org domain, better keep explorer subdomain.

xaur:
There is a consistency concern regarding singular vs plural form.

Popular explorers use singular form:

https://blockchain.info/block/0000000000000000007f6c84ee3f61cb0d897a928cf29c590645bdef447d0d6f
https://blockchain.info/tx/b1606386935bdad97b017c215ce9a7c35dd4e25fc26de7a5c0a75680c5337070
https://blockchain.info/address/18cBEMRxXHqzWWCxZNtU91F5sbUNKhL5PX

https://mainnet.decred.org/address/DsUBWJgJheWPQynZ4nqRXTTPvVTx3aDkPci
https://mainnet.decred.org/tx/98580b8a092b8258b6e0f8c90b89f402ca58b65730709130ddb2398b70806fb7
https://mainnet.decred.org/address/DsmM7jVwRsBk21rLK4QyUywiNDALZfk54G7

https://www.coinprism.info/block/0000000000000000002dc1a21e0c24645a434b49aed634008ab362c4c74ebbd2
https://www.coinprism.info/tx/9232d9e3cac93cf34f5f9f34293fff32c6188bcb1269880cc8f5d4cb999b4c84
https://www.coinprism.info/address/19y9cs7ffYLpJ15Y299KZXhsRWc8gzfzgC

[...]

PR #291, opened 2017-11-28
Added error pages and reorganized explorer URL paths

RogueElement:
Should address #261.


explorer/explorerroutes.go
```diff
@@ -0,0 +1,304 @@
...
+	str, err := templateExecToString(exp.templates[errorTemplateIndex], "error", struct {
+		ErrorCode   string
+		ErrorString string
+	}{
+		code,
+		message,
+	})
+	if err != nil {
+		log.Errorf("Template execute failure: %v", err)
+		str = "Something went very wrong if you can see this, try refreshing"
```
chappjc: The HTTP error code in this situation should be a 500, not 404.

explorer/websockethandlers.go
```diff
@@ -0,0 +1,162 @@
+package explorer
```
chappjc: Let's add our standard copyright notice on new files going forward.

explorer/explorerroutes.go
```diff
@@ -0,0 +1,304 @@
...
+	// then redirect to the address page if it is
+	address := exp.blockData.GetExplorerAddress(searchStr, 1, 0)
+	if address != nil {
+		http.Redirect(w, r, "/explorer/address/"+searchStr, http.StatusPermanentRedirect)
+		return
+	}
+
+	// Check if the value is a valid hash
+	if _, err = chainhash.NewHashFromStr(searchStr); err != nil {
+		http.Redirect(w, r, "/error/"+searchStr, http.StatusTemporaryRedirect)
```
chappjc: Breaks the websocket.  Did you mean to send this to /explore/error?

chappjc:
I'm also finding that very often when performing a search, instead of getting the response from the search handler, I see the exact same page I was just looking at, but with `?search=blah` appended.

explorer/explorerroutes.go
```diff
@@ -299,6 +289,6 @@ func (exp *explorerUI) errorPage(w http.ResponseWriter, code string, message str
 		str = "Something went very wrong if you can see this, try refreshing"
 	}
 	w.Header().Set("Content-Type", "text/html")
-	w.WriteHeader(http.StatusNotFound)
```
chappjc: Oh, I meant 500 only when the template execution fails.  It can be 404 when the user searches for a non-existent item.

chappjc:
This is up on beta.dcrdata.org.

Search seems to work over the websocket, but only when there is a valid search result.  When you search for "asdf" you see an error in the dcrdata logs, but nothing happens in the browser.

RogueElement:
I haven't added the error handling to the browser side yet, want to merge the home page and explorer first so searching on any page works (searching on the home page doesn't work yet).

RogueElement:
Added commits to address #282

explorer/explorer.go
```diff
@@ -236,6 +253,10 @@ func New(dataSource explorerDataSourceLite, primaryDataSource explorerDataSource
 		"int64Comma": func(v int64) string {
 			return humanize.Comma(v)
 		},
+		"ticketWindowProgress": func(i int) float64 {
+			p := (float64(i) / 144) * 100
```
chappjc: 144 needs to come from chaincfg.Params

explorer/explorer.go
```diff
@@ -370,6 +400,18 @@ func (exp *explorerUI) Store(blockData *blockdata.BlockData, _ *wire.MsgBlock) e
...
+		CoinSupply:       blockData.ExtraInfo.CoinSupply,
+		StakeDiff:        blockData.CurrentStakeDiff.CurrentStakeDifficulty,
+		IdxBlockInWindow: blockData.IdxBlockInWindow,
+		Difficulty:       blockData.Header.Difficulty,
+		NBlockSubsidy: BlockSubsidy{
+			Dev:   blockData.ExtraInfo.NextBlockSubsidy.Developer,
+			PoS:   blockData.ExtraInfo.NextBlockSubsidy.PoS,
+			PoW:   blockData.ExtraInfo.NextBlockSubsidy.PoW,
+			Total: blockData.ExtraInfo.NextBlockSubsidy.Total,
+		},
```
chappjc: If needed, perhaps here is where the chaincfg params stuff can go

explorer/explorertypes.go
```diff
@@ -157,6 +159,29 @@ type AddressBalance struct {
...
+
+// BlockSubsidy
+type BlockSubsidy struct {
+	Total int64 `json:"total"`
+	PoW   int64 `json:"pow"`
+	PoS   int64 `json:"pos"`
+	Dev   int64 `json:"dev"`
+}
+
+// Mempool
```
chappjc: MempoolInfo ...

views/extras.tmpl
```diff
@@ -125,12 +125,9 @@
         el.html(msg + '<div></div>')
     }
 
-
-
     function getSocketURI(loc) {
         var protocol = (loc.protocol === 'https:') ? 'wss' : 'ws'
-        var route = (loc.pathname === "/") ? "/ws" : loc.pathname + '/ws'
-        return protocol + "://" + loc.host + route
+        return protocol + "://" + loc.host + '/ws'
```
chappjc: The is by far my favourite change.

views/extras.tmpl
```diff
@@ -205,27 +202,29 @@
...
-                $("#bsubsidy_pos").html(humanize.decimalParts((b.next_block_subsidy.pos/500000000), false, 8))
-                $("#bsubsidy_dev").html(humanize.decimalParts(b.next_block_subsidy.developer/100000000, false, 8))
-                $("#coin_supply").html(humanize.decimalParts(b.coin_supply/100000000, true, 8))
-                $("#blocksdiff").html(humanize.decimalParts(b.sdiff, false, 8))
-                $("#window_block_index").text(newBlock.stake.window_block_index)
-                $("#pos-window-progess-bar").css({ width: (newBlock.stake.window_block_index/144)*100 + "%" })
+                $("#difficulty").html(humanize.decimalParts(ex.difficulty/100000000, false, 8))
+                $("#bsubsidy_total").html(humanize.decimalParts(ex.subsidy.total/100000000, false, 8))
+                $("#bsubsidy_pow").html(humanize.decimalParts(ex.subsidy.pow/100000000, false, 8))
+                $("#bsubsidy_pos").html(humanize.decimalParts((ex.subsidy.pos/500000000), false, 8))
```
chappjc: I'm OK with assuming 5 votes per block as it's that way on all networks, but perhaps a comment here.

views/extras.tmpl
```diff
@@ -205,27 +202,29 @@
...
+                $("#bsubsidy_dev").html(humanize.decimalParts(ex.subsidy.dev/100000000, false, 8))
+                $("#coin_supply").html(humanize.decimalParts(ex.coin_supply/100000000, true, 8))
+                $("#blocksdiff").html(humanize.decimalParts(ex.sdiff, false, 8))
+                $("#window_block_index").text(ex.window_idx)
+                $("#pos-window-progess-bar").css({ width: (ex.window_idx/144)*100 + "%" })
```
chappjc: The 144 on the other hand needs to come from a variable.

views/extras.tmpl
```diff
@@ -262,7 +259,7 @@
     <div class="container wrapper text-center">
         <a class="nav-item" href="https://github.com/dcrdata/dcrdata" title="dcrdata on GitHub" target="_blank">GitHub</a>
         <a class="nav-item" href="https://github.com/dcrdata/dcrdata#json-rest-api" title="API Endpoints" target="_blank">API</a>
-        <a class="nav-item" href="/explorer/decodetx" title="Decode or send a raw transaction">Decode Tx</a>
+        <a class="nav-item" href="//decodetx" title="Decode or send a raw transaction">Decode Tx</a>
```
chappjc: extra slash

explorer/explorer.go
gozart1: 144 is also hardcoded in [root.tmpl](https://github.com/dcrdata/dcrdata/blob/a51c9e3d5914a13457aa0161e2f082a8d201c1b3/views/root.tmpl#L49) and in [extras.tmpl](https://github.com/dcrdata/dcrdata/blob/f865c40de6d833b6c5f461937058fe0826db5fda/views/extras.tmpl#L218). Somehow the `updateBlockData` js function needs to know the params, setting it as a data attribute in the html seems the simplest approach.

chappjc:
This is looking really good. Just need a rebase.

explorer/explorer.go
RogueElement: Updated them in [extras.tmpl](https://github.com/RogueElement/dcrdata/blob/error-page/views/extras.tmpl#L218) and [home.tmpl](https://github.com/RogueElement/dcrdata/blob/error-page/views/home.tmpl#L47)

explorer/explorer.go
```diff
@@ -83,437 +83,20 @@ type explorerUI struct {
 	wsHub           *WebsocketHub
 	NewBlockDataMtx sync.RWMutex
 	NewBlockData    BlockBasic
+	ExtraInfo       HomeInfo
+	MempoolData     MempoolInfo
+	ChainParams     chaincfg.Params
```
chappjc: This can/should be a pointer.

explorer/explorer.go
```diff
@@ -644,6 +241,10 @@ func New(dataSource explorerDataSourceLite, primaryDataSource explorerDataSource
 			val := a - b
 			return val
 		},
+		"divide": func(n int64, d int64) int64 {
+			val := n / d
+			return val
```
chappjc: please `return n / d`

explorer/explorer.go
```diff
@@ -659,6 +260,10 @@ func New(dataSource explorerDataSourceLite, primaryDataSource explorerDataSource
 		"int64Comma": func(v int64) string {
 			return humanize.Comma(v)
 		},
+		"ticketWindowProgress": func(i int) float64 {
+			p := (int64(i) / exp.ChainParams.StakeDiffWindowSize) * 100
```
chappjc: The conversions here are dubious.  Why divide in integer, (and does it produce the right result)?

explorer/explorer.go
```diff
@@ -793,41 +431,34 @@ func (exp *explorerUI) Store(blockData *blockdata.BlockData, _ *wire.MsgBlock) e
...
-			rd.Get("/ws", exp.rootWebsocket)
-		})
-	})
-	exp.Mux.Route("/decodetx", func(r chi.Router) {
-		r.Get("/", exp.decodeTxPage)
-		r.Get("/ws", exp.rootWebsocket)
-	})
-
-	exp.Mux.With(searchPathCtx).Get("/search/{search}", exp.search)
+	redirect := func(url string) func(http.ResponseWriter, *http.Request) {
```
chappjc: a style thing since the underlying types are the same, but I'd like the return as `http.HandlerFunc` since that's what `chi.(*Mux).Get` takes, plus it's less repetition of the function signature here.

chappjc:
Ah, I actually meant that a pointer for `chaincfg.Params` in `explorerUI`.  All three of the parameters sets defined in chaincfg/params.go will never be changed at runtime, but they are large.

explorer/explorerroutes.go
```diff
@@ -0,0 +1,341 @@
...
+		if err == nil {
+			return "/explorer/block/" + searchStr
+		}
+	}
+
+	// Call GetExplorerAddress to see if the value is an address hash and
+	// then redirect to the address page if it is
+	address := exp.blockData.GetExplorerAddress(searchStr, 1, 0)
+	if address != nil {
+		return "/explorer/address/" + searchStr
```
chappjc: `/explorer` is used throughout this function

chappjc: There are also instances of `/explorer` in extras.tmpl and tx.tmpl

explorer/explorer.go
```diff
@@ -793,41 +430,34 @@ func (exp *explorerUI) Store(blockData *blockdata.BlockData, _ *wire.MsgBlock) e
...
-	exp.Mux.Route("/decodetx", func(r chi.Router) {
-		r.Get("/", exp.decodeTxPage)
-		r.Get("/ws", exp.rootWebsocket)
-	})
-
-	exp.Mux.With(searchPathCtx).Get("/search/{search}", exp.search)
+	redirect := func(url string) http.HandlerFunc {
+		return func(w http.ResponseWriter, r *http.Request) {
+			x := chi.URLParam(r, "x")
+			http.Redirect(w, r, "/"+url+"/"+x, http.StatusPermanentRedirect)
```
chappjc: `if x == ""` then do not use trailing slash

chappjc:
Running 635357a6 on beta.dcrdata.org now and I noticed some issues (in comments above)

PR #292, opened 2017-11-29
change date format to yyyy-mm-dd

chappjc:
To address issue https://github.com/dcrdata/dcrdata/issues/283


chappjc:
@raedah we cool with this format?  Personally I think it's very good.

raedah:
looks good

PR #301, opened 2017-12-16
Address QR code [WIP]

gozart1:
This adds QR code support for address pages, which is handy for mobile wallets.

qr

This is currently just a quick proof of concept. Implementation considerations:
- Generate the image on server or client? ( currently server-side )
- Show by default or hide behind a link/toggle? ( currently shown by default )

I'm leaning towards the current server-side implementation, so as to not load extra js ( smallest qr code lib out there is ~20kb minified ). There is some overhead in generating the image so ideally only the users who actually use this should incur that overhead. Maybe we can just have link to "generate qr code" which reloads the page with "?qr=1" appended to the url to gate the logic.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -151,6 +151,12 @@
 
 [[projects]]
   branch = "master"
+  name = "github.com/skip2/go-qrcode"
+  packages = [".","bitset","reedsolomon"]
+  revision = "80895b4a9eff675bc0f8ea09f60106bca9ac44c8"
+
+[[projects]]
+  branch = "master"
   name = "golang.org/x/crypto"
   packages = ["nacl/secretbox","pbkdf2","poly1305","ripemd160","salsa20/salsa","scrypt"]
   revision = "9419663f5a44be8b34ca85f08abc5fe1be11f8a3"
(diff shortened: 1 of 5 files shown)


chappjc:
This is really nice looking.

I think we can start with server generated images, if: (1) we just have some golang benchmarks to give us an idea of us/op, and (2) we don't generate the code by default (need query).

Having `?qr=1` like you suggested would be good, except that crawlers would just load that up if it were a regular link.  On the other hand, if we put it in JS, I'm not sure if that would degrade gracefully, but that's probably fine for the QR code.  Anyway, I think it would be a better UX if a click brought it up in the same window as an overlay or whatever, like lightbox.

chappjc:
Actually, an API endpoint that returns the QR (binary PNG data, not JSON) would be valuable for a number of reasons, so we should do that regardless.

raedah:
http://chart.apis.google.com/chart?cht=qr&chs=300x300&chl=test&chld=H|0

raedah:
Id use a small qr icon, which when clicked either expands and loads a new div, or zooms to be large front and center. Would only load on demand. There could be graceful fallback if JS is disabled, such as pointing to the QR image url directly.

chappjc:
Superseded by https://github.com/dcrdata/dcrdata/pull/312

Issue #255, opened 2017-10-26
Searching for address with no transaction should not be error


chappjc:
When searching for an address that has not had any transaction should still take you to the address page, not the /error page.

In fact, it would be nice if the address page would use a websocket to update when/if a transaction comes in for that address.

RogueElement:
I could work this into #253

chappjc:
@RogueElement Nah, let's do one page at a time.  Go ahead with 253 and we'll get to this.

About the error, I noticed that I somehow managed to pass a whitespace character all the way to the DecodeAddress part of the code.  We should strip all leading and trailing whitespaces from searches.

McEdward:
@chappjc , please could you give an example of addresses that has not had any transaction.

chappjc:
For example, you can use `getnewaddress` with to get a fresh address from your wallet.  The address itself is valid (would pass through the address validation functions since the network and checksum are OK), but there are no tx yet.  Ideally, we would respond with 0.0 DCR, no transactions, etc. but stay on the tx page so the websocket can update it when a relevant tx comes in.

raedah:
More needs to be done here. Currently, searching still first takes the user to
https://explorer.dcrdata.org/search?search=ADDRESS
and then after at that URL they doing another search will redirect them to
https://explorer.dcrdata.org/address/ADDRESS
They should end up at the /address/ url the first time.
I also see the url alternating on repeated searches.

The other issue is that they are currently getting this returned
"Error: Something went wrong...
and it's not your fault, try refreshing... that usually fixes things"
They should instead end up at a the regular /address/ page which should be properly displaying 0 transactions. Make sure the template displays in the same format as our other pages for when zero items are being returned.

PR #243, opened 2017-10-20
Summary at the top of the address page

McEdward:
Resolving issue #121

PR #302, opened 2017-12-19
Searching for an address with no tx/websocket implementation on address page

McEdward:
For #255, Still working on how to implement websocket to update when/if a transaction comes in for that address.

### DIFF
--- a/db/dcrsqlite/apisource.go
+++ b/db/dcrsqlite/apisource.go
@@ -1050,16 +1050,7 @@ func (db *wiredDB) GetExplorerAddress(address string, count, offset int64) *expl
 	if err != nil {
 		log.Warnf("GetAddressTransactionsRaw failed for address %s: %v", addr, err)
 		return &explorer.AddressInfo{
-			Address:         address,
-			Limit:           0,
-			Offset:          0,
-			Transactions:    nil,
-			NumFundingTxns:  0,
-			NumSpendingTxns: 0,
-			NumUnconfirmed:  0,
-			TotalReceived:   0,
-			TotalSent:       0,
-			Unspent:         0,
+			Address: address,
 		}
 	}
 
--- a/db/dcrsqlite/apisource.go
+++ b/db/dcrsqlite/apisource.go
@@ -1049,6 +1049,11 @@ func (db *wiredDB) GetExplorerAddress(address string, count, offset int64) *expl
 		int(offset), int(count), true, true, nil)
 	if err.Error() == "-32603: No Txns available" {
 		log.Warnf("GetAddressTransactionsRaw failed for address %s: %v", addr, err)
+
+		if !ValidateNetworkAddress(addr, db.params) {
+			log.Warnf("Address %s is not valid for this network", address)
+			return nil
+		}
 		return &explorer.AddressInfo{
 			Address: address,
 		}
@@ -1101,4 +1106,7 @@ func (db *wiredDB) GetExplorerAddress(address string, count, offset int64) *expl
 		TotalSent:       totalsent,
 		Unspent:         totalreceived - totalsent,
 	}
+}
+func ValidateNetworkAddress(address dcrutil.Address, p *chaincfg.Params) bool {
+	return address.IsForNet(p)
 }
--- a/db/dcrsqlite/apisource.go
+++ b/db/dcrsqlite/apisource.go
@@ -1098,7 +1098,12 @@ func (db *wiredDB) GetExplorerAddress(address string, count, offset int64) *expl
 		}
 	}
 	numberMaxOfTx := int64(len(txs))
-
+	var numFundingTxns int64
+	if numberMaxOfTx < count {
+		numFundingTxns = numberMaxOfTx
+	} else {
+		numFundingTxns = count
+	}
 	balance := &explorer.AddressBalance{
 		Address:      address,
 		NumSpent:     numSpending,
@@ -1111,7 +1116,7 @@ func (db *wiredDB) GetExplorerAddress(address string, count, offset int64) *expl
 		Limit:            count,
 		Offset:           offset,
 		Transactions:     addressTxs,
-		NumFundingTxns:   count,
+		NumFundingTxns:   numFundingTxns,
 		KnownFundingTxns: numberMaxOfTx,
 		NumSpendingTxns:  numSpending,
 		NumUnconfirmed:   numUnconfirmed,
--- a/db/dcrsqlite/apisource.go
+++ b/db/dcrsqlite/apisource.go
@@ -1098,11 +1098,9 @@ func (db *wiredDB) GetExplorerAddress(address string, count, offset int64) *expl
 		}
 	}
 	numberMaxOfTx := int64(len(txs))
-	var numFundingTxns int64
+	var numFundingTxns = count
 	if numberMaxOfTx < count {
 		numFundingTxns = numberMaxOfTx
-	} else {
-		numFundingTxns = count
 	}
 	balance := &explorer.AddressBalance{
 		Address:      address,
--- a/db/dcrsqlite/apisource.go
+++ b/db/dcrsqlite/apisource.go
@@ -1047,11 +1047,14 @@ func (db *wiredDB) GetExplorerAddress(address string, count, offset int64) *expl
 
 	txs, err := db.client.SearchRawTransactionsVerbose(addr,
 		int(offset), int(count), true, true, nil)
-	if err != nil {
+	if err.Error() == "-32603: No Txns available" {
 		log.Warnf("GetAddressTransactionsRaw failed for address %s: %v", addr, err)
 		return &explorer.AddressInfo{
 			Address: address,
 		}
+	} else if err != nil {
+		log.Warnf("GetAddressTransactionsRaw failed for address %s: %v", addr, err)
+		return nil
 	}
 
 	addressTxs := make([]*explorer.AddressTx, 0, len(txs))
--- a/db/dcrsqlite/apisource.go
+++ b/db/dcrsqlite/apisource.go
@@ -1045,8 +1045,11 @@ func (db *wiredDB) GetExplorerAddress(address string, count, offset int64) *expl
 		return nil
 	}
 
+	maxcount := explorer.MaxAddressRows
 	txs, err := db.client.SearchRawTransactionsVerbose(addr,
 		int(offset), int(count), true, true, nil)
+	maxLimitOfTxs, err := db.client.SearchRawTransactionsVerbose(addr,
+		int(offset), int(maxcount), true, true, nil)
 	if err != nil && err.Error() == "-32603: No Txns available" {
 		log.Warnf("GetAddressTransactionsRaw failed for address %s: %v", addr, err)
 
@@ -1094,6 +1097,7 @@ func (db *wiredDB) GetExplorerAddress(address string, count, offset int64) *expl
 		}
 	}
 	numberOfTx := int64(len(txs))
+	numberMaxOfTx := int64(len(maxLimitOfTxs))
 
 	balance := &explorer.AddressBalance{
 		Address:      address,
@@ -1107,9 +1111,8 @@ func (db *wiredDB) GetExplorerAddress(address string, count, offset int64) *expl
 		Limit:            count,
 		Offset:           offset,
 		Transactions:     addressTxs,
-		NumFundingTxns:   numReceiving,
-		NumTransactions:  numberOfTx,
-		KnownFundingTxns: numberOfTx,
+		NumFundingTxns:   numberOfTx,
+		KnownFundingTxns: numberMaxOfTx,
 		NumSpendingTxns:  numSpending,
 		NumUnconfirmed:   numUnconfirmed,
 		TotalReceived:    totalreceived,
--- a/db/dcrsqlite/apisource.go
+++ b/db/dcrsqlite/apisource.go
@@ -1097,7 +1097,7 @@ func (db *wiredDB) GetExplorerAddress(address string, count, offset int64) *expl
 	balance := &explorer.AddressBalance{
 		Address:      address,
 		NumSpent:     numSpending,
-		NumUnspent:   int64(numReceiving),
+		NumUnspent:   numReceiving,
 		TotalSpent:   int64(totalsent),
 		TotalUnspent: int64(totalreceived - totalsent),
 	}
--- a/db/dcrsqlite/apisource.go
+++ b/db/dcrsqlite/apisource.go
@@ -1047,8 +1047,6 @@ func (db *wiredDB) GetExplorerAddress(address string, count, offset int64) *expl
 
 	maxcount := explorer.MaxAddressRows
 	txs, err := db.client.SearchRawTransactionsVerbose(addr,
-		int(offset), int(count), true, true, nil)
-	maxLimitOfTxs, err := db.client.SearchRawTransactionsVerbose(addr,
 		int(offset), int(maxcount), true, true, nil)
 	if err != nil && err.Error() == "-32603: No Txns available" {
 		log.Warnf("GetAddressTransactionsRaw failed for address %s: %v", addr, err)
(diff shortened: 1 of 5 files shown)


views/address.tmpl
```diff
@@ -48,6 +48,10 @@
                             {{template "decimalParts" (amountAsDecimalParts .Balance.TotalSpent true)}}<span class="pl-1 unit">DCR</span>
                         </td>
                     </tr>
+                    {{else}}
+                    <td class="mono nowrap text-right">
+                        <span class="pl-1 unit">No Transactions are availabe in this address yet</span>
+                    </td>
```
chappjc: How about this instead:
```html
                    <tr>
                        <td class="text-right pt-2 pr-1">FRESH ADDRESS</td>
                    </tr>
                    <tr>
                        <td class="text-right pr-1">This address has not been observed in any transactions.</td>
                    </tr>
```

db/dcrsqlite/apisource.go
```diff
@@ -1049,7 +1049,9 @@ func (db *wiredDB) GetExplorerAddress(address string, count, offset int64) *expl
 		int(offset), int(count), true, true, nil)
 	if err != nil {
 		log.Warnf("GetAddressTransactionsRaw failed for address %s: %v", addr, err)
-		return nil
+		return &explorer.AddressInfo{
+			Address: address,
```
chappjc: valid addresses for the wrong network will make it to here.  You can use `addr.IsForNet(db.params)` to check, but you might want to make a "ValidateNetworkAddress(string, *chaincfg.Params)` function or something like it.  Otherwise you don't really know why `GetExplorerAddress` returned `nil`

McEdward: What are the major distinctions for detection of a valid network address? for the function ValidateNetworkAddress(string, *chaincfg.Params)

chappjc: Each network had its own values for chaincfg.Params

McEdward: Yes, I mostly meant. where can i find the decred specific values for chaincfg.Params

db/dcrsqlite/apisource.go
```diff
@@ -1047,11 +1047,14 @@ func (db *wiredDB) GetExplorerAddress(address string, count, offset int64) *expl
...
 		int(offset), int(count), true, true, nil)
-	if err != nil {
+	if err.Error() == "-32603: No Txns available" {
 		log.Warnf("GetAddressTransactionsRaw failed for address %s: %v", addr, err)
 		return &explorer.AddressInfo{
 			Address: address,
 		}
+	} else if err != nil {
+		log.Warnf("GetAddressTransactionsRaw failed for address %s: %v", addr, err)
+		return nil
```
chappjc: `return err`

db/dcrsqlite/apisource.go
chappjc: By network I mean the 3 Decred networks: mainnet, testnet, and simnet:  https://godoc.org/github.com/decred/dcrwallet/netparams

chappjc: Set in `wiredDB` on constructions: https://github.com/McEdward/dcrdata/blob/86c601d031ce647d6aa33b7a6ce21eec124ca16b/db/dcrsqlite/apisource.go#L40

McEdward: So in this case I'd be making sure it's on the mainnet

chappjc: Well it's supposed to work with any of them.  When the program starts up, the user sets the network and then `wiredDB.params` is set for you.  We just need to use that variable to validate for the network that dcrdata is running on.

db/dcrsqlite/apisource.go
chappjc: (this is a legit malfunction in `SearchRawTransactionsVerbose` for an address that was already validated, so we don't want to continue on with a value for `txs` that is almost certainly `nil`)

McEdward: But the return type for GetExplorerAddress() is *explorer.AddressInfo

chappjc: Ah yeah, sorry.  Nevermind.

McEdward:
screenshot 21

raedah:
Since there are no transactions being displayed we can drop the column headers.
Keeping the style consistant, use the same style as when no tickets are mined in a block 
(example: https://explorer.dcrdata.org/block/197917 ). 
Display just below the "Transactions" header,
"No transactions found for this address."
Remove the text on the top right.

chappjc:
Technical aspects approved.  Good job with the changes.
Please defer to @raedah regarding the page design in this PR.  Having a screenshot at the start of a PR is often helpful to broaden the reviewing audience early

McEdward:
screenshot 22 I kind of looks a little packed together right now @raedah What do you think?

raedah:
Looks like TOTAL UNSPENT, RECEIVED, SPENT  are missing in master. These totals should be being displayed with 0.0 as the value.  There should be proper spacing between the address and the Transactions header.

McEdward:
screenshot 25
screenshot 24

chappjc:
Looks nice.  You'll just need to address this little fix: db/dcrsqlite/apisource.go:1100:22:warning: unnecessary conversion (unconvert)

raedah:
Found some edge cases with sstxs that are being resolved.

chappjc:
@raedah What were the edge cases?

raedah:
http://127.0.0.1:7777/address/DsQxuVRvS4eaJ42dhQEsCXauMWjvopWgrVg
http://127.0.0.1:7777/address/Dsg8kKustN7RiJhG11qrSiGwJa46HEd9Xpq
-totals are failing in lite mode, "RetrieveAddressSpentUnspent(pgb.db, address) it's now dependent on the database"

http://127.0.0.1:7777/address/Dsp7GoensoXTk2ny9m757y7XhPZukX1sST7
http://127.0.0.1:7777/address/Dsoo43jpcyEJiYtChefRS91mBzCAVSspHfY
-pagination selector is present even though there is not a full page of items.
-"funding transactions 1 — 2 of 3"  ...count is erroneous.  We could add counts of spending transactions, receiving transactions, ticket transactions to the header, and keep the total transaction count in pagination where it matches up.
-lacks an indicator for transaction type sstx, does not show a credit or debit. Put sstxcommitment under the credit column.

chappjc:
That is true, we cannot do address totals.  Outpoint spending info and address page stuff were the main items addressed by a complete database.

@McEdward Even in full mode (non-lite) DsQxuVRvS4eaJ42dhQEsCXauMWjvopWgrVg does not show a received value.  And for Dsg8kKustN7RiJhG11qrSiGwJa46HEd9Xpq, there's a template exec error:

> [INF] EXPR: Unable to get address Dsg8kKustN7RiJhG11qrSiGwJa46HEd9Xpq history: ReduceAddressHistory failed
> [ERR] EXPR: Template execute failure: template: address.tmpl:150:66: executing "address" at <index $heights $i>: error calling index: index out of range: 0


@raedah I'm not seeing the pagination selector for those last two links you gave.  Maybe hard refresh?

chappjc:
This table seems to be a relic:

https://github.com/dcrdata/dcrdata/pull/302/files#diff-466b86ad29575a3440b760df0f4461a1R14

chappjc:
```
diff --git a/views/address.tmpl b/views/address.tmpl
index ff9574f..bfeb41a 100644
--- a/views/address.tmpl
+++ b/views/address.tmpl
@@ -11,19 +11,6 @@
             <div class="col-md-8 col-sm-6">
                 <h4>Address</h4>
                 <div class="mono">{{.Address}}</div>
-                <table class="table-centered-1rem">
-                    {{if .Balance}}
-                    {{if not .KnownFundingTxns}}
-                    <tr>
-                        <td class="text-right pr-2 h1rem p03rem0">The {{.NumFundingTxns}} most recent transactions are tabulated below.</td>
-                    </tr>
-                    {{end}}
-                    <!-- <tr>
-                        <td class="text-right pr-2 h1rem p03rem0">UNCONFIRMED</td>
-                        <td>{{.NumUnconfirmed}}</td>
-                    </tr> -->
-                    {{end}}
-                </table>
             </div>
             <div class="col-md-4 col-sm-6 d-flex pb-3">
                 <table>
@@ -37,9 +24,8 @@
                     <tr>
                         <td class="text-right pr-2">RECEIVED</td>
                         <td class="mono nowrap text-right">
-                            {{with $received := add .Balance.TotalSpent .Balance.TotalUnspent}}
+                            {{$received := add .Balance.TotalSpent .Balance.TotalUnspent}}
                             {{template "decimalParts" (amountAsDecimalParts $received true)}}<span class="pl-1 unit">DCR</span>
-                            {{end}}
                         </td>
                     </tr>
                     <tr>
@@ -147,7 +133,7 @@
                                 <span data-age="{{.Time}}"></span>
                             {{end}}
                             </td>
-                            <td data-confirmation-block-height="{{index $heights $i}}">{{.Confirmations}}</td>
+                            <td data-confirmation-block-height="{{index $.ConfirmHeight $i}}">{{.Confirmations}}</td>
                             <td>{{.FormattedSize}}</td>
                             {{end}}
                         </tr>
```

chappjc:
So I'd like us to look at this more carefully, as @raedah is doing.

As soon as this is in we can:
1. tag v1.2.0
2. create a release-1.2 maintenance branch for patches
3. test and merge the stake tables PR on master (versioned as 2.0.0)
4. begin converging on v2.0.0 release (backporting fixes as needed to release-1.2 branch)

raedah:
>we cannot do address totals (reliably) in lite mode

This previously worked. Just limited to a sane number of items. Most addresses wont have many transactions.

chappjc:
Haven't run lite mode recently.  But that's what I meant by reliably -- only for limited numbers of txns.  I do recall it was addressed by specifying that the limit was exceeded if that was the case for the address in question.  I agree we need to restore that

McEdward:
Could anyone please tell me the difference between KnownFundingTxns and NumFundingTxns in explorer.AddressInfo @chappjc @raedah. I currently trying to make the summary at the top of the page tailored to the mode it's running on so lite mode looks that in #243

chappjc:
I'll be online in 2 hours, but know is the actual transaction list.  I believe you only get this in full node

chappjc:
*full mode

chappjc:
So `KnownFundingTxns` and `NumFundingTxns` are both counts.  See `explorer.ReduceAddressHistory`.  Now see the `if exp.liteMode {` ... `} else` block in `explorer.AddressPage`.

Compare the two code paths.  The lite mode path does not set `KnownFundingTxns`.  In "full" mode, this variable represents all *known* funding transactions, while `NumFundingTxns` represents, well... see here: https://github.com/dcrdata/dcrdata/blob/master/db/dcrpg/pgblockchain.go#L239-L270

chappjc:
(We do not have a count of total funding transactions for an address without the full db)

McEdward:
explorer.ReduceAddressHistory doesn't seem to show type sstxchange so when there is only one sstxchange type transaction it shows a template error page, not yet sure how to go about this yet. the paginator is still bugged showing the total number of receiving txs on the page, It also doesn't know how many txs are there in total

chappjc:
I'm afraid I don't have a quick answer for you -- I would have to go back and study the code.  Just jump in to the go code, trace the flow of code and try debugging with VSCode.

Be sure to not break full mode though.  There is opportunity for clean up though, so if you want to change the way things are shown and tallied up, that might be a welcome change too

McEdward:
Paginator is working fine now, I set the a new variable txs that uses MaxAddressRows instead of count, since count changes from the template page every time the user tries to load more transactions Maximum number of txs that can be seen on the address page remains 1000 to save loading time
screenshot 27

chappjc:
Hmm, that looks good.  Will review, but one comment: It would be nice to have something to say that the 1000 is actually an upper limit.  Like with an "*" and a bit of text at the bottom.

db/dcrsqlite/apisource.go
```diff
@@ -1045,9 +1045,22 @@ func (db *wiredDB) GetExplorerAddress(address string, count, offset int64) *expl
 		return nil
 	}
 
+	maxcount := explorer.MaxAddressRows
 	txs, err := db.client.SearchRawTransactionsVerbose(addr,
 		int(offset), int(count), true, true, nil)
-	if err != nil {
+	maxLimitOfTxs, err := db.client.SearchRawTransactionsVerbose(addr,
```
chappjc: I'm not OK with calling this RPC twice.  Looks like this is only being used to get the count `int64(len(maxLimitOfTxs))`, and `maxLimitOfTxs` is otherwise not used.

We have a hard limit (`maxcount`), so this is not even getting the real total count.

Think about if `len(txs)` were less than `count`... the second RPC is pointless.

Also, you could get by with just the second RPC with the larger `maxcount`, and then only use `count` of the transactions to build the `[]*explorer.AddressTx` with `makeExplorerAddressTx` below.

explorer/explorerroutes.go
```diff
@@ -205,8 +205,27 @@ func (exp *explorerUI) AddressPage(w http.ResponseWriter, r *http.Request) {
 		addrHist, balance, errH := exp.explorerSource.AddressHistory(
 			address, limitN, offsetAddrOuts)
 		if errH != nil {
-			log.Errorf("Unable to get address %s history: %v", address, errH)
-			exp.ErrorPage(w, "Something went wrong...", "could not find that address's history", false)
+			log.Infof("Unable to get address %s history: %v", address, errH)
+			addrData := exp.blockData.GetExplorerAddress(address, limitN, offsetAddrOuts)
```
chappjc: If this actually finds addresses, but the PG query did not, there needs to be a stern `log.Error` message.

chappjc:
Can you clarify the issue with sstxchange?

McEdward:
I don't know how to yet, It'll be helpful if i can get some tips

db/dcrsqlite/apisource.go
```diff
@@ -1111,7 +1111,7 @@ func (db *wiredDB) GetExplorerAddress(address string, count, offset int64) *expl
 		Limit:            count,
 		Offset:           offset,
 		Transactions:     addressTxs,
-		NumFundingTxns:   numberOfTx,
+		NumFundingTxns:   count,
```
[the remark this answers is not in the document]
McEdward: oh, I'll correct that immediately based on the size of numberOfTx and count

explorer/explorerroutes.go
chappjc: No, the one above should still be `Infof`.  I meant after the `if addrData == nil { ... }` block below, add a new Error message (just dcrdata console, not the web page) because this should not happen. It means the PG DB has malfunctioned

db/dcrsqlite/apisource.go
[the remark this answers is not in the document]
McEdward: This should fix that

db/dcrsqlite/apisource.go
```diff
@@ -1098,7 +1098,12 @@ func (db *wiredDB) GetExplorerAddress(address string, count, offset int64) *expl
 		}
 	}
 	numberMaxOfTx := int64(len(txs))
-
+	var numFundingTxns int64
+	if numberMaxOfTx < count {
+		numFundingTxns = numberMaxOfTx
+	} else {
```
chappjc: This should do it.  But note that more idiomatic Go would be to avoid the `else` by initializing with `numFundingTxns := count` rather than zero initializing.

chappjc:
In some brief testing, this all seems to be working.  However, I think a superscript asterisk needs to be next to the "1000", indicating to look for the footnote that you added.  For example:

image

chappjc:
BTW, is there still a crash with sstxchange?  I can't reproduce one

McEdward:
pertaining sstxchange, i couldn't find any issues with it any more

views/address.tmpl
```diff
@@ -173,11 +179,6 @@
                 {{end}}
 
                 {{if lt .NumFundingTxns .KnownFundingTxns}}
-                {{if not .Fullmode}}
-                <div>
-                    *1000 is the upper limit for total number of transactions shown
```
chappjc: This needs to stay.  The idea of the superscript asterisk is to indicate that there is a footnote to read.  After this, we're all set.  :smile:

raedah:
I am seeing display issues still. Will address in a new PR.

PR #303, opened 2017-12-20
Put version in footer, rearrange items, format.

chappjc:
Built with setting `main.CommitHash`:

image

Built without setting `main.CommitHash` (default):
image

### DIFF
--- a/explorer/explorer.go
+++ b/explorer/explorer.go
@@ -85,6 +85,7 @@ type explorerUI struct {
 	ExtraInfo       *HomeInfo
 	MempoolData     *MempoolInfo
 	ChainParams     *chaincfg.Params
+	Version         string
 }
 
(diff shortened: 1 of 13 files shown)


chappjc:
@dcrdata/explorers How about we try on this look for size?  The primary purpose of this PR was to put the version in the footer, but I ended up tweaking the text and layout too.

chappjc:
Up on https://beta.dcrdata.org/ for testing.

RogueElement:
Seems to work just fine

chappjc:
Just found a template exec error on tx and address pages

RogueElement:
Yeah, just saw that seems the template line is still in the txinfo or addressinfo scope, changing to `{{ template "footer" $.Version }}` should fix it for both

chappjc:
I moved stuff around instead, but you're right that we could have used `$.`

PR #306, opened 2017-12-26
Changed search to use the url

RogueElement:
This should solve most of #261 (just not the update search with websocket part).

### DIFF
--- a/explorer/explorerroutes.go
+++ b/explorer/explorerroutes.go
@@ -110,6 +110,7 @@ func (exp *explorerUI) Block(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	w.Header().Set("Content-Type", "text/html")
+	w.Header().Set("Turbolinks-Location", "/block/"+hash)
 	w.WriteHeader(http.StatusOK)
 	io.WriteString(w, str)
 }
@@ -167,6 +168,7 @@ func (exp *explorerUI) TxPage(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	w.Header().Set("Content-Type", "text/html")
+	w.Header().Set("Turbolinks-Location", "/tx/"+hash)
 	w.WriteHeader(http.StatusOK)
 	io.WriteString(w, str)
 }
@@ -261,6 +263,7 @@ func (exp *explorerUI) AddressPage(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	w.Header().Set("Content-Type", "text/html")
+	w.Header().Set("Turbolinks-Location", "/address/"+address)
 	w.WriteHeader(http.StatusOK)
 	io.WriteString(w, str)
 }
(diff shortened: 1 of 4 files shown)


raedah:
As it was with websockets search bar was not a good user experience. Red coloring for search failure was unclear and didnt tell the user what was happening. Also didnt put the user on a page where they could simply refresh to search again, or get the eventual websocket update to happen when the result becomes valid.

RogueElement:
@chappjc I can't seem to see how it breaks the websocket, the websocket stays connected when you search and get the error page and also when you search and it redirects you to the desired page. Can you get an instance of when the websocket breaks?

chappjc:
I'm rebuilding my db to try this again, so I can't give an exact example for an hour or so, but what I am seeing is from a user perspective the websocket is always connected, but the websocket hub looses the connection and has to establish a new one.  What I was seeing yesterday in the dcrdata log with this change was that a search would cause the original websocket connection to be abandoned.

Where the `action=/search` part of the search form is used, the browser opens that page with the query.  This is a regular page load vs. the way turbolinks loads the pages with XHR (javascript).  When you initial switched the search to send on the websocket, that allowed the websocket connection to be reused.

I'll check this out again in the dcrdata terminal, and in the browser's developer tools (the Network tab in particular).

RogueElement:
Oh, I see, I'll look into making it use xhr.

chappjc:
Since it's not a UX issue, but rather an architectural issue, I suppose the PR is solid as is.  I just wanted to identify a more elegant way to handle the websocket connections than having the client drop and server eventually realize it after a ping times out.

When this might be helpful is when the WebsocketHub needs to send client-specific messages, as is likely to be the case for the address page. (Say we implement subscription to address/whatever notifications). Right now it's broadcast only.  I suppose some other more persistent identifier like cookies could be involved there however.

RogueElement:
I added something to make it use XHR (via turbolinks) but the "Connected" status text disappears if the search fails and if the search works the URL still shows `/search?search=x` and I have no idea how to fix either of those. If @gozart1 could chip it with some info on that, I'll appreciate it.

raedah:
Function looks good. After merge it would be good for @gozart1 to do a face lift of the error pages. We probably do not even need to call it an ERROR like something went wrong, but could be more gentle about it with something along the lines of "Sorry, that transactions could not be found." As it is a normal expected occurrence when using the tool. I have opened https://github.com/dcrdata/dcrdata/issues/307 for tracking autorefresh when search items become available.

chappjc:
@RogueElement did you see this: https://github.com/turbolinks/turbolinks/blob/master/README.md#following-redirects

chappjc:
Since `Search` uses redirects in some cases now, it might be needed to set that location header.  Not sure about why the connected text disappears

RogueElement:
Added lines to set the location header, now the url changes to the appropriate one (the connected text is back too but I have no idea why).

chappjc:
Good stuff.  Looks about ready to go.  Will test and merge ASAP

chappjc:
This solves the issue of being able to refresh failed searches, so merging.

But searches still drop the existing websocket connection, and a new one is created when the result page loads.  (TODO)

PR #311, opened 2018-01-05
Desktop Notifications

chappjc:
Implement desktop notifications (new block, transaction opened current page, etc.) using the HTML5 Notifications API.  Use github.com/alexgibson/notify.js

On Linux, the Firefox (top) and Chrome (notifications) look like this:

image

We can put a little dcrdata avatar/icon like other services do for album covers and stuff.  I'm not sure about that other html on the chrome notification...

Refs:
- https://www.w3.org/TR/notifications/
- https://notifications.spec.whatwg.org/#image-url


chappjc:
Updating to show icon:

image

still don't know what does the html in the Chrome notification

chappjc:
@gozart1 Any experience with the HTML5 Notifications API?  I don't get why the Chrome one has the messed up href HTML on it.

RogueElement:
screenshot
The messed up href doesn't show on mine and I use Windows so it might be how Chrome itself handles it for different platforms/versions.

chappjc:
Ah but it does say localhost:7777.
Thanks for the screenshot.  Didn't realize the icon is too big.  I'll have to make a new one with alpha channel too.

chappjc:
Anyway, I'm going to let this one sit while the issues with 1.2, and the stake tables for 2.0 get sorted out.

gozart1:
Here's how it looks for me ( OSX / Chrome )
screen shot 2018-01-06 at 10 58 36 pm
Looks like it used the 32x32 icon, perhaps there's a way to make it more high res. Not too familiar w/ the web notification api, will have to brush up on it.

chappjc:
Minified notify.js, updated icon/image, and now asks for permission on page load.

gozart1:
Seems there's not much else that can be done to style the notifications, since they are native element of the browser platform similar to an alert. So what we have here looks good to me.

PR #312, opened 2018-01-07
js based address qrcode

gozart1:
This is a lightweight, purely client side approach to let a user generate an address qr code by clicking on a qr code icon.

This is an alternative implementation to https://github.com/dcrdata/dcrdata/pull/301
`$.getScript` is used to load the qr code lib only if needed.

screen shot 2018-01-06 at 10 13 12 pm
screen shot 2018-01-06 at 10 13 23 pm

https://github.com/dcrdata/dcrdata/pull/301


chappjc:
Yeah, bit thumbs up.  Will take a closer look at the implementation, but I think this checks all the boxes.

Not that this is new to the PR, but what do the ttf and woff files provide?  Is that in case the browser can't use SVG?

Will test...

gozart1:
tff and woff are just other font formats ( for cross browser support https://www.w3schools.com/css/css3_fonts.asp )

I used https://icomoon.io to create the custom icon font.
By the way I bought us a license for the "Icomoon - Ultimate" pack which is where the qrcode icon is from.

raedah:
No qr code nor qr icon is showing up for me.

chappjc:
I'll cherry-pick the commit from https://github.com/dcrdata/dcrdata/pull/326#issuecomment-361036369

raedah:
Works for me after @RogueElement 's fix. I can suggest that the display of the QR code is made toggle. Currently the icon disappears when clicked and then the large qr code is displayed. I suggest that when the large qr code is clicked, it disappears and returns back to the icon. Also, id make the icon stay black instead of blue. It seems obvious that is is there to be clicked, and the blue looks odd because qr codes are expected to be black.

chappjc:
I agree about having it toggle.  Blue does suggest it's clickable, but I agree that's pretty clear it's meant to be clicked for a QR.

gozart1:
I've incorporated your feedback. Ready to merge.

chappjc:
tACK.
Love it.

PR #315, opened 2018-01-18
Full ticket pool database type and API endpoints

chappjc:
Adds full ticket pool routes, and new `TicketPool` type to `package stakedb`.

Endpoints:
1. /stake/pool/full
2. /stake/pool/b/{idxorhash}/full

Notes:
 - github.com/asdine/storm is now vendored by dep.
 - TicketPool uses storm for persistent storage of ticket pool diffs (in/out as blocks connected)
 - StakeDatabase maintains a TicketPool, growing and trimming it when blocks are connected or disconnected.

Added:
- PoolAtHeight and PoolAtHash to StakeDatabase
-  GetPool and GetPoolByHash to APIDataSource interface
-  GetPool and GetPoolByHash impl. to wiredDB (TODO: dcrpg/ChainDB)
- (*appContext).getTicketPool
- BlockIndexOrHashPathCtx middleware

Bonus:
- Add pprof HTTP profiling option to dcrdata.


apirouter.go
```diff
@@ -102,7 +102,9 @@ func newAPIRouter(app *appContext, userRealIP bool) apiMux {
 		})
 		r.Route("/pool", func(rd chi.Router) {
 			rd.With(app.BlockIndexLatestCtx).Get("/", app.getTicketPoolInfo)
+			rd.With(app.BlockIndexLatestCtx).Get("/full", app.getTicketPool)
 			rd.With(BlockIndexPathCtx).Get("/b/{idx}", app.getTicketPoolInfo)
+			rd.With(BlockIndexOrHashPathCtx).Get("/b/{idxorhash}/full", app.getTicketPool)
```
marcopeereboom: b?  SHouldn't we have something more descriptive?

apiroutes.go
```diff
@@ -923,6 +925,23 @@ func (c *appContext) getBlockRangeSteppedSummary(w http.ResponseWriter, r *http.
...
 
+func (c *appContext) getTicketPool(w http.ResponseWriter, r *http.Request) {
+	// getBlockHeightCtx falls back to try hash if height fails
+	idx := c.getBlockHeightCtx(r)
+	if idx < 0 {
+		http.Error(w, http.StatusText(422), 422)
+		return
+	}
+
+	tp, err := c.BlockData.GetPool(int(idx))
```
marcopeereboom: We should use the correct type instead of casting.

stakedb/stakedb.go
```diff
@@ -394,6 +456,19 @@ func (db *StakeDatabase) Open() error {
 	return err
 }
 
+// Close closes the database.
+func (db *StakeDatabase) Close() error {
```
marcopeereboom: This is super confusing to read.

stakedb/stakedb.go
```diff
@@ -454,6 +529,20 @@ func (db *StakeDatabase) PoolSize() int {
 	return db.BestNode.PoolSize()
 }
 
+// PoolAtHeight gets the entire list of live tickets at the given chain height.
+func (db *StakeDatabase) PoolAtHeight(height int64) ([]chainhash.Hash, error) {
```
marcopeereboom: What happens here during a fork?

stakedb/ticketpool.go
```diff
@@ -0,0 +1,326 @@
...
+		return nil, fmt.Errorf("failed (*storm.DB).AllByIndex: %v", err)
+	}
+	diffs := make([]PoolDiff, len(poolDiffs))
+	for i := range poolDiffs {
+		diffs[i] = poolDiffs[i].PoolDiff
+	}
+
+	// Construct TicketPool with loaded diffs and diff DB
+	return &TicketPool{
+		RWMutex: new(sync.RWMutex),
```
marcopeereboom: meh, not needed.

stakedb/ticketpool.go
```diff
@@ -0,0 +1,326 @@
...
+// Cursor returns the current cursor, the location of the next unapplied diff.
+func (tp *TicketPool) Cursor() int64 {
+	tp.RLock()
+	defer tp.RUnlock()
+	return tp.cursor
+}
+
+// append grows the diffs slice and advances the tip height.
+func (tp *TicketPool) append(diff *PoolDiff) {
+	tp.tip++
```
marcopeereboom: flip this with line 94.

stakedb/ticketpool.go
```diff
@@ -0,0 +1,326 @@
...
+	tp.diffs = append(tp.diffs, *diff)
+}
+
+// trim is the non-thread-safe version of Trim.
+func (tp *TicketPool) trim() int64 {
+	if tp.tip == 0 || len(tp.diffs) == 0 {
+		return tp.tip
+	}
+	tp.tip--
```
marcopeereboom: Is this the right spot? I ask because there is an error path underneath it.

stakedb/stakedb.go
chappjc: The stake database always follows what the node thinks is the main chain.  On a reorg, stakedb.go trims back the ticket pool db and appends the new main chain.  The node doesn't convey any information about forks that don't cause a reorg, so we don't have a way to look up those blocks even if we wanted to.

stakedb/ticketpool.go
chappjc: What's not needed?  Thread-safing this is needed, so do you mean the `new`?  It's a pointer to the mutex so it has to be allocated manually.

stakedb/ticketpool.go
chappjc: Yes, here's why.  The main function of `trim` is to drop the last element from `diffs`, which is essentially the opposite of `append` (so, `tip--` and `tp.diffs = tp.diffs[:len(tp.diffs)-1]`), but if the cursor would be invalidated, then we also have to retreat (apply the about-to-disappear diffs in reverse to the pool map).  Now the reason `retreatTo` comes before `tp.diffs[:len(tp.diffs)-1]` is because retreating may require the diff being chopped.  About `tip--`, `retreatTo` checks that the destination height is valid with `tip`.  So actually, the error should be a `panic` because this shouldn't happen and there's no way to recover aside from deleting the db and starting fresh.

apirouter.go
chappjc: perhaps.  We were using `b` for block and `r` for range to keep the path length under control

apiroutes.go
chappjc: agreed, `GetPool` should take `int64` since it does a conversion too lol

chappjc:
@marcopeereboom Regarding mainnet validation, one of the tests uses a .db file from mainnet to detect some regressions, but we're only checking pool size rather than the full contents.  Should do this.

chappjc:
Although this is slated for 1.3, this is going in now, before the rest of the 1.2 stuff.  PRs for 1.2 will go to master and then cherry-picked onto a 1.2 release branch.

PR #318, opened 2018-01-19
block range fixes

chappjc:
The stepped endpoint could insert a comma in the Json array, making it invalid for parsers.

Also, the result of GetSummary was not checked so a bad request could continue to execute queries for many non existing blocks


raedah:
whats this?

chappjc:
going in now since this fixes a potential abuse

PR #319, opened 2018-01-20
Amendments to the address page summary

McEdward:
sstx transactions appear in address without a credit or debit into the address, to show this types of transactions we are displaying 'sstxcommitment' in the credit column of these transactions for visibility. Also if the transactions exceed the maximum number of transactions the totals at the top become very inaccurate, to fixed this it is replaced with unavailable.

### DIFF
--- a/db/dcrsqlite/apisource.go
+++ b/db/dcrsqlite/apisource.go
@@ -1082,7 +1082,9 @@ func (db *wiredDB) GetExplorerAddress(address string, count, offset int64) *expl
 			if len(y.ScriptPubKey.Addresses) != 0 {
 				if address == y.ScriptPubKey.Addresses[0] {
 					t, _ := dcrutil.NewAmount(y.Value)
-					totalreceived += t
+					if t > 0 {
+						totalreceived += t
+					}
 					numReceiving++
 				}
 			}
@@ -1091,7 +1093,9 @@ func (db *wiredDB) GetExplorerAddress(address string, count, offset int64) *expl
 			if u.PrevOut != nil && len(u.PrevOut.Addresses) != 0 {
 				if address == u.PrevOut.Addresses[0] {
 					t, _ := dcrutil.NewAmount(*u.AmountIn)
-					totalsent += t
+					if t > 0 {
+						totalsent += t
+					}
 					numSpending++
 				}
 			}
--- a/db/dcrsqlite/apisource.go
+++ b/db/dcrsqlite/apisource.go
@@ -1112,6 +1112,7 @@ func (db *wiredDB) GetExplorerAddress(address string, count, offset int64) *expl
 	return &explorer.AddressInfo{
 		Address:          address,
 		Limit:            count,
+		MaxLimit:         maxcount,
 		Offset:           offset,
 		Transactions:     addressTxs,
 		NumFundingTxns:   numFundingTxns,
--- a/db/dcrsqlite/apisource.go
+++ b/db/dcrsqlite/apisource.go
@@ -1098,9 +1098,9 @@ func (db *wiredDB) GetExplorerAddress(address string, count, offset int64) *expl
 		}
 	}
 	numberMaxOfTx := int64(len(txs))
-	var numFundingTxns = count
+	var numTxns = count
 	if numberMaxOfTx < count {
-		numFundingTxns = numberMaxOfTx
+		numTxns = numberMaxOfTx
 	}
 	balance := &explorer.AddressBalance{
 		Address:      address,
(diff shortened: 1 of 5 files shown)


chappjc:
Could you edit the description with detail on the changes?  For instance, please explain reasoning behind the sstxcommitment edit.  Thanks!

raedah:
Some of the transactions relating to an address are sstx transactions that use the address as an sstxcommitment. Thats why they have no credits or debits. Displaying 'sstxcommitment' next to the transaction was the easiest way to expose that information to the user.

chappjc:
Ah, makes sense.  Thanks @raedah !

raedah:
Tested this, found errors.

http://127.0.0.1:7777/address/DsQxuVRvS4eaJ42dhQEsCXauMWjvopWgrVg
UNCONFIRMED    11
should be
UNCONFIRMED    11 transactions

http://127.0.0.1:7777/address/Dsg8kKustN7RiJhG11qrSiGwJa46HEd9Xpq
Total Unspent says unavailable even though it should be available. Looks like the template is simply testing for 0, same issue on Sent and Received. These amounts are only unavailable if we are hitting the 1000 tx constraint in lite mode.

raedah:
Tested again and errors are resolved.

views/address.tmpl
```diff
@@ -86,7 +100,7 @@
                     <h5 class="mr-auto mb-0">Transactions</h5>
                     {{if lt .NumFundingTxns .KnownFundingTxns}}
                     <div div class="d-flex flex-wrap-reverse align-items-center justify-content-end">
-                        <span class="fs12 nowrap text-right">funding transactions {{intComma (add .Offset 1)}} &mdash; {{intComma (add .Offset .NumFundingTxns)}} of {{intComma .KnownFundingTxns}}
+                        <span class="fs12 nowrap text-right"> {{intComma (add .Offset 1)}} &mdash; {{intComma (add .Offset .NumFundingTxns)}} transactions of {{intComma .KnownFundingTxns}}
```
chappjc: OK, but there is reason to say "funding transactions" and not just "transactions"
IMO it's going to be confusing to see a table with 3 rows (say 2 paying to the address and 1 spending), but this text saying "1 - 2 transactions of 2"
I think it would be better to say "1 - 2 of 2 transactions paying to this address"

chappjc: nvm.  Please also test with postgresql (full mode)

raedah:
Looks like this needs to be fixed as well,
https://github.com/dcrdata/dcrdata/commit/cfb3084e606db0baf46360bb81d504ebf6681148
"If .MaxLimit is 1000, you want ge instead of eq"

views/address.tmpl
```diff
@@ -56,24 +56,38 @@
                         </tr>
                         {{end}}
                     {{else}}
-                    <tr>
-                        <td class="pr-2 lh1rem vam text-right xs-w91">UNCONFIRMED</td>
-                        <td class="fs28 mono nowrap fs16-decimal d-flex align-items-center justify-content-end">
-                            {{.NumUnconfirmed}}
-                        </td>
+                    <tr  class="h2rem">
+                        <td class="pr-2 lh1rem vam text-right xs-w91">TOTAL UNSPENT</td>
```
raedah: TOTAL UNSPENT is in the code twice here because the {{if .Fullmode}} loop is covering an unnecessarily large area. Should only be covering the code that is needed and not duplicating the output.

views/address.tmpl
```diff
@@ -56,24 +56,38 @@
...
                         <td class="mono nowrap text-right">{{.TotalSent}}</td>
+                        {{end}}
                     </tr>
-                    <tr>
-                        <td class="text-right pr-2">UNSPENT</td>
-                        <td class="mono nowrap text-right">{{.Unspent}}</td>
-                    </tr>
+                    {{if ne .NumUnconfirmed 0}}
+                        <tr>
+                            <td class="text-right pr-2">UNCONFIRMED</td>
```
raedah: This is only in the lite mode loop but should apply to both modes. Only needs to be coded once.

db/dcrsqlite/apisource.go
```diff
@@ -1110,20 +1110,43 @@ func (db *wiredDB) GetExplorerAddress(address string, count, offset int64) *expl
...
+		Transactions:      addressTxs,
+		NumTransactions:   numTxns,
+		KnownTransactions: numberMaxOfTx,
+		KnownFundingTxns:  numReceiving,
+		NumSpendingTxns:   numSpending,
+		NumUnconfirmed:    numUnconfirmed,
+		TotalReceived:     totalreceived,
+		TotalSent:         totalsent,
+		Unspent:           totalreceived - totalsent,
+		Balance:           balance,
```
chappjc: please add a new line before ValidateNetworkAddress

db/dcrsqlite/apisource.go
```diff
@@ -1110,20 +1110,43 @@ func (db *wiredDB) GetExplorerAddress(address string, count, offset int64) *expl
...
 	}
 }
 func ValidateNetworkAddress(address dcrutil.Address, p *chaincfg.Params) bool {
 	return address.IsForNet(p)
 }
+
+// SearchRawTransactionsForUnconfirmedTransactions returns the number of unconfirmed transactions,
```
chappjc: ...the number of unconfirmed transactions *involving the specified address*,...

db/dcrsqlite/apisource.go
```diff
@@ -1110,20 +1110,43 @@ func (db *wiredDB) GetExplorerAddress(address string, count, offset int64) *expl
...
+// given a maximum possible unconfirmed
+func (db *wiredDB) SearchRawTransactionsForUnconfirmedTransactions(address string, offset int64, maxUnconfirmedPossible int64) (numUnconfirmed int64, err error) {
```
chappjc: please rename to `CountUnconfirmedTransactions`.

At a glance, I would have expected more output from this function given the original name.  It calls SearchRawTransactionsVerbose, but that's just how the function gets the count, not the purpose of the function

chappjc: The `offset` argument does not make much sense here, unless you are planning to call the function repeatedly in the case where it returns `numUnconfirmed == maxUnconfirmedPossible`

db/dcrsqlite/apisource.go
```diff
@@ -1110,20 +1110,43 @@ func (db *wiredDB) GetExplorerAddress(address string, count, offset int64) *expl
...
+	addr, err := dcrutil.DecodeAddress(address)
+	if err != nil {
+		log.Infof("Invalid address %s: %v", address, err)
+		return 0, err
```
chappjc: `return`

db/dcrsqlite/apisource.go
```diff
@@ -1110,20 +1110,43 @@ func (db *wiredDB) GetExplorerAddress(address string, count, offset int64) *expl
...
+	}
+	txs, err := db.client.SearchRawTransactionsVerbose(addr, int(offset), int(maxUnconfirmedPossible), true, true, nil)
+	if err != nil {
+		log.Warnf("GetAddressTransactionsRaw failed for address %s: %v", addr, err)
+		return 0, err
```
chappjc: `return`

explorer/explorertypes.go
```diff
@@ -134,21 +134,23 @@ type BlockInfo struct {
...
-	Unspent          dcrutil.Amount
-	Balance          *AddressBalance
-	Path             string
-	Fullmode         bool
+	Address           string
+	Limit             int64
+	MaxLimit          int64
+	Offset            int64
+	Transactions      []*AddressTx
+	NumFundingTxns    int64
```
chappjc: I take a lot of blame for ambiguous naming in this section, but I think we could use some comments on these counts.  e.g. 

        NumFundingTxns    int64  // The number of transactions paying to the address

down to `NumUnconfirmed` please

chappjc:
Testing and noticed this:

image

The bogus spend I think is coming from the unconfirmed transaction that *spends* "from" the address"

McEdward:
I have trying to replicate this problem on my computer with no success

chappjc:
Try to find a tx w/ similar uconfirmed credit and debits like the screenshot I posted.

But more importantly, the explorer rendered a negative value.  We need to lookback through the code from explorer to the functions that get the amounts.  Please have a look and see how it's possible.  Are we testing values == 0 instead of <= 0?  I've seen outputs of the dcrd RPC calls that will give bogus negatives.  I think we are just accepting one of those and not checking for that.

McEdward:
a values <= 0 check has been added to check for bogus values

chappjc:
OK

raedah:
tACK

PR #326, opened 2018-01-28
Remove qr functionality from conditional block

RogueElement:
^^


raedah:
Yes, this fixes the breakage for me.

PR #337, opened 2018-02-07
improve fresh start error handling and logging

chappjc:
Special handling for error message that should not be shown on fresh sync.
Add error return to several functions, including `dcrsqlite.NewDB`.
Fix returned height from `GetBlockSummaryHeight` when db is empty.
Do not log scanning progress for block 1 when just did genesis.


chappjc:
Any 1 review would be good to confirm nothing is broken.

PR #341, opened 2018-02-07
fixing the search for addresses with no transactions

McEdward:
This should fix the new issues with #255

### DIFF
--- a/db/dcrsqlite/apisource.go
+++ b/db/dcrsqlite/apisource.go
@@ -1090,7 +1090,8 @@ func (db *wiredDB) GetExplorerAddress(address string, count, offset int64) *expl
 			return nil
 		}
 		return &explorer.AddressInfo{
-			Address: address,
+			Address:  address,
+			MaxLimit: maxcount,
 		}
 	} else if err != nil {
 		log.Warnf("GetAddressTransactionsRaw failed for address %s: %v", addr, err)
--- a/db/dcrsqlite/apisource.go
+++ b/db/dcrsqlite/apisource.go
@@ -1090,8 +1090,8 @@ func (db *wiredDB) GetExplorerAddress(address string, count, offset int64) *expl
 			return nil
 		}
 		return &explorer.AddressInfo{
-			Address:  address,
-			MaxLimit: maxcount,
+			Address:    address,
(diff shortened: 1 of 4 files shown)


db/dcrsqlite/apisource.go
```diff
@@ -1090,7 +1090,8 @@ func (db *wiredDB) GetExplorerAddress(address string, count, offset int64) *expl
 			return nil
 		}
 		return &explorer.AddressInfo{
-			Address: address,
+			Address:  address,
+			MaxLimit: maxcount,
```
chappjc: I'd like for us to start using more descriptive variable names. E.g. what is this a limit on? Also, how are Max and limit not redundant?  I'm guilty of this too btw

explorer/explorerroutes.go
```diff
@@ -217,15 +217,19 @@ func (exp *explorerUI) AddressPage(w http.ResponseWriter, r *http.Request) {
...
 			pageData := struct {
 				Data          *AddressInfo
 				ConfirmHeight []int64
+				Version       string
 			}{
 				addrData,
-				nil,
+				confirmHeights,
+				exp.Version,
 			}
```
chappjc: Good catch

db/dcrsqlite/apisource.go
McEdward: Yes, limit and maxlimit look redundant but limits value changes from the template not Maxlimit which is always the upper limit of transactions that's can be possible in litemode

chappjc: The key word there is "transactions".  What is it a limit on?

explorer/explorerroutes.go
```diff
@@ -271,6 +275,8 @@ func (exp *explorerUI) AddressPage(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
+	log.Info(addrData.KnownFundingTxns, addrData.MaxLimit)
```
chappjc: Please remove

chappjc:
good for merge if anyone gives a tACK

RogueElement:
full mode:
screen shot 2018-02-09 at 19 31 31\
lite mode:
screen shot 2018-02-09 at 19 32 39

Issue #230, opened 2017-10-10
Latest transactions via WebSocket on root explorer page


chappjc:
This would bring the network to life and make the root page more interesting.

RogueElement:
Would this nullify the need for #129 ?

raedah:
The two can probably be combined with a dedicated mempool page as well as minimized mempool view for the front page. Better with more of our own special sauce so it doesnt look too much like the copay explorer.

Issue #129, opened 2017-09-08
Add real time mempool viewer


raedah:
Home page currently shows Tickets in Mempool.  
We can also show a count of Transactions in the Mempool.
The word Mempool can be clickable, and go to /mempool , which would be a page that lists all transaction in the mempool and resuses the templates partials for the block view.
getrawmempool rpc should be used.

chappjc:
package mempool needs updating for tracking and reporting tx types other than votes and tickets. (TODO @chappjc)

PR #343, opened 2018-02-09
Live Mempool Page

RogueElement:
For issues #230 and #129 
Todo:
- [x] Mini live transaction viewer on front page
- ~~Extra mempool page for votes and revocations~~
Mempool page is now similar to block page

### DIFF
--- a/db/dcrsqlite/apisource.go
+++ b/db/dcrsqlite/apisource.go
@@ -6,7 +6,6 @@ package dcrsqlite
 import (
 	"database/sql"
 	"encoding/hex"
-	"errors"
 	"fmt"
 	"sort"
 	"strings"
@@ -409,10 +408,10 @@ func (db *wiredDB) DecodeRawTransaction(txhex string) (*dcrjson.TxRawResult, err
 }
 
 func (db *wiredDB) SendRawTransaction(txhex string) (string, error) {
-	msg := txhelpers.MsgTxFromHex(txhex)
-	if msg == nil {
-		log.Errorf("SendRawTransaction failed: could not decode hex")
-		return "", errors.New("Could not decode hex")
+	msg, err := txhelpers.MsgTxFromHex(txhex)
+	if err != nil {
+		log.Errorf("SendRawTransaction failed: could not decode tx")
+		return "", err
 	}
 	hash, err := db.client.SendRawTransaction(msg, true)
 	if err != nil {
@@ -853,8 +852,8 @@ func makeExplorerBlockBasic(data *dcrjson.GetBlockVerboseResult) *explorer.Block
 
 	// Count the number of revocations
 	for i := range data.RawSTx {
-		msgTx := txhelpers.MsgTxFromHex(data.RawSTx[i].Hex)
-		if msgTx == nil {
+		msgTx, err := txhelpers.MsgTxFromHex(data.RawSTx[i].Hex)
+		if err != nil {
 			log.Errorf("Unknown transaction %s", data.RawSTx[i].Txid)
 			continue
 		}
@@ -973,9 +972,9 @@ func (db *wiredDB) GetExplorerBlock(hash string) *explorer.BlockInfo {
 	tickets := make([]*explorer.TxBasic, 0, block.FreshStake)
 
 	for _, tx := range data.RawSTx {
-		msgTx := txhelpers.MsgTxFromHex(tx.Hex)
-		if msgTx == nil {
-			log.Errorf("Unknown transaction %s", tx.Txid)
+		msgTx, err := txhelpers.MsgTxFromHex(tx.Hex)
+		if err != nil {
+			log.Errorf("Unknown transaction %s: %v", tx.Txid, err)
 			return nil
 		}
 		switch stake.DetermineTxType(msgTx) {
@@ -994,7 +993,11 @@ func (db *wiredDB) GetExplorerBlock(hash string) *explorer.BlockInfo {
 
 	txs := make([]*explorer.TxBasic, 0, block.Transactions)
 	for _, tx := range data.RawTx {
-		exptx := makeExplorerTxBasic(tx, txhelpers.MsgTxFromHex(tx.Hex), db.params)
+		msgTx, err := txhelpers.MsgTxFromHex(tx.Hex)
+		if err != nil {
+			continue
+		}
+		exptx := makeExplorerTxBasic(tx, msgTx, db.params)
 		for _, vin := range tx.Vin {
 			if vin.IsCoinBase() {
 				exptx.Fee, exptx.FeeRate = 0.0, 0.0
@@ -1053,9 +1056,9 @@ func (db *wiredDB) GetExplorerTx(txid string) *explorer.TxInfo {
 		log.Errorf("GetRawTransactionVerbose failed for: %v", txhash)
 		return nil
 	}
-	msgTx := txhelpers.MsgTxFromHex(txraw.Hex)
-	if msgTx == nil {
-		log.Errorf("Cannot create MsgTx for tx %v", txhash)
+	msgTx, err := txhelpers.MsgTxFromHex(txraw.Hex)
+	if err != nil {
+		log.Errorf("Cannot create MsgTx for tx %v: %v", txhash, err)
 		return nil
 	}
 	txBasic := makeExplorerTxBasic(*txraw, msgTx, db.params)
@@ -1264,4 +1267,60 @@ func (db *wiredDB) CountUnconfirmedTransactions(address string, maxUnconfirmedPo
 		}
 	}
 	return
+}
+
+// GetMepool gets all transactions from the mempool for explorer
+// and adds the total out for all the txs and vote info for the votes
+func (db *wiredDB) GetMempool() []explorer.MempoolTx {
+	mempooltxs, err := db.client.GetRawMempoolVerbose(dcrjson.GRMAll)
+	if err != nil {
+		return nil
+	}
+
+	txs := make([]explorer.MempoolTx, 0, len(mempooltxs))
+
+	for hash, tx := range mempooltxs {
+		rawtx, hex := db.getRawTransaction(hash)
+		total := 0.0
+		if rawtx == nil {
+			continue
+		}
+		for _, v := range rawtx.Vout {
+			total += v.Value
+		}
+		msgTx, err := txhelpers.MsgTxFromHex(hex)
+		if err != nil {
+			continue
+		}
+		var voteInfo *explorer.VoteInfo
+
+		if ok := stake.IsSSGen(msgTx); ok {
+			validation, version, bits, choices, err := txhelpers.SSGenVoteChoices(msgTx, db.params)
+			if err != nil {
+				log.Debugf("Cannot get vote choices for %s", hash)
+
+			} else {
+				voteInfo = &explorer.VoteInfo{
+					Validation: explorer.BlockValidation{
+						Hash:     validation.Hash.String(),
+						Height:   validation.Height,
+						Validity: validation.Validity,
+					},
+					Version: version,
+					Bits:    bits,
+					Choices: choices,
+				}
+			}
+		}
+		txs = append(txs, explorer.MempoolTx{
+			Hash:     hash,
+			Time:     tx.Time,
+			Size:     tx.Size,
+			TotalOut: total,
+			Type:     txhelpers.DetermineTxTypeString(msgTx),
+			VoteInfo: voteInfo,
+		})
+	}
+
+	return txs
 }
--- a/explorer/explorer.go
+++ b/explorer/explorer.go
@@ -25,7 +25,6 @@ import (
 	"github.com/decred/dcrd/wire"
 	"github.com/decred/dcrdata/blockdata"
 	"github.com/decred/dcrdata/db/dbtypes"
-	"github.com/decred/dcrdata/mempool"
 	humanize "github.com/dustin/go-humanize"
 	"github.com/go-chi/chi"
 	"github.com/go-chi/chi/middleware"
@@ -35,6 +34,7 @@ import (
 const (
 	homeTemplateIndex int = iota
 	rootTemplateIndex
+	mempoolTemplateIndex
 	blockTemplateIndex
 	txTemplateIndex
 	addressTemplateIndex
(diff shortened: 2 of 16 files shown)


raedah:
- Add 'Type' field on the Latest Transactions to show Vote, Ticket, Revoke, Transaction
 - On mempool page, should say Mempool as a top header. 
 - On mempool page, swap the boxes so Tickets list is first on the left and then Tickets next.
 - Instead of "More transactions and tickets..." could say "More mempool..."  
 - Number of rows in Latest transactions is growing dynamically. Should be statically fixed, make it 5.
 - Make the grey header bar not be double tall that would look better.
 - Get the hash row to the size so it wraps to 2 lines instead of leaving white space.
 - Horizontally align latest transactions with latest blocks. Put transactions on the top, blocks on the bottom.
 - Utilize more space on the width of that home page since we are showing more there. 
 - Make latest blocks 5 rows too.

RogueElement:
So far I've:
- added a type field,
- changed "More transactions and tickets..." to "More mempool..."
- made number of latest transaction rows static and put it on top of latest blocks
- made latest blocks 5 rows

I don't know how to do the rest cause they require styling.

raedah:
This is working well, up until these errors occur and the server stops responding,
2018-02-15 06:33:00.426 [DBG] EXPR: Registering new websocket client
2018-02-15 06:35:04.359 [WRN] DATD: expNewTxChan buffer full!
2018-02-15 06:36:19.813 [WRN] EXPR: websocket client receive error: EOF
2018-02-15 06:36:19.813 [WRN] EXPR: unknown client

Other things we can add to this..
TOTAL SENT  (sum dcr out of all tx in mempool)
SIZE  (sum size of all tx in the mempool)
BEST BLOCK  (block # and time ago)
For consistancy, put these fields in roughly the same spot they are on the block page.

To be consistent with the block page, when no transactions of a type are in the mempool, we would not show the headers and show the "No tickets in the mempool." msg, sub for whatever type it is.

Also for consistency, use Total DCR instead of Total Out, and we should standardize on either Transactions ID or Hash.

For the Votes we can display the same additional columns of data which are on the block page.

Move the 'Time in Mempool' column to be the last column on the right.

Use a capital M in the HTML page title.

From the best block page, NEXT BLOCK could link to mempool, and for the text say 'mempool' instead of dash. https://github.com/dcrdata/dcrdata/blob/master/views/block.tmpl#L57

raedah:
tACK

chappjc:
Sweet.  Will review

db/dcrsqlite/apisource.go
```diff
@@ -1235,3 +1235,54 @@ func (db *wiredDB) CountUnconfirmedTransactions(address string, maxUnconfirmedPo
 	}
 	return
 }
+
+func (db *wiredDB) GetMempool() []explorer.MempoolTx {
```
chappjc: Please doc.

explorer/mempool.go
```diff
@@ -0,0 +1,183 @@
+package explorer
```
chappjc: standard copyright

explorer/explorertypes.go
```diff
@@ -277,3 +295,12 @@ type TicketPoolInfo struct {
 	Target        uint16  `json:"target"`
 	PercentTarget float64 `json:"percent_target"`
 }
+
+type MempoolTx struct {
```
chappjc: doc

chappjc:
Please rebase and resolve conflicts.  Just checked... nothing tricky.

RogueElement:
Rebased and resolved.

raedah:
Are all known instances of the sockets hanging fixed?

raedah:
Total Sent is slightly out of alignment to the right.

raedah:
Note, we can also add links to the mempool api, though doesnt need to be in this PR.

ntfnhandlers.go
```diff
@@ -225,8 +240,56 @@ func makeNodeNtfnHandlers(cfg *config) (*rpcclient.NotificationHandlers, *collec
...
+		// 		log.Warn("newTxChan buffer full!")
+		// 	}
+		// 	//log.Trace("Transaction accepted to mempool: ", hash, amount)
+		// },
+
+		// OnTxAcceptedVerbose is invoked same as OnTxAccepted but is used here
+		// for the explorer mempool monitor to avoid an extra call to dcrd for
+		// the tx details
+		OnTxAcceptedVerbose: func(txDetails *dcrjson.TxRawResult) {
+			msgTx := txhelpers.MsgTxFromHex(txDetails.Hex)
```
chappjc: In general, we want to do as little as possible in these On* handlers since while they are being executed, the RPC client is blocked.  This is why I try to send the bare minimum on a channel to the actual handlers downstream, with a hefty buffer.

So sending the msgTx might be a good idea, then leave the function calls (IsSSGen, SSGenVoteChoices, etc.) to whatever function reads the channel.

db/dcrsqlite/apisource.go
```diff
@@ -1262,3 +1262,56 @@ func (db *wiredDB) CountUnconfirmedTransactions(address string, maxUnconfirmedPo
 	}
 	return
 }
+
+// GetMepool gets all transactions from the mempool for explorer
```
chappjc: Sometime in the future we need to go through and remove all "Get" prefixes where it's absolutely not necessary.  This is good for now, just a heads up.

db/dcrsqlite/apisource.go
```diff
@@ -1262,3 +1262,56 @@ func (db *wiredDB) CountUnconfirmedTransactions(address string, maxUnconfirmedPo
...
+	for hash, tx := range mempooltxs {
+		rawtx, hex := db.getRawTransaction(hash)
+		total := 0.0
+		if rawtx == nil {
+			continue
+		}
+		for _, v := range rawtx.Vout {
+			total += v.Value
+		}
+		msgTx := txhelpers.MsgTxFromHex(hex)
```
chappjc: Need to check if its nil, or better, refactor MsgTxFromHex to return its error.

chappjc:
I suspect duplicate txns are being sent to our handlers via OnTxAcceptedVerbose.  Perhaps we should check for duplicate votes, or look into speeding up the new tx handers somehow.

chappjc:
please rebase and squash at least partially.

ntfnhandlers.go
chappjc: If we're still seeing issues with hanging and backed up buffered channels, we should look carefully at these handlers.

RogueElement:
Rebased and squashed

db/dcrsqlite/apisource.go
```diff
@@ -973,8 +972,8 @@ func (db *wiredDB) GetExplorerBlock(hash string) *explorer.BlockInfo {
 	tickets := make([]*explorer.TxBasic, 0, block.FreshStake)
 
 	for _, tx := range data.RawSTx {
-		msgTx := txhelpers.MsgTxFromHex(tx.Hex)
-		if msgTx == nil {
+		msgTx, err := txhelpers.MsgTxFromHex(tx.Hex)
+		if err != nil {
```
chappjc: please include `err` in the log message below since we're not returning it from this functions

db/dcrsqlite/apisource.go
```diff
@@ -1053,8 +1056,8 @@ func (db *wiredDB) GetExplorerTx(txid string) *explorer.TxInfo {
 		log.Errorf("GetRawTransactionVerbose failed for: %v", txhash)
 		return nil
 	}
-	msgTx := txhelpers.MsgTxFromHex(txraw.Hex)
-	if msgTx == nil {
+	msgTx, err := txhelpers.MsgTxFromHex(txraw.Hex)
+	if err != nil {
 		log.Errorf("Cannot create MsgTx for tx %v", txhash)
```
chappjc: log the `err` here too please

explorer/explorer.go
```diff
@@ -427,6 +438,15 @@ func New(dataSource explorerDataSourceLite, primaryDataSource explorerDataSource
 	}
 	exp.templates = append(exp.templates, explorerTemplate)
 
+	mempoolTemplate, err := template.New("mempool").Funcs(exp.templateHelpers).ParseFiles(
+		exp.templateFiles["mempool"],
+		exp.templateFiles["extras"],
+	)
+	if err != nil {
+		log.Errorf("Unable to create new html template: %v", err)
```
chappjc: Please use `noTemplateError(err)` now.  (This changed when you rebased)

explorer/explorerroutes.go
```diff
@@ -16,10 +16,10 @@ import (
 func (exp *explorerUI) Home(w http.ResponseWriter, r *http.Request) {
 	height := exp.blockData.GetHeight()
 
-	blocks := exp.blockData.GetExplorerBlocks(height, height-6)
+	blocks := exp.blockData.GetExplorerBlocks(height, height-5)
 
 	exp.NewBlockDataMtx.Lock()
-	exp.MempoolData.RLock()
+	exp.MempoolData.Lock()
```
chappjc: Why can't this be `RLock`?

explorer/explorerroutes.go
```diff
@@ -135,6 +135,28 @@ func (exp *explorerUI) Block(w http.ResponseWriter, r *http.Request) {
 	io.WriteString(w, str)
 }
 
+// Mempool is the page handler for the "/mempool" path
+func (exp *explorerUI) Mempool(w http.ResponseWriter, r *http.Request) {
+	exp.MempoolData.Lock()
```
chappjc: Why can't this be `RLock`?

explorer/mempool.go
```diff
@@ -0,0 +1,228 @@
...
+}
+
+func (exp *explorerUI) storeMempoolInfo() {
+
+	defer func(start time.Time) {
+		log.Debugf("storeMempoolInfo() completed in %v",
+			time.Since(start))
+	}(time.Now())
+
+	exp.MempoolData.Lock()
```
chappjc: I'm thinking we should move this closer to where we actually need it locked.  Even after the exp.NewBlockDataMtx.Lock() block as far as I can tell

explorer/mempool.go
```diff
@@ -0,0 +1,228 @@
...
+}
+
+func (txs byTime) Swap(i, j int) {
+	txs[i], txs[j] = txs[j], txs[i]
+}
+
+type byType []MempoolTx
+
+func (txs byType) Less(i, j int) bool {
+	weights := map[string]int{
```
chappjc: this gets constructed every time `Less` is run.  Lets make it a global or deal with it some other way.

chappjc: for instance, we could ditch the `sort.Interface` approach used her, and use[ `sort.Slice`](https://golang.org/pkg/sort/#Slice) where the specified closure (`less`) would capture the map if it were defined as a meber of `explorerUI`.  Just a though

ntfnhandlers.go
```diff
@@ -34,9 +36,13 @@ func registerNodeNtfnHandlers(dcrdClient *rpcclient.Client) *ContextualError {
 	// }
 
 	// Register for tx accepted into mempool ntfns
-	if err = dcrdClient.NotifyNewTransactions(false); err != nil {
-		return newContextualError("new transaction "+
-			"notification registration failed", err)
+	// if err = dcrdClient.NotifyNewTransactions(false); err != nil {
```
chappjc: please remove this bit of old code

ntfnhandlers.go
```diff
@@ -225,8 +238,33 @@ func makeNodeNtfnHandlers(cfg *config) (*rpcclient.NotificationHandlers, *collec
...
+		// 	}
+		// 	//log.Trace("Transaction accepted to mempool: ", hash, amount)
+		// },
+
+		// OnTxAcceptedVerbose is invoked same as OnTxAccepted but is used here
+		// for the explorer mempool monitor to avoid an extra call to dcrd for
+		// the tx details
+		OnTxAcceptedVerbose: func(txDetails *dcrjson.TxRawResult) {
+			select {
+			case ntfnChans.expNewTxChan <- &explorer.NewMempoolTx{
```
chappjc: 👍

chappjc:
Looking pretty darn good now.  I haven't reviewed any JS or HTML yet though.

explorer/mempool.go
RogueElement: Removed `ByType` entirely, it is no longer needed since the transactions are already sorted into their respective slices but I'll keep in mind the `sort.Slice` approach.

explorer/explorerroutes.go
RogueElement: That slipped by, I was testing something. Changed back now.

raedah:
tACK

chappjc:
This tests really well.  Nice any clean implementation.
Only thing is that this doesn't really provide any useful information, just confusion:

VOTES IN MEMPOOL | 28
-- | --

I think we need to filter out votes that are not for the tip block, and combine votes spending the same ticket, and only show a count from 0-5.

explorer/websocket.go
```diff
@@ -121,21 +124,28 @@ func (wsh *WebsocketHub) run() {
...
 			switch hubSignal {
 			case sigNewBlock:
 				log.Infof("Signaling new block to %d websocket clients.", len(wsh.clients))
 			case sigPingAndUserCount:
 				log.Tracef("Signaling ping/user count to %d websocket clients.", len(wsh.clients))
 			case sigMempoolUpdate:
 				log.Infof("Signaling mempool update to %d websocket clients.", len(wsh.clients))
+			case sigNewTx:
+				newtx = <-wsh.NewTxChan
+				log.Infof("Signaling new tx %s to %d websocket clients.", newtx.Hash, len(wsh.clients))
```
chappjc: This happens too often for `Info` level logging.  Please make Debug

chappjc:
In the future, I'd like to consider sending txns in batch to websocket clients.  So after say a 5 second delay, send whatever new txns we have, or if there are >N txns waiting, send them regardless of the delay.

chappjc:
Now running on https://beta.dcrdata.org

chappjc:
The votes table wraps the tx hash, but it could be on one line with small tweaks.  Window is full size here:

image

gozart1:
I've working on a flexbox based layout for these tables that allows for some more fine grained control over table row dimensions and truncates the hash to the space available. Also planning to add some animation to delineate new items being added.

The PR is still WIP, so maybe best to merge this without worrying too much about getting layout perfect and I'll make a separate PR for my frontend stuff.
https://github.com/gozart1/dcrdata-explorer/commit/ff1bc315ebd3acbf5dffe5a90925c442488b2b7f
Should have it ready in a few days.

chappjc:
ok

PR #346, opened 2018-02-11
Nil ptr check getexploreraddress

chappjc:
`GetExplorerAddress` ran w/o a nil pointer check in the AddressPage handler.

Also add a `Gosched()` in `updateDevFundBalance` since it's low priority and CPU intense.


chappjc:
causing headaches for me on beta site, so I'm merging w/o review.

Issue #280, opened 2017-11-18
Show block reward transaction as its own type


raedah:
Looking at a block reward transaction shows it is being detected as type Regular. This could instead be its own type, Coinbase.

PR #357, opened 2018-02-23
Show block reward transaction as its own type "Coinbase"

McEdward:
resolving issue #280

### DIFF
--- a/db/dcrsqlite/apisource.go
+++ b/db/dcrsqlite/apisource.go
@@ -1080,9 +1080,9 @@ func (db *wiredDB) GetExplorerTx(txid string) *explorer.TxInfo {
 	}
 	if tx.Type == "Coinbase" {
 		if tx.Confirmations < int64(db.params.CoinbaseMaturity) {
-			tx.Mature = "True"
-		} else {
 			tx.Mature = "False"
+		} else {
+			tx.Mature = "True"
 		}
 	}
 	if tx.Type == "Vote" || tx.Type == "Ticket" {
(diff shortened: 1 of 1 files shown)


db/dcrsqlite/apisource.go
```diff
@@ -1075,6 +1075,9 @@ func (db *wiredDB) GetExplorerTx(txid string) *explorer.TxInfo {
 		})
 	}
 	tx.Vin = inputs
+	if tx.Vin[0].IsCoinBase() {
+		tx.Type = "Coinbase"
+	}
 	if tx.Type == "Vote" || tx.Type == "Ticket" || tx.Vin[0].IsCoinBase() {
```
chappjc: Why does this block evaluate for coinbase?

Coinbase tx types should determine maturity against this parameter:

https://github.com/decred/dcrd/blob/7647b2594de0cd2a43c2d69a53c8365a6284b08f/chaincfg/params.go#L405-L407

McEdward: Oh, i wasn't trying to evaluate for coinbase maturity, Just wanted block reward to have it own type called coinbase instead of being classified as a regular transaction

chappjc: Right, I get that this wasn't your bit, but I noticed that the block below your edit is evaluated for coinbase although that doesn't make sense.  Is it clear what needs to be changed so the rest of the logic in `GetExplorerTx` makes sense?

chappjc:
No worries.  Thanks for the fix.

Issue #281, opened 2017-11-19
Additional ticket data to show on the ticket page


raedah:
Could show time till expiration, as well as how close it is to average win time.
chaincfg/params.go:     TicketExpiry:            40960

The ticket is valid until it is chosen or for 40,960 blocks (target of 142.22 days)(~4.7 months), after which time it expires. There is around a 99% chance the ticket will be selected before expiring.

Statistically the average win time is 28 days (~28.444) (8192 / 144 / 2)  (8192 blocks)

raedah:
-Check and make sure we are honoring ticket maturity periods.
-Show bar corresponding to ticket log() probability.

mikefullman:
The Decred docs specify that 50% of the time a ticket will be voted after its maturity reaches 8192 blocks.   I cannot find the exact source of this calculation - just that the system was designed this way.  I posted the question to the forums, but its still waiting to be moderated.  I'll include the full text below as it goes through the math.

In the meantime I analyzed the existing tickets in the database.

| Item | Blocks |
|----------|-------------:|
| Records Analyzed | 1,115,572 |
| Median blocks until voted | 5875 |
| Average blocks until voted | 8116 |
| Minimum blocks until voted | 257 |
| Maximum blocks until voted | 41216 |

Since no tickets voted in under 257 blocks I surmise that the ticket maturity is included in my analysis, but the spending maturity is not.  To doublecheck I analyzed one of the records with the minimum blocks that was purchased on block 19758 and voted on block 20015.  

The documentation outlines the following two items: (https://docs.decred.org/mining/proof-of-stake/)
- average vote time of 28 days
- The chance of a ticket voting is based on a Poisson distribution with a mean of 28 days. After 28 days a ticket has a 50% chance to have already voted.

Based on this nomenclature, the proper statistic to use is the median.  Luckily this matches almost exactly to my calculation below of 19 days (5677 + 256) = 5933.  So my conclusion is that the documentation is *wrong* and the proper number of blocks to put in the code for our calculation is:

5677 + 256 + 256 = 6189 or 21.48 days

Furthermore I would recommend we go ahead and do the full math including the current ticket pool size and therefore:

 ROI_DAYS= [[ ln(50%) / ln(1-(5/ACTUAL_TICKET_POOL)) ] + 256 * 2 ] / 288 

28 days is literally all over the documentation so Its a big discrepancy.  I don't want to conclude that the documentation is wrong so please help me understand what I'm missing?  


[...]

PR #361, opened 2018-02-25
ticket page: luck and additional data

McEdward:
resolving issue #281

### DIFF
--- a/db/dcrsqlite/apisource.go
+++ b/db/dcrsqlite/apisource.go
@@ -1324,3 +1324,19 @@ func (db *wiredDB) GetMempool() []explorer.MempoolTx {
 
 	return txs
 }
+
+// TxHeight gives the block height of the transaction id specified
+func (db *wiredDB) TxHeight(txid string) (height int64) {
+	txhash, err := chainhash.NewHashFromStr(txid)
+	if err != nil {
+		log.Errorf("Invalid transaction hash %s", txid)
+		return 0
+	}
+	txraw, err := db.client.GetRawTransactionVerbose(txhash)
+	if err != nil {
+		log.Errorf("GetRawTransactionVerbose failed for: %v", txhash)
+		return 0
+	}
+	height = txraw.BlockHeight
+	return
+}
--- a/explorer/explorerroutes.go
+++ b/explorer/explorerroutes.go
@@ -202,6 +202,39 @@ func (exp *explorerUI) TxPage(w http.ResponseWriter, r *http.Request) {
 					tx.TicketInfo.PoolStatus = poolStatus.String()
 				}
 				tx.TicketInfo.SpendStatus = spendStatus.String()
+				tx.TicketInfo.TicketPoolSize = int64(exp.ChainParams.TicketPoolSize) * int64(exp.ChainParams.TicketsPerBlock)
+				tx.TicketInfo.TicketExpiry = int64(exp.ChainParams.TicketExpiry)
+				expirationInDays := (exp.ChainParams.TargetTimePerBlock.Hours() * float64(exp.ChainParams.TicketExpiry)) / 24
+				maturityInDay := (exp.ChainParams.TargetTimePerBlock.Hours() * float64(tx.TicketInfo.TicketMaturity)) / 24
+				tx.TicketInfo.TimeTillMaturity = ((float64(exp.ChainParams.TicketMaturity) - float64(tx.Confirmations)) / float64(exp.ChainParams.TicketMaturity)) * maturityInDay
+				ticketExpiryBlocksLeft := int64(exp.ChainParams.TicketExpiry) - tx.Confirmations
+				tx.TicketInfo.TicketExpiryDaysLeft = (float64(ticketExpiryBlocksLeft) / float64(exp.ChainParams.TicketExpiry)) * expirationInDays
+				if tx.TicketInfo.SpendStatus == "Voted" {
+					tx.TicketInfo.ShortConfirms = exp.blockData.TxHeight(tx.SpendingTxns[0].Hash) - tx.BlockHeight
+				} else if tx.Confirmations >= int64(exp.ChainParams.TicketExpiry) {
+					tx.TicketInfo.ShortConfirms = int64(exp.ChainParams.TicketExpiry)
+				} else {
+					tx.TicketInfo.ShortConfirms = tx.Confirmations
+				}
+				voteRounds := (tx.TicketInfo.ShortConfirms - tx.TicketMaturity)
+				tx.TicketInfo.BestLuck = tx.TicketInfo.TicketExpiry / int64(exp.ChainParams.TicketPoolSize)
+				tx.TicketInfo.AvgLuck = tx.TicketInfo.BestLuck - 1
+				tx.TicketInfo.VoteLuck = float64(tx.TicketInfo.BestLuck) - (float64(voteRounds) / float64(exp.ChainParams.TicketPoolSize))
+				if tx.TicketInfo.VoteLuck >= float64(tx.TicketInfo.BestLuck-(1/int64(exp.ChainParams.TicketPoolSize))) {
+					tx.TicketInfo.LuckStatus = "*Perfection*"
+				} else if tx.TicketInfo.VoteLuck > (float64(tx.TicketInfo.BestLuck) - 0.25) {
+					tx.TicketInfo.LuckStatus = "Very Lucky!"
+				} else if tx.TicketInfo.VoteLuck > (float64(tx.TicketInfo.BestLuck) - 0.75) {
+					tx.TicketInfo.LuckStatus = "Good Luck"
+				} else if tx.TicketInfo.VoteLuck > (float64(tx.TicketInfo.BestLuck) - 1.25) {
+					tx.TicketInfo.LuckStatus = "Normal"
+				} else if tx.TicketInfo.VoteLuck > (float64(tx.TicketInfo.BestLuck) * 0.50) {
+					tx.TicketInfo.LuckStatus = "Bad Luck"
+				} else if tx.TicketInfo.VoteLuck > 0 {
+					tx.TicketInfo.LuckStatus = "Horrible Luck!"
+				} else if tx.TicketInfo.VoteLuck == 0 {
+					tx.TicketInfo.LuckStatus = "No Luck"
+				}
 			}
 		}
 	}
--- a/explorer/explorerroutes.go
+++ b/explorer/explorerroutes.go
@@ -221,7 +221,7 @@ func (exp *explorerUI) TxPage(w http.ResponseWriter, r *http.Request) {
 				tx.TicketInfo.AvgLuck = tx.TicketInfo.BestLuck - 1
 				tx.TicketInfo.VoteLuck = float64(tx.TicketInfo.BestLuck) - (float64(voteRounds) / float64(exp.ChainParams.TicketPoolSize))
 				if tx.TicketInfo.VoteLuck >= float64(tx.TicketInfo.BestLuck-(1/int64(exp.ChainParams.TicketPoolSize))) {
-					tx.TicketInfo.LuckStatus = "*Perfection*"
+					tx.TicketInfo.LuckStatus = "Perfection"
 				} else if tx.TicketInfo.VoteLuck > (float64(tx.TicketInfo.BestLuck) - 0.25) {
 					tx.TicketInfo.LuckStatus = "Very Lucky!"
 				} else if tx.TicketInfo.VoteLuck > (float64(tx.TicketInfo.BestLuck) - 0.75) {
--- a/explorer/explorertypes.go
+++ b/explorer/explorertypes.go
@@ -67,9 +67,18 @@ type TxInfo struct {
 }
 
 type TicketInfo struct {
-	TicketMaturity int64
-	PoolStatus     string
-	SpendStatus    string
+	TicketMaturity       int64
+	TimeTillMaturity     float64 // Time before a particular ticket reaches maturity
+	PoolStatus           string
+	SpendStatus          string
+	TicketPoolSize       int64   // Total number of ticket in the pool
+	TicketExpiry         int64   // Total number of blocks before a ticket expires
+	TicketExpiryDaysLeft float64 // Approximate days left before the given ticket expires
+	ShortConfirms        int64   // Total number of confirms up until the point the ticket votes or expires
+	BestLuck             int64   // Best possible Luck for voting
+	AvgLuck              int64   // Average Luck for voting
+	VoteLuck             float64 // Actual Luck for voting on a ticket
+	LuckStatus           string  // Short discription based on the VoteLuck
 }
 
 // TxInID models the identity of a spending transaction input
--- a/views/tx.tmpl
+++ b/views/tx.tmpl
@@ -89,7 +89,7 @@
                 </tr>
                 {{end}}
                 {{if eq .Type "Ticket"}}
-                {{if .TicketInfo.VoteLuck}}
+                {{if (and (.TicketInfo.VoteLuck) (eq .Mature "True"))}}
                 <tr>
                     <td class="text-right pr-2 h1rem p03rem0">LUCK</td>
                     <td>
--- a/views/tx.tmpl
+++ b/views/tx.tmpl
@@ -56,7 +56,7 @@
                 <tr>
                     <td width="90" class="text-right pr-2 h1rem p03rem0 xs-w91">POOL STATUS</td>
                     <td>
-                        {{if ne .TicketInfo.SpendStatus "voted"}} {{.TicketInfo.PoolStatus}}  / {{end}} {{if (index .SpendingTxns 0).Hash}}<a href="/tx/{{(index .SpendingTxns 0).Hash}}">{{.TicketInfo.SpendStatus}}</a>{{else}}{{.TicketInfo.SpendStatus}}{{end}}
+                        {{if ne .TicketInfo.SpendStatus "Voted"}} {{.TicketInfo.PoolStatus}}  / {{end}} {{if (index .SpendingTxns 0).Hash}}<a href="/tx/{{(index .SpendingTxns 0).Hash}}">{{.TicketInfo.SpendStatus}}</a>{{else}}{{.TicketInfo.SpendStatus}}{{end}}
                     </td>
                 </tr>
                 {{end}}
@@ -79,7 +79,7 @@
                                         aria-valuemax="256"
                                     >
                                         <span style="color:#383b41" class="nowrap pl-1 pr-1">
-                                            Immature, {{ if eq .Type "Ticket"}}eligible to vote{{else}}spendable{{end}} in {{ if eq .Confirmations .TicketInfo.TicketMaturity }}next block{{else}}{{subtract (add .TicketInfo.TicketMaturity 1) .Confirmations}} blocks{{end}}
+                                            Immature, {{ if eq .Type "Ticket"}}eligible to vote{{else}}spendable{{end}} in {{ if eq .Confirmations .TicketInfo.TicketMaturity }}next block{{else}}{{subtract (add .TicketInfo.TicketMaturity 1) .Confirmations}} blocks ({{printf "%.1f" .TicketInfo.TimeTillMaturity}} days remaining){{end}}
                                         </span>
                                     </div>
                                 </div>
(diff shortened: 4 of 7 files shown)


explorer/explorerroutes.go
```diff
@@ -180,6 +180,13 @@ func (exp *explorerUI) TxPage(w http.ResponseWriter, r *http.Request) {
 					tx.TicketInfo.PoolStatus = poolStatus.String()
 				}
 				tx.TicketInfo.SpendStatus = spendStatus.String()
+				var averageWinTimeInBlocks int64 = 8192
```
chappjc: I think this depends on the network. Pool size is different on testnet for example

explorer/explorertypes.go
```diff
@@ -67,9 +67,13 @@ type TxInfo struct {
...
 type TicketInfo struct {
-	TicketMaturity int64
-	PoolStatus     string
-	SpendStatus    string
+	TicketMaturity                int64
+	PoolStatus                    string
+	SpendStatus                   string
+	TicketExpiryBlocksLeft        int64
+	TicketExpiryDaysLeft          float64
+	ClosureToAverageWinDaysLeft   float64
```
chappjc: Please document these.  It's not clear what this represents

views/tx.tmpl
```diff
@@ -59,6 +59,16 @@
                         {{if ne .TicketInfo.SpendStatus "voted"}} {{.TicketInfo.PoolStatus}}  / {{end}} {{if (index .SpendingTxns 0).Hash}}<a href="/tx/{{(index .SpendingTxns 0).Hash}}">{{.TicketInfo.SpendStatus}}</a>{{else}}{{.TicketInfo.SpendStatus}}{{end}}
                     </td>
                 </tr>
+                {{if .TicketInfo.TicketExpiryBlocksLeft}}
+                <tr>
+                    <td class="text-right pr-2 h1rem p03rem0">TICKET EXPIRY</td>
+                    <td>{{.TicketInfo.TicketExpiryBlocksLeft}} blocks or (~{{printf "%.2f" .TicketInfo.TicketExpiryDaysLeft}} Days Left)</td>
+                </tr>
+                <tr>
+                    <td class="text-right pr-2 h1rem p03rem0">CLOSURE TO AVERAGE WIN TIME</td>
```
chappjc: I think the word closure here is confusing.  Does this mean remaining time?

explorer/explorerroutes.go
raedah: @McEdward https://github.com/decred/dcrd/blob/master/chaincfg/params.go

views/tx.tmpl
McEdward: I use closure here to depict how close the ticket is to the average win time, the remain time till it gets there

McEdward:
screenshot 35
Also here what it looks like currently

views/tx.tmpl
McEdward: Couldn't really come up with a better word there

explorer/explorerroutes.go
McEdward: will rectify

McEdward:
screenshot 36
made some changes

chappjc:
I'm not sure what word to use, but I think "closure" is going to confuse a lot of people, myself included.

This shows progress to the average number of blocks until a mature ticket votes. What happens to the progress bar when a ticket has been live for twice this period?

If the progress meter hits 100%, what should the user think?  Really that means there is just a 50% chance of that your ticket would have voted by that number of blocks, and if it hasn't, well, unlucky owner.  What about when the progress meter is half way, 2x?

McEdward:
screenshot 37
screenshot 38
This is currently what it does when it is over a 100%, might still want to remove it entirely when the ticket expires

McEdward:
Also would it make sense to remove the progress to average win time after the ticket has voted, and is ok to display confirmations/ticketpoolsize% avg when it is over 100% like the 1st screenshot above

chappjc:
Yeah, that makes sense, except average time to vote should be ticketpoolsize/numvotesperblock

db/dcrsqlite/apisource.go
```diff
@@ -1262,3 +1262,17 @@ func (db *wiredDB) CountUnconfirmedTransactions(address string, maxUnconfirmedPo
 	}
 	return
 }
+func (db *wiredDB) GetNumberOfTicketsPerBlock(idx int) (numberOfTicketsPerBlock int64, err error) {
+	data := db.GetBlockVerbose(idx, true)
+	if data == nil {
+		err = errors.New("Unable to get block for block index " + string(idx))
```
chappjc: Please begin `error` messages without caps

db/dcrsqlite/apisource.go
```diff
@@ -1262,3 +1262,17 @@ func (db *wiredDB) CountUnconfirmedTransactions(address string, maxUnconfirmedPo
 	}
 	return
 }
+func (db *wiredDB) GetNumberOfTicketsPerBlock(idx int) (numberOfTicketsPerBlock int64, err error) {
```
chappjc: Newline and docs please

db/dcrsqlite/apisource.go
```diff
@@ -1262,3 +1262,17 @@ func (db *wiredDB) CountUnconfirmedTransactions(address string, maxUnconfirmedPo
...
+	data := db.GetBlockVerbose(idx, true)
+	if data == nil {
+		err = errors.New("Unable to get block for block index " + string(idx))
+		return
+	}
+	for _, tx := range data.RawSTx {
```
chappjc: Please refactor this loop into `txhelpers` package, but I think something like this exists already

db/dcrsqlite/apisource.go
chappjc: Just `NumberOfTicketsPerBlock` please.  We need to change them all in the future.

McEdward:
screenshot 39

explorer/explorerroutes.go
```diff
@@ -180,6 +180,20 @@ func (exp *explorerUI) TxPage(w http.ResponseWriter, r *http.Request) {
 					tx.TicketInfo.PoolStatus = poolStatus.String()
 				}
 				tx.TicketInfo.SpendStatus = spendStatus.String()
+				tx.TicketInfo.TicketPoolSize = int64(exp.ChainParams.TicketPoolSize)
+				tx.TicketInfo.TicketExpiry = int64(exp.ChainParams.TicketExpiry)
+				averageExpirationDays := (exp.ChainParams.TargetTimePerBlock.Hours() * float64(exp.ChainParams.TicketExpiry)) / 24
```
chappjc: averageExpirationDays -> expirationDays

explorer/explorerroutes.go
```diff
@@ -180,6 +180,20 @@ func (exp *explorerUI) TxPage(w http.ResponseWriter, r *http.Request) {
...
+				averageWinTimeInDay := (exp.ChainParams.TargetTimePerBlock.Hours() * float64(tx.TicketInfo.TicketPoolSize)) / 24
+				averageMaturityInDay := (exp.ChainParams.TargetTimePerBlock.Hours() * float64(tx.TicketInfo.TicketMaturity)) / 24
```
chappjc: averageMaturityInDay -> maturityInDay

explorer/explorerroutes.go
```diff
@@ -180,6 +180,20 @@ func (exp *explorerUI) TxPage(w http.ResponseWriter, r *http.Request) {
...
+				tx.TicketInfo.TimeTillMaturity = ((float64(exp.ChainParams.TicketMaturity) - float64(tx.Confirmations)) / float64(exp.ChainParams.TicketMaturity)) * averageMaturityInDay
+				ticketExpiryBlocksLeft := int64(exp.ChainParams.TicketExpiry) - tx.Confirmations
+				tx.TicketInfo.TicketExpiryDaysLeft = (float64(ticketExpiryBlocksLeft) / float64(exp.ChainParams.TicketExpiry)) * averageExpirationDays
+				tx.TicketInfo.ClosureToAverageWinDaysLeft = averageWinTimeInDay - (averageExpirationDays - tx.TicketInfo.TicketExpiryDaysLeft)
+				tx.TicketInfo.NumberOfVotesPerBlock, err = exp.blockData.GetNumberOfTicketsPerBlock(int(tx.BlockHeight))
+				tx.TicketInfo.ConfirmsToPoolSizePercentage = float64((tx.Confirmations / tx.TicketInfo.TicketPoolSize)) * 100
+				if err != nil {
```
chappjc: what error is this checking?

explorer/explorertypes.go
```diff
@@ -67,9 +67,16 @@ type TxInfo struct {
 }
 
 type TicketInfo struct {
-	TicketMaturity int64
-	PoolStatus     string
-	SpendStatus    string
+	TicketMaturity               int64
+	TimeTillMaturity             float64 //TimeTillMaturity represents the time before a particular ticket reaches maturity
```
chappjc: For consistency with the rest of the codebase and , please use a space after the slashes in the comment

explorer/explorertypes.go
```diff
@@ -67,9 +67,16 @@ type TxInfo struct {
...
+	PoolStatus                   string
+	SpendStatus                  string
+	TicketPoolSize               int64   //TicketPoolSize represents the total number of block before the average win
```
chappjc: Not what TicketPoolSize necessarily.    When taken from chaincfg.Params, it's actually 8192 (40960 / 5), where the "ticket pool" size in tickets is 40960.  The meaning of 8192 is the "number of different ticket numbers".  The idea being that in the ticket pool, 5 tickets at a time are drawn, so you can conceptualize the pool of 40960 tickets as being in 8192 buckets.

If that's not clear, we should consider making `TicketPoolSize` equal to `exp.ChainParams.TicketPoolSize * TicketsPerBlock`, and then compute average time to vote as `TicketPoolSize / TicketsPerBlock`.  It's extra math, but the value of `TicketsPerBlock` is perhaps more intuitive that way.

explorer/explorertypes.go
```diff
@@ -67,9 +67,16 @@ type TxInfo struct {
...
+	NumberOfVotesPerBlock        int64   //NumberOfTicketPerBlock represents the total number of votes in the block
+	TicketExpiry                 int64   //TicketExpiry represents the total number of blocks before a ticket expires
+	ConfirmsToPoolSizePercentage float64 //ConfirmsToPoolSizePercentage represents the percentile average of the confirmations and TicketPoolSize
+	TicketExpiryDaysLeft         float64 //TicketExpiryDaysLeft represents the approximate days left before the given ticket expires
+	ClosureToAverageWinDaysLeft  float64 //ClosureToAverageWinDaysLeft represent approximately how close the ticket is to the average win in days
```
chappjc: I'm still flummoxed by the word "closure" here. @decred/dcrdata suggestions?

explorer/explorertypes.go
McEdward: I'm confused here wouldn't this evaluate back to exp.ChainParams.TicketPoolSize, and what do you mean by average time to vote is it the same as average win time

db/dcrsqlite/apisource.go
McEdward: thinking of removing the function entirely, It currently serves no purpose if there exist this exp.ChainParams.TicketsPerBlock already

explorer/explorertypes.go
McEdward: Because average win time is the how close the ticket is to voting in days

chappjc: Yes, it would evaluate back to exp.ChainParams.TicketPoolSize.  But what I'm pointing out is just that `TicketPoolSize` in the `TicketInfo` struct is either named wrong (should be avg blocks to vote) or the value should be set to exp.ChainParams.TicketPoolSize * exp.ChainParams.TicketsPerBlock.  The units of exp.ChainParams.TicketPoolSize are not in number of tickets, so it's a really misleading variable.

On mainnet, the target ticket pool size is 40960 tickets.  We can estimate number of votes as 5/block, not counting misses.  So it works out that the average time to vote is 8192 = 40960/5 blocks.

chappjc: But a related issue I brought up with @raedah is that there's a different point where your ticket has a 50% chance to vote by. which is not the same as this.  It's a Poisson distribution, not a Gaussian.  So over all tickets, the average time to vote may be 8192, but a given ticket has a 50% change of voting by a time closer to block 5700.

explorer/explorertypes.go
```diff
@@ -72,11 +72,10 @@ type TicketInfo struct {
 	PoolStatus                   string
 	SpendStatus                  string
 	TicketPoolSize               int64   //TicketPoolSize represents the total number of block before the average win
-	NumberOfVotesPerBlock        int64   //NumberOfTicketPerBlock represents the total number of votes in the block
 	TicketExpiry                 int64   //TicketExpiry represents the total number of blocks before a ticket expires
 	ConfirmsToPoolSizePercentage float64 //ConfirmsToPoolSizePercentage represents the percentile average of the confirmations and TicketPoolSize
 	TicketExpiryDaysLeft         float64 //TicketExpiryDaysLeft represents the approximate days left before the given ticket expires
-	ClosureToAverageWinDaysLeft  float64 //ClosureToAverageWinDaysLeft represent approximately how close the ticket is to the average win in days
+	AveWinTime                   float64 //ClosureToAverageWinDaysLeft represent approximately how close the ticket is to the average win in days
```
chappjc: `AveWinTimeETA` perhaps?  be sure to update the comment accordingly

explorer/explorertypes.go
McEdward: I don't seem to understand how to calculate it, Could you please explain the formula

explorer/explorertypes.go
```diff
@@ -67,15 +67,15 @@ type TxInfo struct {
...
-	ConfirmsToPoolSizePercentage float64 //ConfirmsToPoolSizePercentage represents the percentile average of the confirmations and TicketPoolSize
-	TicketExpiryDaysLeft         float64 //TicketExpiryDaysLeft represents the approximate days left before the given ticket expires
-	AveWinTime                   float64 //ClosureToAverageWinDaysLeft represent approximately how close the ticket is to the average win in days
+	TicketMaturity          int64
+	TimeTillMaturity        float64 //TimeTillMaturity represents the time before a particular ticket reaches maturity
+	PoolStatus              string
+	SpendStatus             string
+	TicketPoolSize          int64   //TicketPoolSize represents the total number of ticket in the pool
+	TicketExpiry            int64   //TicketExpiry represents the total number of blocks before a ticket expires
+	ConfirmsToAveTimeToVote float64 //ConfirmsToAveTimeToVote represents the percentile average of the confirmations and AveTimeToVote
```
chappjc: not "percentile".  This is the number progress, as a percent, toward the average number of block until voting

explorer/explorertypes.go
```diff
@@ -67,15 +67,15 @@ type TxInfo struct {
...
+	TicketExpiryDaysLeft    float64 //TicketExpiryDaysLeft represents the approximate days left before the given ticket expires
+	AveTimeToVote           int64   //AveTimeToVote represents the total number of block before the average win
```
chappjc: "the average number of blocks tickets are live before voting"

explorer/explorertypes.go
```diff
@@ -67,15 +67,15 @@ type TxInfo struct {
...
-	TimeTillMaturity             float64 //TimeTillMaturity represents the time before a particular ticket reaches maturity
-	PoolStatus                   string
-	SpendStatus                  string
-	TicketPoolSize               int64   //TicketPoolSize represents the total number of block before the average win
-	TicketExpiry                 int64   //TicketExpiry represents the total number of blocks before a ticket expires
-	ConfirmsToPoolSizePercentage float64 //ConfirmsToPoolSizePercentage represents the percentile average of the confirmations and TicketPoolSize
-	TicketExpiryDaysLeft         float64 //TicketExpiryDaysLeft represents the approximate days left before the given ticket expires
-	AveWinTime                   float64 //ClosureToAverageWinDaysLeft represent approximately how close the ticket is to the average win in days
+	TicketMaturity          int64
+	TimeTillMaturity        float64 //TimeTillMaturity represents the time before a particular ticket reaches maturity
```
chappjc: in general, I think comments on the same line like this can omit the variable name part and just start with the description (e.g. "the time before..")

explorer/explorertypes.go
```diff
@@ -67,15 +67,15 @@ type TxInfo struct {
...
+	PoolStatus              string
+	SpendStatus             string
+	TicketPoolSize          int64   //TicketPoolSize represents the total number of ticket in the pool
```
chappjc: well, this is the target ticket pool size, right?  We're not actually putting the current ticket pool size in this variable.

explorer/explorerroutes.go
```diff
@@ -180,19 +180,15 @@ func (exp *explorerUI) TxPage(w http.ResponseWriter, r *http.Request) {
 					tx.TicketInfo.PoolStatus = poolStatus.String()
 				}
 				tx.TicketInfo.SpendStatus = spendStatus.String()
-				tx.TicketInfo.TicketPoolSize = int64(exp.ChainParams.TicketPoolSize)
+				tx.TicketInfo.TicketPoolSize = int64(exp.ChainParams.TicketPoolSize) * int64(exp.ChainParams.TicketsPerBlock)
+				tx.TicketInfo.AveTimeToVote = int64(exp.ChainParams.TicketPoolSize) / int64(exp.ChainParams.TicketsPerBlock)
```
chappjc: My last comment was referring to:
`tx.TicketInfo.TicketPoolSize / int64(exp.ChainParams.TicketsPerBlock)`
As you said, this does reduce to `int64(exp.ChainParams.TicketPoolSize)` because of the odd semantics used in `chaincfg`.

I just though it was important to have `tx.TicketInfo.TicketPoolSize` be 40960, and `tx.TicketInfo.AveTimeToVote` be 8192.

Initially you had `tx.TicketInfo.TicketPoolSize = int64(exp.ChainParams.TicketPoolSize)`, which was confusing because this was 8192, what we wanted to call AveTimeToVote.

db/dcrsqlite/apisource.go
```diff
@@ -1262,3 +1262,19 @@ func (db *wiredDB) CountUnconfirmedTransactions(address string, maxUnconfirmedPo
 	}
 	return
 }
+
+//HeightOfSpendingTxForVotes gives the block height of the transaction id specified
```
chappjc: In all comments, put a space after the slashes (`// Height..`)

db/dcrsqlite/apisource.go
```diff
@@ -1262,3 +1262,19 @@ func (db *wiredDB) CountUnconfirmedTransactions(address string, maxUnconfirmedPo
...
+func (db *wiredDB) HeightOfSpendingTxForVotes(txid string) (height int64) {
```
chappjc: This should be renamed as it has nothing to do with votes or spending txns.  `TxHeight` would be more appropriate.

explorer/explorertypes.go
```diff
@@ -67,15 +67,18 @@ type TxInfo struct {
...
-	TimeTillMaturity        float64 //TimeTillMaturity represents the time before a particular ticket reaches maturity
-	PoolStatus              string
-	SpendStatus             string
-	TicketPoolSize          int64   //TicketPoolSize represents the total number of ticket in the pool
-	TicketExpiry            int64   //TicketExpiry represents the total number of blocks before a ticket expires
-	ConfirmsToAveTimeToVote float64 //ConfirmsToAveTimeToVote represents the percentile average of the confirmations and AveTimeToVote
-	TicketExpiryDaysLeft    float64 //TicketExpiryDaysLeft represents the approximate days left before the given ticket expires
-	AveTimeToVote           int64   //AveTimeToVote represents the total number of block before the average win
+	TicketMaturity       int64
+	TimeTillMaturity     float64 //represents the time before a particular ticket reaches maturity
```
chappjc: `// the time before...`
(add space, remove "represents" as that is implied)

chappjc:
Also needs a rebase on master.  All the imports have changed so when you finish with rebase, remember to do `dep ensure` to get your vendors in order.

chappjc:
OK

chappjc:
Please:

1. remove "dcrdata.exe~" from the PR w/ `git rm dcrdata.exe~`.
2. squash (`git rebase -i 3503ec7f5`) "pick" only the first commit and "fix" the rest.  Note this commit comes from `git merge-base origin/master HEAD`.
3. rebase on master (`git rebase origin/master`)

chappjc:
Looks like it needs special handling for immature tickets:

image

Also, for everyone's info, here's the luck on a voted ticket:

image

Would be cool to have some kind of LUCK for an "expired / revoked" ticket, like "0, crappy".  But that can be a future PR.

explorer/explorerroutes.go
```diff
@@ -202,6 +202,39 @@ func (exp *explorerUI) TxPage(w http.ResponseWriter, r *http.Request) {
...
+				if tx.TicketInfo.VoteLuck >= float64(tx.TicketInfo.BestLuck-(1/int64(exp.ChainParams.TicketPoolSize))) {
+					tx.TicketInfo.LuckStatus = "Perfection"
+				} else if tx.TicketInfo.VoteLuck > (float64(tx.TicketInfo.BestLuck) - 0.25) {
+					tx.TicketInfo.LuckStatus = "Very Lucky!"
+				} else if tx.TicketInfo.VoteLuck > (float64(tx.TicketInfo.BestLuck) - 0.75) {
+					tx.TicketInfo.LuckStatus = "Good Luck"
+				} else if tx.TicketInfo.VoteLuck > (float64(tx.TicketInfo.BestLuck) - 1.25) {
+					tx.TicketInfo.LuckStatus = "Normal"
+				} else if tx.TicketInfo.VoteLuck > (float64(tx.TicketInfo.BestLuck) * 0.50) {
+					tx.TicketInfo.LuckStatus = "Bad Luck"
```
chappjc: Having tested and taken a closer look here, I think we need to reconsider the luck categories proposed by @raedah here: https://github.com/decred/dcrdata/issues/375#issuecomment-369789909

In [this ticket](http://127.0.0.1:7777/tx/ef38d477563797320294e0f374b6a40b961c1cadaa95faf018fd1b4de3955a57), it says:

> LUCK | 3.6572                          final, Bad Luck                          (Best: 5 Avg: 4)

For an average of 4.0, I wouldn't consider 3.65722 to be "Bad Luck".  But this number is between 3.75 and 2.5.

I suggest we change the category boundaries to be linearly spaced from 0 to 5.  So we can just define each as a multiple of (5) best luck.  To the boundaries:

<pre> 0,  >0          1          2         3         4         <5, 5
no,bad      |   fair    |  ave.   |   good  |      great, perfect</pre>

chappjc:
Another attempt at those boundaries.

Say we have 6 luck labels: {'great', 'good', 'average', 'fair', 'bad', 'terrible'}
The absolute ends could be tagged 'perfect' and 'zip'.
Spacing linearly (except for ends) on **probability of not having voted by a given block** (`p`), and computing the live block count at those values (`x`):

    p = 0.0067359   0.1666667   0.3333333   0.5000000   0.6666667   0.8333333   0.9998779
    x = log(bounds)./log(8191/8192)
     ~=   40960       14677      8999.3      5677.9      3321.4      1493.5         1

image

Or 5 bins:
image

Essentially, for a given ticket at live block `t`, we could:

1. use the computed `x` values as thresholds on `t` for the luck labels
2. compute `p = (8191/8192)^t`, and use that as luck (e.g. a vote at live block #1 would have a luck of 0.9998779, while at the last live block before expiration the luck is 0.0067359)

chappjc:
And if the word "average" hints at a statistical significance, we could use "ok" or "meh".

explorer/explorerroutes.go
raedah: testnet and simnet go to 6

raedah: "I wouldn't consider 3.65722 to be Bad Luck"
Why is that?  Its quite a ways beyond the 28.44 day point at which half the tickets have voted. Looks like it took about 38.19 days for this ticket.

chappjc: Yeah, scratch this.  Have a look at my comment with they plots and let me know what you think, please.

raedah:
The last live block before expiration does not have any special significance. The worse case is expiration, not voting at all.

chappjc:
@raedah The only significance of the last block before expiration is just that it's the ticket's last chance to vote.  I'm giving a number there because its the boundary of the live block range we're concerned with.

raedah:
Using log looks like the better way to do it, then we can flush out the label wording.

chappjc:
as per slack discussion, merging now with intent of refining in the future.

PR #379, opened 2018-03-06
Add websocket tx send buffer

RogueElement:
Follow up to #343:
This add a buffer `WebsocketHub.newTxBuffer` of size `newTxBufferSize`
for the websocket new tx sends. The websocket clients are now of type
`client` (`[]*MempoolTx` and `sync.RwMutex`) to accommodate the bulk sends.
A ticker function, `WebsocketHub.periodicBufferSend()`, is sent on a
goroutine at the start of `WebsocketHub.run()` that sends the buffer every
`bufferTickerInterval` seconds. The buffer is cleared on every send.

### DIFF
--- a/explorer/websocket.go
+++ b/explorer/websocket.go
@@ -9,10 +9,14 @@ import (
 )
 
 const (
-	wsWriteTimeout           = 10 * time.Second
-	wsReadTimeout            = 12 * time.Second
-	pingInterval             = 12 * time.Second
-	sigNewBlock    hubSignal = iota
+	wsWriteTimeout     = 10 * time.Second
+	wsReadTimeout      = 12 * time.Second
+	pingInterval       = 12 * time.Second
+	tickerSigReset int = iota
+	tickerSigStop
+	bufferTickerInterval           = 5
+	newTxBufferSize                = 5
+	sigNewBlock          hubSignal = iota
 	sigMempoolUpdate
 	sigPingAndUserCount
 	sigNewTx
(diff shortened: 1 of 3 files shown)


explorer/websocket.go
```diff
@@ -164,3 +184,58 @@ func (wsh *WebsocketHub) run() {
...
+		return
+	}
+	txs := make([]*MempoolTx, len(wsh.newTxBuffer))
+	copy(txs, wsh.newTxBuffer)
+	wsh.newTxBuffer = make([]*MempoolTx, 0, newTxBufferSize)
+	wsh.RUnlock()
+	log.Debugf("Signaling %d new tx to %d clients", len(txs), len(wsh.clients))
+	for signal, client := range wsh.clients {
+		client.RLock()
+		client.newTxs = txs
```
chappjc: All clients get the same slice of transactions, so i suggest a different approach:  use the `WebsocketHub`'s `run()` loop to handle a lot of these concurrency issues.  That is, follow the pattern like `RegisterClient`/`registerClient` where the former sends on a channel only read by `run`, and the latter is called when `run` gets to it.

The main problem that needs to be addresses with this PR is that `(wsh *WebsocketHub) sendNewTxs()`, which ranges over `wsh.clients`, is launched as a goroutine and runs concurrently with the main `run` loop, which registers and unregisters clients in `wsh.clients`.

I'm open to any solution, it doesn't need to be the all-event-loop idea I mentioned here, but concurrent map access (clients) needs to be addressed.

explorer/websocket.go
```diff
@@ -126,12 +132,6 @@ func safeClose(cc hubSpoke) {
 func (wsh *WebsocketHub) Stop() {
 	// end the run() loop, allowing in progress operations to complete
 	wsh.quitWSHandler <- struct{}{}
-	// end the buffer interval send loop
-	wsh.bufferTickerChan <- tickerSigStop
-	// unregister all clients
-	for client := range wsh.clients {
```
RogueElement: This ranges over `wsh.clients` , is only called from the main go routine and so causes a data race. Moved it to the main run loop.

chappjc:
utACK.  Needs an independent test and it's good to go

chappjc:
tOK

PR #380, opened 2018-03-06
explorer/mempool: Ignore mempool votes not for the tip block

RogueElement:
Another follow up to #343. Still [WIP] because I have no idea how identify votes spending the same ticket so I can combine them. Currently reduces vote count to around 5-7.

### DIFF
--- a/explorer/mempool.go
+++ b/explorer/mempool.go
@@ -15,6 +15,14 @@ import (
 
 func (exp *explorerUI) mempoolMonitor(txChan chan *NewMempoolTx) {
 	exp.storeMempoolInfo()
+	exp.NewBlockDataMtx.Lock()
+	lastBlockHash, err := exp.blockData.GetBlockHash(exp.NewBlockData.Height)
+	if err != nil {
+		log.Warnf("Could not get bloch hash for last block, mempool vote filter disabled")
+	}
+	lastBlockTime := exp.NewBlockData.BlockTime
+	exp.NewBlockDataMtx.Unlock()
+
 	for {
 		ntx, ok := <-txChan
 		if !ok {
@@ -30,15 +38,18 @@ func (exp *explorerUI) mempoolMonitor(txChan chan *NewMempoolTx) {
 		// A tx with an empty hex is the new block signal
 		if ntx.Hex == "" {
 			exp.storeMempoolInfo()
+			exp.NewBlockDataMtx.Lock()
+			lastBlockHash, err = exp.blockData.GetBlockHash(exp.NewBlockData.Height)
+			if err != nil {
+				log.Warnf("Could not get bloch hash for last block")
+			}
+			lastBlockTime = exp.NewBlockData.BlockTime
+			exp.NewBlockDataMtx.Unlock()
 			exp.wsHub.HubRelay <- sigMempoolUpdate
 			continue
 		}
 
 		// Ignore this tx if it was received before the last block
-		exp.NewBlockDataMtx.Lock()
-		lastBlock := exp.NewBlockData.Height
-		lastBlockTime := exp.NewBlockData.BlockTime
-		exp.NewBlockDataMtx.Unlock()
 
 		if ntx.Time < lastBlockTime {
 			continue
@@ -54,7 +65,7 @@ func (exp *explorerUI) mempoolMonitor(txChan chan *NewMempoolTx) {
 		var voteInfo *VoteInfo
 		if ok := stake.IsSSGen(msgTx); ok {
 			validation, version, bits, choices, err := txhelpers.SSGenVoteChoices(msgTx, exp.ChainParams)
-			if validation.Height != lastBlock {
+			if validation.Hash.String() != lastBlockHash && lastBlockHash != "" {
 				continue
 			}
 			if err != nil {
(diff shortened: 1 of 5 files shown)


explorer/mempool.go
```diff
@@ -53,6 +54,9 @@ func (exp *explorerUI) mempoolMonitor(txChan chan *NewMempoolTx) {
 		var voteInfo *VoteInfo
 		if ok := stake.IsSSGen(msgTx); ok {
 			validation, version, bits, choices, err := txhelpers.SSGenVoteChoices(msgTx, exp.ChainParams)
+			if validation.Height != lastBlock {
```
chappjc: Need to compare hash.  side chain may be at same height

chappjc:
To get the ticket hash being spent, you need to check `msgTx.TxIn[1].PreviousOutpoint.Hash`.  I have it getting this for debug purposes in the `TxHandler` run loop:

https://github.com/decred/dcrdata/blob/f0bd65ff4de03d407deb8e373a9bc0b2e49db7c0/mempool/mempool.go#L156-L158

In `explorer`'s mempool loop, this should be fairly easy to add. However, we still want to show all votes spending the same ticket.  It's just that it would be helpful to show which ticket the votes are spending.  I think it would even be helpful if the 5 winning tickets were just assigned numbers from 1-5 and this number shown somewhere for each vote.  So multiple vote txn rows could have the same ticket index.

RogueElement:
Votes now show ticket spent and index.
These two votes...
screenshot 33
spend the same ticket.
screenshot 32

chappjc:
Perfect.  I love the hover text solution.

explorer/mempool.go
```diff
@@ -15,6 +15,14 @@ import (
 
 func (exp *explorerUI) mempoolMonitor(txChan chan *NewMempoolTx) {
 	exp.storeMempoolInfo()
+	exp.NewBlockDataMtx.Lock()
+	lastBlockHash, err := exp.blockData.GetBlockHash(exp.NewBlockData.Height)
+	if err != nil {
+		log.Warnf("Could not get bloch hash for last block, mempool vote filter disabled")
```
chappjc: bloch --> block

explorer/mempool.go
```diff
@@ -30,14 +38,18 @@ func (exp *explorerUI) mempoolMonitor(txChan chan *NewMempoolTx) {
 		// A tx with an empty hex is the new block signal
 		if ntx.Hex == "" {
 			exp.storeMempoolInfo()
+			exp.NewBlockDataMtx.Lock()
+			lastBlockHash, err = exp.blockData.GetBlockHash(exp.NewBlockData.Height)
+			if err != nil {
+				log.Warnf("Could not get bloch hash for last block")
```
chappjc: Spelling

explorer/mempool.go
```diff
@@ -147,11 +170,31 @@ func (exp *explorerUI) storeMempoolInfo() {
...
+	lastBlockTime := exp.NewBlockData.BlockTime
+	exp.NewBlockDataMtx.Unlock()
+
+	txindexes := make(map[string]int)
 	for _, tx := range memtxs {
 		switch tx.Type {
 		case "Ticket":
 			tickets = append(tickets, tx)
 		case "Vote":
+			if lastBlockHash != "" && tx.VoteInfo.Validation.Hash != lastBlockHash {
```
chappjc: A comment here explaining intentions please

explorer/mempool.go
```diff
@@ -147,11 +172,31 @@ func (exp *explorerUI) storeMempoolInfo() {
 	var totalOut float64
 	var totalSize int32
 
+	exp.NewBlockDataMtx.Lock()
```
chappjc: Consider whether a full read-write lock (as with `Lock()`) is needed or if a read lock is good here (`RLock()`).  And please move the unlock as close as possible to the lock.  That is, can it go before the call to `GetBlockHash`, which may have slow IO.

explorer/mempool.go
```diff
@@ -147,11 +172,31 @@ func (exp *explorerUI) storeMempoolInfo() {
...
+	lastBlock := exp.NewBlockData.Height
+	lastBlockHash, err := exp.blockData.GetBlockHash(lastBlock)
+	if err != nil {
+		log.Warnf("Could not get block hash for last block, mempool vote filter disabled")
+	}
+	lastBlockTime := exp.NewBlockData.BlockTime
+	exp.NewBlockDataMtx.Unlock()
```
chappjc: Since the above block of new code is identical to the code in `mempoolMonitor` (except for the message), please make a new `(exp *explorerUI)` method to get the lastBlock* stuff.

explorer/mempool.go
```diff
@@ -187,7 +227,19 @@ func (exp *explorerUI) storeMempoolInfo() {
 		NumRevokes:         len(revs),
 		LatestTransactions: latest,
 		FormattedTotalSize: humanize.Bytes(uint64(totalSize)),
+		TicketIndexes:      txindexes,
+	}
+}
+
+func voteForLastBlock(blockHash, validationHash string, blockHeight, validationHeight int64) bool {
```
chappjc: Only compare hash, not height.  Equal height does not mean it's the same block.

explorer/mempool.go
```diff
@@ -147,11 +172,31 @@ func (exp *explorerUI) storeMempoolInfo() {
...
+	lastBlockTime := exp.NewBlockData.BlockTime
+	exp.NewBlockDataMtx.Unlock()
+
+	txindexes := make(map[string]int)
 	for _, tx := range memtxs {
 		switch tx.Type {
 		case "Ticket":
 			tickets = append(tickets, tx)
 		case "Vote":
+			if !voteForLastBlock(lastBlockHash, tx.VoteInfo.Validation.Hash, lastBlock, tx.VoteInfo.Validation.Height) {
```
chappjc: As in the other comment, I do not think a height comparison is correct, only hash.  So, `voteForLastBlock` can be removed.

If `lastBlockHash` is empty/nil, skip all this and just do the `votes` append.

PR #383, opened 2018-03-06
name the go-humanize import

chappjc:
Some linters are confused by the go-humanize import when it is missing
an explicit name of the package, humanize. Fix this by naming it
humanize in the source where it's not already done like that.


chappjc:
Actually I think I ran into trouble with `dep init` because of this.

Issue #377, opened 2018-03-02
Address view filter


raedah:
On the address view, include a filter selector.
All
Credits
Debits
SSTX

Function in lite mode and full mode.

chappjc:
💯
One I'd like to hit up, given a chance.  I'll see

chappjc:
PR #385 does it for full mode.  Lite mode needs an entirely different approach since all we can do is searchrawtransactions, not arbitrary queries.

raedah:
For lite mode I suggest pulling up to x (1000) and filtering those. This would be in line with existing policy and warnings displayed to users, and would be useful for the large majority of cases that have less then 1000 txs in an address. Though I dont consider extending features on litemode a high priority.

chappjc:
That would work, but it would hobble the address page.  Doing a search for with a limit of 1000 it slow when there are at least 1000 to look through.

For addresses w/ many funding transactions and not so many spending transactions, it would be bad no matter how clever we got with starting small and increasing the limit since it would have to repeatedly fetch them all just to see if there are spending transactions, not just funding.  Eventually it would have to search up to 1000 just to see that there aren't any.

Issue #376, opened 2018-03-02
Pagination params in url disappear on address page


raedah:
These should stay persistent so that a user can link to a specific view.

chappjc:
Related to issues encountered with address filter PR #385

chappjc:
dup of issue #https://github.com/decred/dcrdata/issues/408, and addressed in PR #430

PR #385, opened 2018-03-07
address page: credit/debit txn filter

chappjc:
This provides a major overhaul to the full-mode address page.   The main introduced change is, a new drop-down listbox on the address page with the following options: "All", "Credit", and "Debit".

Other related changes:

The page size logic in the address html template is also simplified.
The problem of URL query params disappearing was related to the solution we used for handling redirects from the search page.  Instead of:

```go
w.Header().Set("Turbolinks-Location", "/address/"+address)
```

we should use:

```go
w.Header().Set("Turbolinks-Location", r.URL.RequestURI())
```

This will include the path, as before, but also include the URL queries (e.g. `?txntype=credit&n=20&start=0`).

Resolves issue #377 .

### DIFF
--- a/db/dbtypes/types.go
+++ b/db/dbtypes/types.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"strconv"
+	"strings"
 
 	"github.com/decred/dcrdata/db/dbtypes/internal"
 )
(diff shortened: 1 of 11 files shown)


chappjc:
@gozart1 It should be running well enough now to play with the URL/turbolinks/drop-down issues that I can't figure out.  Thanks!

chappjc:
Updated PR description and title.  Ready for review.

chappjc:
Running this on beta.dcrdata.org

chappjc:
hmm, this needs to handle unconfirmed txns better

chappjc:
The last commit adds a number of changes to better support unconfirmed transactions in full-mode.  The main PG table only includes confirmed transactions, so `getrawmempool` is used to fetch unconfirmed transactions and new functions in `txhelpers` are used to scan the inputs and outputs.

chappjc:
This resolves the following mostly-related issues:  https://github.com/decred/dcrdata/issues/377, https://github.com/decred/dcrdata/issues/408 (on the address page), https://github.com/decred/dcrdata/issues/376, https://github.com/decred/dcrdata/issues/393,

chappjc:
Removed unconfirmed txn machinery, moved to PR #422 .  This PR is just the cred/debit filter again.

Issue #387, opened 2018-03-11
check dcrd network id against dcrdata config


chappjc:
when connecting to dcrd, the expected network should be verified.

For instance, it is possible to start with `dcrdata --testnet --dcrdserv 127.0.0.1:9109` and dcrdata will try treat the network as testnet and try to operated.  As noted by port 9109, this is likely a mainnet node.  This is an easy mistake to make as the dcrdata.conf may contain the `dcrdserv` setting, while the user attempts to change to testnet with just the `--testnet` flag.

To prevent this, dcrdata should use `getcurrentnet` to ensure that the network matches the expected net.

chappjc:
Resolved by PR #388

PR #388, opened 2018-03-11
main: check network is the expected one

chappjc:
This prevents dcrdata from using a network node for a network (mainnet, testnet, simnet) that does not match the network specified with dcrdata's flags (e.g. --mainnet).

Previously, one could use `--testnet`, while the config file specifed `dcrdserv=host:19109` (i.e. likely a testnet node) and dcrdata would do it's best, eventually throwing hard to diagnose errors.

Now, dcrdata refuses to operate:

```
[CRT] DATD: Network of connected node, TestNet2, does not match expected network, MainNet.
[INF] DATD: Closing connection to dcrd.
[INF] DATD: Bye!
[ERR] DATD: expected network TestNet2, got MainNet
```

This resolves issue #387 .

### DIFF
--- a/main.go
+++ b/main.go
@@ -102,6 +102,12 @@ func mainCore() error {
 	log.Infof("Connected to dcrd (JSON-RPC API v%s) on %v",
 		nodeVer.String(), curnet.String())
 
+	if curnet != activeNet.Net {
+		log.Criticalf("Network of connected node, %s, does not match expected "+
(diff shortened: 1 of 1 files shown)


chappjc:
@dajohi can you have a look when you get a chance?  BTW, what's good etiquette for maintainers requesting review?

raedah:
utACK

PR #395, opened 2018-03-15
explorer: refactor the explorer templates so they are easier to manage

RogueElement:
Simple refactoring to make it easier to add templates since the current system is more complicated that it should be.

### DIFF
--- a/explorer/templates.go
+++ b/explorer/templates.go
@@ -1,0 +1,221 @@
+// Copyright (c) 2017, The dcrdata developers
+// See LICENSE for details.
+
+package explorer
+
+import (
+	"bytes"
+	"fmt"
+	"html/template"
+	"math"
+	"path/filepath"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/decred/dcrd/chaincfg"
+	humanize "github.com/dustin/go-humanize"
+)
+
+type pageTemplate struct {
+	file     string
+	template *template.Template
+}
+
+type templates struct {
+	templates map[string]pageTemplate
+	defaults  []string
+	folder    string
+	helpers   template.FuncMap
+}
+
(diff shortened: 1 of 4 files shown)


explorer/templates.go
```diff
@@ -0,0 +1,221 @@
...
+		helpers:   helpers,
+	}
+}
+
+func (t templates) addTemplate(name string) error {
+	fileName := filepath.Join(t.folder, name+".tmpl")
+	files := append(t.defaults, fileName)
+	temp, err := template.New(name).Funcs(t.helpers).ParseFiles(files...)
+	if err == nil {
+		t.templates[name] = pageTemplate{
```
chappjc: value receiver OK in this function?

explorer/templates.go
```diff
@@ -0,0 +1,221 @@
...
+	case int16:
+		return int64(vt)
+	case uint16:
+		return int64(vt)
+	default:
+		return math.MinInt64
+	}
+}
+
+func makeTempHelpers(params *chaincfg.Params) template.FuncMap {
```
chappjc: just a nitpick, but I think it would be clearer if the function were named `makeTemplateFuncMap`.

explorer/templates.go
RogueElement: I don't follow

chappjc: I'm referring to the use of `(t templates)` instead of `(t *templates)`

RogueElement: Oh, yeah... a pointer receiver would be better since it could mutate `t.templates`. Changed all `t` receivers for consistency.

chappjc: BTW, I think this would have worked fine because of the "true" nature of maps in go, but I think it was right to change it to a pointer receiver.

FYI https://dave.cheney.net/2017/04/30/if-a-map-isnt-a-reference-variable-what-is-it

Issue #390, opened 2018-03-12
rebuilddb2: drop tables flag loads the ticket pool db first


RogueElement:
I don't think it needs to do this or connect to dcrd just to drop table.

chappjc:
resolved by pr #396

PR #396, opened 2018-03-16
ChainDB/rebuilddb2: add to ChainDB the ability to set a StakeDatabase

chappjc:
Modify rebuilddb2 to load the `StakeDatabase` (and ticket pool db) *after*
creating the `ChainDB` so that it is not necessary to load the ticket pool
db when only using `ChainDB` to drop tables.

Add "rebuild_data/" to .gitignore.


chappjc:
Resolves Issue #390

chappjc:
Hmm, I'll check out the version flag

chappjc:
Version flag should be fixed now.

Issue #308, opened 2017-12-29
Chain Parameter page


raedah:
Create a page for displaying data about the current chain parameters.
Yup, the stuff in chaincfg/params.go
Should not focus on the simple variable settings, but instead display the information that is the most useful for understanding the consensus rules in a simple human understandable way.
This information almost never changes, but does differ between mainnet, testnet, simnet.

PR #397, opened 2018-03-17
Creating a chainparameters page

McEdward:
This creates a new page @"/chainparameters" which is for displaying all the chain parameters, values and description.
resolving #308


McEdward:
screenshot 52
What it currently looks like

chappjc:
Remove:

- Default port
- Ticket pool size weight, unless we have a much better explanation
- (maybe) Retarget adjustment factor

Rename:
- target time span  --> PoW retarget interval
- SStx change maturity --> ticket purchase change maturity
- maximum block sizes --> maximum block size

Other changes:
- "ticket pool size" should show 40960, so compute it as the current value mult. by tickets per  block.  No one will understand the buckets part.
- move the maturity parameters next to each other
- clock time formatting needs work (e.g. 5m0s -> 5 min.)
- only show the first block size in the slice (the pre-vote block size).  Ideally, we'd look at the results of any agenda affecting this and show the current state, but this can be put off for another day.

In general, please work on making the descriptions **less verbose**.  For example, MAXIMUM TRANSACTION SIZE could be described as "size of largest allowable transaction".  Do not start a description with "defined", "this is", "they are", "is", etc.  They are all descriptions.

Instead of "desired", say "target".
"Maturity for spending" is redundant.  Saying just "maturity" carries the same meaning.
When describing maturity for tickets, also explain that maturity is when the ticket is considered "live".

chappjc:
Also add:
- block/stake/tax reward proportions.
- HD coin type
- Organization PkScript and decoded address
- link to genesis block (0)
- link to block 1, which includes the airdrop

raedah:
>target time span --> PoW retarget interval
It is also the PoS retarget interval.

Widen the value row so that information stays on one line.

explorer/explorerroutes.go
```diff
@@ -497,10 +498,31 @@ func (exp *explorerUI) NotFound(w http.ResponseWriter, r *http.Request) {
...
+	cp.MaximumBlockSize = int64(exp.ChainParams.MaximumBlockSizes[0])
+	cp.CoinbaseMaturity = int64(exp.ChainParams.CoinbaseMaturity)
+	cp.SStxChangeMaturity = int64(exp.ChainParams.SStxChangeMaturity)
+	cp.TicketMaturity = int64(exp.ChainParams.TicketMaturity)
+	cp.TicketPoolSize = int64(exp.ChainParams.TicketPoolSize * exp.ChainParams.TicketsPerBlock)
+	cp.TicketsPerBlock = int64(exp.ChainParams.TicketsPerBlock)
+	cp.TicketExpiry = int64(exp.ChainParams.TicketExpiry)
+	cp.StakeRewardProportion = int64(exp.ChainParams.StakeRewardProportion)
+	cp.HDCoinType = int64(exp.ChainParams.HDCoinType)
+	cp.OrganizationPkScript = exp.ChainParams.OrganizationPkScript
```
chappjc: For the above kind of initialization, please do:

```
cp := ChainParams{
    Net:           exp.ChainParams.Net,
    MaxTxSize:     int64(exp.ChainParams.MaxTxSize),
    ...
}
```

views/chainparameters.tmpl
```diff
@@ -29,70 +29,85 @@
...
-                                <td class="lh1rem">RetargetAdjustmentFactor is the adjustment factor used to limit the minimum and maximum amount of adjustment that can occur between difficulty retargets.</td>
+                                <td class="lh1rem">Amount of time to generate each block.</td>
                             </tr>
                             <tr>
-                                <td class="text-right pr-2 lh1rem nowrap p03rem0">MAXIMUM BLOCK SIZES</td>
-                                <td class="lh1rem">{{.MaximumBlockSizes}} b</td>
-                                <td class="lh1rem">The maximum size a block can get to</td>
+                                <td class="text-right pr-2 lh1rem nowrap p03rem0">MAXIMUM BLOCK SIZE</td>
+                                <td class="lh1rem">{{.MaximumBlockSize}} b</td>
+                                <td class="lh1rem">Size a block can get to</td>
```
chappjc: Largest allowable block size

views/chainparameters.tmpl
```diff
@@ -29,70 +29,85 @@
...
                             <tr>
-                                <td class="text-right pr-2 lh1rem nowrap p03rem0">SStxChange MATURITY</td>
+                                <td class="text-right pr-2 lh1rem nowrap p03rem0">TICKET PURCHASE CHANGE MATURITY</td>
                                 <td class="lh1rem">{{.SStxChangeMaturity}}</td>
                                 <td class="lh1rem">Maturity for spending SStx change outputs</td>
                             </tr>
+                            <tr>
+                                <td class="text-right pr-2 lh1rem nowrap p03rem0">TICKET MATURITY</td>
+                                <td class="lh1rem">{{.TicketMaturity}} blocks</td>
+                                <td class="lh1rem">Number of blocks for tickets to mature</td>
```
chappjc: ..to mature (enter the live ticket pool)

views/chainparameters.tmpl
```diff
@@ -29,70 +29,85 @@
...
+                            </tr>
                             <tr>
                                 <td class="text-right pr-2 lh1rem nowrap p03rem0">TICKET POOL SIZE</td>
                                 <td class="lh1rem">{{.TicketPoolSize}}</td>
-                                <td class="lh1rem">This denotes the number of possible buckets/number of different ticket numbers</td>
+                                <td class="lh1rem">Number of possible number of different ticket numbers</td>
```
chappjc: As I mentioned before, we really want to show the value of TicketPoolSize x TicketsPerBlock.  No one will get the ticket numbers part.  With this being 40960, it can simply be described as "Target size of ticket pool"

raedah:
[ERR] EXPR: Unable to create new html template: open views/chainParameters.tmpl: no such file or directory
views/chainparameters.tmpl

raedah:
-Need a link to the chain parameters page, we've been putting these on the footer.
-Looks like we are using a different font. Should be the same font we use on the rest of the site.
-Descriptions are sentences and should end with a period.
-Since we are showing Stake Reward Proportion, we should WorkRewardProportion and BlockTaxProportion too.
-Lots of additional params can be shown that have not yet been included. We can include subsections for chain parameters, subsidy parameters, etc to correspond with what is shown in the params.go file. Several params exist for voting RuleChange.
-Many of the description are not clear and will be updated.

chappjc:
If gozart's design gets in, there will be plenty of space in the menu.  The footer is super full.

explorer/explorertypes.go
```diff
@@ -343,3 +325,74 @@ type NewMempoolTx struct {
...
+	var SimnetPrefixes = []string{"Sk", "Ss", "Se", "SS", "Sc", "Ps", "sprv", "spub"}
+	name := params.Name
+	if name == "mainnet" {
+		for i, desc := range Descriptions {
+			addrPrefix = append(addrPrefix, AddrPrefix{
+				Name:        Name[i],
+				Description: desc,
+				Prefix:      MainnetPrefixes[i],
+			})
+		}
```
chappjc: Note that only `Prefix` is different, so please simplify:

```go
var netPrefixes []string
if name == "mainnet" {
    netPrefixes = MainnetPrefixes
} else if strings.HasPrefix(name, "testnet") {
    netPrefixes = TestnetPrefixes
} else if name == simnet" {
    netPrefixes = SimnetPrefixes
} else {
    return nil
}
```

so that you just have one loop:
```go
for i := range Descriptions {
	addrPrefix = append(addrPrefix, AddrPrefix{
		Name:        Name[i],
		Description: Descriptions[i],
		Prefix:      netPrefixes[i],
	})
}
```

explorer/explorertypes.go
```diff
@@ -343,3 +325,74 @@ type NewMempoolTx struct {
...
+		"First 2 letters of a secp256k1 Schnorr P2PKH address",
+		"First 2 letters of a P2SH address",
+		"First 2 letters of a WIF private key",
+		"BIP32 hierarchical deterministic extended key magics",
+		"BIP32 hierarchical deterministic extended key magics",
+	}
+
+	addrPrefix := make([]AddrPrefix, 0, len(Descriptions))
+
+	var Name = []string{"PubKeyAddrID",
```
chappjc: `Name := ...`

explorer/explorertypes.go
```diff
@@ -343,3 +325,74 @@ type NewMempoolTx struct {
...
+		"First 2 letters of a P2PKH address",
+		"First 2 letters of an Edwards P2PKH address",
+		"First 2 letters of a secp256k1 Schnorr P2PKH address",
+		"First 2 letters of a P2SH address",
+		"First 2 letters of a WIF private key",
+		"BIP32 hierarchical deterministic extended key magics",
+		"BIP32 hierarchical deterministic extended key magics",
+	}
+
+	addrPrefix := make([]AddrPrefix, 0, len(Descriptions))
```
chappjc: move the init of `addrPrefix` to just before loop

explorer/explorertypes.go
```diff
@@ -343,3 +325,74 @@ type NewMempoolTx struct {
...
+// AddrPrefix represent the address name it's prefix and description
+type AddrPrefix struct {
+	Name        string
+	Prefix      string
+	Description string
+}
+
+// AddressPrefixes generates an array AddrPrefix by using chaincfg.Params
+func AddressPrefixes(params *chaincfg.Params) []AddrPrefix {
+	var Descriptions = []string{"First 2 letters of a P2PK address",
```
chappjc: `Descriptions := ...`

explorer/explorertypes.go
```diff
@@ -343,3 +325,74 @@ type NewMempoolTx struct {
...
+
+	var Name = []string{"PubKeyAddrID",
+		"PubKeyHashAddrID",
+		"PKHEdwardsAddrID",
+		"PKHSchnorrAddrID",
+		"ScriptHashAddrID",
+		"PrivateKeyID",
+		"HDPrivateKeyID",
+		"HDPublicKeyID",
+	}
```
chappjc: I'm not convinced we need the `Name` column since the description is more clear than the variable name, and if you want to know the variable name for any reason, you are probably already looking at the code.

raedah: By keeping it for now, the table columns will continue to align up.

explorer/explorertypes.go
```diff
@@ -323,3 +325,74 @@ type NewMempoolTx struct {
...
+	var MainnetPrefixes = []string{"Dk", "Ds", "De", "DS", "Dc", "Pm", "dprv", "dpub"}
+	var Testnet2Prefixes = []string{"Tk", "Ts", "Te", "TS", "Tc", "Pt", "tprv", "tpub"}
+	var SimnetPrefixes = []string{"Sk", "Ss", "Se", "SS", "Sc", "Ps", "sprv", "spub"}
```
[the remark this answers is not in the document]
chappjc: Yes, but it seemed to be getting unnecessarily complex.  We got through the P2PKH addresses.  Then for the others the length of the data input to the base45.Encode function was different, and it just started to seem silly when none of these would change.  It's doable if there's good reason.

raedah:
- BaseSubsidy is showing 3119582664 DCR , but it is actually atoms that need to be converted to DCR,  31.19582664 DCR
 -  TicketPoolSize says "8192 actual"   Should be "8192 (40960 actual)"  use the params to derive this, no hard coding.  TicketPoolSize * TicketsPerBlock
- NetworkAddressPrefix is in the wrong table, should be moved down into Address parameters.
- Getting chappjc's suggestion of link to block 1 for the airdrop in there somehow? Could add on a row after GenesisBlock... "PremineBlock, 1, Premine block of the chain". I think that would only apply to mainnet though.

explorer/explorerroutes.go
```diff
@@ -596,3 +596,32 @@ func (exp *explorerUI) ErrorPage(w http.ResponseWriter, code string, message str
 func (exp *explorerUI) NotFound(w http.ResponseWriter, r *http.Request) {
 	exp.ErrorPage(w, "Not found", "Cannot find page: "+r.URL.Path, true)
 }
+
+// ParametersPage is the page handler for the "/ChainParameters" path
```
raedah: update url in comment

raedah:
This looks good to go to me. Just needs an entry added on the new menu to access it.

raedah:
ACK

Issue #374, opened 2018-03-01
Stop asking users to allow notifications


noahpierau:
![image](https://user-images.githubusercontent.com/16287891/36855922-0a7199e6-1d75-11e8-99c8-347a65147625.png)

Many websites are asking me to allow notifications. It annoys me a lot! 

Why does dcrdata need this? 

Can you make this opt-in, instead of bothering everyone with this pop-up?

chappjc:
This is the opt-in prompt.  You only need to click they little down arrow and then "never for this site" or whatever your browser has.  It will *never ask your again*.

What this provides is pretty cool IMO: New block notifications.  When transacting, I find it really handy to have it pop up and tell me a block was mined.  No more staring at dcrwallet/dcrd.

In the future, I hope to have the ability to register for notifications for certain addresses, tx hashes, etc.  With ability to set and clear these from a settings page maybe.  Just brainstorming...

I suspect what would make this more palatable would be a little text on the page that says something like "Would you like to enable desktop notifications? Click *here* to enable.  Or *dismiss*.

raedah:
I feel the same way as @noahpierau, and think this should be made opt in as a setting the user can enable, but not be prompted for otherwise.

chappjc:
At some point we could have a settings page, but a quick change would be to have a webmail style subtle text asking if you would like to enable, it dismiss the message.  It could be at the bottom, and non-blocking off courses.

Alternatively, there could be a little subtle icon in search bar, far left or right, that you could click to get a prompt if you want it.

raedah:
The issue with prompting is that it requires a cookie to save the users choice. If a user has cookies disabled, they will get prompted every time they visit the site.

chappjc:
Good point, although that's probably not many users.

In any case, just a little place to click to activate is probably best then.

chappjc:
Resolved by PR #398

PR #398, opened 2018-03-17
html templates: Do not automatically request to enable desktop ntfns

chappjc:
To enable desktop notifications, the user must click the connection
indicator at the bottom right.

This resolves issue #374


chappjc:
Now running on beta.dcrdata.org

To test, it is necessary to reset the acceptance/rejection of notifications like so:
image

PR #403, opened 2018-03-22
update colors, table styles and header/footer in move towards new design

gozart1:
This is an initial pass to align current theme closer to the new design being worked on.
https://projects.invisionapp.com/share/K6GFONQ5AU7

Changes include table styling, background and text colors, and top nav search styling
screen shot 2018-03-22 at 1 10 05 am

(winging the dark theme here, I'm sure DZ will come up w/ something nicer!)
screen shot 2018-03-22 at 1 10 18 am

borderless search input
screen shot 2018-03-22 at 1 11 16 am
screen shot 2018-03-22 at 1 11 02 am

### DIFF
--- a/public/css/main.css
+++ b/public/css/main.css
@@ -256,6 +256,97 @@ body.darkBG .navbar-fixed-bottom .nav-item {
   background: transparent !important;
 }
 
+#hamburger-menu {
+  z-index: 10;
+  position: fixed;
+  width: 100%;
+  top: 0px;
+  right: 0px;
+  left: 0px;
+}
+#hamburger-menu a {
+  text-decoration: none;
+  color: #232323;
+  transition: color 0.3s ease;
+}
+#hamburger-menu a:hover {
+  color: #2e75ff;
+}
+#menuToggle {
+  display: block;
+  top: 7px;
+  right: 8px;
+  position: absolute;
+  z-index: 1;
+  -webkit-user-select: none;
+  user-select: none;
+}
+#menuToggle input{
+  display: block;
+  width: 40px;
+  height: 36px;
+  position: absolute;
+  top: -6px;
+  left: -3px;
+  cursor: pointer;
+  opacity: 0;
+  padding: 0px;
+  z-index: 2;
+  -webkit-touch-callout: none;
+}
+#menuToggle span {
+  display: block;
+  width: 33px;
+  height: 4px;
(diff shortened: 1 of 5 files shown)


raedah:
Need to make sure the footer is not taking up too much screen space on mobile. Its an existing issue on the current site since we've been adding to the footer.

chappjc:
We're going to need a menu to handle the footer items.

gozart1:
Ok, lets try consolidating the nav items in a hamburger menu as outlined in the new design mocks and implement it in pure CSS, similar to this one: https://codepen.io/erikterwan/pen/EVzeRP

chappjc:
Yeah, 100%.  That's a great menu, and even better since it's pure CSS.

chappjc:
@gozart1 BTW, the URL query part of the URL paths seem to disappear on several pages, like when clicking "next" on the address page's transactions table, or when changing the address tx type filter in my other PR.  I really don't get what's happening at all.  Any insight would be appreciated.

gozart1:
Did some initial tinkering...

screen shot 2018-03-22 at 11 38 52 pm

Expanded: 

screen shot 2018-03-22 at 11 38 57 pm

gozart1:
@chappjc I went ahead and opened an issue for the url param dropping, along with possible solution

chappjc:
That menu is looking nice.  Just a couple suggestions:

- perhaps the hamburger and the X could be a hair smaller
- It would be great to give each menu item an icon, like the sun/moon thing for Night Mode. (the rest can get them eventually, if needed)

Regardign the url parameter dropping, thanks for the suggestion with `Turbolinks-Location`.  I had forgotten that was a thing.  We had to apply that for other stuff like the search link.  I'll see if that helps.

chappjc:
The only thing left on this PR is the `href`s in the menu, AFAIK.

@raedah I think this is a nice subtle improvement.  Yet I think it would be great to punctuate the otherwise under-the-hood 2.0 release.

@gozart1 Did you want to do anything else in this besides the menu links?

gozart1:
Sure, let me do a final pass to wire up the nav and do some cross browser/device QA. ETA ~24hrs.

gozart1:
Tested on ios and android.
For sake of ergonomics when opening the menu, esp on mobile, I suggest we keep the menu toggle at least as big as it currently is.

PR #405, opened 2018-03-22
[WIP] Add support for insight socket.io api

RogueElement:
Implements the websocket api described [here](https://github.com/bitpay/insight-api#web-socket-api). The objects used in this pr however are based on the results of the current insight server on mainnet.decred.org.
Currently implemented:
- [x] `tx` event in `inv` room.
- [x] `block` event in `inv` room.
- [ ] `<dcrAddress>` event in `<dcrAddress>` room.
- [ ] `status` event in `sync` room.

This pr can be tested by loading [this](https://gist.github.com/RogueElement/9ac2fd59049dd19516616f96662f673f) up in a browser.


api/insight/socket.io.go
```diff
@@ -0,0 +1,91 @@
...
+				vouts[ntx.Vouts[i].ScriptPubKey.Addresses[0]] = v.Value
+			}
+		}
+		tx := WebSocketTx{
+			Hash:     msgTx.TxHash().String(),
+			Size:     len(ntx.Hex) / 2,
+			TotalOut: total,
+			Vout:     vouts,
+		}
+		soc.BroadcastTo("inv", "tx", tx)
```
chappjc: I like this concept of "rooms" that socket.io brings.  Much easier than our hand-rolled websocket solution.

Please be sure to try working with the go-socket.io v1.4 branch as recommended by their README.md as apparently the project is abandoned by the maintainer.

RogueElement:
Tried working with the v1.4 branch and I don't think that's what we need. The v1.4 branch has no concept of rooms only namespaces which isn't what is specified by the api so I think it'll be better to just stick with master.

raedah:
I will defer API review to chappjc.

RogueElement:
This is ready

chappjc:
Ok, status room deferred for future then?

RogueElement:
Yeah

main.go
```diff
@@ -492,6 +501,10 @@ func mainCore() error {
 		insightApp := insight.NewInsightContext(dcrdClient, chainDBRPC, cfg.IndentJSON)
 		insightMux := insight.NewInsightApiRouter(insightApp, cfg.UseRealIP)
 		webMux.Mount("/insight-api", insightMux.Mux)
+
+		if insightSocketServer != nil {
+			webMux.Get("/socket.io/", insightSocketServer.ServeHTTP)
```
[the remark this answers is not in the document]
RogueElement: `"socket.io"` seems to be the conventional path.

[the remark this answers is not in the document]
chappjc: k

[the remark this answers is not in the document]
chappjc: So I think we should still put this under a common `/insight` router.  We should use whatever paths are expected for the Insight API and the Insight socket.io endpoint, but let's make them accessible via a prefix.  After all, we're just supporting Insight, we are not Insight.

RogueElement:
Changed the insight path to `insight/api` and the insight websocket path to `insight/socket.io`. Also changed the gist to reflect this.

chappjc:
Please `dep ensure` and git add the Gopkg.lock

chappjc:
@RogueElement could you also please include your gist as api/insight/test_socket.html?

chappjc:
I subscribed to "Dcur2mcGjmENx4DhNqDctW5wJCVyT3Qeqkx" and never received anything (using https://gist.github.com/chappjc/b1116518aecf46f5d087846c7f03774d).

I realized that it's because the coinbase never enters mempool.  It's put in the block by the miner.

So my question is, how is the <address> room supposed to work?  Should it broadcast the tx hash involving address in which cases:

1.  As the relevant tx enters mempool?  (PR currently)
2.  When a relevant tx is mined?
3.  Both?

notification/ntfnhandlers.go
```diff
@@ -234,6 +236,16 @@ func MakeNodeNtfnHandlers() (*rpcclient.NotificationHandlers, *collectionQueue)
...
 			}
 
+			select {
+			case NtfnChans.InsightNewTxChan <- &insight.NewTx{
+				Time:  time.Now().Unix(),
+				Hex:   txDetails.Hex,
+				Vouts: txDetails.Vout,
+			}:
+			default:
+				log.Warn("InsightNewTxChan buffer full!")
```
chappjc: This is going to be very noising (and misleading) when running in lite mode, when `NtfnChans.InsightNewTxChan` is `nil`.

public/js/clipboard.min.js
```diff
@@ -1,7 +0,0 @@
-/*!
```
chappjc: Please be sure we aren't using this.  I though we had clipboard functionality in use somewhere.

public/js/Chart.min.js
```diff
@@ -1,14 +0,0 @@
-/*!
```
chappjc: I'm pretty sure we aren't charting or anything yet, but please be sure this isn't needed.

api/insight/socket.io.go
```diff
@@ -0,0 +1,143 @@
...
+		if err != nil {
+			continue
+		}
+		hash := msgTx.TxHash().String()
+		vouts := make(map[string]int64)
+		var total int64
+		isStake := txhelpers.IsStakeTx(msgTx)
+		for i, v := range msgTx.TxOut {
+			total += v.Value
+			if !isStake && len(ntx.Vouts[i].ScriptPubKey.Addresses) != 0 {
```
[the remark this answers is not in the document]
RogueElement: The insight server on mainet.decred.org ignored stake txs so I just followed suit.

public/js/clipboard.min.js
RogueElement: We had it a while ago, don't anymore. If we need need to add it again, can probably do that without use of this library.

RogueElement:
Ignore the `Gopkg.lock` diff, still trying to update dep

chappjc:
@RogueElement looks like you're running an old version of the `dep` tool, judging by the packages put on the same line.  You'll want to `go get -u -v github.com/golang/dep/cmd/dep` to get it updated.

main.go
```diff
@@ -504,7 +515,11 @@ func mainCore() error {
 		chainDBRPC, _ := dcrpg.NewChainDBRPC(auxDB, dcrdClient)
 		insightApp := insight.NewInsightContext(dcrdClient, chainDBRPC, cfg.IndentJSON)
 		insightMux := insight.NewInsightApiRouter(insightApp, cfg.UseRealIP)
-		webMux.Mount("/insight-api", insightMux.Mux)
+		webMux.Mount("/insight/api", insightMux.Mux)
+
+		if insightSocketServer != nil {
+			webMux.Get("/insight/socket.io/", insightSocketServer.ServeHTTP)
```
chappjc: I suspect we'll want to have this handled still, just with a helpful message.

chappjc:
This is tested and working, but I'd like to understand why the server logs spew this for every connection:

    2018/05/22 18:26:32 http: response.WriteHeader on hijacked connection

RogueElement:
googollee/go-socket.io/issues/108
googollee/go-engine.io/issues/35

Seems to be an issue with the `master` branch, just not sure if its fixed in `v1.4`.

PR #407, opened 2018-03-23
main: configure http.Server with ReadTimeout and WriteTimeout

chappjc:
Instead of using the default `http.ListenAndServe`, configure the `http.Server` with sensible (not infinite) read and write deadlines, and use `(*http.Server).ListenAndServe` (similarly for `ListenAndServeTLS`).

Thanks @raedah 
https://blog.cloudflare.com/the-complete-guide-to-golang-net-http-timeouts/


chappjc:
well, it doesn't seem to break anything...

Issue #360, opened 2018-02-25
Page size selector should only display valid options


raedah:
The page selector on the address page will show Page Size options for 20, 100, 1000 items per page. It shows this when only 15 rows are being returns.  
If 15 rows are being returned, then the page size selector should now be shown. 
If 47 rows were being returned, page size options should be 20, 47.  (47 is the max rows available)
If 522 rows were being returned, page size options should be 20, 100, 522.
If over 1000 items are being returned, then show 20, 100, 1000. (1000 is the max allowed rows displayed)

raedah:
The selector should default to the currently selected choice.

chappjc:
Resolved by PR #411

PR #411, opened 2018-03-25
Page size selector feature

McEdward:
Rounding up issue #360


McEdward:
Also i noticed that after selecting any of the All (Numtx) the page size doesn't seem to move lower. I don't seem to understand why

PR #417, opened 2018-03-31
add corresponding type names to API docs

chappjc:
README: begin adding types to API endpoint tables
api/types: apitypes.BlockTransactionCounts
api: use apitypes.BlockTransactionCountsin getBlockTransactionsCount


chappjc:
@marcopeereboom does this work?

Issue #392, opened 2018-03-14
mempool continued


raedah:
In continuation from mempool PR https://github.com/decred/dcrdata/pull/380 ,
some issues remain.
- Instead of hiding votes that are not for the tip, we should show whats in the mempool. This can be done with a column that specifies what block the vote is for. Votes can be sorted so the most recent block votes are at the top. If needed, there can be additional graphical distinguishments, perhaps a grey background on the rows that are for earlier blocks.
- Vote #, currently displayed in the hover, is useful information for comparison that should be viewable at a glance. It can have a column. Can also be a link to the ticket itself.

Current work is in https://github.com/decred/dcrdata/pull/561

raedah:
This was closed by https://github.com/decred/dcrdata/pull/634

PR #419, opened 2018-04-03
explorer: Change mempool vote filter

RogueElement:
for pr #392:
- The mempool page no longer ignores votes for the last block, now it just adds a red background.
- Mempool vote columns changed to TxID, Vote Block, Vote Index, Vote Version, Last Block Valid, Total DCR, Size and time.
- Mempool vote table is now sorted by the vote block.
votes-unfilter


chappjc:
Why remove the vote agenda/issue and choice for the votes on the block page too?  Better shown by the actual vote transaction page?

RogueElement:
@raedah suggested removing the vote agenda and choice for simplicity

explorer/mempool.go
```diff
@@ -208,11 +194,24 @@ func (exp *explorerUI) storeMempoolInfo() (lastBlockHash string, lastBlock int64
...
+	} else {
+		idxs := make(map[string]int)
+		idxs[tx.VoteInfo.TicketSpent] = 1
+		txindexes[tx.VoteInfo.Validation.Height] = idxs
+		tx.VoteInfo.MempoolTicketIndex = 1
 	}
-	return true
+}
+func voteForLastBlock(blockHash, validationHash string) bool {
+	return !(blockHash != "" && validationHash != blockHash)
```
chappjc: This is a little hard to read with the double negatives.  Perhaps an OR instead.

     blockHash == "" || validationHash == blockHash

explorer/mempool.go
```diff
@@ -208,11 +194,24 @@ func (exp *explorerUI) storeMempoolInfo() (lastBlockHash string, lastBlock int64
 	return
 }
 
-func voteForLastBlock(blockHash, validationHash string) bool {
-	if blockHash != "" && validationHash != blockHash {
-		return false
+func addVoteIndex(tx *MempoolTx, txindexes map[int64]map[string]int) {
```
chappjc: Despite being non-exported, it would be helpful to have a description of this function to the intent is clear.  Looks like it is updating `tx.VoteInfo.MempoolTicketIndex` and/or the txindexes map.  Actually, perhaps it should take a `*VoteInfo` instead of a `*MempoolTx`.

views/mempool.tmpl
```diff
@@ -61,36 +61,24 @@
...
-                            <th>Last Block Valid</th>
-                            <th>Issue ID</th>
-                            <th>Choice ID</th>
-                            <th class="text-right">Total DCR</th>
-                            <th>Size</th>
-                            <th class="text-right">Time in Mempool</th>
+                            <th>Vote Block</th>
+                            <th>Vote Index</th>
+                            <th>Vote Version</th>
+                            <th>Vote Valid</th>
```
chappjc: To me "Vote Valid" seems like it's somehow the validity of the vote itself.  I don't feel strongly about this, but perhaps "Vote Validation" would be clearer?

raedah: Vote Choice

raedah:
Removed the agenda because there was not enough horizontal room for all the columns, and the existing agenda display also had issues because it was not built to support multiple agendas. So we passed on trying to make multiple agendas viewable here and just moved that to the Vote page to view the agendas. We will also have more agenda information available on the upcoming Agenda page view.

I suggest we make the same changes to the vote table on the Block page, so that the columns and headers match the mempool page. Could be done in this same PR.

raedah:
Looks like there is a bug that will clear the list of votes even though a new block is not yet found. On page refresh, all the votes appear again.

Vote table is currently being sorted by Vote Block, but also needs to be secondarily sorted by Vote Index. Note: There will be duplicate votes with the same index in the mempool. This is normal for vote transactions.

The row bgcolor shading does not appear to always be being used on the correct rows. Rows for last block are shaded.

We can show a count at the top of the page for VOTE CHOICE  X / 5 (Valid)  where x is the number  of votes received for this round, 5 should be set via chain param, and Valid or Invalid is the result based on the majority vote choice received that reaches 3 (chain param).  Only displayed when determined.

RogueElement:
Still looking into how to fix the votes clearing on new block bug.

chappjc:
Where are we at with this one?

raedah:
decred2

After running for a short period of time, it becomes a single line.

chappjc:
bump

chappjc:
@raedah Does this need a new owner?

raedah:
I see duplicate transaction hashes listed. A single hash should be not be showing in the mempool more then once. I believe this bug is happening in dcrdata and is not caused by the dcrd mempool.

raedah:
@chappjc yes, would be good to get this finished off since @RogueElement has other outside commitments.

papacarp:
I'll work on this.

raedah:
After running for a short period of time, it becomes a single line. Something is causing it to also not clear any further with new votes, because the one row being displayed is a very old block.

papacarp:
Just to re-iterate.  now that I'm back on the tip with my db, I'm finally working on this one as of now.

chappjc:
Great.  Thanks.

Issue #393, opened 2018-03-15
address page: refactor CountUnconfirmedTransactions


chappjc:
CountUnconfirmedTransactions uses searchrawtransactions (verbose).  For some addresses with very large transactions, this kills any performance we got out of using PostgreSQL for the confirmed txn database.

CountUnconfirmedTransactions needs to be changed so it does not use searchrawtransactions.  Perhaps getrawmempool?

chappjc:
closed by PR #422

Issue #277, opened 2017-11-17
address view does not show unconfirmed transactions


raedah:

papacarp:
Some sleuthing guidance for this one:
1) if an address has no transactions in it previously, then addressHistory fails and it falls back to RPC to display the transactions which includes unconfirmed transactions.

https://github.com/decred/dcrdata/blob/e22ff164f2e24e11c2b2fecd712821d008db7a43/explorer/explorerroutes.go#L353-L358

Thus we are able to see unconfirmed transactions sometimes.
![image](https://user-images.githubusercontent.com/11194546/40158909-c16810ea-595b-11e8-9444-d8c6c76c6d7c.png)

2) if an address *does* have transactions in it previously, then AddressHistory succeeds, unconfirmed transactions are retrieved, but never merged into the resulting output.  

https://github.com/decred/dcrdata/blob/e22ff164f2e24e11c2b2fecd712821d008db7a43/explorer/explorerroutes.go#L441

They are just left in a the var uctxn and ignored.  The fix is to merge these into the addrData.Transactions list, and likely re-slice that list to adhere to the start/limit inputs.

_This is my theory as I haven't noticed an unconfirmed transaction go through that has previous history in it to verify and my testnet is not up right now._

If this is not  urgent then due to the rework happening on addresses table this should likely be fixed after we sort out the disposition of #427 by either @dmigwi or I.  If urgent I can put a PR in for it based on existing architecture.

chappjc:
Yes, this is absolutely right.  I had started a branch fixing these issues, but never finished up.  https://github.com/chappjc/dcrdata/commits/addr-page-fixes

I must have stashed teh changes related to the above issues though.

chappjc:
Please feel free to pick this up of course.  But do not wait on PR #426 .  It's big and breaking, so we're not going to rush it through.  Although it's getting closer.

dmigwi:
Since the unconfirmed transactions are already available in the frontend and the only problem is when they are displayed why not add a link to the UNCONFIRMED metric value such that when the link is clicked the unconfirmed txs appear. 

![hyperlink](https://user-images.githubusercontent.com/22055953/40200494-1dbd8be4-5a25-11e8-8f44-bf9f4554db13.png)

I suggest we add the link to the box in red. Whats your take on this @papacarp?

chappjc:
Correction: I meant do not wait on *PR #427 *

raedah:
[...]

PR #422, opened 2018-04-04
Unconfirmed transaction helpers, reworked CountUnconfirmedTransactions

chappjc:
This is extracted from the swelling address-filter PR.

Add func UnconfirmedTxnsForAddress.
Updated AddressInfo type.
Modify explorer.AddressPage to get unconfirmed txns and store them.  They are not displayed at all.  Treat this as an example.

### DIFF
--- a/db/dcrsqlite/apisource.go
+++ b/db/dcrsqlite/apisource.go
@@ -1,3 +1,4 @@
+// Copyright (c) 2018, The Decred developers
 // Copyright (c) 2017, Jonathan Chappelow
 // See LICENSE for details.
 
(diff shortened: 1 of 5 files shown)


chappjc:
This addresses issue #https://github.com/decred/dcrdata/issues/393.

This also provides the required tooling for unconfirmed support on the addres view when in full/PG mode: https://github.com/decred/dcrdata/issues/277

@RogueElement Would you mind reviewing this?  There should be no user-visible changes.  `UnconfirmedTxns *AddressTransactions` is added to the `AddressInfo` type, but this would likely not be used in this way since we want to have the unconfirmed transactions merged with the list of confirmed transactions.

PR #424, opened 2018-04-04
restyle progress bars

gozart1:
Updates progress bar appearance so to be more like the design comps. The gray background is a bit darker than the comp and a slight drop shadow has been added to the white text to improve readability when it overflows the green bar.

design comp:
screen shot 2018-04-04 at 1 08 23 am

PR:
screen shot 2018-04-04 at 12 53 16 am

screen shot 2018-04-04 at 1 15 46 am

### DIFF
--- a/public/css/main.css
+++ b/public/css/main.css
@@ -640,7 +640,19 @@ body.darkBG #menu {
   padding: 1px 5px;
 }
 
+.progress-bar {
+  background: #41BF53;
+  color: #fff;
+  height: 20px;
+  line-height: 20px;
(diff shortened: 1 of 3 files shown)


chappjc:
OK with me.  I like it!

raedah:
Not legible when I tested.
unreadable

Also breaks from existing dark view.
dark

chappjc:
@raedah That's what it looked like for me at first too.  Need to clear cache.  CTRL+SHIFT+R on FF

raedah:
tACK

PR #425, opened 2018-04-06
multi: devops

githubsands:
1. 'Dockerfile' is created, but a container/image still needs to be pushed on decred/dcrdata 
2. Add the ability to pull using docker in 'run_tests.sh'

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,57 @@
+#dcrdata-1
+
+FROM golang:1.10 
+
+LABEL description="Dcrdata"
+LABEL version="1.0"
+LABEL maintainer vracek@protonmail.com
+
+ENV TERM linux
+ENV USER build
+
+# create user
+RUN adduser --disabled-password --gecos '' $USER
+
+# update base distro & install  build tooling
+ENV DEBIAN_FRONTEND noninteractive
+
+RUN apt-get update && \
+    apt-get install -qy rsync 
+
+# create directory for build artifacts, adjust user permissions 
+RUN mkdir /release && \
+    chown $USER /release
+
+# create directory to get source from
+RUN mkdir /src && \
+    chown $USER /src && \
+    mkdir -p /go/src/github.com/decred/dcrdata && \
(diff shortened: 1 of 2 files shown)


chappjc:
I think to get rid of the merge commits you'll have to git reset b59f8e89, git add (everything), git commit, force push.
Or whatever seems right.  Maybe back up first though.  :)

chappjc:
Thanks for fixing that up @githubsands.  Could you offer a brief description of what goes on in run_tests.sh in terms of docker and what exactly it does (improvements?).  Thanks!

@decred/dcrdata Anyone experienced with docker wanna lend their eyes for a review?

chappjc:
@dajohi Could you scan through this if you have a minute?

orthomind:
To get these tests to run, I think you need to at least modify the .travis.yml file to load docker.  See, for example:

https://github.com/decred/dcrweb/blob/master/.travis.yml

You can then also have travis build the image itself before running the tests.  However, looking at the script, it's not clear to me that getting travis to build the image first will work with the way the script has been written here with the cache.

orthomind:
Also, I'm thinking that maybe you shouldn't use the same Dockerfile for tests as for running the production service.  Tests require certain things in the image that aren't required for a production run (or, at least, I assume so!).

If we have separate Dockerfile's for each of testing and production, then production can use a multistage Dockerfile to keep it as small as possible.

I think we should also always build the testing image as part of the travis test process, rather than pulling it, so that the testing Dockerfile also gets tested each time.  If we rely on a pull, then it could be that changes to the Dockerfile that are part of that pull request would have caused tests to fail, but the tests didn't fail because the tests ran against the Dockerfile that's in the image registry rather than the new image that's requested to be merged (or vice versa: tests fail that wouldn't fail with the new proposed Dockerfile).

If I'm not making sense, let me know.

chappjc:
Thank you both very much for your work on this.  This PR is now superseded by PR https://github.com/decred/dcrdata/pull/497/files

Issue #135, opened 2017-09-08
Chart of long term data


raedah:
-Ticket price chart
-Ticket pool size (by tickets)
-Ticket pool value (# of tickets * ticket price)
-Percentage of coins staked (ticket pool value / coin supply)
-PoW difficulty chart

dmigwi:
The task title says that a chart for ticket price history `(Ticket pool size (by tickets) history chart)` should be created. The task description also mentions several more types of history charts i.e.

>Coinsupply and stake pool value history chart (can be done in one chart)
>Percentage of staked coins for each block height
>Difficulty history chart
>Hashrate history chart (data source?)

Are this part of the ticket price history chart or they are separate individual charts @raedah

raedah:
Separate charts

raedah:
https://blockchain.info/charts
Blockchain size,
average block size, 
Txns per block, 
median tx confirmation time/fee
unique addresses, 
txns per day

https://blockchain.info/charts/utxo-count
Number of Unspent Transaction Outputs

raedah:
Much of this was taken care of in https://github.com/decred/dcrdata/issues/463 . Keeping open for the few ideas that still remain.

PR #426, opened 2018-04-06
[WIP] collect data for ticket price chart

rmukhamet:
Api for  generating ticket price chart.
1) Added table chartblocks to DB 
2) Model ChartBlock to dbtypes.
3) in sync every 144 block saves to chartblocks (height, sbits, time).
4) in  explorer and apiroutes added method to get all data from chartblocks.
5) For explorerUI  handler to get all data in  json.

- [ ] getting data with parameters: from date, to date.


raedah:
This is a WIP for https://github.com/decred/dcrdata/issues/135

rmukhamet:
image

chappjc:
I'm also not convinced we need a new PostgreSQL table for this.  All the stakediff data is available in lite mode (via SQLite) and the existing PostgreSQL `blocks` table.  Why is a new table needed?

rmukhamet:
@chappjc,
- [ X ] The explorerUI type is for the web interface, whereas the api package is for serving JSON.
- [ ]  The name "chartblock" is confusing. The "chart" part is referring to the intended use, but the table contains stakediff interval data. Please propose an more descriptive name.  
  - "ticketprice"  will be ok? 
- [ ] I'm also not convinced we need a new PostgreSQL table for this. All the stakediff data is available in lite mode (via SQLite) and the existing PostgreSQL blocks table. Why is a new table needed?
  - the reason of performance.  Do you want performance testing?

dmigwi:
@rmukhamet I am supposed to use your changes to build the frontend for the ticket price.

Kindly help understand the following issues:
```go
// ChartBlock models data for charts.
type ChartBlock struct {
	Height    int64   `db:"height" json:"height"`
	SBits     float64 `db:"sbits" json:"sbits"`
	BlockTime int64   `db:"time" json:"time"`
}
```
- The above is struct defines what chartblock table should hold. The goal with ticket price chart is that we need the pair of `{ticket price : block time}` sent to the frontend in form of an array. Your struct doesn't define a ticket price at least from the comments, how were you planning to get the ticket from the tickets table?

- How and when will the table be populated?

I may be wrong on this so kindly enlighten me.

dmigwi:
I am also convinced that since the ticket price data may not change that oftenly unless there is a decred ticket price change. Also two or more users viewing the ticket price from different machines at the same time will more or less be viewing the same data set, in my opinion caching the data once there is a price change makes more sense to me than introducing a new table that will increase the data migration time. 

Caching the data set might deliver more performance than a new table. Once again I might be wrong, so kindly enlighten me.

dmigwi:
@rmukhamet 

Here is a sample format.
```json
[
   {"date": "2012-07-27", "price": 13}, 
   {"date": "2012-07-28", "price": 11},
   {"date": "2012-07-29", "price": 15},
   {"date": "2012-07-30", "price": 16}
]
```

Plus much of the ticket price data is constant. I mean the ticket price from now till when it was first recorded doesn't change, what may change is the future ticket price. If we found a way to fetch all the constant data at once when the programs starts and then we keep appending the changes as they happen, this would mean that its possible to make a real time tickets chart so easily.

dmigwi:
https://github.com/decred/dcrdata/pull/462

Here is a different implementation of the same task.

Gopkg.lock
```diff
@@ -260,6 +260,6 @@
 [solve-meta]
   analyzer-name = "dep"
   analyzer-version = 1
-  inputs-digest = "2bfe6c04715a8ce07c4c67aeaa959e5781c1fd325c9add797b1a3cf9cb5036ed"
+  inputs-digest = "27710db38f3b670776c48bb7fa0c79f158be5889d5ba2c8f5f0e03e98598934b"
```
chappjc: Please reset this change.  dep ensure should not change this for this PR

api/apirouter.go
```diff
@@ -35,6 +35,8 @@ func NewAPIRouter(app *appContext, userRealIP bool) apiMux {
 
 	mux.HandleFunc("/status", app.status)
 
+	mux.Get("/ticket_price", app.TicketPrices)
```
chappjc: There is a `/stake` router with a `/diff` subrouter. Within that router, there is already the following endpoint:

`Get("/r/{idx0}/{idx}", app.getStakeDiffRange)`

What differentiates this new endpoint from that one?  Move and rename the new endpoint accordingly.  (It shows the ticket price intervals).

api/apiroutes.go
```diff
@@ -1062,6 +1063,18 @@ func (c *appContext) getAddressTransactionsRaw(w http.ResponseWriter, r *http.Re
 	writeJSON(w, txs, c.getIndentQuery(r))
 }
 
+// TicketPrices is the handler for charts.
```
chappjc: What does it show to the user?  It is not just the handler for the charts, although it happens to be used that way.  It handles returns ticket prices for all of the stake diff intervals.

db/dbtypes/types.go
```diff
@@ -372,3 +373,16 @@ type BlockDataBasic struct {
...
+type ChartBlock struct {
+	Height    int64   `db:"height" json:"height"`
+	SBits     float64 `db:"sbits" json:"sbits"`
+	BlockTime int64   `db:"time" json:"time"`
+}
+
+// TicketPrice front charts requested data
+type TicketPrice struct {
+	Price    float64   `json:"price"`
+	DateTime time.Time `json:"date"`
```
chappjc: In line with what I said about intervals, this data only makes sense as a price on a range of blocks.  What is that range?  A StartBlock and EndBlock are needed.

db/dbtypes/types.go
```diff
@@ -372,3 +373,16 @@ type BlockDataBasic struct {
...
+// ChartBlock models data for charts.
+type ChartBlock struct {
+	Height    int64   `db:"height" json:"height"`
+	SBits     float64 `db:"sbits" json:"sbits"`
+	BlockTime int64   `db:"time" json:"time"`
+}
+
+// TicketPrice front charts requested data
+type TicketPrice struct {
+	Price    float64   `json:"price"`
```
chappjc: We've decided to return all amounts as atoms (int64).

db/dcrpg/internal/chartblockstmts.go
```diff
@@ -0,0 +1,25 @@
+package internal
+
+const (
+	// InsertChartBlock holds DML for Chart block data inserting.
+	InsertChartBlock = `INSERT INTO chartblocks VALUES ($1, $2, $3);`
```
chappjc: The work "chart" should not appear anywhere in this PR, except perhaps in a comment.  This PR needs to be serving a certain kind of data on one or more API endpoints, which may or may not be used for charting.

db/dcrpg/internal/chartblockstmts.go
```diff
@@ -0,0 +1,25 @@
...
+
+	// SelectChartBlockByTimeRangeSQL holds the query string for selecting chart data by a time range.
+	SelectChartBlockByTimeRangeSQL = `SELECT * FROM chartblocks WHERE time BETWEEN $1 AND $2 ORDER BY time LIMIT $3;`
+
+	// SelectAllChartBlocks holds the query string for selecting chart data.
+	SelectAllChartBlocks = `SELECT * FROM chartblocks ORDER BY time;`
+
+	// CreateChartBlockTable holds DDL string for creating the chart data table.
+	CreateChartBlockTable = `CREATE TABLE IF NOT EXISTS chartblocks (
```
chappjc: I still don't get why we need a table for this.  For two reasons it's a problem: (1) The data is already in existing tables, and (2) this functionality won't work in lite mode.

db/dcrpg/pgblockchain.go
```diff
@@ -1735,3 +1751,21 @@ func ticketpoolStatusSlice(ss dbtypes.TicketPoolStatus, N int) []dbtypes.TicketP
 	}
 	return S
 }
+
+// TicketPrices returns slice of ticket prices and dates
```
chappjc: these dates correspond to what?

raedah:
Closing this, as it is superseded by https://github.com/decred/dcrdata/pull/427

PR #427, opened 2018-04-14
dcrpg: new adresses table format

dmigwi:
Goal: Allow proper sorting vout spends/funds on address page.

**New Addresses Table Schema Proposal**

<table>
  <tr>
    <td></td>
    <td>Columns</td>
    <td>Data Type</td>
    <td>Description</td>
  </tr>
  <tr>
    <td>1.</td>
    <td>id</td>
    <td>SERIAL8 PRIMARY KEY</td>
    <td>This is the primary key of the addresses table</td>
  </tr>
  <tr>
    <td>2.</td>
    <td>address</td>
    <td>TEXT</td>
    <td>Defines the actual address used to reference the actual transactions made.</td>
  </tr>
  <tr>
    <td>3.</td>
    <td>tx_hash</td>
    <td>TEXT</td>
    <td>Defines the funding/spending transaction hash as listed in the transactions table</td>
  </tr>
  <tr>
    <td>4.</td>
    <td>matching_tx_hash<br><del>in_out_row_id</del></td>
    <td> TEXT<br><del>INT8</del></td>
    <td>Defines a value that is is bi-directional i.e.
For a spending tx the value stored here should be the row id of the funding tx that credited it.
For a funding tx the value stored here should be the tx output that spends the amount credited by the funding tx.</td>
  </tr>
  <tr>
    <td>5.</td>
    <td>value</td>
    <td>INT</td>
    <td>Defines the actual amount that was either credited or debited by a given transaction input or output.</td>
  </tr>
  <tr>
    <td>6.</td>
    <td>block_time</td>
    <td>INT8</td>
    <td>- Define the time when the actual block was discovered.
- Listed here in help order the transaction in a meaningful without using sql joins
- Will reduce the number of queries made in order to display all the data that should appear on the `/address/{transaction}` route.  (History Page)</td>
  </tr>
  <tr>
    <td>7.</td>
    <td>is_funding</td>
    <td>BOOLEAN
</td>
    <td>- Helps differentiate between funding and spending transactions.</td>
  </tr>
  <tr>
    <td>***8.</td>
    <td>tx_vin_vout_index <br> (Should be dropped)</td>
    <td>INT8</td>
    <td>Contains a value that references the value as listed in tx_index or prev_tx_index rows for vins table or just tx_index row for the vouts table.
</td>
  </tr>
  <tr>
    <td>***9.</td>
    <td>tx_vin_vout_row_id <br> (Should be dropped)</td>
    <td>INT8</td>
    <td>Contains a value that references the value as listed in id row for the vins table or the vouts table. </td>
  </tr>
</table>

**Additional Notes:**

Unless persuaded otherwise, I am convinced that the **tx_vin_vout_index** and **tx_vin_vout_row_id**  (Columns with ***) are not that necessary. Tx_vin_vout_index is column merged from the formerly **spending_tx_vin_index** and **funding_tx_vout_index** columns. Tx_vin_vout_row_id is also a column merged from the formerly **vout_row_id** and **vin_row_id**.

Consider the following points:

* Vin_row_id and Vout_row_id need not to be referenced in the new table as either spending_tx_hash or funding_tx_hash that will now appear in the tx_hash can be used to fetch either of the details from the vins table or the vouts table.

* The specific tx_hash also be used to fetch the tx_index and prev_tx_index values  from vins table or just tx_hash from vouts table that match what exactly is  to be stored in this table.

**Conclusion.**

Unless fetching the values to be stored in both **tx_vin_vout_index** and **tx_vin_vout_row_id** from either vins or vouts table might result to slow queries, we should do away with them in the addresses table as they store unnecessary data.

We should index column **block_time** so as to improve the queries performance. Most queries will either search, group or even order the queries output by the field. For unindexed search by field,  linear search is used but for indexed field binary search is used. I.e.

Big O notation for Unindexed  = N 
for 1 million records with each record taking 1ms to visit then the whole query should take:
**Time taken : **1,000,000 x 1ms = 1,000,000ms

 Big O notation for Indexed  = Log 2N  
**Time taken :** (Log 2 1,000,000) x 1ms = 19.931569 x 1ms = 19.931569ms    Appox. 20 ms

**Caveat**

* An extra set of data is added to table for every indexed column thus increasing the space consumed by the table.
* Because two write operations are made one to the index (heap postgres and binary Trees for mysql) and the other to the table every write operation is slower

**Appendix/Reference**

* What is Column indexing: [https://stackoverflow.com/questions/1108/how-does-database-indexing-work](https://stackoverflow.com/questions/1108/how-does-database-indexing-work)

**HOW TO TEST block_time INDEXING**

Access the command line version of postgres via `psql`  in your machine and run the following commands:
```sql
CREATE TABLE indexed_addr (id SERIAL8 PRIMARY KEY, address TEXT, tx_hash TEXT, in_out_row_id INT8, value INT8, block_time INT8 NOT NULL, is_funding BOOLEAN);

CREATE INDEX block_time_index ON indexed_addr (block_time);

CREATE TABLE unindexed_addr (id SERIAL8 PRIMARY KEY, address TEXT, tx_hash TEXT, in_out_row_id INT8, value INT8, block_time INT8 NOT NULL, is_funding BOOLEAN);
```

run `\timing` to activate query timing in postgres for that session.

Get the sample data (I can't paste my zipped file with 2 million records here as its larger than 10 MB)
Alternatively generate the sample data from the exist testnet data in the postgres by running this:

```sql
\copy (select a.id, a.address, a.funding_tx_hash, t.id, a.value, t.block_time, case a.value when 0 then false else true end from addresses a, transactions t where a.funding_tx_hash = t.tx_hash limit 2000000) TO ~/sample_data.csv (format CSV, delimiter ",");
```

Load the sample data into indexed_addr and unindexed_addr tables
```sql
\copy indexed_addr from ~/sample_data.csv (format CSV, delimiter ",")

\copy unindexed_addr from ~/sample_data.csv (format CSV, delimiter ",")
```

Test the index by running the following queries
```sql
select count(*) from indexed_addr where block_time = 1512437355; => took 0.531 ms

select count(*) from unindexed_addr where block_time = 1512437355; => took 352.423 ms
```
screen shot 2018-04-14 at 03 17 40

Remember to drop the test tables after running tests on them.
```sql
drop table indexed_addr, unindexed_addr;
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,6 +93,30 @@ First, update the repository (assuming you have `master` checked out):
 Look carefully for errors with `git pull`, and reset locally modified files if
 necessary.
 
+## Upgrading Instructions 
+
+*__Only necessary while upgrading from v1.3.2 or below.__*
+
+1. Drop the old dcrdata database and create a new empty dcrdata database.
+ ```sql
+ // drop the old database
+ DROP DATABASE dcrdata;
+
+// create a new database with the same user set in the dcrdata.conf e.g user dcrdata
+CREATE DATABASE dcrdata WITH OWNER dcrdata;
+
+// grant all permissions to user dcrdata
+GRANT ALL PRIVILEGES ON DATABASE dcrdata to dcrdata;
+ ```
+
+2. Delete the data folder in dcrdata folder where dcrdata.conf file is located
+  - Mac :  `~/Library/Application Support/Dcrdata/data`
+  - Linux :  `~/Dcrdata/data`
+  - Window :  `C:\Users\<your-username>\AppData\Local\Dcrdata\data` 
+  
+3. Restart the data migration again by running the dcrdata and dcrd binaries
+  - run the downloaded dcrd binary first then the generated dcrdata binary
+
 ## Getting Started
 
 ### Configure PostgreSQL (IMPORTANT)
--- a/db/dbtypes/types.go
+++ b/db/dbtypes/types.go
@@ -248,16 +248,14 @@ type Vout struct {
 // AddressRow represents a row in the addresses table
 type AddressRow struct {
 	// id int64
-	Address            string
-	FundingTxDbID      uint64
-	FundingTxHash      string
-	FundingTxVoutIndex uint32
-	VoutDbID           uint64
-	Value              uint64
-	SpendingTxDbID     uint64
-	SpendingTxHash     string
-	SpendingTxVinIndex uint32
-	VinDbID            uint64
+	Address        string
+	InOutRowID     uint64
+	IsFunding      bool
+	TxBlockTime    uint64
+	TxHash         string
+	TxVinVoutIndex uint32
+	Value          uint64
+	VinVoutDbID    uint64
 }
 
 // ScriptPubKeyData is part of the result of decodescript(ScriptPubKeyHex)
--- a/db/dbtypes/types.go
+++ b/db/dbtypes/types.go
@@ -249,9 +249,9 @@ type Vout struct {
 type AddressRow struct {
 	// id int64
 	Address string
-	// InOutRowID is the row ID that provides the relationship
+	// MatchingTxHash that provides the relationship
 	// between spending tx inputs and funding tx outputs
-	InOutRowID     uint64
+	MatchingTxHash string
 	IsFunding      bool
 	TxBlockTime    uint64
 	TxHash         string
--- a/db/dbtypes/types.go
+++ b/db/dbtypes/types.go
@@ -249,8 +249,8 @@ type Vout struct {
 type AddressRow struct {
 	// id int64
 	Address string
-	// MatchingTxHash that provides the relationship
-	// between spending tx inputs and funding tx outputs
+	// MatchingTxHash provides the relationship between spending tx inputs and
+	// funding tx outputs.
 	MatchingTxHash string
 	IsFunding      bool
 	TxBlockTime    uint64
--- a/db/dbtypes/types.go
+++ b/db/dbtypes/types.go
@@ -248,7 +248,9 @@ type Vout struct {
 // AddressRow represents a row in the addresses table
 type AddressRow struct {
 	// id int64
-	Address        string
+	Address string
+	// InOutRowID is the row ID that provides the relationship
+	// between spending tx inputs and funding tx outputs
 	InOutRowID     uint64
 	IsFunding      bool
 	TxBlockTime    uint64
--- a/db/dcrpg/internal/addrstmts.go
+++ b/db/dcrpg/internal/addrstmts.go
@@ -83,8 +83,9 @@ const (
 		` FROM addresses WHERE address=$1)
         SELECT * FROM these order by block_time desc limit $2 offset $3;`
 
-	SelectAddressByTxHash = `select id, address, value from addresses 
-	   where tx_hash=$1 and is_funding = true ORDER BY block_time DESC LIMIT 1;`
+	SelectAddressByTxHash = `SELECT id, address, value FROM addresses 
+	   WHERE tx_hash = $1 AND tx_vin_vout_index = $2 and is_funding = TRUE
+	   ORDER BY block_time DESC LIMIT 1;`
 
 	SelectAddressDebitsLimitNByAddress = `SELECT ` + columnNames + `
 		FROM addresses WHERE address=$1 and is_funding = FALSE
@@ -103,12 +104,11 @@ const (
 	SetAddressFundingForMatchingTxHash = `UPDATE addresses SET matching_tx_hash=$1 
 	    WHERE id=$2 and is_funding = true;`
 
-	IndexBlockTimeOnTableAddress = `CREATE INDEX block_time_index ON addresses (block_time);`
-
+	IndexBlockTimeOnTableAddress   = `CREATE INDEX block_time_index ON addresses (block_time);`
 	DeindexBlockTimeOnTableAddress = `DROP INDEX block_time_index;`
 
-	IndexMatchingTxHashOnTableAddress = `CREATE INDEX matching_tx_hash_index ON addresses (matching_tx_hash);`
-
+	IndexMatchingTxHashOnTableAddress = `CREATE INDEX matching_tx_hash_index 
+	    ON addresses (matching_tx_hash);`
 	DeindexMatchingTxHashOnTableAddress = `DROP INDEX matching_tx_hash_index;`
 
 	IndexAddressTableOnAddress = `CREATE INDEX uix_addresses_address
--- a/db/dcrpg/internal/addrstmts.go
+++ b/db/dcrpg/internal/addrstmts.go
@@ -35,7 +35,7 @@ const (
 		value INT8,
 		block_time INT8 NOT NULL,
 		is_funding BOOLEAN,
-		tx_vin_vout_index INT8,
+		tx_vin_vout_index INT4,
 		tx_vin_vout_row_id INT8
 		);`
 
@@ -56,38 +56,29 @@ const (
 	SelectAddressSpentCountAndValue = `SELECT COUNT(*), SUM(value) FROM addresses 
 	    WHERE address = $1 and is_funding = FALSE and matching_tx_hash != '';`
 
-	SelectAddressUnspentWithTxn = `SELECT
-									addresses.address,
-									addresses.tx_hash,
-									addresses.value,
-									transactions.block_height,
-									addresses.block_time,
-									tx_vin_vout_index,
-									pkscript
-									FROM addresses 
-									JOIN transactions ON 
-									addresses.tx_hash = transactions.tx_hash
-									JOIN vouts on addresses.tx_hash = vouts.tx_hash 
-									and addresses.tx_vin_vout_index=vouts.tx_index
-									WHERE 
-									addresses.address=$1 
-									AND 
-									addresses.is_funding = FALSE order by addresses.block_time desc`
+	SelectAddressUnspentWithTxn = `SELECT addresses.address, addresses.tx_hash, addresses.value,
+			transactions.block_height, addresses.block_time, tx_vin_vout_index, pkscript
+		FROM addresses 
+		JOIN transactions ON 
+			addresses.tx_hash = transactions.tx_hash
+		JOIN vouts on addresses.tx_hash = vouts.tx_hash AND addresses.tx_vin_vout_index=vouts.tx_index
+			WHERE addresses.address=$1 AND addresses.is_funding = FALSE 
+			ORDER BY addresses.block_time DESC;`
 
-	columnNames = `id, address, matching_tx_hash, tx_hash, tx_vin_vout_index, block_time, tx_vin_vout_row_id, value, is_funding`
+	addrsColumnNames = `id, address, matching_tx_hash, tx_hash, tx_vin_vout_index, block_time, tx_vin_vout_row_id, value, is_funding`
 
-	SelectAddressLimitNByAddress = `SELECT ` + columnNames + ` FROM addresses
-	    WHERE address=$1 order by block_time desc limit $2 offset $3;`
+	SelectAddressLimitNByAddress = `SELECT ` + addrsColumnNames + ` FROM addresses
+	    WHERE address=$1 ORDER BY block_time DESC LIMIT $2 OFFSET $3;`
 
-	SelectAddressLimitNByAddressSubQry = `WITH these as (SELECT ` + columnNames +
+	SelectAddressLimitNByAddressSubQry = `WITH these as (SELECT ` + addrsColumnNames +
 		` FROM addresses WHERE address=$1)
-        SELECT * FROM these order by block_time desc limit $2 offset $3;`
+        SELECT * FROM these ORDER BY block_time DESC LIMIT $2 OFFSET $3;`
 
-	SelectAddressDebitsLimitNByAddress = `SELECT ` + columnNames + `
+	SelectAddressDebitsLimitNByAddress = `SELECT ` + addrsColumnNames + `
 		FROM addresses WHERE address=$1 and is_funding = FALSE
 		ORDER BY block_time DESC LIMIT $2 OFFSET $3;`
 
-	SelectAddressCreditsLimitNByAddress = `SELECT ` + columnNames + `
+	SelectAddressCreditsLimitNByAddress = `SELECT ` + addrsColumnNames + `
 		FROM addresses WHERE address=$1 and is_funding = TRUE
 		ORDER BY block_time DESC LIMIT $2 OFFSET $3;`
 
--- a/db/dcrpg/internal/addrstmts.go
+++ b/db/dcrpg/internal/addrstmts.go
@@ -109,6 +109,8 @@ const (
 
 	IndexBlockTimeOnTableAddress = `CREATE INDEX block_time_index ON addresses (block_time);`
 
+	DeindexBlockTimeOnTableAddress = `DROP INDEX block_time_index;`
+
 	IndexAddressTableOnAddress = `CREATE INDEX uix_addresses_address
 		ON addresses(address);`
 	DeindexAddressTableOnAddress = `DROP INDEX uix_addresses_address;`
@@ -119,4 +121,5 @@ const (
 
 	IndexAddressTableOnTxHash = `CREATE INDEX uix_addresses_funding_tx
 		ON addresses(tx_hash);`
+	DeindexAddressTableOnTxHash = `DROP INDEX uix_addresses_funding_tx;`
 )
--- a/db/dcrpg/internal/addrstmts.go
+++ b/db/dcrpg/internal/addrstmts.go
@@ -79,7 +79,7 @@ const (
 	SelectAddressLimitNByAddressSubQry = `WITH these as (SELECT * FROM addresses WHERE address=$1)
 		SELECT * FROM these order by block_time desc limit $2 offset $3;`
 
-	SelectAddressDebitsLimitNByAddress = `SELECT id, in_out_row_id, tx_hash, tx_vin_vout_index, block_time, tx_vin_vout_row_id, value
+	SelectAddressDebitsLimitNByAddress = `SELECT id, address, in_out_row_id, tx_hash, tx_vin_vout_index, block_time, tx_vin_vout_row_id, value
 	FROM addresses
 	WHERE address=$1 and is_funding = FALSE
 	ORDER BY block_time DESC LIMIT $2 OFFSET $3;`
--- a/db/dcrpg/internal/addrstmts.go
+++ b/db/dcrpg/internal/addrstmts.go
@@ -41,7 +41,7 @@ const (
 
 	SelectAddressAllByAddress = `SELECT * FROM addresses WHERE address=$1 order by block_time desc;`
 	SelectAddressRecvCount    = `SELECT COUNT(*) FROM addresses WHERE address=$1;`
-	SelectAddressesAllTxn = `SELECT tx_hash, block_time as tx_time, ftxd.block_height as height 
+	SelectAddressesAllTxn     = `SELECT tx_hash, block_time as tx_time, ftxd.block_height as height 
 		from addresses left join transactions as ftxd on funding_tx_row_id=ftxd.id 
 		where address = $1 order by tx_time desc;`
 
--- a/db/dcrpg/internal/addrstmts.go
+++ b/db/dcrpg/internal/addrstmts.go
@@ -87,6 +87,8 @@ const (
 	SelectAddressByTxHash = `select id, address, value from addresses 
 	   where tx_hash=$1 and is_funding = true`
 
+	SelectAddressByRowID = `select tx_hash FROM addresses WHERE id = $1;`
+
 	SelectAddressDebitsLimitNByAddress = `SELECT ` + columnNames + `
 		FROM addresses WHERE address=$1 and is_funding = FALSE
 		ORDER BY block_time DESC LIMIT $2 OFFSET $3;`
--- a/db/dcrpg/internal/addrstmts.go
+++ b/db/dcrpg/internal/addrstmts.go
@@ -1,20 +1,19 @@
 package internal
 
 const (
-	insertAddressRow0 = `INSERT INTO addresses (address, in_out_row_id, tx_hash,
+	insertAddressRow0 = `INSERT INTO addresses (address, matching_tx_hash, tx_hash,
 		tx_vin_vout_index, tx_vin_vout_row_id, value, block_time, is_funding)
 		VALUES ($1, $2, $3, $4, $5, $6, $7, $8) `
 
 	InsertAddressRow = insertAddressRow0 + `RETURNING id;`
-	// InsertAddressRowChecked = insertAddressRow0 +
-	// 	`ON CONFLICT (vout_row_id, address) DO NOTHING RETURNING id;`
+
 	UpsertAddressRow = insertAddressRow0 + `ON CONFLICT (tx_vin_vout_row_id, address, is_funding) DO UPDATE 
 		SET address = $1, tx_vin_vout_row_id = $5 RETURNING id;`
 	InsertAddressRowReturnID = `WITH inserting AS (` +
 		insertAddressRow0 +
 		`ON CONFLICT (tx_vin_vout_row_id, address, is_funding) DO UPDATE
-			SET address = NULL WHERE FALSE
-			RETURNING id
+		SET address = NULL WHERE FALSE
+		RETURNING id
 		)
 		SELECT id FROM inserting
 		UNION  ALL
@@ -32,7 +31,7 @@ const (
 		id SERIAL8 PRIMARY KEY,
 		address TEXT,
 		tx_hash TEXT,
-		in_out_row_id INT8,
+		matching_tx_hash TEXT,
 		value INT8,
 		block_time INT8 NOT NULL,
 		is_funding BOOLEAN,
@@ -52,10 +51,10 @@ const (
 		address = $1 and tx_hash=$2;`
 
 	SelectAddressUnspentCountAndValue = `SELECT COUNT(*), SUM(value) FROM addresses 
-	    WHERE address = $1 and is_funding = TRUE and in_out_row_id < 1;`
+	    WHERE address = $1 and is_funding = TRUE and matching_tx_hash = '';`
 
 	SelectAddressSpentCountAndValue = `SELECT COUNT(*), SUM(value) FROM addresses 
-	    WHERE address = $1 and is_funding = FALSE and in_out_row_id > 0;`
+	    WHERE address = $1 and is_funding = FALSE and matching_tx_hash != '';`
 
 	SelectAddressUnspentWithTxn = `SELECT
 									addresses.address,
@@ -75,7 +74,7 @@ const (
 									AND 
 									addresses.is_funding = FALSE order by addresses.block_time desc`
 
-	columnNames = `id, address, in_out_row_id, tx_hash, tx_vin_vout_index, block_time, tx_vin_vout_row_id, value, is_funding`
+	columnNames = `id, address, matching_tx_hash, tx_hash, tx_vin_vout_index, block_time, tx_vin_vout_row_id, value, is_funding`
 
 	SelectAddressLimitNByAddress = `SELECT ` + columnNames + ` FROM addresses
 	    WHERE address=$1 order by block_time desc limit $2 offset $3;`
@@ -85,9 +84,7 @@ const (
         SELECT * FROM these order by block_time desc limit $2 offset $3;`
 
 	SelectAddressByTxHash = `select id, address, value from addresses 
-	   where tx_hash=$1 and is_funding = true ORDER BY block_time ASC LIMIT 1;`
-
-	SelectAddressByRowID = `select tx_hash FROM addresses WHERE id = $1;`
+	   where tx_hash=$1 and is_funding = true ORDER BY block_time DESC LIMIT 1;`
 
 	SelectAddressDebitsLimitNByAddress = `SELECT ` + columnNames + `
 		FROM addresses WHERE address=$1 and is_funding = FALSE
@@ -103,7 +100,7 @@ const (
 	SelectAddressIDByVoutIDAddress = `SELECT id FROM addresses WHERE address=$1 and 
 	    tx_vin_vout_row_id=$2 and is_funding = true ORDER BY block_time DESC;`
 
-	SetAddressFundingForInOutID = `UPDATE addresses SET in_out_row_id=$1 
+	SetAddressFundingForMatchingTxHash = `UPDATE addresses SET matching_tx_hash=$1 
 	    WHERE id=$2 and is_funding = true;`
 
 	IndexBlockTimeOnTableAddress = `CREATE INDEX block_time_index ON addresses (block_time);`
--- a/db/dcrpg/internal/addrstmts.go
+++ b/db/dcrpg/internal/addrstmts.go
@@ -42,24 +42,20 @@ const (
 
 	SelectAddressAllByAddress = `SELECT * FROM addresses WHERE address=$1 order by block_time desc;`
 	SelectAddressRecvCount    = `SELECT COUNT(*) FROM addresses WHERE address=$1;`
-	SelectAddressesAllTxn     = `WITH these as (SELECT funding_tx_hash as tx_hash, ftxd.time as tx_time, ftxd.block_height as height
-		from addresses left join transactions as ftxd on funding_tx_row_id=ftxd.id
-		where address = ANY($1)
-		UNION
-		SELECT DISTINCT spending_tx_hash as tx_hash, stxd.time as tx_time, stxd.block_height as height from addresses  
-		left join transactions as stxd on spending_tx_hash=stxd.tx_hash  
-		where address = ANY($1) and spending_tx_hash IS NOT NULL) select tx_hash, height from these order by tx_time desc;`
+	SelectAddressesAllTxn = `SELECT tx_hash, block_time as tx_time, ftxd.block_height as height 
+		from addresses left join transactions as ftxd on funding_tx_row_id=ftxd.id 
+		where address = $1 order by tx_time desc;`
 
-	SelectAddressesTxnByFundingTx = `SELECT funding_tx_vout_index, spending_tx_hash, spending_tx_vin_index, 
+	SelectAddressesTxnByFundingTx = `SELECT tx_vin_vout_index, tx_hash, tx_vin_vout_index, 
 		block_height FROM addresses LEFT JOIN 
-		transactions on transactions.tx_hash=spending_tx_hash WHERE 
-		address = ANY ($1) and funding_tx_hash=$2;`
+		transactions on transactions.tx_hash=tx_hash and is_funding = FALSE WHERE 
+		address = $1 and tx_hash=$2;`
 
 	SelectAddressUnspentCountAndValue = `SELECT COUNT(*), SUM(value) FROM addresses 
-	    WHERE address=$1, is_funding = FALSE and in_out_row_id = 0;`
+	    WHERE address = $1 and is_funding = TRUE and in_out_row_id < 1;`
 
 	SelectAddressSpentCountAndValue = `SELECT COUNT(*), SUM(value) FROM addresses 
-	    WHERE address=$1, is_funding = FALSE and in_out_row_id > 0;`
+	    WHERE address = $1 and is_funding = FALSE and in_out_row_id > 0;`
 
 	SelectAddressUnspentWithTxn = `SELECT
 									addresses.address,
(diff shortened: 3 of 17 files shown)


chappjc:
Excellent write-up of the design

To merge we'll just need to ensure all address queries are functional.

chappjc:
oops

dmigwi:
If the design is okay. I can go ahead and fix all the other queries that use the addresses table since even the table columns have also changed.

chappjc:
Yes the design does look OK.  Please do make sure the other queries can be fixed without much trouble or performance regressions.

dmigwi:
@chappjc It seems that I have to add column 8 and 9 in the proposed design due to this https://github.com/decred/dcrdata/blob/master/db/dcrpg/internal/addrstmts.go#L85-L96

chappjc:
Many of these may not actually be wired up.  If they aren't used and we can't think of a need, they can be removed

dmigwi:
Here is the final proposal of how the frontend should look while taking full advantage of the new addresses table. 

final-shot

`Matched Tx` is new column that defines the relationship between the spent tx and the respective tx that funded it.

chappjc:
As I said in Slack, I think that Matched Tx column addresses my concerns completely.  We get sensible sorting without the losing the ability to see where the funds flow

dmigwi:
Here is the updated version of how the new address page should look.

screen shot 2018-04-23 at 01 12 44

dmigwi:
This PR should be ready for review and testing.

explorer/explorertypes.go
```diff
@@ -51,14 +51,17 @@ type AddressTx struct {
 	FormattedTime string
 	ReceivedTotal float64
 	SentTotal     float64
+	IsFunding     bool
+	MatchedTx     uint64
```
chappjc: This seems to be the source of several of the issues I raised.  For locating an input's/output's corresponding output/input, the row ID in the address table would indeed be needed.  I think this is what you are storing in `MatchedTx` now.  However, for just linking to the transaction page for the transaction of the corresponding in/out, we just need the tx_hash or row ID _in the transactions_ table.

main.go
```diff
@@ -484,6 +484,7 @@ func mainCore() error {
 	webMux.Mount("/explorer", explore.Mux)
 	webMux.Get("/blocks", explore.Blocks)
 	webMux.Get("/mempool", explore.Mempool)
+	webMux.With(explorer.AddressRowIDCtx).Get("/tx/addressid/{id}", explore.TxByAddressRowID)
```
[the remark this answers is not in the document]
dmigwi: This reason why I choose to use `/tx/addressid/{id}` instead of `/tx/{txid}` is because we are using `in_out_row_id` value that is stored in the database to match one tx to another. `in_out_row_id` is the row id of the matching tx. I initial thought of using `in_out_row_id` to fetch the `tx_hash` (in the respective queries) or simply the `txid` that can be used with `/tx/{txid}` but it felt that reusing `/tx/{txid}` would be very inefficient and reusing what is readily available from the database would have been better (`in_out_row_id`).

db/dcrpg/pgblockchain.go
```diff
@@ -1418,24 +1431,26 @@ func (pgb *ChainDB) storeTxns(msgBlock *MsgBlockPG, txTree int8,
 		}
 	}
 
-	// Store tx Db IDs as funding tx in AddressRows and rearrange
+	// Store tx Db block time as block time in AddressRows and rearrange
+	// Also set is_funding to true since this are funding tx inputs
```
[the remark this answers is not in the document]
dmigwi: Noted.

db/dbtypes/types.go
```diff
@@ -248,16 +248,14 @@ type Vout struct {
...
-	FundingTxHash      string
-	FundingTxVoutIndex uint32
-	VoutDbID           uint64
-	Value              uint64
-	SpendingTxDbID     uint64
-	SpendingTxHash     string
-	SpendingTxVinIndex uint32
-	VinDbID            uint64
+	Address        string
+	InOutRowID     uint64
```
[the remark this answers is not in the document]
dmigwi: Noted

db/dcrpg/queries.go
```diff
@@ -492,19 +464,79 @@ func SetSpendingForVinDbID(db *sql.DB, vinDbID uint64) (int64, error) {
...
 }
 
-func SetSpendingForFundingOP(db *sql.DB,
-	fundingTxHash string, fundingTxVoutIndex uint32,
-	spendingTxDbID uint64, spendingTxHash string, spendingTxVinIndex uint32,
-	vinDbID uint64) (int64, error) {
-	res, err := db.Exec(internal.SetAddressSpendingForOutpoint,
-		fundingTxHash, fundingTxVoutIndex,
-		spendingTxDbID, spendingTxHash, spendingTxVinIndex, vinDbID)
+// SetSpendingForFundingOP inserts a new spending tx row
```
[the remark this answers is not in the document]
dmigwi: Noted

db/dcrpg/queries.go
```diff
@@ -1351,19 +1350,34 @@ func InsertVouts(db *sql.DB, dbVouts []*dbtypes.Vout, checked bool) ([]uint64, [
...
-		dbA.FundingTxHash, dbA.FundingTxVoutIndex, dbA.VoutDbID,
-		dbA.Value).Scan(&id)
+	err := db.QueryRow(sqlStmt, dbA.Address, dbA.InOutRowID, dbA.TxHash,
+		dbA.TxVinVoutIndex, dbA.VinVoutDbID, dbA.Value, dbA.TxBlockTime,
+		dbA.IsFunding).Scan(&id)
 	return id, err
 }
 
-func InsertAddressOuts(db *sql.DB, dbAs []*dbtypes.AddressRow, dupCheck bool) ([]uint64, error) {
+// InsertAddressRows can insert for both inpoints and outpoints from the respective transactions
```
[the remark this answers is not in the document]
dmigwi: Noted

explorer/explorertypes.go
dmigwi: To fix this issue we might need to change the addresses table schema so that instead of storing the row id of the matching tx we can store the tx_hash for that specific input/output.

db/dcrpg/queries.go
```diff
@@ -1351,19 +1350,34 @@ func InsertVouts(db *sql.DB, dbVouts []*dbtypes.Vout, checked bool) ([]uint64, [
...
+	return &data, err
+}
+
+// RecieveTxHashByRowID fetches the tx_hash associated with a given row id
+func RecieveTxHashByRowID(db *sql.DB, id uint64) (data string, err error) {
+	err = db.QueryRow(internal.SelectAddressByRowID, id).Scan(&data)
+	return
+}
+
+// InsertAddressRow can insert an inpoint and outpoint for the respective transaction.
```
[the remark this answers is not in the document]
dmigwi: Noted

db/dcrpg/queries.go
```diff
@@ -1351,19 +1350,34 @@ func InsertVouts(db *sql.DB, dbVouts []*dbtypes.Vout, checked bool) ([]uint64, [
 	return ids, addressRows, dbtx.Commit()
 }
 
-func InsertAddressOut(db *sql.DB, dbA *dbtypes.AddressRow, dupCheck bool) (uint64, error) {
+func RecieveAddressByTxHash(db *sql.DB, tx_hash string) (*dbtypes.AddressRow, error) {
+	var data dbtypes.AddressRow
+	err := db.QueryRow(internal.SelectAddressByTxHash, tx_hash).Scan(&data.InOutRowID, &data.Address, &data.Value)
```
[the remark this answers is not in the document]
dmigwi: Sure we need discuss on this. I thought that every tx_hash is unique??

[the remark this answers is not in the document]
chappjc: `SelectAddressByTxHash` is a query in to the `addresses` table, which indexes occurrences of addresses in tx inputs and outputs.  Of of course transactions may have numerous inputs and outputs, so in the following query,

```go
SelectAddressByTxHash = `select id, address, value from addresses 
    where tx_hash=$1 and is_funding = true`
```

This loads all inputs and outputs for the transaction with that hash.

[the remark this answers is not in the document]
dmigwi: but `is_funding = true` make a small difference. Question: by default do we have unique `tx_hash` value or the get repeated once in a while?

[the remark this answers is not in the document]
chappjc: On the blockchain, you will only have 1 valid transaction with a given tx_hash.  However, there may be multiple occurrences of the same transaction as a result of blocks being invalidated by stake votes.  A regular tx that was invalidated will normally be picked up and mined in a different block, but the chain is unbroken so that tx will appear in both places, just only one is valid.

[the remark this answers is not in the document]
chappjc: But in the addresses table, a specific `tx_hash` value may be repeated many times for the reasons I mentioned.

db/dcrpg/internal/txstmts.go
```diff
@@ -61,6 +61,8 @@ const (
 	SelectTxByHash       = `SELECT id, block_hash, block_index, tree FROM transactions WHERE tx_hash = $1;`
 	SelectTxsByBlockHash = `SELECT id, tx_hash, block_index, tree FROM transactions WHERE block_hash = $1;`
 
+	SelectTxBlockTimeByHash = `SELECT block_time FROM transactions where tx_hash = $1`
```
dmigwi: This query should also be wrong too @chappjc. How do I best select the block time here taking into consideration that tx_hash may not be unique?

chappjc: Take the latest block time.  :)

dmigwi: ok.

chappjc: LIMIT 1 ORDER BY block_time DESC

or something like that

dmigwi: The same solution should perfectly work here 
```SelectAddressByTxHash = select id, address, value from addresses 
    where tx_hash=$1 and is_funding = true;```

Comment (dmigwi) [reply db/dcrpg/internal/txstmts.go]: This should work seamlessy 
```SelectTxBlockTimeByHash = SELECT block_time FROM transactions where tx_hash = $1 ORDER BY block_time DESC LIMIT 1;
```

db/dcrpg/queries.go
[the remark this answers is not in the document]
dmigwi: I think here, picking the details of the row with the earliest block time for multiple rows case scenario in the result makes more sense to me here.

db/dcrpg/internal/txstmts.go
chappjc: But this is a query in the addresses table, not the transactions table.  You'd need the output index to get a single row, the one you want I should say

chappjc: Ah, yes.  I see you changed to the transactions table.  I gotta go, but we'll chat later

db/dcrpg/queries.go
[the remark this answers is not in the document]
dmigwi: My current solution.
```
SelectAddressByTxHash = select id, address, value from addresses 
	   where tx_hash=$1 and is_funding = true ORDER BY block_time ASC LIMIT 1;
```

dmigwi:
### To Successfully Test this PR
1. Drop the old dcrdata database and create a new empty dcrdata database.
2. Delete the data folder in dcrdata folder where dcrdata.conf file is located.
3. Restart the data migration again by running the dcrdata and dcrd binaries.

Everything should work seamlessly as if you are doing a fresh install. (was added to the README.md file as **Upgrading Instructions**)

README.md
```diff
@@ -93,6 +93,30 @@ First, update the repository (assuming you have `master` checked out):
 Look carefully for errors with `git pull`, and reset locally modified files if
 necessary.
 
+## Upgrading Instructions 
+
+*__Only necessary while upgrading from upgrading from v1.3.2.0.0.0.0.0.0.1 or below.__*
```
chappjc: "from v1.3.2 or below" (pretend like the other one doesn't exist. It was a joke)

README.md
```diff
@@ -93,6 +93,30 @@ First, update the repository (assuming you have `master` checked out):
...
+
+1. Drop the old dcrdata database and create a new empty dcrdata database.
```
chappjc: In db/dcrpg/tables.go, please bump `tableMajor` to 3 for this change.

dmigwi: Noted

README.md
dmigwi: Noted

chappjc:
Using dcrdata for the firs time stalls here:

```
2018-05-02 04:56:00.739 [INF] PSQL: Indexing addresses table on tx hash...
2018-05-02 04:56:54.395 [INF] PSQL: Indexing addresses table on matching tx hash...
2018-05-02 04:56:59.648 [INF] PSQL: Indexing addresses table on block time...
2018-05-02 04:57:08.171 [INF] PSQL: Indexing tickets table on ticket hash...
2018-05-02 04:57:14.191 [INF] PSQL: Indexing tickets table on transaction Db ID...
2018-05-02 04:57:15.162 [INF] PSQL: Populating spending tx info in address table...
2018-05-02 04:57:28.589 [INF] PSQL: Updating spending tx info for 10495511 addresses...
2018-05-02 04:57:28.590 [INF] PSQL: Updating from vins 0 to 249999...
```

SIGABRT shows two stuck goroutines:

```
goroutine 1 [chan receive, 175 minutes]:
main.waitForSync(0xc4202ce4e0, 0xc4202ce540, 0xc42514c101, 0xc42014c600, 0xdbde00, 0xc42561a0e0, 0xc42014c600, 0xc425010001)
	/home/jon/go/src/github.com/decred/dcrdata/main.go:537 +0x174
main.mainCore.func4(0x10001, 0x0, 0x2, 0x2, 0x3, 0xc42560a980)
	/home/jon/go/src/github.com/decred/dcrdata/main.go:321 +0x25b
main.mainCore(0x0, 0x0)
	/home/jon/go/src/github.com/decred/dcrdata/main.go:323 +0x142a
main.main()
	/home/jon/go/src/github.com/decred/dcrdata/main.go:522 +0x26

goroutine 20 [syscall]:
os/signal.signal_recv(0xdb8740)
	/usr/lib/go/src/runtime/sigqueue.go:139 +0xa6
os/signal.loop()
	/usr/lib/go/src/os/signal/signal_unix.go:22 +0x22
created by os/signal.init.0
	/usr/lib/go/src/os/signal/signal_unix.go:28 +0x41

goroutine 21 [chan receive, 425 minutes]:
github.com/decred/dcrdata/notification.(*collectionQueue).ProcessBlocks(0xc4201435c0)
	/home/jon/go/src/github.com/decred/dcrdata/notification/ntfnhandlers.go:88 +0x29f
created by github.com/decred/dcrdata/notification.MakeNodeNtfnHandlers
	/home/jon/go/src/github.com/decred/dcrdata/notification/ntfnhandlers.go:141 +0x96
```

dmigwi:
I have managed to fast forward the data migration to that point you are experiencing errors using a database dump I had kept earlier with only funding tx output added.

```
2018-05-02 22:58:05.110 [INF] PSQL: Indexing misses table...
2018-05-02 22:58:05.206 [INF] PSQL: Indexing addresses table on tx hash...
2018-05-02 22:58:25.109 [INF] PSQL: Indexing addresses table on matching tx hash...
2018-05-02 22:58:32.561 [INF] PSQL: Indexing addresses table on block time...
2018-05-02 22:58:39.959 [INF] PSQL: Indexing tickets table on ticket hash...
2018-05-02 22:58:43.925 [INF] PSQL: Indexing tickets table on transaction Db ID...
2018-05-02 22:58:45.490 [INF] PSQL: Populating spending tx info in address table...
2018-05-02 22:58:53.545 [INF] PSQL: Updating spending tx info for 8104093 addresses...
2018-05-02 22:58:53.546 [INF] PSQL: Updating from vins 0 to 249999...
```

On the backend `In the addresses table` things seems to be happening
```
dcrdata=# select count(*) from addresses where matching_tx_hash != '';
 count
-------
 86714
(1 row)

dcrdata=# select count(*) from addresses where matching_tx_hash != '';
 count
-------
 90523
(1 row)

dcrdata=# select count(*) from addresses where matching_tx_hash != '';
 count
-------
 97314
(1 row)
```

chappjc:
Then we need to understand why this is getting stuck for so long.  It sat for almost 4 hours.  This step usually finishes in maybe 45 minutes.

dmigwi:
@chappjc How were you able to check and see that the two goroutine were stuck?

I take it that the above error happened when you were migrating the mainnet. By the way, In this PR I have added extra indexes to the addresses table that should slow the data writing (Caveats listed above in this PR) but speed up the data reading.

Since I can't replicate your error with the testnet data, I will be setting up a vm in gcloud to do data migration on mainnet.

chappjc:
@dmigwi There was no error output by the program.  It was just stuck.  What looks like an error is part of the stack trace I obtained by `killall -ABRT dcrdata`, which tells the go runtime to dump a stack trace and exit.

But what you said about the addresses table indexes made me realize the problem.  For the exact reason you mentioned (indexes slowing down write operations), I had delayed creation indexes for the addresses until _after_ the spending info updates.  Note the comment on `UpdateSpendingInfoInAllAddresses`

https://github.com/decred/dcrdata/blob/d3288e7ea631f3ad9fdadaec9005530f8083c435/db/dcrpg/pgblockchain.go#L1571-L1578

And how the sync functions will remove the address indexes before running the above function:

https://github.com/decred/dcrdata/blob/d3288e7ea631f3ad9fdadaec9005530f8083c435/db/dcrpg/sync.go#L224-L238

https://github.com/decred/dcrdata/blob/d3288e7ea631f3ad9fdadaec9005530f8083c435/cmd/rebuilddb2/rebuilddb2.go#L375-L389

So addressing this may be just a matter of updating this stuff:

https://github.com/decred/dcrdata/blob/d3288e7ea631f3ad9fdadaec9005530f8083c435/db/dcrpg/pgblockchain.go#L1016-L1046

chappjc:
update comment above

dmigwi:
So the delaying creation of addresses indexes might be a probable solution. Let work on that now.

dmigwi:
for 
```
2018-05-02 22:58:53.546 [INF] PSQL: Updating from vins 0 to 249999...
2018-05-02 23:12:03.834 [INF] PSQL: Updating from vins 250000 to 499999...
2018-05-02 23:21:04.067 [INF] PSQL: Updating from vins 500000 to 749999...
2018-05-02 23:30:29.418 [INF] PSQL: Updating from vins 750000 to 999999...
2018-05-02 23:40:01.484 [INF] PSQL: Updating from vins 1000000 to 1249999...
2018-05-02 23:55:51.884 [INF] PSQL: Updating from vins 1250000 to 1499999...
2018-05-03 00:03:41.708 [INF] PSQL: Updating from vins 1500000 to 1749999...
2018-05-03 00:14:51.742 [INF] PSQL: Updating from vins 1750000 to 1999999...
2018-05-03 00:22:06.799 [INF] PSQL: Updating from vins 2000000 to 2249999...
2018-05-03 00:29:33.009 [INF] PSQL: Updating from vins 2250000 to 2499999...
```

What if we also reduced the ranges to e.g. from `0 to 249999...` to `0 to 49999` This will work too in the worst case scenario where a user can see more progress in the updating of the spending tx inputs

chappjc:
yeah that's fine

dmigwi:
I have finished updating those functions here is the commit https://github.com/decred/dcrdata/pull/427/commits/21e64aa52a47a0d7626b3149db0725916c8f0118

dmigwi:
@chappjc You plan has worked again, we have managed to reduce time taken to `Updating from vins 0 to 249999...` (using testnet) from 14 minutes to 6 minutes. More than 50% improvement.

```
2018-05-03 02:26:36.840 [INF] PSQL: Updating from vins 0 to 49999...
2018-05-03 02:27:49.853 [INF] PSQL: Updating from vins 50000 to 99999...
2018-05-03 02:29:00.044 [INF] PSQL: Updating from vins 100000 to 149999...
2018-05-03 02:30:14.893 [INF] PSQL: Updating from vins 150000 to 199999...
2018-05-03 02:31:32.159 [INF] PSQL: Updating from vins 200000 to 249999...
2018-05-03 02:32:49.232 [INF] PSQL: Updating from vins 250000 to 299999...
```
The latest update drops indexes on `block_time` and `matching_tx_hash` that existed before but maintains an index on `tx_hash` as it improves the speed of the query the fetches the repeated fields between an input and an output i.e. address and value of the tx output.

chappjc:
nice.  Retesting.

dmigwi:
Just finished data loading on mainnet into my gcloud vm instance with machine type: n1-standard-2 (`2 vCPUs(Haswell), 7.5 GB memory`) and it took approximately 12 hr to do it from scratch.

```
2018-05-03 15:18:24.310 [INF] SQLT: Rescan finished successfully at height 235479.
2018-05-03 15:18:24.310 [INF] DATD: SQLite sync ended at height 235479
2018-05-03 15:18:29.159 [INF] PSQL: Avg. speed: 3 tx/s, 13 vout/s
2018-05-03 15:18:29.159 [INF] PSQL: Sync finished at height 235479. Delta: 70 blocks, 1148 transactions, 3007 ins, 4549 outs
2018-05-03 15:18:29.159 [INF] DATD: PostgreSQL sync ended at height 235479
2018-05-03 15:18:29.159 [INF] DATD: All ready, at height 235479.
2018-05-03 15:18:29.358 [INF] EXPR: Signaling new block to 0 websocket clients.
2018-05-03 15:18:31.561 [INF] EXPR: Now serving explorer on http://127.0.0.1:7777/
2018-05-03 15:18:31.561 [INF] JAPI: Now serving API on http://127.0.0.1:7777/
2018-05-03 15:21:40.364 [INF] PSQL: Dcur2mcGjmENx4DhNqDctW5wJCVyT3Qeqkx: 266346 spent totalling 690320.885434 DCR, 1380 unspent totalling 3110.273563 DCR
2018-05-03 15:21:40.364 [INF] PSQL: Caching address receive count for address Dcur2mcGjmENx4DhNqDctW5wJCVyT3Qeqkx: count = 267726 at block 235479.
2018-05-03 15:24:41.465 [INF] SKDB: Connected block 235480 to stake DB.
2018-05-03 15:24:41.470 [INF] BLKD: Block height 235480 connected. Collecting data...
```

chappjc:
OMG!  that's like 6x too slow.  Did you appropriate tune postgresql.conf first?

chappjc:
This stuff: https://github.com/decred/dcrdata/blob/master/db/dcrpg/postgresql-tuning.conf
Especially synchronous_commit=off

dmigwi:
Not really although its much better than when I was setting up dcrdata for the first time in my machine. If there is a way to fine tune it, I will look it to it.  

It actually took more that 12 hrs: Approximately (15 hrs)
```
2018-05-03 01:38:13.087 [INF] SKDB: Advancing ticket pool DB to tip via diffs...
2018-05-03 01:38:13.089 [INF] SKDB: Pre-populating live ticket cache...
2018-05-03 01:38:13.089 [INF] DATD: SQLite DB successfully opened: dcrdata.sqlt.db
2018-05-03 01:38:13.096 [INF] PSQL: Creating the "vin_t" type.
2018-05-03 01:38:13.101 [INF] PSQL: Creating the "vout_t" type.
```
to 

```
2018-05-03 15:18:29.159 [INF] DATD: All ready, at height 235479.
2018-05-03 15:18:29.358 [INF] EXPR: Signaling new block to 0 websocket clients.
2018-05-03 15:18:31.561 [INF] EXPR: Now serving explorer on http://127.0.0.1:7777/
2018-05-03 15:18:31.561 [INF] JAPI: Now serving API on http://127.0.0.1:7777/
2018-05-03 15:21:40.364 [INF] PSQL: Dcur2mcGjmENx4DhNqDctW5wJCVyT3Qeqkx: 266346 spent totalling 690320.885434 DCR, 1380 unspent totalling 3110.273563 DCR
2018-05-03 15:21:40.364 [INF] PSQL: Caching address receive count for address Dcur2mcGjmENx4DhNqDctW5wJCVyT3Qeqkx: count = 267726 at block 235479.
```

chappjc:
For reference, average processing speed should be around 250 tx/s, 700 vouts/s (vs. 3 tx/s, 13 vout/s), so I suspect `synchronous_commit` was `on` for you.

But I'm starting another fresh sync because again it got stuck for hours updating spend info from vins.  It's possible I restarted it before I got the indexing fixes.

dmigwi:
Here is the configuration I used. https://github.com/dmigwi/dcrdata/blob/proposed_addresses_table/db/dcrpg/postgresql-tuning.conf#L15-L18

 `(3 tx/s, 13 vout/s)` belongs to the second batch of new blocks discovered after updating vins entries that were added by the first batch of blocks that had average processing speed of `Avg. speed: 144 tx/s, 531 vout/s`

dmigwi:
I have figured out a bug on this PR thanks to @papa Carp testing this PR. I am fixing ASAP.

db/dcrpg/internal/addrstmts.go
```diff
@@ -34,76 +30,80 @@ const (
...
-	SelectAddressIDByVoutIDAddress = `SELECT id FROM addresses
-		WHERE address=$1 and vout_row_id=$2
-		ORDER BY id DESC;`
-
-	SetAddressSpendingForID = `UPDATE addresses SET spending_tx_row_id = $2, 
-		spending_tx_hash = $3, spending_tx_vin_index = $4, vin_row_id = $5 
-		WHERE id=$1;`
-	SetAddressSpendingForOutpoint = `UPDATE addresses SET spending_tx_row_id = $3, 
-		spending_tx_hash = $4, spending_tx_vin_index = $5, vin_row_id = $6 
-		WHERE funding_tx_hash=$1 and funding_tx_vout_index=$2;`
```
dmigwi: Is it possible that `SetAddressSpendingForOutpoint` would update multiple rows often?

chappjc: Extremely rarely, and it would not be incorrect to do so.

papacarp: @dmigwi should i have it rebuild my entire db up again with this fix so I can keep testing?

dmigwi: @papacarp What I did to save on time was, to dump the state of the postgres db immediately after all the blocks were loaded. Not much was changed on how blocks are loaded thus its safe to restart the data migration from that state with all blocks loaded but not yet started updating addresses table from vins table content.

Command to dump and load the database

```bash
// Dump the postgres database on the shell
pg_dump -U {db_user} {db_name} > dbexport.pgsql

// Command to load the db
psql -U  {db_user} {db_name} < dbexport.pgsql
```

chappjc:
Sync timing on same machine (Intel [Q9650](https://ark.intel.com/products/35428/Intel-Core2-Quad-Processor-Q9650-12M-Cache-3_00-GHz-1333-MHz-FSB) <-- circa 2008, 8GB DDR2 RAM, SSD):

```
master
19:18:30 - start
00:03:05 - blocks end - Avg. speed: 245 tx/s, 904 vout/s
00:56:00 - end

proposed address table (8fc15158)
18:22:43 - start
23:06:50 - blocks end - Avg. speed: 246 tx/s, 907 vout/s
02:38:28 - end
```

The block sync that comes before indexing is essentially identical, while the steps that follow (namely "Updating spending tx info for 10549064 addresses...") take much longer with the new address table.  This is one place that needs some investigation.

The other area of concern is time to add new blocks.  The data I have indicates that adding blocks is extremely slow (~1 tx/s, ~7 vout/s).  This is a bigger problem unfortunately, which needs to be addressed.

I suggest timing ChainDB's Store functions (Debug level), identifying the bottleneck, and attempting to speed it up.

dmigwi:
>while the steps that follow (namely "Updating spending tx info for 10549064 addresses...") take much longer with the new address table. This is one place that needs some investigation.

The extra query is the first place I will visit.

>The other area of concern is time to add new blocks. The data I have indicates that adding blocks is extremely slow (<1 tx/s). This is a bigger problem unfortunately, which needs to be addressed.

I suspect that we are not deindexing the tables again while we are loading the blocks the second time.

Overall, you data loading time was quite fast possibly because my hard disk might not be SDD.

chappjc:
Regarding time to add new blocks, I'm referring to adding in two cases:

1. Regular startup sync, not initial rebuild.  Say you got it syncd, shut it down for a day, and started backup.  Sync in this case does not drop indexes.
2. New mined blocks while running.  Blocks must be processed with minimal latency.

dmigwi:
>The other area of concern is time to add new blocks. The data I have indicates that adding blocks is extremely slow (<1 tx/s). This is a bigger problem unfortunately, which needs to be addressed.

Regarding above I went on to use the `go tool trace` to find out what is happening under the hood.

go profiling

The screen shot  shows some gaps (in red circles) while some OS threads are running. My initial conclusion was that specific processes where taking too long to finish but now I think the red circles mark instances when specific OS threads were  a sleep and not running anything.

Mind you the above image is a snapshot  of the go routines running in a span of 7 minutes while the following was being logged on the terminal.
```
2018-05-12 03:32:29.919 [INF] PSQL: >>>>> Starting the go tool trace.
2018-05-12 03:32:29.946 [INF] SQLT: Current best block (chain server):     237932
2018-05-12 03:32:29.946 [INF] SQLT: Current best block (sqlite block DB):  237869
2018-05-12 03:32:29.947 [INF] SQLT: Current best block (stakedb):          237869
2018-05-12 03:32:32.467 [INF] PSQL: Processing blocks 237870 to 237932...
2018-05-12 03:32:33.296 [INF] SQLT: Scanning blocks 237870 to 237932 (41345 live)...
2018-05-12 03:33:30.074 [INF] PSQL: (  0 blk/s,    3 tx/s,    7 vin/sec,    9 vout/s)
2018-05-12 03:33:51.114 [INF] PSQL: (  0 blk/s,    2 tx/s,    5 vin/sec,    9 vout/s)
2018-05-12 03:34:21.830 [INF] PSQL: (  0 blk/s,    5 tx/s,   31 vin/sec,   17 vout/s)
2018-05-12 03:34:38.692 [INF] PSQL: (  0 blk/s,    7 tx/s,   15 vin/sec,   33 vout/s)
2018-05-12 03:34:50.234 [INF] PSQL: (  0 blk/s,    2 tx/s,    6 vin/sec,    8 vout/s)
2018-05-12 03:35:13.005 [INF] PSQL: (  0 blk/s,    4 tx/s,   12 vin/sec,   21 vout/s)
2018-05-12 03:35:36.078 [INF] PSQL: (  0 blk/s,    1 tx/s,    2 vin/sec,    4 vout/s)
2018-05-12 03:35:59.066 [INF] PSQL: (  0 blk/s,    1 tx/s,    2 vin/sec,    4 vout/s)
2018-05-12 03:37:07.067 [INF] PSQL: (  0 blk/s,    7 tx/s,   25 vin/sec,   28 vout/s)
2018-05-12 03:37:21.623 [INF] PSQL: (  0 blk/s,    1 tx/s,    4 vin/sec,    6 vout/s)
2018-05-12 04:17:03.696 [INF] PSQL: (  0 blk/s,    3 tx/s,   10 vin/sec,   12 vout/s)
``

**Comment by dmigwi** [comment]
By modifying the slow query to fetch data from vouts table instead of addresses table seems to have fixed the performance issue we were initially facing.

```
2018-05-14 03:23:55.165 [INF] SQLT: Scanning blocks 238501 to 238502 (40799 live)...
2018-05-14 03:23:55.563 [INF] SQLT: Rescan finished successfully at height 238502.
2018-05-14 03:23:55.563 [INF] DATD: SQLite sync ended at height 238502
2018-05-14 03:23:56.665 [INF] PSQL: Avg. speed: 32 tx/s, 165 vout/s
2018-05-14 03:23:56.665 [INF] PSQL: Sync finished at height 238502. Delta: 4 blocks, 65 transactions, 328 ins, 330 outs
2018-05-14 03:23:56.665 [INF] DATD: PostgreSQL sync ended at height 238502
```

**Comment by dmigwi** [comment]
@chappjc I think have addressed most of the critical performance issues you raised about this PR.

This PR is a little bit slower during data migration process than earlier versions because; initially to update a spent tx only one sql query was needed but not that the format has changed a minimum of 3 queries are need to implement the atomic transaction that updates the spent tx input.

**Comment by chappjc** [comment]
This is excellent progress.  My DB is rebuilding

**Comment by dmigwi** [comment]
Just discovered a small bug that was not formatting the address for the spent tx as required. I have fixed the issue and all is well. 

On can now rebuild the db and check if things add up as expected.

**Comment by chappjc** [comment]
The post-indexing sync step is sped up a lot, but there is an error with `SetSpendingForFundingOP`:

```
2018-05-15 22:38:16.943 [INF] PSQL: Processing blocks 238979 to 239000...
2018-05-15 22:38:17.991 [INF] SQLT: Scanning blocks 238979 to 239000 (40732 live)...
2018-05-15 22:38:37.587 [INF] PSQL: (  0 blk/s,   11 tx/s,  276 vin/sec,   40 vout/s)
2018-05-15 22:38:40.292 [INF] PSQL: Processing blocks 239001 to 239151...
2018-05-15 22:38:40.325 [INF] SQLT: Scanning blocks 239001 to 239151 (40783 live)...
2018-05-15 22:38:42.254 [ERR] PSQL: SetSpendingForFundingOP: RowsAffected: InsertAddressRow: pq: duplicate key value violates unique constraint "uix_addresses_vout_id" + <nil> (rollback)
2018-05-15 22:38:56.861 [INF] PSQL: (  5 blk/s,  101 tx/s,  237 vin/sec,  395 vout/s)
2018-05-15 22:39:04.304 [INF] SQLT: Rescan finished successfully at height 239151.
2018-05-15 22:39:04.304 [INF] DATD: SQLite sync ended at height 239151
2018-05-15 22:39:04.382 [INF] PSQL: Avg. speed: 66 tx/s, 262 vout/s
```

**Comment by dmigwi** [comment]
I think we have duplicates in the vins table that were not deleted.

my new query 
```
SelectAddressByTxHash = `SELECT script_addresses, value FROM vouts
		WHERE tx_hash = $1 AND tx_index = $2 and tx_tree = $3;`
```
depends on index to ensure all entries to vouts table are unique. https://github.com/decred/dcrdata/blob/master/db/dcrpg/internal/vinoutstmts.go#L120-L121

**Comment by dmigwi** [comment]
Or Something is wrong somewhere else:
```
2018-05-15 22:38:40.325 [INF] SQLT: Scanning blocks 239001 to 239151 (40783 live)...
2018-05-15 22:38:42.254 [ERR] PSQL: SetSpendingForFundingOP: RowsAffected: InsertAddressRow: pq: duplicate key value violates unique constraint "uix_addresses_vout_id" + <nil> (rollback)
```
Notice it has happened in the most recent blocks i.e. `239001 to 239151`

**Comment by dmigwi** [comment]
I figured out the issue. I am fixing it now.

**Comment by dmigwi** [comment]
For a couple of weeks now, I have been wondering why do the same query on the psql cmd interface take different times to finish executing. According to this https://stackoverflow.com/questions/7142335/how-does-postgresql-cache-statements-and-data and https://madusudanan.com/blog/understanding-postgres-caching-in-depth/ queries are cached especially when PREPARED statements are used.

It therefore goes without saying that prepared statements while doing data migration are bad news. I will get deeper on this and try to see how much efficiency can be achieved by eliminating the prepared statements while rebuilding the db.

**Comment by chappjc** [comment]
Now that the table format seems to have settled, please consider if there's any way to support an automated upgrade. If there is no way to upgrade automatically there must be some clear instructions rather than a cryptic error

File: README.md [hunk 13]
```diff
@@ -93,6 +93,30 @@ First, update the repository (assuming you have `master` checked out):
...
+ ```
+
+2. Delete the data folder in dcrdata folder where dcrdata.conf file is located
+  - Mac :  `~/Library/Application Support/Dcrdata/data`
+  - Linux :  `~/Dcrdata/data`
+  - Window :  `C:\Users\<your-username>\AppData\Local\Dcrdata\data` 
+  
+3. Restart the data migration again by running the dcrdata and dcrd binaries
+  - run the downloaded dcrd binary first then the generated dcrdata binary
+
```
dmigwi: @chappjc I added this Upgrading instructions to address the last issue you just raised.

chappjc: I missed that it contained the `DROP DATABASE dcrdata` instruction so that's good.  In addition to this, it should be clear enough.

https://github.com/decred/dcrdata/pull/427/files#diff-60206eba7205ea27cd7a8137d424f61cR269

It's unfortunately there's no possibility for upgrade, only rebuild, so we'll need to put in a few PRs ahead of this.  But it's in good shape I think.

dmigwi: The link in the above post redirects to this page. Was that the intention?

dmigwi:
Just fixed the merge conflicts here. I wish @papacarp could cross check if his changes are still intact.

papacarp:
I can take a look later today.  :)

papacarp:
It looks like the function you deleted is not used in the insight api code.  When you actually merge with master there will be some changes that need to be made.  specifically the queries we created for insight will have to be rewritten slightly.  Those should be straightforward to straighten out.  Once you merge I can go through the full test suite again and confirm things are working on insight still.

chappjc:
Please rebase on master.  I'm mostly happy with this, but I'm going to push some small edits and merge.  However, without an auto-upgrade, this has to go in 3.0, not 2.1

chappjc:
@dmigwi I've made a PR (https://github.com/dmigwi/dcrdata/pull/1) to supplement this PR.  Please review, comment, merge at your convenience.  After that please rebase this on master.

dmigwi:
Let me rebase now.

chappjc:
FYI, when you merged the PR you used the regular merge option, but we stick to "rebase and merge" or squash.  This should be OK though

dmigwi:
Done

PR #430, opened 2018-04-17
do not drop URL query params

### LINKED ISSUE #408: turbolinks dropping url params
When clicking "next" in the address table, query params such as "?n=20&start=80" get dropped after the page loads.

Looks like the cause might be the "Turbolinks-Location" header which is related to redirects
https://github.com/turbolinks/turbolinks#following-redirects

So maybe including the limit and offset params here where the header gets set will prevent the params getting dropped. https://github.com/decred/dcrdata/blob/master/explorer/explorerroutes.go#L392

chappjc:
Use `r.URL.RequestURI` to set the `Turbolinks-Location` html header correctly on the Block and Transaction page.  This was already addressed on the Address page.

### DIFF
--- a/views/address.tmpl
+++ b/views/address.tmpl
@@ -95,7 +95,7 @@
             <div class="col">
                 <div class="d-flex flex-wrap align-items-center justify-content-end mb-1">
                     <h5 class="mr-auto mb-0">History</h5>
-                    {{if lt .NumTransactions .KnownTransactions}}
+                    {{if lt .NumTransactions $TxnCount}}
                     <div div class="d-flex flex-wrap-reverse align-items-center justify-content-end">
                         <span class="fs12 nowrap text-right">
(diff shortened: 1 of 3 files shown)


chappjc:
Resolves issue #408

views/address.tmpl
```diff
@@ -38,7 +38,7 @@
                             {{if ge .KnownTransactions .MaxTxLimit}}
                             <td class="fs28 mono nowrap d-flex align-items-center justify-content-end">unavailable</td>
                             {{else}}
-                            <td class="fs28 mono nowrap fs16-decimal d-flex align-items-center justify-content-end">{{.Unspent}}</td>
+                            <td class="fs28 mono nowrap fs16-decimal d-flex align-items-center justify-content-end">{{.AmountUnspent.ToCoin}}</td>
```
chappjc: also fixes this issue with lite mode

Issue #327, opened 2018-01-30
Full transaction list in address view with postgres


raedah:
Jon Chappelow [11:15 PM]
In full mode, with psql, we get the transactions for a certain address by querying the `addresses` table, which is structured as follows...
each row is an occurence of an address in a certain transaction output
So there are multiple rows in this table for a given address
Thus, getting all the rows for a given address provides you right away with all the *funding transactions*
A spending transaction is defined as a transaction that spends one of these outpoints.  Because we want to simulate the concept of a balance for an address, we record these spending transactions in the addresses table as additional columns (edited)
So each row has 0 or 1 spending transactions, just as a transaction outpoint may only be spent once
If the spending transactions _columns_ are still `NULL`, then it is a UTXO (unspend transaction outpoint)
If there is spending txn data in those columns for some row, then that outpoint (funded via some address we're interested in) is spent
It's nice because when you get a row, you potentially get two transactions: the one that funded that outpoint (w/ an address) and one that spent that outpoint
This is very logical to me and reflects how coins actually flow in Decred (and Bitcoin, etc.)
With the `searchrawtransactions` RPC, it pulls any txns but it's really confusing.

I'm looking at how we use `SearchRawTransactionsVerbose` for a solution, but there's nothing simple or quick
If the 10 most recent transactions spend outpoints w/ the given address, those will show first (edited)
But the full mode focuses on funding transactions, which in this example are older
so it would show the 10 most recent funding transactions with any spendingtransactions right below the funding one (paired), and possibly none of the 10 spending transactions that `SearchRawTransactionsVerbose` returned (edited)
The psql addresses table does not even store tx heights, just hash and vout index.  The ordering comes from the query ordering by row id, which is the same ordering as the height that the funding tx would give
2.0 will bring a db scheme change requiring a rebuild of psql, so if we want to add spend height (and recieve/fund height) to the addresses table, we can do that
that would allow us to query w/ "order by spend_height limit 10", for example (edited)
not sure how far that would get us though...
I'd only like to add data to the tables that is actually useful

raedah [3:02 AM]
Each row is an occurrence of an address in a certain transaction output, and we are simply querying the table and getting the corresponding txhash. The same rows being returned also contain the txhash of the spending transaction (which is all the transactions that use that address as the input.) Sounds like we already ahve the complete return set we need, it would just be a matter of including the spending txhash set in the list of transactions we are displaying.
[...]

PR #433, opened 2018-04-18
[WIP] Address history update - add tx_time to existing address table

mikefullman:
This PR demonstrates one possible solution for #327 

Database changes are:
1.  TxTime added to vins
2.  SpendingTxTime added to addresses
3.  FundingTxTime added to addresses

Relationship to PR #427: 
This PR gives an alternative strategy to the full rebuild of the addresses table.  However, the rendering side makes use of the same struct for address rows (AddressRow_new) that I imagine will be used with that methodology as well and thus at least this part is reusable.

What is not complete:
1.  API has not been updated
2.  rebuilddb has not bee updated
3.  Only works with pg db.  no testing has been done with litedb

A rebuild of the database is required to add in the TxTimes but it takes about a full day on my computer.  Easiest way to accomplish that is to just update the name of the db in your dcrdata.conf to a new db and let the system re-initialize.

It *may* also be possible to use the following query to manually update addresses table so the performance can be evaluated.  To try a manual update make the necessary modifications to the table structure and then run the following queries:  (This query timed out every time I tried to run):

`UPDATE addresses as a1 SET spending_tx_time=transactions.time FROM addresses as a2 LEFT JOIN transactions ON transactions.tx_hash=a2.spending_tx_hash WHERE a2.spending_tx_hash IS NOT NULL;`

`UPDATE addresses as a1 SET funding_tx_time=transactions.time FROM addresses as a2 LEFT JOIN transactions ON transactions.id=a2.funding_tx_row_id`

ToDo- Metrics:
I plan to do some metrics to compare performance between this version and the existing addresses implementation.  For the following test cases I would like to rig up page load times.  If anyone would like to see other metrics then please send them my way.

| URL |      Proposed(ms)|  Current(ms) |
|----------|-------------:|------:|
|address/Dcur2mcGjmENx4DhNqDctW5wJCVyT3Qeqkx | | |
| address/Dcur2mcGjmENx4DhNqDctW5wJCVyT3Qeqkx?n=20&start=200000&txntype=all | | |
| address/Dcur2mcGjmENx4DhNqDctW5wJCVyT3Qeqkx?txntype=credit&n=1000&start=0 | | |
|address/Dcur2mcGjmENx4DhNqDctW5wJCVyT3Qeqkx?txntype=debit&n=1000&start=0| | |

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,4 @@ MANIFEST-*
 ffldb_stake*/
 *.pprof
 testnet/
+dcrdata.exe*
--- a/db/dbtypes/extraction.go
+++ b/db/dbtypes/extraction.go
@@ -116,6 +116,7 @@ func processTransactions(msgBlock *wire.MsgBlock, tree int8,
 				BlockHeight: txin.BlockHeight,
 				BlockIndex:  txin.BlockIndex,
 				ScriptHex:   txin.SignatureScript,
+				TxTime:      uint64(blockTime),
 			})
 		}
 
@@ -132,6 +133,7 @@ func processTransactions(msgBlock *wire.MsgBlock, tree int8,
 				Value:        uint64(txout.Value),
 				Version:      txout.Version,
 				ScriptPubKey: txout.PkScript,
+				TxTime:       uint64(blockTime),
 			}
 			scriptClass, scriptAddrs, reqSigs, err := txscript.ExtractPkScriptAddrs(
 				vout.Version, vout.ScriptPubKey, chainParams)
--- a/db/dbtypes/types.go
+++ b/db/dbtypes/types.go
@@ -243,6 +243,7 @@ type Vout struct {
 	Version          uint16           `json:"version"`
 	ScriptPubKey     []byte           `json:"pkScriptHex"`
 	ScriptPubKeyData ScriptPubKeyData `json:"pkScript"`
+	TxTime           uint64           `json:"tx_time"`
 }
 
 // AddressRow represents a row in the addresses table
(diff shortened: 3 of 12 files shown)


chappjc:
Good start.  Thanks for looking into this.  I don't care for wasted effort and duplicated work, but I think exploring the options and coordinating with @dmigwi is absolutely worth it.

mikefullman:
I don't like to waste or duplicate effort either, but I started the process before I saw @dmigwi's pull request and based on that and your encouragement on #327 comments I just took it to a conclusion.

Besides, value of learning Golang and decred:  Priceless.  

I'll work on other things now and we can come back to this if we need to.

See comparisons below:
	
| URL |      Proposed(ms)|  Current(ms) |
|----------|-------------:|------:|
|address/Dcur2mcGjmENx4DhNqDctW5wJCVyT3Qeqkx|3795|1839 |
|address/Dcur2mcGjmENx4DhNqDctW5wJCVyT3Qeqkx?n=20&start=200000&txntype=all|5136|3425|
|address/Dcur2mcGjmENx4DhNqDctW5wJCVyT3Qeqkx?txntype=credit&n=1000&start=0 |4051|3096|
|address/Dcur2mcGjmENx4DhNqDctW5wJCVyT3Qeqkx?txntype=debit&n=1000&start=0|2684|4531|

Methodology: 
1.  block updates off to avoid other thread delays
2. install timer code at start and end of address function:

> 	t := time.Now().UnixNano()
	//
	t2 := time.Now().UnixNano()
	te := (t2 - t) / 1000000
	log.Infof("%s render time: %v ms", r.URL.RequestURI(), te)

PR #436, opened 2018-04-20
Ticket Probability Fixes 

mikefullman:
Proposed fixes for #281 

the ROI calculation for the main page was adjusted to take into consideration 50% likelihood days based on the following Poisson calculation of probabilities:

> Chance for a ticket to be voted in a given time frame:
> 1 - (1 - P)^N
> 
> Where:
> P is the probability of a vote in one block.
> N is the number of blocks before ticket expiry.

**Main Page**
ROI_DAYS= [[ ln(50%) / ln(1-(5/ACTUAL_TICKET_POOL)) ] + TICKET_MATURITY +COINBASE_MATURITY ] / BLOCKS_PER_DAY

**Transaction Page**
Added probability bar chart for this ticket based on time to expiry.  I'm not sure if this is what @raedah was referring to in https://github.com/decred/dcrdata/issues/281#issuecomment-374038922 or not but I felt it would be useful on this page.

Please let me know if you have any thoughts on the luck section.

**Additional Issues Noted**
The expiry time is incorrect.  Expiry starts at the end of the ticket maturity and therefore actual expiry is 40960+256 blocks after ticket purchase.  I would like to adjust that chart so its only visible for mature tickets and starts at block 0 right when the ticket matures, and 100% is at 40960.

**Screenshots**
image

image

image

**Todo**

1. Fix Expiry on transactions page
2. Confirm boundary cases display the proper info for probability and expiry bar charts: immaturity -> maturity -> expiry.  Voted, revoked, expired, missed.
3. Check for divide by zero problems or at least see how Go handles them

### DIFF
--- a/explorer/explorerroutes.go
+++ b/explorer/explorerroutes.go
@@ -220,23 +220,23 @@ func (exp *explorerUI) TxPage(w http.ResponseWriter, r *http.Request) {
 					float64(exp.ChainParams.TicketExpiry)) * expirationInDays
 				if tx.TicketInfo.SpendStatus == "Voted" {
 					// Blocks from eligible until voted (actual luck)
-					tx.TicketInfo.ShortConfirms = exp.blockData.TxHeight(tx.SpendingTxns[0].Hash) -
+					tx.TicketInfo.TicketLiveBlocks = exp.blockData.TxHeight(tx.SpendingTxns[0].Hash) -
 						tx.BlockHeight - int64(exp.ChainParams.TicketMaturity) - 1
 				} else if tx.Confirmations >= int64(exp.ChainParams.TicketExpiry+
 					uint32(exp.ChainParams.TicketMaturity)) { // Expired
 					// Blocks ticket was active before expiring (actual no luck)
-					tx.TicketInfo.ShortConfirms = int64(exp.ChainParams.TicketExpiry)
+					tx.TicketInfo.TicketLiveBlocks = int64(exp.ChainParams.TicketExpiry)
 				} else { // Active
 					// Blocks ticket has been active and eligible to vote
-					tx.TicketInfo.ShortConfirms = blocksLive
+					tx.TicketInfo.TicketLiveBlocks = blocksLive
 				}
 				tx.TicketInfo.BestLuck = tx.TicketInfo.TicketExpiry / int64(exp.ChainParams.TicketPoolSize)
 				tx.TicketInfo.AvgLuck = tx.TicketInfo.BestLuck - 1
-				if tx.TicketInfo.ShortConfirms == int64(exp.ChainParams.TicketExpiry) {
+				if tx.TicketInfo.TicketLiveBlocks == int64(exp.ChainParams.TicketExpiry) {
 					tx.TicketInfo.VoteLuck = 0
 				} else {
 					tx.TicketInfo.VoteLuck = float64(tx.TicketInfo.BestLuck) -
-						(float64(tx.TicketInfo.ShortConfirms) / float64(exp.ChainParams.TicketPoolSize))
+						(float64(tx.TicketInfo.TicketLiveBlocks) / float64(exp.ChainParams.TicketPoolSize))
 				}
 				if tx.TicketInfo.VoteLuck >= float64(tx.TicketInfo.BestLuck-(1/int64(exp.ChainParams.TicketPoolSize))) {
 					tx.TicketInfo.LuckStatus = "Perfection"
--- a/explorer/explorerroutes.go
+++ b/explorer/explorerroutes.go
@@ -7,6 +7,7 @@ package explorer
 import (
 	"database/sql"
 	"io"
+	"math"
 	"net/http"
 	"strconv"
 
@@ -240,6 +241,7 @@ func (exp *explorerUI) TxPage(w http.ResponseWriter, r *http.Request) {
 				} else if tx.TicketInfo.VoteLuck == 0 {
 					tx.TicketInfo.LuckStatus = "No Luck"
 				}
+				tx.TicketInfo.Probability = (1 - math.Pow((1-(float64(exp.ChainParams.TicketsPerBlock)/float64(exp.ExtraInfo.PoolInfo.Size))), (float64(exp.ChainParams.TicketExpiry)-float64(tx.Confirmations)-float64(exp.ChainParams.TicketMaturity)))) * 100
 			}
 		}
 	}
--- a/explorer/explorerroutes.go
+++ b/explorer/explorerroutes.go
@@ -208,27 +208,35 @@ func (exp *explorerUI) TxPage(w http.ResponseWriter, r *http.Request) {
 					tx.TicketInfo.PoolStatus = poolStatus.String()
 				}
 				tx.TicketInfo.SpendStatus = spendStatus.String()
+				blocksLive := tx.Confirmations - int64(exp.ChainParams.TicketMaturity)
 				tx.TicketInfo.TicketPoolSize = int64(exp.ChainParams.TicketPoolSize) * int64(exp.ChainParams.TicketsPerBlock)
 				tx.TicketInfo.TicketExpiry = int64(exp.ChainParams.TicketExpiry)
 				expirationInDays := (exp.ChainParams.TargetTimePerBlock.Hours() * float64(exp.ChainParams.TicketExpiry)) / 24
 				maturityInDay := (exp.ChainParams.TargetTimePerBlock.Hours() * float64(tx.TicketInfo.TicketMaturity)) / 24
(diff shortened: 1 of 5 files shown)


chappjc:
Should the ROI period be the average vote time or the time where a ticket has a 50% chance of voting?  Say you have a large number of tickets.  It's the former.

chappjc:
I approve the expiry fix.  Never noticed that.  :)

The probability issue @raedah was referring to was from this: https://github.com/decred/dcrdata/pull/361#issuecomment-372167721.  So yes, it looks like you got that right.

chappjc:
Actually, I'd argue we want to compute ROI per month, since we're humans.  ;)  So I'd actually scale up the ROI that you get for the ~28 day period, and state "(monthly)".

But I still don't see a case for using the solution for p=0.5 for one ticket.  In aggregate, the average vote time, is relevant.

Consider computing the p at 30 days. It's not 0.5 because probability is not linearly related to time.  There is a slowly decaying tail out to 4.7 months, pulling the average up.

mikefullman:
As per offline dialog with developers
1.  Main page ROI was adjusted to use average + ticket maturity + coinbase maturity.
2.  Transaction pages were adjusted to have probability of NOT voting rather than probability of voting
3.  Expiry graph was fixed to go from 1 to 40960

The differences between existing beta explorer and this PR



| STATUS | OLD | NEW |
|---------|-------|-------|
| Just after buy | image | image |
| Just before mature | image | image |
| First mature block | image | image |
| Unvoted | image  | image |
| Voted and Mature | image | image | 
| Expired | image | image |

explorer/explorerroutes.go
```diff
@@ -241,7 +245,15 @@ func (exp *explorerUI) TxPage(w http.ResponseWriter, r *http.Request) {
...
-				tx.TicketInfo.Probability = (1 - math.Pow((1-(float64(exp.ChainParams.TicketsPerBlock)/float64(exp.ExtraInfo.PoolInfo.Size))), (float64(exp.ChainParams.TicketExpiry)-float64(tx.Confirmations)-float64(exp.ChainParams.TicketMaturity)))) * 100
+
+				// Chance for a ticket to NOT be voted in a given time frame:
+				// C = (1 - P)^N
+
+				// Where:
+				// P is the probability of a vote in one block. (votes per block / current ticket pool size)
+				// N is the number of blocks before ticket expiry. (ticket expiry in blocks - (number of blocks since ticket purchase - ticket maturity))
+				// C is the probability (chance)
+				tx.TicketInfo.Probability = (math.Pow((1 - (float64(exp.ChainParams.TicketsPerBlock) / float64(exp.ExtraInfo.PoolInfo.Size))), (float64(exp.ChainParams.TicketExpiry) - (float64(tx.Confirmations) - float64(exp.ChainParams.TicketMaturity))))) * 100
```
chappjc: By splitting this up, it would be easier to follow.

e.g.:
```
blocksLive := float64(tx.Confirmations) - float64(exp.ChainParams.TicketMaturity)
pVote := float64(exp.ChainParams.TicketsPerBlock) / float64(exp.ExtraInfo.PoolInfo.Size)
// or just pVote := 1 / exp.ChainParams.TicketPoolSize
...

File: views/tx.tmpl [hunk 2]
```diff
@@ -53,110 +53,110 @@
...
+                                {{if (or (lt .TicketInfo.TicketExpiry .TicketInfo.ShortConfirms) (ne .TicketInfo.SpendStatus "Voted"))}} possible, {{.TicketInfo.LuckStatus}} {{else}} final, {{.TicketInfo.LuckStatus}} {{end}}
+                                (Best: {{.TicketInfo.BestLuck}} Avg: {{.TicketInfo.AvgLuck}})
+                            </td>
+                        </tr>
+                    {{end}}
+                    
+                    {{if and (and (eq .Mature "True")  (ne .TicketInfo.SpendStatus "Voted")) (eq .TicketInfo.PoolStatus "live")}}
+                        {{if ge .TicketInfo.TicketExpiry .TicketInfo.ShortConfirms}}
+                            <tr>
+                                <td class="text-right pr-2 h1rem p03rem0">PROBABILITY</td>
```
chappjc: I'm starting to think this would be more accessible to the user phrased like "Chance of expiry"

chappjc:
Need to fix this actually:

    explorer/explorer.go:1::warning: file is not gofmted with -s (gofmt)

raedah:
Luck on Expired should say 'final' instead of 'possible'.

raedah:
LGTM. Maybe make the c in chance lower case.

mikefullman:
It didn't make sense to me at first either.  My first thought was to agree with you, but since we already have another var that is similar, _blocksLive_ which is close but not quite what we want with ShortConfirms, I would recommend we keep the variable name but update the definition comments a bit to make it easier and more precise to understand.  Happy to do whatever you want though.

old:
```go
ShortConfirms        int64   // Total number of confirms up until the point the ticket votes or expires
```

new:
```go
ShortConfirms        int64   // Total number of confirms after maturity up until the point the ticket votes or expires
```

mikefullman:
Great feedback!  Happy to look at that.  There are three outstanding issues now as far as I know.  I wasn't sure how much to keep pulling into this PR.

1.  I could go ahead and fix #440. (immature coinbase transactions do not show correct bar chart)
2.  I noted with @raedah that the existing design makes use of things like ``if tx.Type == "Ticket" {`` and also ``if tx.TicketInfo.SpendStatus == "Voted" {`` and in the template `` {{if and (and (eq .Mature "True")  (ne .TicketInfo.SpendStatus "Voted")) (eq .TicketInfo.PoolStatus "live")}}``.  These types of things can cause problems later when English words are used in comparisons to control program functionality.  If we decide to ever do a translation or even change one of those things from "live" to "Live" then it can cause bugs.  I would recommend using const's instead of english word comparisons for program flow and pull the English words in last second in the view.  @raedah recommended a separate PR.  I just haven't gotten to putting the issue in.
3.  Missed/revoked ticket rendering.  I recommend that I put in a change to remove the luck line item for missed tickets regardless of their revoked/unspent status.  The expiry chart is already not present.  The only other thing to consider is revoked ticket maturity status.  I need to explore if ticket funds are immediately available (as I see when I look at some transactions) or if they need 256 blocks of maturity as the docs specify.    https://docs.decred.org/faq/proof-of-stake/voting-tickets/#6-what-happens-if-a-ticket-i-have-missed-a-vote

Here is existing status of missed/revoked tickets:
(note, expired doesn't mean much for a missed ticket, but I included it below because the existing design renders an expiry chart)

| STATUS | OLD | NEW |
|---------|-------|-------|
| Missed / Unspent |image | image |
| Missed / Revoked / _unexpired_ |image | image |
| Missed / Revoked / _expired_ |image | image |

chappjc:
Regarding (2), the use of string literals, I absolutely agree we should address this, but it's not necessary for this PR.  However, I'd suggest an approach like I used here:

https://github.com/decred/dcrdata/blob/e1c4f606aeab37e6171ee2677f701b89970c4c69/db/dbtypes/types.go#L39-L104

For missed, I believe there is an argument to be made for still showing luck since the ticket was still called in a lottery, but (possibly) the owner of the ticket misconfigured something, unplugged their router or whatever.  So this is sort of a lucky but incompetent situation.  That said, I agree we could hide luck stuff since it has the potential to be confusing.  Being missed changes things.

Regarding maturity of revoked tickets, I'm pretty sure it does not apply since there is no subsidy payed by the revocation transaction.

chappjc:
BTW, thanks for checking all the missed / spent statuses and showing how they appear.  This is always very helpful for PRs.  I agree with removing the expiry bar in all three cases.

chappjc:
If the changes are small, I'd like to see the coinbase maturity (1) fixed in this PR.  It's fresh in your mind and you know what to do.

mikefullman:
Ok, I think we are good to go!
1.  I changed `ShortConfirms ` to `TicketLiveBlocks`
2.  I removed Luck from all missed tickets
3.  I fixed the coinbase transactions to properly reflect maturity after coinbasematurity blocks.
4.  I noticed the days until mature was not working for voted blocks - fixed that.
5.  I went back and spot checked previous test cases to make sure nothing changed.

**LUCK REMOVED**

| STATUS | OLD | NEW |
|---------|-------|-------|
| Missed / Unspent |image | image |
| Missed / Revoked / _unexpired_ |image | image  |
| Missed / Revoked / _expired_ |image | image  |

**COINBASE MATURITY**

| STATUS | OLD | NEW |
|---------|-------|-------|
|Vote, 3 confirmations | image | image |
|coinbase, 255 confirms | image | image |

chappjc:
tACK

mikefullman:
Great!

PR #443, opened 2018-04-26
darken progress bar background for dark theme

gozart1:
old 
screen shot 2018-04-25 at 10 18 35 pm
new
screen shot 2018-04-25 at 10 18 44 pm


chappjc:
OK

Issue #423, opened 2018-04-04
GUI issues


raedah:
On the current https://beta.dcrdata.org/ the options box at the top right is flying way off to the right, looks like it was suppose to be aligned next to the search bar though. Having it on the far right is strange on a wide screen monitor.

At https://beta.dcrdata.org/block/000000000000000526926958a320be6b5a12761236356467e49b28c9f5eb362f  where it says "No missed votes in this block." there is an underline where there shouldnt be.

Change the color of the green progress bars to something that matches the color scheme.

chappjc:
It would also be good to be able to close the menu with either the Escape key or clicking off the menu.  Now it's just the X.

Regarding the underline below "No missed votes..."  I think this is an improvement as the text on its own looks a little funny.  Perhaps it doesn't need to be a line all the way across, but I think it needs something to punctuate it.

PR #454, opened 2018-04-29
refine hamburger menu style & animation

gozart1:
This addresses menu toggle alignment raised in https://github.com/decred/dcrdata/issues/423 and refines styling to match the design spec.

screen shot 2018-04-29 at 3 31 21 am

screen shot 2018-04-29 at 3 00 44 am
screen shot 2018-04-29 at 3 00 55 am

### DIFF
--- a/public/css/main.css
+++ b/public/css/main.css
@@ -146,6 +146,7 @@ body.darkBG #hamburger-menu a:hover {
 }
 body.darkBG #menu {
   background:  #2a2a2a;
+  box-shadow: 0px 0px 0px 1px #ffffff33;
 }
 body.darkBG .progress {
   background-color: #757575;
@@ -278,8 +279,8 @@ body.darkBG .progress {
 }
 
 #hamburger-menu {
+  flex: 0 40px;
   z-index: 10;
-  position: fixed;
   width: 100%;
   top: 0px;
   right: 0px;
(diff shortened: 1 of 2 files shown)


raedah:
tested, works great.

chappjc:
@raedah That space looks intentional, or at least it seems right.  Note that the right edge of the menu button is aligned with the right side of the Output table.

chappjc:
Please rebase.  The new parameters link just needs to land in the menu.  BTW, if you check the box to allow edits from maintainer, I should be able to resolve this stuff.

raedah:
To be more specific, the spacing is correct, but i expect the search bar (blue underline) to extend equally on the left and right side so that it includes the area shown in the red box.

chappjc:
Ah.  I see now.
@gozart1, would this change and the rebase be something you could hit up today or tomorrow?  Otherwise, it's totally cool for me to handle it, if you permit edits from maintainers

gozart1:
Oh its np, I'll fix it up now. I was also considering adding that off-menu click to close menu interaction, so we can wrap up https://github.com/decred/dcrdata/issues/423 ( close hotkey makes more sense as a separate PR focused on keyboard navigation comprehensively )

gozart1:
- rebased with latest master
- has balanced search input padding
- has off menu click close handler

chappjc:
tACK

Issue #381, opened 2018-03-06
Add APR calc


raedah:
Along side front page ticket ROI calculation, calculate out the annual percentage return.
+1.59% per ~28.44 days
The current ~monthly would get drawn out to 365 days.
Take into account the time remaining till the next block rewards adjustment, and the subsequent reward adjustments that happen every x (21.33 mainnet) days.

raedah:
While we are at it, need to make sure we are including ticket maturation time in the ROI calc.

raedah:
Lets add to the home page where it says  
TICKET ROI	+1.43% per ~28.44 days
it would become
TICKET ROI	+1.43% per ~28.44 days  (14.97% APR)

PR #455, opened 2018-05-02
APR Calculation on home page

papacarp:
This Proposed PR addresses APR calculations as referenced in #381
image

**Guidance is requested on the following items:**

1.  A new function, simulateAPR,  was added to explorer.go.  Architecturally I'm not sure if this is the best place for this.  Viable alternatives may be:  explorermiddleware.go or txhelpers.go
2.  StakeRewardAtBlock was originally created using a calculation and then replaced with an RPC call.  Either works fine so its just a matter of preference.  Seems like an RPC call would be less efficient, but likely is a better architecture to use a function if it exists.
3.  MaxCoinSupplyAtBlock is a calculation.  I was hoping to find an RPC call to provide a theoretical coinbase projection for an arbitrary future block, but it does not seem to exist.  I created a 4th order poly to fit to the expected emissions.  The calculation is here:  https://docs.google.com/spreadsheets/d/1uzIBF9q__JYx4EWeAJAdviB4B-Y8IVNEUGmSww3HRX8/edit?usp=sharing
4.  Should the apr simulation be executed on a separate thread much like we have done with `go exp.updateDevFundBalance()` to help improve rendering time?  

**Brief Explanation of Solution:**
1.  We have a function to forecast reward at an arbitrary block number
2.  We have a function to forecast max coin supply at an arbitrary block number
3.  We have a function to forecast theoretical ticket price at an arbitrary block number.  Assuming stake pool percent of total coin supply stays constant, we can project forward what the ticket price will be as the coin supply increases.
4.  We normalize our ticket prices and coin supply to match existing actuals to further refine our projections.
5.  We then start a loop where we simulate a purchase at the current block, a reward after the "average" reward time in blocks, and coinbase maturity to get the funds back in our wallet.  We repeat this cycle enough times to get through a full year.
6.  Once we are through at least 365 days we then reduce the ROI down to exactly 365 days to get a APR for a year rather than for the simulated number of days.

**Simulation Options:**
1.  You can select to use integer ticket numbers or float ticket numbers.  This was done to make the ROI independent of actual tickets purchased, and assuming 100% of funds are in use.  It is our vision that in the future a separate "detailed APR" calculation page can be implemented where the user can enter their actual DCR and simulate actual possible ticket purchases.
2.  The simulation outputs a simple text table that can be leveraged into a future implementation where the details of the simulation are presented to the user.  Rather than remove it from the prototyping I did, I left it so its easier to add back later.  You can see it in action here:  https://play.golang.org/p/y6exAOH9yQD

### DIFF
--- a/README.md
+++ b/README.md
@@ -282,6 +282,7 @@ parsing more efficient for the client.
 | Other | Path | Type |
 | --- | --- | --- |
 | Status | `/status` | `types.Status` |
+| Coin Supply | `/supply` | `types.CoinSupply` |
 | Endpoint list (always indented) | `/list` | `[]string` |
 | Directory | `/directory` | `string` |
 
--- a/README.md
+++ b/README.md
@@ -95,6 +95,15 @@ necessary.
 
 ## Getting Started
 
+### Configure PostgreSQL (IMPORTANT)
+
+If you intend to run dcrdata in "full" mode (i.e. with the `--pg` switch), which
+uses a PostgreSQL database backend, it is crucial that you configure your
+PostgreSQL server for your hardware and the dcrdata workload.
+
+Read [postgresql-tuning.conf](./db/dcrpg/postgresql-tuning.conf) carefully for
+details on how to make the necessary changes to your system.
+
 ### Create configuration file
 
 Begin with the sample configuration file:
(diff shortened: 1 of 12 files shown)


chappjc:
Please rebase, resolve conflicts, and let me know when ready for review (remove WIP).

chappjc:
Looks like the rebase didn't go quite right.  I think you did a merge instead.

But to clean this up. please do an interactive rebase an squash all the commits that aren't part of this PR.  With your apr_update branch checked out:  `git rebase -i master`.   Before you to git rebase master, make sure you pull latest master.

Issue #373, opened 2018-03-01
Front page dev fund balance goes to 0 on websocket refresh


raedah:

chappjc:
Indeed, although I described it much less succinctly.

https://github.com/decred/dcrdata/issues/348

mikefullman:
Just dropping this info here because I need to come back to this later.

The data from the websocket update has dev_fund = 0

> Received newblock message {"block":{"height":235487,"size":5980,"valid":false,"votes":5,"tx":3,"tickets":8,"revocations":0,"time":1525362808,"formatted_time":"2018-05-03 15:53:28","formatted_bytes":"6.0 kB"},"extra":{"coin_supply":766403701994539,"sdiff":87.58268129,"window_idx":48,"reward_idx":2015,"difficulty":325473380.9549447,**"dev_fund":0,**"dev_address":"Dcur2mcGjmENx4DhNqDctW5wJCVyT3Qeqkx","roi":1.4642569753644041,"roi_period":"28.44 days","subsidy":{"total":2137392543,"pow":1282435529,"pos":641217760,"dev":213739254},"params":{"window_size":144,"reward_window_size":6144,"target_pool_size":0,"target_block_time":300000000000},"pool_info":{"size":40860,"value":3543347.75147725,"valavg":86.71923033473446,"percent":46.23343731581425,"target":40960,"percent_target":99.755859375}}}

It looks like we have a race condition since the call the "update dev fund" is asynchronous
https://github.com/decred/dcrdata/blob/d3288e7ea631f3ad9fdadaec9005530f8083c435/explorer/explorer.go#L280-L290

A simple solution may be to add DevFund to the ExtraInfo initialization so it maintains whatever the last value was in there until the async call updates.  dev fund will be one block behind, but I think that is preferable to holding up the page rendering for an async call.

```go
exp.ExtraInfo = &HomeInfo{
		CoinSupply:        blockData.ExtraInfo.CoinSupply,
		StakeDiff:         blockData.CurrentStakeDiff.CurrentStakeDifficulty,
		IdxBlockInWindow:  blockData.IdxBlockInWindow,
		IdxInRewardWindow: int(newBlockData.Height % exp.ChainParams.SubsidyReductionInterval),
		DevAddress:        exp.ExtraInfo.DevAddress,
->		DevFund:           exp.ExtraInfo.DevFund,
		Difficulty:        blockData.Header.Difficulty,```

mikefullman:
[...]

PR #461, opened 2018-05-04
dev funds zero

papacarp:
This should address #373 .  Per the dialog there it looks like we have a race condition causing this problem.  There are two potential solutions as discussed in the bug.

The problem is difficult to replicate on my system - likely because my server architecture behaves differently than the server.  However, I was able to quite easily replicate the issue locally by inserting a small delay in the async call as shown here:

```go
func (exp *explorerUI) updateDevFundBalance() {
	time.Sleep(10000) //sleep for 10,000 ns
        // yield processor to other goroutines
	runtime.Gosched()
	exp.NewBlockDataMtx.Lock()
	defer exp.NewBlockDataMtx.Unlock()

	devBalance, err := exp.explorerSource.DevBalance()
	if err == nil && devBalance != nil {
		exp.ExtraInfo.DevFund = devBalance.TotalUnspent
	} else {
		log.Errorf("explorerUI.updateDevFundBalance failed: %v", err)
	}
}
```

The issue happens frequently (although not 100% of the time on beta.dcrdata.org) where the performance of the server likely exacerbates the issue.

### DIFF
--- a/explorer/explorer.go
+++ b/explorer/explorer.go
@@ -172,7 +172,7 @@ func New(dataSource explorerDataSourceLite, primaryDataSource explorerDataSource
 		log.Warnf("explorer.New: %v", err)
 	}
 	log.Debugf("Organization address: %s", devSubsidyAddress)
-	
+
 	// Set default static values for ExtraInfo
 	exp.ExtraInfo = &HomeInfo{
 		DevAddress: devSubsidyAddress,
@@ -254,7 +254,7 @@ func (exp *explorerUI) Store(blockData *blockdata.BlockData, _ *wire.MsgBlock) e
 		PosSubPerVote := dcrutil.Amount(blockData.ExtraInfo.NextBlockSubsidy.PoS).ToCoin() / float64(exp.ChainParams.TicketsPerBlock)
 		return percentage(PosSubPerVote, blockData.CurrentStakeDiff.CurrentStakeDifficulty)
 	}()
-	 	
+
 	// If there are ticketpoolsize*TicketsPerBlock total tickets and
 	// TicketsPerBlock are drawn every block, and assuming random selection
 	// of tickets, then any one ticket will, on average, be selected to vote
--- a/explorer/explorer.go
+++ b/explorer/explorer.go
@@ -172,8 +172,18 @@ func New(dataSource explorerDataSourceLite, primaryDataSource explorerDataSource
 		log.Warnf("explorer.New: %v", err)
 	}
 	log.Debugf("Organization address: %s", devSubsidyAddress)
+	
+	// Set default static values for ExtraInfo
 	exp.ExtraInfo = &HomeInfo{
 		DevAddress: devSubsidyAddress,
+		Params: ChainParams{
+			WindowSize:       exp.ChainParams.StakeDiffWindowSize,
+			RewardWindowSize: exp.ChainParams.SubsidyReductionInterval,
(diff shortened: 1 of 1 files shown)


chappjc:
OK, but I would like Issue #373 to stay open since the race condition still exists (access to `ExtraInfo.DevFund` is not thread-safe) and it may be the value from the previous block.

My suggestion for a permanent solution that does not block the web page is a websocket event just for the dev fund balance update.  The current PR's solution is better than showing 0.0000, but we need to address it eventually.

explorer/explorer.go
```diff
@@ -226,6 +226,7 @@ func (exp *explorerUI) Store(blockData *blockdata.BlockData, _ *wire.MsgBlock) e
 		IdxBlockInWindow:  blockData.IdxBlockInWindow,
 		IdxInRewardWindow: int(newBlockData.Height % exp.ChainParams.SubsidyReductionInterval),
 		DevAddress:        exp.ExtraInfo.DevAddress,
+		DevFund:           exp.ExtraInfo.DevFund,
```
chappjc: But how is `exp.ExtraInfo.DevFund` initially set?

papacarp: line 175 initializes the value to 0 when explorerUI is initialized.  Since startup values of 0 until the first block should not be a problem I figured this was ok.

https://github.com/decred/dcrdata/blob/1bb5fb11b2efa701639666e41b7678e315c3b656/explorer/explorer.go#L170-L180

chappjc: That's what I'm going after.  How about updating exp.ExtraInfo instead of allocating a new one?  That way the update dev fund go routine will not lose it's changes.

chappjc: I haven't tried this, but it strikes me that reallocating it might be party of the issue...

papacarp: @chappjc good idea.  I can try that!

Also, regarding thread safe I've been trying to understand thread safe in go.  In the current design the mutex locks it up before updating and unlocks it after dev fund is kicked off which also locks and unlocks. - therefore I *think* its thread safe.    Having said that, Its not clear to me how those two routines would  not occasionally overlap locks or try to unlock something that was already unlocked.

chappjc: you are right, that mutex serves the purpose.

papacarp: updating ExtraInfo rather than a new instance seems to work great.  I can update this PR so you can test as well.

Also, what would you think about changing this:

```go
	if !exp.liteMode {
		go exp.updateDevFundBalance()
	}

	exp.NewBlockDataMtx.Unlock()
```

to this:

```go
        exp.NewBlockDataMtx.Unlock()

	if !exp.liteMode {
		go exp.updateDevFundBalance()
	}
```

Seems cleaner and insures the lock is released before we try to relock for the dev fund update.

chappjc: Great, please update with the update so I can have a look.

The suggested lock change is probably fine, but note that since `updateDevFundBalance` runs as a goroutine there's no deadlock.  Rather, if the machine has enough cores, `updateDevFundBalance` will wait on the lock, and run as soon as `Store` unlocks.  If instead we unlock first, there's likely going to be a longer window where the lock is free and when `updateDevFundBalance` gets it.  Anyway, if this were really a requirement, I would have made a comment and explained why, but I didn't, so it's no big deal.  I just wanted to let you know the thinking behind the current ordering.

PR #463, opened 2018-05-10
Add charts

dmigwi:
This PR added 8 Charts implementations:

Namely:
1. Ticket price chart   - plot `price(sbits) vs block time` - picks 144th block price
2. Ticket pool size  - plot `pool_size(Number of Tickets) vs block time`
3. Ticket pool value  - plot `pool value(pool_size *bits) vs block time`
4. Average block size - plots `block size vs block time`

screen shot 2018-05-10 at 09 28 10

5. Blockchain size chart - plots  `Sumation of block size vs Block time`
6. Txns per day chart - plots `Transactions count vs Day`
7. Txns per block chart - plots `Transactions count vs Block time`
8. PoW difficulty chart - plots `difficulty vs block time`

screen shot 2018-05-10 at 09 28 26


raedah:
Look like Ticket Pool Value is not correct. Appears to be number of tickets in the pool at the current ticket price, but the tickets in the pool are bought in different price windows so will need to fetch the ticket price on all the tickets in the pool individually. Note, there is also the question of immature tickets, which could be included on the same chart but distinguished.

dmigwi:
That is attributed to the fact that pool value data is fetched from the blocks table that contains all the ticket price whether used or not. After our conversation with @chappjc I realized that the pool value data may be accurately captured when fetched from tickets table. 

The ticket price contains all the ticket prices whether used or not (data fetched from blocks table).

dmigwi:
Updated Ticket Pool Value chart (screenshot)

screen shot 2018-05-17 at 04 16 52

chappjc:
Really cool.  Funny story, the pool size and value at each block was the main reason I started coding on dcrdata, back before it was an explorer.  It's one bit of data not easily accessible anywhere else AFAIK.  😀

chappjc:
BTW, depending on backend load impact of repeated SQLite queries for the same entire range, we may want to add in caching.  That should be saved for another PR though, unless performance is seriously impacted by this.

dmigwi:
You got a point @chappjc `/charts` page display is almost freezing my machine once in a while. I suggest that the caching part be done for all charts at regular intervals especially when the system is idle (lazy loading) to minimize the possibility of things freezing when one in running the system on a resource constrained machine.

dmigwi:
I will first add all other pending charts and then focus on optimizing the performance part. I have about  5 more to go.

raedah:
Appending new data to the tables can be done when a new block is received.

dmigwi:
Here is the screenshot of the latest chart. (Coin Supply Chart)

screen shot 2018-05-22 at 06 41 31

The `/charts` page is taking too long to load since we've got too many charts on one page (page loading time is almost 20 seconds). I think that if we split the charts the situation might improve.

dmigwi:
Current version of `/charts` page display.

screen shot 2018-05-30 at 16 35 06

dmigwi:
Ready for review.

chappjc:
There are no conflicts but would you mind rebasing?  I'm going to put this up on a test site so we can evaluation.

dmigwi:
>Do you think there is a possibility for an automatic online upgrade of the db version?

The vins table has changed i.e. some new columns exists. @chappjc How do you suggest I make the migration smoother?

chappjc:
You can add columns to the table.  Then you just need a way of filling out the columns for existing rows when startup detects a version change that requires an upgrade.

dmigwi:
@chappjc I have implemented and tested a way to automatically upgrade the vins table once this PR is merged and run for the first time.

screen shot 2018-06-06 at 19 02 42

chappjc:
Great, thanks.  It's important we keep a smooth upgrade path when possible

db/dcrpg/tableupgrades.go
```diff
@@ -0,0 +1,172 @@
...
+	} else {
+		return false, nil
+	}
+
+	switch {
+	case upgradeInfo[0].UpgradeType != "upgrade":
+		return false, nil
+
+	// When the required table version is 2.x.0 where x is greater than or equal to 3
+	case version.major == 2 && version.minor > 2 && version.patch == 0:
```
chappjc: This needs to work for upgrades over multiple versions.   So if there is a 2.4.0, and the software is just updated from 2.0.0, it will need to run each upgrade step in sequence.  I *think* this approach can be adapted to work that way, but give it some thought too please.

db/dcrpg/tableupgrades.go
```diff
@@ -0,0 +1,172 @@
...
+
+	log.Infof(" %v rows in the vins table (Coin Supply Upgrade) were successfully upgraded.",
+		rowsUpdated)
+
+	return nil
+}
+
+// addNewColumns checks if the new columns all ready exists
+// and add them if they are missing.
+func addNewColumns(db *sql.DB) (int, error) {
```
chappjc: Will need a more version A -> version B specific name.  Consider what I said about chaining upgrades.

dmigwi:
@chappjc I have implemented the changes you requested.

chappjc:
Thanks Migwi.  I'll try the upgrade again.  I ran it by accident with an empty data directory, but some blocks in postgresql, and there were some odd errors

dmigwi:
Ok. Noted

chappjc:
@gozart1 Would you mind having a look through the html/js edits here?  There a bunch of new scripts, which are now embedded in the html.  I suspect we'll want to add them as a static resource for browser caching, etc.  Any other style and/or best practices comments would be very helpful too!

gozart1:
Running into a snag in the db upgrade phase, it's stuck on this last line:
```
2018-06-15 12:01:23.902 [INF] SKDB: Loading ticket pool DB. This may take a minute...
2018-06-15 12:02:08.640 [DBG] SKDB: Opened existing stake db.
2018-06-15 12:02:08.640 [INF] SKDB: Advancing ticket pool DB to tip via diffs...
2018-06-15 12:02:09.134 [INF] SKDB: Pre-populating live ticket cache...
2018-06-15 12:02:24.353 [INF] DATD: SQLite DB successfully opened: dcrdata.sqlt.db
2018-06-15 12:02:25.191 [INF] PSQL: Pre-loading unspent ticket info for InsertVote optimization.
2018-06-15 12:02:27.812 [INF] PSQL: Storing data for 43637 unspent tickets in cache.
2018-06-15 12:02:27.907 [DBG] PSQL: Table addresses: v2.0.0
2018-06-15 12:02:27.908 [DBG] PSQL: Table transactions: v2.0.0
2018-06-15 12:02:27.908 [DBG] PSQL: Table misses: v2.0.0
2018-06-15 12:02:27.908 [DBG] PSQL: Table blocks: v2.0.0
2018-06-15 12:02:27.908 [DBG] PSQL: Table vins: v2.0.0
2018-06-15 12:02:27.908 [DBG] PSQL: Table votes: v2.0.0
2018-06-15 12:02:27.908 [DBG] PSQL: Table vouts: v2.0.0
2018-06-15 12:02:27.908 [DBG] PSQL: Table block_chain: v2.0.0
2018-06-15 12:02:27.908 [DBG] PSQL: Table tickets: v2.0.0
2018-06-15 12:02:28.075 [INF] PSQL: Found the best block at height: 248134
2018-06-15 12:02:28.092 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 0 to 4999 
32 execute past end of script 801679e98561ada96caec2949a5d41c4cab3851eb740d951c10ecbcf265c1fd9
```

gozart1:
hmm, I left it running and it is progressing now
```
2018-06-15 12:05:30.395 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 5000 to 9999 
2018-06-15 12:08:34.161 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 10000 to 14999 
2018-06-15 12:11:31.193 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 15000 to 19999 
2018-06-15 12:14:19.643 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 20000 to 24999 
```
so, lets see

chappjc:
I think we need to look into optimizing the coin supply upgrade.  This will be a bad surprise when people update.

views/charts.tmpl
```diff
@@ -0,0 +1,254 @@
...
+                digitsAfterDecimal: 8,
+                showRangeSelector: true,
+                legend: "follow",
+                ylabel: "Data loading...",
+                xlabel: "Data loading...",
+                title: "Charts Loading Please Wait ...",
+                labelsSeparateLines: true
+            });
+
+        var data = {{.Data}};
```
gozart1: Sending the chart data inline with in the HTML incurs a large performance penalty because it forces the browser to store it twice. Once as text inside the DOM, and again in memory once the script is parsed. On my machine I'm seeing ~2 seconds just to parse the html. During this time the browser will essentially be frozen, or in UX terminology, the end user experiences "jank".

Chrome dev tools has a great Performance tab for doing this kind of profiling:
 
screen shot 2018-06-15 at 4 24 38 pm

As first step to "dejank" this, the data instead should come from an api endpoint and fetched via xhr ( use  [$.ajax()](http://api.jquery.com/jquery.ajax/) ). That way the page can load smoothly and a loading animation can be shown to the end user while the data is requested. 

Furthermore, each dataset should have it's own endpoint, as the current payload is 25mb over the wire. And parsing that network request will again create jank. 

Another trick for transporting large datasets is to use csv format instead of json, since that way we're not repeating the column names for each row of data. So I given the nature of these datasets, we might as well go that route.

papacarp: Good feedback @gozart1.  Also you might as well make the endpoint have a param to only give you the datapoints that are useful and visible, like daily or weekly or something.  that will squish the data even more.  reload the data for zoom functions.  or just lazy load the detailed data once you have the broad framework in place if you have dynamic zooms.

gozart1: Ok had a chance to look into this PR a bit more now and I see that dygraphs expects an unparsed csv string as input. So we should definitely just return csv from the backend so it can be passed directly to the chart instead of having to transcode first.

Interactive charts backed by large datasets up to a million or so records **are** viable, but only if things are streamlined. A good example is on the homepage of [crossfilter](http://square.github.io/crossfilter/) which is a d3 based charting lib. It's quite responsive even as you interact with it to apply multi-dimensional filters on a 5 column dataset with 230k records.

dmigwi: @gozart1 The current implementation loads all the data needed from the backend to the frontend, the data is pulled once and reused severally thereby reducing the effects of network latency if it occurs while one is trying to load one of the many expected graphs. A work around on this would be making ajax requests for each graph while its being plotted but this will increase the data being sent from the backend to the frontend especially if once tries to view several graphs in short period of time.

dmigwi:
Hey guys, I have made an update that should make the slowest chart (Coin Supply chart) in loading pretty fast. Kindly please pull the latest changes and let me know if you notice the difference.

gozart1:
I'm getting 
```
2018-06-16 23:25:20.338 [INF] EXPR: Pre-populating the charts data. This may take a minute...
2018-06-16 23:25:42.328 [ERR] EXPR: Invalid PG data found: RetrieveCoinSupply: sql: Scan error on column index 0: converting driver.Value type <nil> ("<nil>") to a int64: invalid syntax
2018-06-16 23:25:54.186 [INF] EXPR: Done Pre-populating the charts data
```
I think the vins indexing may not have gone smoothly, probably due to my computer entering sleep mode during the upgrade process. I get a lot of data errors of undefined attributes when I try switching between charts, and the one chart that works, the values look off:
screen shot 2018-06-16 at 11 46 15 pm

Troubleshooting this would also be more straightforward if the datasets separate endpoints with good names. But for now I've added some console logs to share the current state I'm in.
screen shot 2018-06-16 at 11 54 04 pm
screen shot 2018-06-16 at 11 53 44 pm

Any recommendations on how to redo the indexing? Do I just need to drop the vins table and then start up dcrdata again?

dmigwi:
Just drop "is_valid", "block_time", "value_in" columns in vins table. Also change the value 3 to a higher value https://github.com/dmigwi/dcrdata/blob/1fafe75ea125d15818309f9ce5524b62fad1387e/db/dcrpg/tables.go#L42-L50 (remember to reset the value back to 3)
The upgrade should pickup automatically.

dmigwi:
Before deleting the columns try first hard reloading the charts page.

gozart1:
After dropping the columns and re-upgrading, I have the same state as the screenshots posted earlier. I'm curious if anyone else has tried running this branch and how it works for them, I'm trying to determine if what I'm experiencing is related to my local data ( which I have not noticed any issues with otherwise ) or something in this PR itself.

chappjc:
I reran the upgrade last night with you profiling.  I'll review the logs before I delve into optimization.

chappjc:
Here's the upgrade log for me (no errors, but about 45 minutes to upgrade the vins table):

```2018-06-16 09:23:20.794 [INF] DATD: Log folder:  /home/jon/.dcrdata/logs/mainnet
2018-06-16 09:23:20.794 [INF] DATD: Config file: /home/jon/.dcrdata/dcrdata.conf
2018-06-16 09:23:20.794 [INF] DATD: dcrdata version 2.0.0
2018-06-16 09:23:20.794 [INF] DATD: Running in full-functionality mode with PostgreSQL backend enabled.
2018-06-16 09:23:20.801 [DBG] RPCC: Attempting to connect to dcrd RPC localhost:9109 as user jdcrd (no TLS)
2018-06-16 09:23:20.802 [INF] RPCC: Established connection to RPC server localhost:9109
2018-06-16 09:23:20.803 [INF] DATD: Connected to dcrd (JSON-RPC API v3.3.0) on MainNet
2018-06-16 09:23:20.805 [INF] SKDB: Loading ticket pool DB. This may take a minute...
2018-06-16 09:23:44.403 [DBG] SKDB: Opened existing stake db.
2018-06-16 09:23:44.403 [INF] SKDB: Advancing ticket pool DB to tip via diffs...
2018-06-16 09:23:44.629 [INF] SKDB: Pre-populating live ticket cache and computing pool value...
2018-06-16 09:23:48.981 [INF] DATD: SQLite DB successfully opened: dcrdata.sqlt.db
2018-06-16 09:23:49.750 [INF] PSQL: Pre-loading unspent ticket info for InsertVote optimization.
2018-06-16 09:23:50.367 [INF] PSQL: Storing data for 43508 unspent tickets in cache.
2018-06-16 09:23:50.383 [DBG] PSQL: Table votes: v2.0.0
2018-06-16 09:23:50.383 [DBG] PSQL: Table transactions: v2.0.0
2018-06-16 09:23:50.383 [DBG] PSQL: Table vouts: v2.0.0
2018-06-16 09:23:50.383 [DBG] PSQL: Table addresses: v2.0.0
2018-06-16 09:23:50.383 [DBG] PSQL: Table vins: v2.0.0
2018-06-16 09:23:50.383 [DBG] PSQL: Table tickets: v2.0.0
2018-06-16 09:23:50.383 [DBG] PSQL: Table misses: v2.0.0
2018-06-16 09:23:50.383 [DBG] PSQL: Table blocks: v2.0.0
2018-06-16 09:23:50.383 [DBG] PSQL: Table block_chain: v2.0.0
2018-06-16 09:23:50.516 [INF] PSQL: Found the best block at height: 248177
2018-06-16 09:23:50.517 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 0 to 4999 
2018-06-16 09:24:26.647 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 5000 to 9999 
2018-06-16 09:25:03.770 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 10000 to 14999 
2018-06-16 09:25:41.800 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 15000 to 19999 
2018-06-16 09:26:20.327 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 20000 to 24999 
2018-06-16 09:27:04.253 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 25000 to 29999 
2018-06-16 09:27:50.988 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 30000 to 34999 
2018-06-16 09:28:34.365 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 35000 to 39999 
2018-06-16 09:29:19.834 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 40000 to 44999 
2018-06-16 09:30:05.761 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 45000 to 49999 
2018-06-16 09:30:49.000 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 50000 to 54999 
2018-06-16 09:31:32.144 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 55000 to 59999 
2018-06-16 09:32:14.172 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 60000 to 64999 
2018-06-16 09:32:58.148 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 65000 to 69999 
2018-06-16 09:33:47.404 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 70000 to 74999 
2018-06-16 09:34:36.340 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 75000 to 79999 
2018-06-16 09:35:19.233 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 80000 to 84999 
2018-06-16 09:36:13.559 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 85000 to 89999 
2018-06-16 09:37:11.844 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 90000 to 94999 
2018-06-16 09:38:09.659 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 95000 to 99999 
2018-06-16 09:39:06.746 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 100000 to 104999 
2018-06-16 09:40:02.553 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 105000 to 109999 
2018-06-16 09:40:54.528 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 110000 to 114999 
2018-06-16 09:41:42.724 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 115000 to 119999 
2018-06-16 09:42:42.187 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 120000 to 124999 
2018-06-16 09:43:43.840 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 125000 to 129999 
2018-06-16 09:44:50.663 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 130000 to 134999 
2018-06-16 09:46:02.290 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 135000 to 139999 
2018-06-16 09:47:11.069 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 140000 to 144999 
2018-06-16 09:48:21.203 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 145000 to 149999 
2018-06-16 09:49:29.506 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 150000 to 154999 
2018-06-16 09:50:31.931 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 155000 to 159999 
2018-06-16 09:51:46.325 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 160000 to 164999 
2018-06-16 09:52:50.998 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 165000 to 169999 
2018-06-16 09:53:47.724 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 170000 to 174999 
2018-06-16 09:54:46.232 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 175000 to 179999 
2018-06-16 09:55:44.779 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 180000 to 184999 
2018-06-16 09:56:43.789 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 185000 to 189999 
2018-06-16 09:57:52.459 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 190000 to 194999 
2018-06-16 09:58:58.277 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 195000 to 199999 
2018-06-16 10:00:00.773 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 200000 to 204999 
2018-06-16 10:01:09.490 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 205000 to 209999 
2018-06-16 10:02:07.853 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 210000 to 214999 
2018-06-16 10:03:07.417 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 215000 to 219999 
2018-06-16 10:04:08.049 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 220000 to 224999 
2018-06-16 10:04:58.330 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 225000 to 229999 
2018-06-16 10:05:49.878 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 230000 to 234999 
2018-06-16 10:06:44.796 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 235000 to 239999 
2018-06-16 10:07:43.457 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 240000 to 244999 
2018-06-16 10:08:40.273 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 245000 to 248177 
2018-06-16 10:09:16.118 [INF] PSQL:  11109432 rows in the vins table (Coin Supply Upgrade) were successfully upgraded.
2018-06-16 10:09:16.119 [INF] PSQL: Modified the blocks table version to 2.3.0
2018-06-16 10:09:16.120 [INF] PSQL: Modified the transactions table version to 2.3.0
2018-06-16 10:09:16.121 [INF] PSQL: Modified the vouts table version to 2.3.0
2018-06-16 10:09:16.121 [INF] PSQL: Modified the block_chain table version to 2.3.0
2018-06-16 10:09:16.121 [INF] PSQL: Modified the addresses table version to 2.3.0
2018-06-16 10:09:16.121 [INF] PSQL: Modified the vins table version to 2.3.0
2018-06-16 10:09:16.121 [INF] PSQL: Modified the tickets table version to 2.3.0
2018-06-16 10:09:16.122 [INF] PSQL: Modified the votes table version to 2.3.0
2018-06-16 10:09:16.122 [INF] PSQL: Modified the misses table version to 2.3.0
2018-06-16 10:09:16.497 [DBG] DATD: Setting ticket pool cache capacity to 235 blocks
2018-06-16 10:09:16.497 [DBG] SQLT: Pre-loading pool info for 3 blocks ([248175, 248177]) into cache.
2018-06-16 10:09:16.497 [DBG] EXPR: Organization address: Dcur2mcGjmENx4DhNqDctW5wJCVyT3Qeqkx
2018-06-16 10:09:16.500 [INF] EXPR: Mean Voting Blocks calculated: 7860
2018-06-16 10:09:16.519 [INF] EXPR: Pre-populating the charts data. This may take a minute...
2018-06-16 10:09:30.820 [INF] EXPR: Done Pre-populating the charts data
2018-06-16 10:09:30.821 [INF] EXPR: Starting WebsocketHub run loop.
2018-06-16 10:09:30.821 [DBG] SQLT: Setting block gate height to 248178
2018-06-16 10:09:30.821 [INF] SQLT: Current best block (chain server):     248392
2018-06-16 10:09:30.821 [INF] SQLT: Current best block (sqlite block DB):  248177
2018-06-16 10:09:30.821 [INF] SQLT: Current best block (stakedb):          248177
2018-06-16 10:09:30.917 [INF] PSQL: Processing blocks 248178 to 248250...
2018-06-16 10:09:31.402 [INF] SQLT: Scanning blocks 248178 to 248250 (40877 live)...
2018-06-16 10:09:38.557 [INF] PSQL: Processing blocks 248251 to 248392...
2018-06-16 10:09:38.584 [INF] SQLT: Scanning blocks 248251 to 248392 (40842 live)...
2018-06-16 10:09:48.553 [INF] SQLT: Rescan finished successfully at height 248392.
2018-06-16 10:09:48.553 [INF] DATD: SQLite sync ended at height 248392
2018-06-16 10:09:48.591 [INF] PSQL: Avg. speed: 205 tx/s, 838 vout/s
2018-06-16 10:09:48.591 [INF] PSQL: Sync finished at height 248392. Delta: 215 blocks, 3500 transactions, 8643 ins, 14246 outs
2018-06-16 10:09:48.591 [INF] DATD: PostgreSQL sync ended at height 248392
2018-06-16 10:09:48.591 [INF] DATD: All ready, at height 248392.
2018-06-16 10:09:48.600 [DBG] BLKD: Collector.Collect() completed in 8.516345ms
2018-06-16 10:09:48.608 [DBG] EXPR: Got new block 248392 for the explorer.
2018-06-16 10:09:48.608 [INF] EXPR: Signaling new block to 0 websocket clients.
2018-06-16 10:09:48.615 [DBG] MEMP: mempoolDataCollector.Collect() completed in 6.828015ms
2018-06-16 10:09:50.618 [INF] EXPR: Now serving explorer on http://127.0.0.1:7777/
2018-06-16 10:09:50.618 [INF] JAPI: Now serving API on http://127.0.0.1:7777/```

Other than bumping the major version of dcrdata to 3, I don't know what we can do to prepare users of master.  This is what I am thinking we should do actually.  Insight and other minor things for 2.1, then `master` will be working toward 3.0

**Comment by chappjc** [comment]
BTW, the profiling data provided no insight.  It was just waiting for SQL

**Comment by gozart1** [comment]
Ok so I went ahead and spun up a remote server to reindex dcrdata from scratch and that got this PR running with all the charts drawing correctly now, except for "Total Fee per Block" and "Duration Between Blocks".

Those two don't draw any chart and show a js error by the dygraph parseCSV_ function:
`CSV is out of order; order it correctly to speed loading.`

I noticed that we can just pass the data into dygraph in its native format ( javascript array ). That would be much more efficient than passing it a CSV which it has to parse before it can make use of the data. For example, in this profile where I switch to the "Total Coin Supply" chart, dygraphs spent 985ms just parsing the csv. 

screen shot 2018-06-23 at 11 48 35 am

Note the dygpraph docs call out that this is not efficient and should be avoided http://dygraphs.com/data.html#array

> If you'll be constructing your data set from a server-side program (or from JavaScript) then you're better off producing an array than CSV data. This saves the cost of parsing the CSV data and also avoids common parser errors.

Right now the data transcodes 3 times
1. Parsing the html to convert the inlined data into native javascript array 
2. Converting native javascript array into csv
3. Converting csv back into native javascript array.

If we decide it's ok to closely couple the charts api to dygraphs' format, then we could whittle this down to a single step:
1. Parsing the JSON body of api response into native javascript array

But if we just want to stick with the existing api response, then let's at least skip converting to csv and just convert the data directly into the dygraph native array format.

With a bit streamlining and modularizing the data handling, we can get these charts loading reasonable fast on a median powered mobile device. We've made a conscious effort to keep the explorer usable for smart phones and low bandwidth scenarios and think we should continue to aim for that even for data intensive visualizations.

**Comment by dmigwi** [comment]
@gozart1 Great work on hosting a live instance of dcrdata. 

I would like you to confirm if you got the latest code from this PR. The last push made to this reduced the occurrences of this errors.
>Those two don't draw any chart and show a js error by the dygraph parseCSV_ function:
>`CSV is out of order; order it correctly to speed loading.`

On the choice of dygraph data input i.e. between CSV and Javascript Array. I see your point and its a valid one. Let me do some extra tests on some graphs that I am currently working on, If they turn out okay, I will update you and also make the changes to this PR.

**Comment by dmigwi** [comment]
@gozart I have replaced the csv data input with a javascript array. You will find that I am still looping through the  fetched dataset because I have to set that date data to `new Date(1468989000)` format as that is what is acceptable when the data source is a javascript array.

That date format cannot be set on the server side as it will almost double the data size sent from server to frontend side.

File: views/charts.tmpl [hunk 4]
```diff
@@ -90,61 +90,59 @@
...
-                if (isNaN(item.value)) {
-                    return;
-                }
-                time_btw_blocks = time_btw_blocks +  item.value + "," + (isNaN(item.valuef) ? 0 : item.valuef) + "\n";
+            block_size = [];
+            blockchain_size = [];
+            tx_per_block = [];
+            time_btw_blocks = [];
+            blockSize.forEach(function(n){
+                block_size.push([new Date(n.time*1000), n.size]);
```
gozart1: we can cache the result to avoid parsing same date more than once
eg `var date = new Date(n.time*1000)` then
```
block_size.push([date, n.size]);
blockchain_size.push([date, size]);
tx_per_block.push([date, n.count]);
```

views/charts.tmpl
```diff
@@ -0,0 +1,252 @@
...
+            func()
+            resolve('resolved');
+            }, 2000);
+        });
+        }
+
+        var plotGraph = async function (value){
+            switch(value){
+                case "1": // price graph
+                    if (tickets_price === undefined) { await resolvePromise(ticketsFunc); }
```
gozart1: What is the purpose of using asynchronous primitives here? Or added any sort of delay at all?
Why not just execute the transcoding step immediately like so?
```
if (tickets_price === undefined) { ticketsFunc() }
```

views/charts.tmpl
```diff
@@ -0,0 +1,252 @@
...
+                break;
+
+                case "10": // block fee graph
+                    if (fee_per_block === undefined) { await resolvePromise(blockFeeFunc);}
+                    drawGraph(fee_per_block, ["Block Height", "Total Fee"], false, 'Total Fee (DCR)', 'Block Height', 
+                    'Total Fee Per Block Chart', true, false)
+                break;
+
+                case "11": // Duration between blocks graph
+                    if (time_btw_blocks === undefined) { await resolvePromise(sizefunc);}
```
gozart1: Please use camelCase for all client side naming. It's simply a convention amongst all major javascript codebases. So `time_btw_blocks` should be `timeBtwBlocks`

views/charts.tmpl
```diff
@@ -0,0 +1,252 @@
...
+        function resolvePromise(func) {
+        return new Promise(resolve => {
+            setTimeout(() => {
+            func()
+            resolve('resolved');
+            }, 2000);
+        });
+        }
+
+        var plotGraph = async function (value){
```
gozart1: seems like this should just be synchronous

gozart1:
Cool, all the charts are rendering for me now!
I played around with this a bit locally and after i removed all the async logic and the artificial 2000ms delay in [resolvePromise](https://github.com/decred/dcrdata/pull/463/files#diff-1913d14c3d3efcab9e79e18096104427R163) the UI is a lot more usable.

Besides the comments I just made, here are the remaining big things to be addressed:
1. Each chart should have a url backed state ( ie make it possible to share a link to a given chart )
2. Data should be fetched as needed instead of all at once. 
3. It should be possible to View Source, without the browser tab crashing because of too much text to parse ( by getting fetching data via xhr instead of inlining it )

I think these 3 points are all related and it looks like you had started down the right path by making `plotGraph` async. Right now, that doesn't make sense to me because all the code in that function is currently synchronous. But it would make sense for it be async if that function did something more like this:
1. choose chart
2. await xhr response for chart data ( can be cached )
3. draw chart

views/charts.tmpl
dmigwi: Noted.

views/charts.tmpl
dmigwi: Noted.

dmigwi:
I am adding extra 4 charts here and your suggestion of making xhr requests to the backend for each graph sound scalable. My current implementation cannot support extra charts.

I will address all those concerns you've raised and get back you on it.

chappjc:
thanks for the frontend review @gozart1!

gozart1:
I'm looking forward to the new endpoints!

I've been doing a bit of research and have come to the conclusion that Dygraph is not efficient enough at rendering the larger datasets, and that should find something better. The best library I've found so far is [plotly.js](https://plot.ly/javascript/).

Also the Dygraph github has not had a commit in almost 2 years, vs 12 hours for plotly.

I've done some benchmarking here on a 1 million data point time series chart.

[Plotly Benchmark](https://1million-plotly.netlify.com) 1,625ms to draw chart
screen shot 2018-06-28 at 1 22 11 am

[Dygraph Benchmark](https://1million-dygraph.netlify.com/) 9,560ms to draw chart
screen shot 2018-06-28 at 1 22 45 am

Not only is plotly ~6x as fast, it also uses ~1/4 of the memory .
( I'm seeing 223MB javascript heap size vs 929MB javascript heap size once page comes to rest )

You can checkout the source code and javascript console on those benchmark pages to see how it performs.

We'll hit 1 million decred blocks sometime around the year 2025. It would be great if the charts we build today still work well at that point! This is decred after all, built to last!

I tested plotly at 5 million data points just to push it to the extreme and it still renders in under 10 seconds, also memory usage and overall browser performance is still good shape once the chart is drawn. On the other hand, Dygraph crashes my browser when attempting to render the benchmarking chart with more than 1.25+ million data points. 

Plotly is also nice because it utilizes [d3](https://d3js.org/), which is by far the most popular data visualization base library for the browser, so we'd be getting a very solid toolkit there for doing all sorts of fancy visualizations.

dmigwi:
Great analysis. I will make the change. Thanks @gozart1

dmigwi:
Just added two charts here and ajax requests for each chart.  The ajax requests make /charts page scalable and faster in the long run.

dmigwi:
PR ready for another round of review.

dmigwi:
@gozart1 and @chappjc I have made a couple of extra changes, Please do have a look at my changes when you get the time.

chappjc:
Will do, thanks!

db/dcrpg/internal/blockstmts.go
```diff
@@ -99,6 +99,15 @@ const (
 
 	SelectBlockChainRowIDByHash = `select block_db_id from block_chain where this_hash = $1;`
 
+	// SelectBlocksTicketsPrice selects the ticket price and difficulty for the 144th block
+	SelectBlocksTicketsPrice = `select sbits, time, difficulty from blocks where height % 144 = 0 order by time;`
+
+	SelectBlocksBlockSize = `Select time, size, numtx, height from blocks order by time;`
+
+	IndexBlocksTableOnHeight = `CREATE INDEX uix_height_index ON blocks(height);`
```
chappjc: Can we call this `uix_block_height` so that it's clear what table it's on from the name, and the `_index` is implied by the `uix` part (unique index)... well, it's not a UNIQUE, but that's fine.

chappjc: Also, please move all the statements on the `blocks` table up to the others.  This area of the file deals with the `block_chain` table.

db/dcrpg/internal/txstmts.go
```diff
@@ -66,11 +66,26 @@ const (
...
+	SelectTxsPerDay = `SELECT to_timestamp(time)::date as date, count(*) FROM transactions GROUP BY date ORDER BY date;`
+
 	SelectFullTxByHash = `SELECT id, block_hash, block_height, block_time, 
 		time, tx_type, version, tree, tx_hash, block_index, lock_time, expiry, 
 		size, spent, sent, fees, num_vin, vin_db_ids, num_vout, vout_db_ids 
 		FROM transactions WHERE tx_hash = $1;`
 
+	SelectTicketsOutputCountByAllBlocks = `SELECT block_height,
+		SUM(CASE WHEN num_vout = 3 THEN 1 ELSE 0 END) as solo,
+		SUM(CASE WHEN num_vout = 5 THEN 1 ELSE 0 END) as pooled
```
chappjc: These cases are not generally correct for identifying solo and pool tickets as either may be created with a large number of outputs, if they are created with the appropriate number of inputs.  3 and 5 are the most common number of outputs for solo and pool tickets, however.

The way to identify pooled tickets would be to identify addresses with P2SH prefixes (pool) that decode to M-of-N multi-sig scripts.

db/dcrpg/internal/vinoutstmts.go
```diff
@@ -58,7 +61,16 @@ const (
 	SelectFundingOutpointByVinID = `SELECT prev_tx_hash, prev_tx_index, prev_tx_tree FROM vins WHERE id=$1;`
 	SelectFundingTxByVinID       = `SELECT prev_tx_hash FROM vins WHERE id=$1;`
 	SelectSpendingTxByVinID      = `SELECT tx_hash, tx_index, tx_tree FROM vins WHERE id=$1;`
-	SelectAllVinInfoByID         = `SELECT * FROM vins WHERE id=$1;`
+	SelectAllVinInfoByID         = `SELECT id, tx_hash, tx_index, tx_tree, is_valid, block_time,
```
chappjc: why get `id` when it's in the `WHERE`?

db/dcrpg/internal/vinoutstmts.go
```diff
@@ -58,7 +61,16 @@ const (
...
+		prev_tx_hash, prev_tx_index, prev_tx_tree, value_in FROM vins WHERE id = $1;`
+	SetIsValidByTxHash = `UPDATE vins SET is_valid = $1 WHERE tx_hash = $2;`
```
chappjc: This will set for transactions with the same `tx_hash` in orphaned side chains.  The block hash and tx_tree need to be included for this to be operating on only the transaction intended by the caller.

db/dcrpg/internal/vinoutstmts.go
```diff
@@ -58,7 +61,16 @@ const (
...
+
+	SetVinsTableCoinSupplyUpgrade = `UPDATE vins SET is_valid = $1, block_time = $2, value_in = $3
+		WHERE tx_hash = $4 and tx_index = $5 and tx_tree = $6;`
```
chappjc: same as above.  This tx hash could be in an orphaned block.

tx_index just says which transaction number within the containing block it is.

db/dcrpg/internal/vinoutstmts.go
```diff
@@ -58,7 +61,16 @@ const (
...
+
+	SelectCoinSupply = `SELECT block_time, sum(value_in) FROM vins WHERE
```
chappjc: A comment explaining that this sums the generated coins for all stakebase and only not-invalidated coinbase transactions.

chappjc: Hmm, which reminds me, it may be complicated if we want to hang on to orphaned blocks in side chains.  For that we'll need a new column called `main_chain` (or similar) and this query will need to include a condition that `mainchain = TRUE`.  This is a note to myself.

db/dcrpg/pgblockchain.go
```diff
@@ -804,6 +817,70 @@ func (pgb *ChainDB) Store(blockData *blockdata.BlockData, msgBlock *wire.MsgBloc
...
+	ticketsByOutputsAllBlocks, err := retrieveTicketByOutputCount(pgb.db, outputCountByAllBlocks)
+	if err != nil {
+		return nil, fmt.Errorf("retrieveTicketByOutputCount by All Blocks: %v", err)
+	}
+
+	ticketsByOutputsTPWindow, err := retrieveTicketByOutputCount(pgb.db, outputCountByTicketPoolWindow)
+	if err != nil {
+		return nil, fmt.Errorf("retrieveTicketByOutputCount by All TP window: %v", err)
+	}
+
```
chappjc: A follow-up PR should create a cache for all of the above queries that do not change until new blocks are mined.

db/dcrpg/queries.go
```diff
@@ -1280,6 +1296,33 @@ func UpdateLastBlock(db *sql.DB, blockDbID uint64, isValid bool) error {
 	return nil
 }
 
+// UpdateLastVins updates the is_valid column of the block specified by the row
```
chappjc: ...of the vins for the block specified by...

db/dcrpg/internal/blockstmts.go
```diff
@@ -99,6 +99,15 @@ const (
 
 	SelectBlockChainRowIDByHash = `select block_db_id from block_chain where this_hash = $1;`
 
+	// SelectBlocksTicketsPrice selects the ticket price and difficulty for the 144th block
+	SelectBlocksTicketsPrice = `select sbits, time, difficulty from blocks where height % 144 = 0 order by time;`
```
chappjc: All 144 occurrences need to be replaced by a parameter set from the `chaincfg.Params.StakeDiffWindowSize`, which is different between mainnet/testnet2/simnet

db/dcrpg/queries.go
```diff
@@ -1290,17 +1333,195 @@ func RetrieveVoutValue(db *sql.DB, txHash string, voutIndex uint32) (value uint6
...
+// closeRows initiated closing of the Rows after reading all the data
+func closeRows(rows *sql.Rows) {
+	if e := rows.Close(); e != nil {
+		log.Errorf("Close of Query failed: %v", e)
+	}
+}
+
+// RetrieveTicketsPriceByHeight fetches the ticket price and its timestamp that are used
+// to display the ticket price variation on ticket price chart. This data is fetched at an interval
+// of 144 blocks.
```
chappjc: fetched at an interval of chaincfg.Params.StakeDiffWindowSize

db/dcrpg/queries.go
```diff
@@ -1290,17 +1333,195 @@ func RetrieveVoutValue(db *sql.DB, txHash string, voutIndex uint32) (value uint6
...
+	for rows.Next() {
+		var timestamp, price uint64
+		var difficulty float64
+		err = rows.Scan(&price, &timestamp, &difficulty)
+		if err != nil {
+			return nil, err
+		}
+
+		items.Time = append(items.Time, timestamp)
+		items.ValueF = append(items.ValueF, float64(price)/100000000)
```
chappjc: Please use `1e8` or `dcrutil.NewAmount`

db/dcrpg/queries.go
```diff
@@ -1290,17 +1333,195 @@ func RetrieveVoutValue(db *sql.DB, txHash string, voutIndex uint32) (value uint6
...
+		items.SizeF = append(items.SizeF, difficulty)
```
chappjc: size for difficulty?

db/dcrpg/queries.go
```diff
@@ -1290,17 +1333,195 @@ func RetrieveVoutValue(db *sql.DB, txHash string, voutIndex uint32) (value uint6
...
+		var value, timestamp int64
+		err = rows.Scan(&timestamp, &value)
+		if err != nil {
+			return nil, err
+		}
+
+		if value < 0 {
+			value = 0
+		}
+		sum += float64(value) / 100000000
```
chappjc: same comment about 1e8 or dcrutil.NewAmount

api/apirouter.go
```diff
@@ -177,6 +177,11 @@ func NewAPIRouter(app *appContext, userRealIP bool) apiMux {
 		})
 	})
 
+	mux.Route("/chart", func(r chi.Router) {
+		r.Get("/", http.NotFound)
```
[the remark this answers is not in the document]
dmigwi: noted

api/apiroutes.go
```diff
@@ -1047,6 +1047,17 @@ func (c *appContext) addressTotals(w http.ResponseWriter, r *http.Request) {
 	writeJSON(w, totals, c.getIndentQuery(r))
 }
 
+func (c *appContext) getChartTypeData(w http.ResponseWriter, r *http.Request) {
+	chartType := m.GetChartTypeCtx(r)
+	chartData, ok := explorer.CacheChartsData[chartType]
+	if !ok {
+		http.Error(w, http.StatusText(422), 422)
```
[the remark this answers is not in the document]
dmigwi: Noted

api/apiroutes.go
```diff
@@ -1047,6 +1047,17 @@ func (c *appContext) addressTotals(w http.ResponseWriter, r *http.Request) {
...
+		log.Warnf("No data matching '%v' chart Type was found: length %v", chartType, len(explorer.CacheChartsData))
```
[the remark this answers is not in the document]
dmigwi: Noted

db/dcrpg/internal/blockstmts.go
dmigwi: ok. Noted

db/dcrpg/queries.go
dmigwi: ok. Noted

db/dcrpg/queries.go
dmigwi: `dcrutil.NewAmount` seems to accept a float and returns an int64. In the above piece of code I am accepting an uint64 and converting to a float64 value so that I don't have to do the int -> float conversion in the frontend at all. I don't understand why I need to use `dcrutil.NewAmount` here..

db/dcrpg/pgblockchain.go
dmigwi: The data is cached (in a global variable) already and updated at intervals of five new blocks. The queries take almost a minute to update thus updating after the cache after every new block is mined may not be that effective.

db/dcrpg/internal/txstmts.go
dmigwi: Can this be easily done on the transactions table or I will have to write additional queries?

db/dcrpg/internal/vinoutstmts.go
dmigwi: ok. Noted

db/dcrpg/internal/vinoutstmts.go
dmigwi: removing it

db/dcrpg/internal/vinoutstmts.go
dmigwi: Where is the `main_chain` be set from? What is its relationship with orphaned blocks?

db/dcrpg/internal/blockstmts.go
dmigwi: ok. Noted

db/dcrpg/internal/blockstmts.go
```diff
@@ -99,6 +99,15 @@ const (
...
+
+	SelectBlocksBlockSize = `Select time, size, numtx, height from blocks order by time;`
```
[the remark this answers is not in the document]
dmigwi: ok. Noted

db/dcrpg/internal/vinoutstmts.go
dmigwi: Currently vins table doesn't have a block_hash, should I add a block_hash just for this or there is way around it? I think this unique index can work in place of the block_hash..
```go
IndexVinTableOnVins = "CREATE UNIQUE INDEX uix_vin ON vins(tx_hash, tx_index, tx_tree);"
```

db/dcrpg/internal/vinoutstmts.go
dmigwi: `tx_hash, tx_index, and tx_tree` are unique for every vin entry.

db/dcrpg/queries.go
chappjc: Right, I read it backwards

db/dcrpg/internal/vinoutstmts.go
chappjc: Ah, yes for the vins table it's ok as is

db/dcrpg/internal/vinoutstmts.go
chappjc: Ok, same as other comment.  I'm there transactions table is not unique for the reason I mentioned, but that doesn't apply.  My mistake

db/dcrpg/internal/txstmts.go
chappjc: Unfortunately it can't be done reliably by looking at the #vins and #vouts, and there's nothing else in the transactions table that can help.

The way to check would be to look at pkScript of the first output.  If it's script hash, then it's pool.  In the `vouts` table, there is `script_type` column.  It should be a string equal to one of these:

https://github.com/decred/dcrd/blob/eec4f96b0d890905bd871d02be498618e1aec745/txscript/standard.go#L49-L64

```go
// scriptClassToName houses the human-readable strings which describe each
// script class.
var scriptClassToName = []string{
	NonStandardTy:     "nonstandard",
	PubKeyTy:          "pubkey",
	PubkeyAltTy:       "pubkeyalt",
	PubKeyHashTy:      "pubkeyhash",
	PubkeyHashAltTy:   "pubkeyhashalt",
	ScriptHashTy:      "scripthash",
	MultiSigTy:        "multisig",
	NullDataTy:        "nulldata",
	StakeSubmissionTy: "stakesubmission",
	StakeGenTy:        "stakegen",
	StakeRevocationTy: "stakerevoke",
	StakeSubChangeTy:  "sstxchange",
}
```

I suggest declaring a ticket a pooled ticket if the type is `"scripthash"` or solo in any other case.

db/dcrpg/internal/vinoutstmts.go
chappjc: This would be part of reorg handling.  orphaned blocks would have main_chain=false

chappjc:
A proper solution for solo vs pooled tickets can wait for another PR.  I'm going to submit a small PR to this branch with some minor edits, and then I'm OK to merge.

db/dcrpg/queries.go
```diff
@@ -1336,6 +1351,32 @@ func UpdateLastBlock(db *sql.DB, blockDbID uint64, isValid bool) error {
...
+		n, err := sqlExec(db, internal.SetIsValidByTxHash,
+			"failed to update last vins tx validity: ", isValid, txHash, timestamps[i], trees[i])
+		if err != nil {
+			return err
+		}
+
+		numRows += int(n)
+	}
+	// Rows updated can be more than the total number of txs fetched because
+	// stakebase and coinbase txs are both updated on vins.
```
chappjc: I'm not following this comment.  All the vins for the transactions in the block are being updated.  I would think the number of updated vins (`numRows`) can exceed the number of transactions since transactions may have more than 1 input.

db/dcrpg/queries.go
jrick: `dcrutil.Amount(price).ToCoin()`

explorer/explorer.go
```diff
@@ -217,9 +223,42 @@ func New(dataSource explorerDataSourceLite, primaryDataSource explorerDataSource
...
 
-func (exp *explorerUI) Store(blockData *blockdata.BlockData, _ *wire.MsgBlock) error {
+// prePopulateChartsData should run in the background the first time the
+// system is initialized and consecutive times when a new block is added.
+// Its called at intervals of 5 new blocks.
+func (exp *explorerUI) prePopulateChartsData() {
+	log.Info("Pre-populating the charts data. This may take a minute...")
+	var err error
+
+	CacheChartsData, err = exp.explorerSource.GetPgChartsData()
```
chappjc: CacheChartsData has to be protected from concurrent writes.

api/apiroutes.go
```diff
@@ -1047,6 +1047,27 @@ func (c *appContext) addressTotals(w http.ResponseWriter, r *http.Request) {
 	writeJSON(w, totals, c.getIndentQuery(r))
 }
 
+func (c *appContext) getTicketPriceChartData(w http.ResponseWriter, r *http.Request) {
+	chartData, ok := explorer.CacheChartsData["ticket-price"]
```
chappjc: data race with the writer

chappjc:
Hit an error on table upgrade:

```2018-07-10 22:02:39.946 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 0 to 4999 
32 execute past end of script 801679e98561ada96caec2949a5d41c4cab3851eb740d951c10ecbcf265c1fd9
2018-07-10 22:03:08.161 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 5000 to 9999 
2018-07-10 22:03:38.534 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 10000 to 14999 
2018-07-10 22:04:08.566 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 15000 to 19999 
2018-07-10 22:04:39.137 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 20000 to 24999 
2018-07-10 22:05:13.800 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 25000 to 29999 
2018-07-10 22:05:49.997 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 30000 to 34999 
2018-07-10 22:06:25.541 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 35000 to 39999 
2018-07-10 22:07:00.493 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 40000 to 44999 
2018-07-10 22:07:35.207 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 45000 to 49999 
2018-07-10 22:08:08.737 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 50000 to 54999 
2018-07-10 22:08:42.634 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 55000 to 59999 
2018-07-10 22:09:15.324 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 60000 to 64999 
2018-07-10 22:09:49.688 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 65000 to 69999 
2018-07-10 22:10:28.236 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 70000 to 74999 
2018-07-10 22:11:05.893 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 75000 to 79999 
2018-07-10 22:11:38.846 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 80000 to 84999 
2018-07-10 22:12:20.620 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 85000 to 89999 
2018-07-10 22:13:09.925 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 90000 to 94999 
2018-07-10 22:13:57.438 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 95000 to 99999 
2018-07-10 22:14:41.412 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 100000 to 104999 
2018-07-10 22:15:25.214 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 105000 to 109999 
2018-07-10 22:16:06.032 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 110000 to 114999 
2018-07-10 22:16:45.004 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 115000 to 119999 
2018-07-10 22:17:34.874 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 120000 to 124999 
2018-07-10 22:18:23.662 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 125000 to 129999 
2018-07-10 22:19:18.839 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 130000 to 134999 
2018-07-10 22:20:18.596 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 135000 to 139999 
2018-07-10 22:21:15.792 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 140000 to 144999 
2018-07-10 22:22:13.182 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 145000 to 149999 
2018-07-10 22:23:11.520 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 150000 to 154999 
2018-07-10 22:24:04.759 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 155000 to 159999 
2018-07-10 22:25:11.128 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 160000 to 164999 
2018-07-10 22:26:11.612 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 165000 to 169999 
2018-07-10 22:27:00.707 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 170000 to 174999 
2018-07-10 22:27:52.421 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 175000 to 179999 
2018-07-10 22:28:41.243 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 180000 to 184999 
2018-07-10 22:29:30.174 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 185000 to 189999 
2018-07-10 22:30:30.094 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 190000 to 194999 
2018-07-10 22:31:29.883 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 195000 to 199999 
2018-07-10 22:32:24.239 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 200000 to 204999 
2018-07-10 22:33:24.076 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 205000 to 209999 
2018-07-10 22:34:14.273 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 210000 to 214999 
2018-07-10 22:35:05.594 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 215000 to 219999 
2018-07-10 22:35:59.012 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 220000 to 224999 
2018-07-10 22:36:44.004 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 225000 to 229999 
2018-07-10 22:37:29.338 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 230000 to 234999 
2018-07-10 22:38:20.961 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 235000 to 239999 
2018-07-10 22:39:16.307 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 240000 to 244999 
2018-07-10 22:40:08.829 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 245000 to 249999 
2018-07-10 22:41:01.489 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 250000 to 254999 
2018-07-10 22:41:52.708 [INF] PSQL: Upgrading the vins table (Coin Supply Upgrade) from height 255000 to 255447 
2018-07-10 22:42:06.932 [INF] DATD: Closing connection to dcrd.
2018-07-10 22:42:06.932 [INF] DATD: Bye!
2018-07-10 22:42:07.183 [ERR] DATD: Expected to update all of 11426912 vins records but -16 records were not updated: 11426928```

db/dcrpg/queries.go
chappjc: yup, I suggested the same in some edits to the branch: https://github.com/dmigwi/dcrdata/pull/4/files#diff-87f7a577b6af45d0284e4c3fbb98ddf5L1446
To avoid magic numbers and make the semantics obvious, it is worth the import.

api/apiroutes.go
dmigwi: ok. Noted

explorer/explorer.go

db/dcrpg/queries.go
dmigwi: Let me improve on it.

dmigwi: Actually It doesn't make sense since by default most transactions have multiple vins entry. So its obvious that the total number of vins entries should be more than the total number of transactions fetched.

dmigwi:
Just finished fixing all the issues with PR, I am not testing if the upgrade will work.

chappjc:
OK, I'll test, but the upgrade is already done so I don't know if that error is fixed.  Do you know what caused that?

dmigwi:
I think its because if the orphaned blocks but I can't tell exactly.

dmigwi:
I finished running the test on my end and things look fine.

chappjc:
Running a fresh sync on mainnet, I got this around block 80,000 or so:

    [ERR] PSQL: UpdateLastVins: failed to update last vins tx validity: pq: syntax error at or near ","

chappjc:
I think it should be changed from:

    SetIsValidByTxHash = `UPDATE vins SET is_valid = $1 WHERE tx_hash = $2, block_time = $3 AND tx_tree = $4;`

to:

    SetIsValidByTxHash = `UPDATE vins SET is_valid = $1 WHERE tx_hash = $2 AND block_time = $3 AND tx_tree = $4;`

PR #466, opened 2018-05-12
Deny search for zero hash P2PHK address

### LINKED ISSUE #358: Deny search for DsQxuVRvS4eaJ42dhQEsCXauMWjvopWgrVg
This is the sstxchange address for 0 value outputs in ticket purchases.  It's used all the time, but never has a balance (unless someone sent money there)

rmukhamet:
resolve #358 
This is the sstxchange address for 0 value outputs in ticket purchases. It's used all the time, but never has a balance (unless someone sent money there)

I've renamed this from deny to "DsQxuVRvS4eaJ42dhQEsCXauMWjvopWgrVg" to deny "zero hash P2PHK address" to emphasize that this needs to work on all nets (mainnet/testnetX/simnet).

### DIFF
--- a/api/apirouter.go
+++ b/api/apirouter.go
@@ -7,7 +7,6 @@ import (
 	"net/http"
 	"strings"
 
-	"github.com/decred/dcrd/chaincfg"
 	m "github.com/decred/dcrdata/middleware"
 	"github.com/go-chi/chi"
 	"github.com/go-chi/chi/middleware"
@@ -141,7 +140,7 @@ func NewAPIRouter(app *appContext, userRealIP bool) apiMux {
 	mux.Route("/address", func(r chi.Router) {
 		r.Route("/{address}", func(rd chi.Router) {
 			rd.Use(m.AddressPathCtx)
-			rd.Use(m.ZeroAddrDenier(&chaincfg.MainNetParams))
+			rd.Use(m.ZeroAddrDenier)
 
 			rd.Get("/totals", app.addressTotals)
 			rd.Get("/", app.getAddressTransactions)
--- a/api/apirouter.go
+++ b/api/apirouter.go
@@ -140,6 +140,7 @@ func NewAPIRouter(app *appContext, userRealIP bool) apiMux {
 	mux.Route("/address", func(r chi.Router) {
 		r.Route("/{address}", func(rd chi.Router) {
 			rd.Use(m.AddressPathCtx)
+			rd.Use(m.DenySpecificAddr)
 			rd.Get("/totals", app.addressTotals)
 			rd.Get("/", app.getAddressTransactions)
 			rd.With((middleware.Compress(1))).Get("/raw", app.getAddressTransactionsRaw)
--- a/api/apirouter.go
+++ b/api/apirouter.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"strings"
 
+	"github.com/decred/dcrd/chaincfg"
 	m "github.com/decred/dcrdata/middleware"
 	"github.com/go-chi/chi"
 	"github.com/go-chi/chi/middleware"
@@ -140,7 +141,7 @@ func NewAPIRouter(app *appContext, userRealIP bool) apiMux {
 	mux.Route("/address", func(r chi.Router) {
 		r.Route("/{address}", func(rd chi.Router) {
 			rd.Use(m.AddressPathCtx)
-			rd.Use(app.ZeroAddrDenier)
+			rd.Use(m.ZeroAddrDenier(&chaincfg.MainNetParams))
 
 			rd.Get("/totals", app.addressTotals)
 			rd.Get("/", app.getAddressTransactions)
--- a/api/apirouter.go
+++ b/api/apirouter.go
@@ -140,7 +140,8 @@ func NewAPIRouter(app *appContext, userRealIP bool) apiMux {
 	mux.Route("/address", func(r chi.Router) {
 		r.Route("/{address}", func(rd chi.Router) {
 			rd.Use(m.AddressPathCtx)
-			rd.Use(m.DenySpecificAddr)
+			rd.Use(app.ZeroAddrDenier)
+
 			rd.Get("/totals", app.addressTotals)
 			rd.Get("/", app.getAddressTransactions)
 			rd.With((middleware.Compress(1))).Get("/raw", app.getAddressTransactionsRaw)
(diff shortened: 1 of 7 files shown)


middleware/apimiddleware.go
```diff
@@ -479,6 +479,24 @@ func AddressPostCtx(next http.Handler) http.Handler {
 	})
 }
 
+// DenySpecificAddr https://github.com/decred/dcrdata/issues/358
+func DenySpecificAddr(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		address := GetAddressCtx(r)
+		if address == "DsQxuVRvS4eaJ42dhQEsCXauMWjvopWgrVg" {
+			apiLog.Info("got match for addr deny condition")
```
chappjc: Trace

middleware/apimiddleware.go
```diff
@@ -479,6 +479,24 @@ func AddressPostCtx(next http.Handler) http.Handler {
 	})
 }
 
+// DenySpecificAddr https://github.com/decred/dcrdata/issues/358
+func DenySpecificAddr(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		address := GetAddressCtx(r)
+		if address == "DsQxuVRvS4eaJ42dhQEsCXauMWjvopWgrVg" {
```
chappjc: See if this address is somewhere in dcrd or dcrwallet code.  I'd prefer not to hard code this

middleware/apimiddleware.go
```diff
@@ -479,6 +479,24 @@ func AddressPostCtx(next http.Handler) http.Handler {
...
+			apiLog.Info("got match for addr deny condition")
+			msg := `Request with DsQxuVRvS4eaJ42dhQEsCXauMWjvopWgrVg was denied. This is the sstxchange address for 0 value outputs in ticket purchases. It's used all the time, but never has a balance`
```
chappjc: Please make a type with json tags

middleware/apimiddleware.go
rmukhamet: deleted log

middleware/apimiddleware.go
rmukhamet: @chappjc  there is no this address. is there are another way to find this address?

middleware/apimiddleware.go
rmukhamet: done

middleware/apimiddleware.go
```diff
@@ -479,6 +480,24 @@ func AddressPostCtx(next http.Handler) http.Handler {
 	})
 }
 
+// DenySpecificAddr https://github.com/decred/dcrdata/issues/358
+func DenySpecificAddr(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		address := GetAddressCtx(r)
+		params := &chaincfg.MainNetParams
```
chappjc: This has to work for all nets.

Need `DenyAddress(address string) func(http.Handler) http.Handler` that captures the address in a closure (the actual middleware).

chappjc: To clarify, `chaincfg.MainNetParams` is going to generate the address DsQxuVRvS4eaJ42dhQEsCXauMWjvopWgrVg .  But testnet has a different address.

chappjc: I've renamed this from deny to "DsQxuVRvS4eaJ42dhQEsCXauMWjvopWgrVg" to deny "zero hash P2PHK address" to emphasize that this needs to work on all nets (mainnet/testnetX/simnet).

api/apirouter.go
```diff
@@ -140,7 +140,8 @@ func NewAPIRouter(app *appContext, userRealIP bool) apiMux {
 	mux.Route("/address", func(r chi.Router) {
 		r.Route("/{address}", func(rd chi.Router) {
 			rd.Use(m.AddressPathCtx)
-			rd.Use(m.DenySpecificAddr)
+			rd.Use(app.ZeroAddrDenier)
```
chappjc: `app.ZeroAddrDenier(params)`

api/apiroutes.go
```diff
@@ -1099,3 +1101,38 @@ func (c *appContext) getBlockHashCtx(r *http.Request) string {
 	}
 	return hash
 }
+
+// ZeroAddrDenier https://github.com/decred/dcrdata/issues/358
+func (c *appContext) ZeroAddrDenier(next http.Handler) http.Handler {
```
chappjc: This can stay in `package middleware`, no need to duplicate code.  More importantly, no need for an RPC call to figure out what net you're running on.  Just pass the params as I suggested in the last review.

So instead of this being a bare middleware, you'll have a factory func:

```go
func (c *appContext) ZeroAddrDenier(params *chaincfg.Params) func(http.Handler) http.Handler {
    deniedAddr := apitypes.NewZeroAddressDenial(params)
    return func(next http.Handler) http.Handler {
        ... // do stuff with deniedAddress
        next.ServeHTTP(w, r.WithContext(r.Context()))
    }
}
```

Then my comment about `app.ZeroAddrDenier(params)` should make sense.

rmukhamet: appContext this is internal  struct of api package. I can not use it in insight package as well as to use it in middleware as you suggested on screen.

Moreover, api imports middleware package. When I try to import api package in middleware, I got cycle import error.

api/apirouter.go
rmukhamet: according docs https://godoc.org/github.com/go-chi/chi#Mux.Use
Use accepts handler without parameters and returns handler without parameters.

api/apiroutes.go
chappjc: Yeah, I forgot to remove the `(c *appContext)` receiver in my example.  The point is you don't need `c`.  It'll be a regular function in the middleware package.

api/apirouter.go
chappjc: Please see the suggested changes.  `m.ZeroAddrDenier(params)` would return such a function:

    func ZeroAddrDenier(params *chaincfg.Params) func(http.Handler) http.Handler {

rmukhamet:
@chappjc So how is one should go about getting current net params without rpc call?

api/apirouter.go
rmukhamet: where you get params from?

chappjc:
Some good work here, but going in circles too.  @decred/dcrdata , anyone to resume this?

PR #467, opened 2018-05-12
add keyboard navigation

gozart1:
As mentioned by @chappjc in https://github.com/decred/dcrdata/issues/423 keyboard navigation can improve the ergonomics of exploring the chain. 

So this is a wip to enhance keyboard navigation beyond the native tabbing functionality of the browser, since that falls short in various ways. 

### Keys Reference
- Arrow keys navigate between links
- "Enter" follow links
- "=" toggles the hamburger menu

Thinking to make this optional so a user would need to activate it via the drop down nav. When enabled, the required js would load and we stop auto-focusing the search input on page loads (  since that takes over the arrow keys ).

Also need to find a way to show a reference of the possible key commands, I'm thinking a small keyboard icon in the footer you can toggle to pop out the cheat sheet.

Lmk what ya'll think!

screen shot 2018-05-12 at 2 06 11 am


chappjc:
utACK. Love the concept.  Do need a cheat sheet as you said.  Maybe activate it with "?" or "h" in addition to a link somewhere.

Issue #453, opened 2018-04-27
Create graphical block display


raedah:
Per the mockup at https://github.com/decred/dcrdata/issues/451
Also include the upcoming block (mempool) and show it being built in real time.
This is probably a good item for @gozart1

denyszayets:
How is this upcoming block built is displayed currently?

raedah:
Upcoming block data is currently being shown at http://explorer.dcrdata.org/mempool , in the same form that blocks are displayed such as http://explorer.dcrdata.org/block/234151

chappjc:
In progress in https://github.com/decred/dcrdata/pull/660 by @dmigwi

PR #468, opened 2018-05-14
APR Calculation on home page

papacarp:
This PR supersedes #455 . 

This Proposed PR addresses APR calculations as referenced in #381
image

**Guidance is requested on the following items:**

1.  A new function, simulateAPR,  was added to explorer.go.  Architecturally I'm not sure if this is the best place for this.  Viable alternatives may be:  explorermiddleware.go or txhelpers.go
2.  StakeRewardAtBlock was originally created using a calculation and then replaced with an RPC call.  Either works fine so its just a matter of preference.  Seems like an RPC call would be less efficient, but likely is a better architecture to use a function if it exists.
3.  MaxCoinSupplyAtBlock is a calculation.  I was hoping to find an RPC call to provide a theoretical coinbase projection for an arbitrary future block, but it does not seem to exist.  I created a 4th order poly to fit to the expected emissions.  The calculation is here:  https://docs.google.com/spreadsheets/d/1uzIBF9q__JYx4EWeAJAdviB4B-Y8IVNEUGmSww3HRX8/edit?usp=sharing
4.  Should the apr simulation be executed on a separate thread much like we have done with `go exp.updateDevFundBalance()` to help improve rendering time?  

**Brief Explanation of Solution:**
1.  We have a function to forecast reward at an arbitrary block number
2.  We have a function to forecast max coin supply at an arbitrary block number
3.  We have a function to forecast theoretical ticket price at an arbitrary block number.  Assuming stake pool percent of total coin supply stays constant, we can project forward what the ticket price will be as the coin supply increases.
4.  We normalize our ticket prices and coin supply to match existing actuals to further refine our projections.
5.  We then start a loop where we simulate a purchase at the current block, a reward after the "average" reward time in blocks, and coinbase maturity to get the funds back in our wallet.  We repeat this cycle enough times to get through a full year.
6.  Once we are through at least 365 days we then reduce the ROI down to exactly 365 days to get a APR for a year rather than for the simulated number of days.

**Simulation Options:**
1.  You can select to use integer ticket numbers or float ticket numbers.  This was done to make the ROI independent of actual tickets purchased, and assuming 100% of funds are in use.  It is our vision that in the future a separate "detailed APR" calculation page can be implemented where the user can enter their actual DCR and simulate actual possible ticket purchases.
2.  The simulation outputs a simple text table that can be leveraged into a future implementation where the details of the simulation are presented to the user.  Rather than remove it from the prototyping I did, I left it so its easier to add back later.  You can see it in action here:  https://play.golang.org/p/y6exAOH9yQD

### DIFF
--- a/explorer/explorer.go
+++ b/explorer/explorer.go
@@ -6,6 +6,7 @@ package explorer
 
 import (
 	"fmt"
+	"math"
 	"net/http"
 	"os"
 	"os/signal"
@@ -51,6 +52,7 @@ type explorerDataSourceLite interface {
 	UnconfirmedTxnsForAddress(address string) (*txhelpers.AddressOutpoints, int64, error)
 	GetMempool() []MempoolTx
 	TxHeight(txid string) (height int64)
+	GetBlockSubsidy(height int64, voters uint16) *dcrjson.GetBlockSubsidyResult
 }
 
 // explorerDataSource implements extra data retrieval functions that require a
@@ -230,6 +232,8 @@ func (exp *explorerUI) Store(blockData *blockdata.BlockData, _ *wire.MsgBlock) e
 		return (a / b) * 100
 	}
 
+	stakePerc := blockData.PoolInfo.Value / dcrutil.Amount(blockData.ExtraInfo.CoinSupply).ToCoin()
+
 	// Update all ExtraInfo with latest data
 	exp.ExtraInfo.CoinSupply = blockData.ExtraInfo.CoinSupply
 	exp.ExtraInfo.StakeDiff = blockData.CurrentStakeDiff.CurrentStakeDifficulty
@@ -243,7 +247,7 @@ func (exp *explorerUI) Store(blockData *blockdata.BlockData, _ *wire.MsgBlock) e
 	exp.ExtraInfo.PoolInfo.Size = blockData.PoolInfo.Size
 	exp.ExtraInfo.PoolInfo.Value = blockData.PoolInfo.Value
 	exp.ExtraInfo.PoolInfo.ValAvg = blockData.PoolInfo.ValAvg
-	exp.ExtraInfo.PoolInfo.Percentage = percentage(blockData.PoolInfo.Value, dcrutil.Amount(blockData.ExtraInfo.CoinSupply).ToCoin())
+	exp.ExtraInfo.PoolInfo.Percentage = stakePerc * 100
 
 	exp.ExtraInfo.PoolInfo.PercentTarget = func() float64 {
 		target := float64(exp.ChainParams.TicketPoolSize * exp.ChainParams.TicketsPerBlock)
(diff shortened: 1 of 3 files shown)


chappjc:
Thanks for the clean up.

papacarp:
Thank you for the feedback.  I'll fix those things.

> I still need to go through simulateAPR carefully. @papacarp are you all done with this PR though (besides addressing review)?

The only thing we might want to do is adjust the "average vote time" to 7860 rather than 8192 as per #471.  That would involve changing this line:
`var AvgTicketBlocks = float64(exp.ChainParams.TicketExpiry) / float64(exp.ChainParams.TicketsPerBlock)`

to something like

`AvgTicketBlocks := 7860`

Along with enough in the comments to explain where that came from.  Or we could make it a calculation based on chain params.

chappjc:
mmm, right.  Expectation: `E[B] = sum(b_i p(b_i)), i=1 to 40960`

image

The infinite sum is 8192, but of course a ticket only gets up to 40960 chances.

papacarp:
Small nomenclature update per requests.  I may roll in the 7860 adjustment here as well so hold off until we resolve that one issue.

image

papacarp:
1.  ASR => / year
2.  7860 is now calculated from mainnet params on explorer init

This is ready for review now.  

image

freethinkingaway:
I've been working on the graphical block display (#453) and after pulling this past week's commits ran into an issue where the websocket appeared to hang (no block signaled to websocket client) and the explorer/API never served to localhost:7777.

After reviewing, compiling, and testing each recent commit, the issue appears to be the added simulateASR function, specifically the for loop at line 402. Perhaps this isn't an issue when running dcrdata for testnet2 or the mainnet, but the loop doesn't appear to escape when running as simnet. I've tried running with zero blocks and I've tried after a quick ~150 block run. Could be an issue on my end or specific to running as simnet.

papacarp:
@freethinkingaway  we never tried it on simnet.  Looks like blocks are 1 second on simnet?  If so, I could see this loop taking ~forever.   I'll look into it.

papacarp:
@freethinkingaway everything is very accelerated on mainnet - block time, maturity, vote times, etc..  I'll confirm with @chappjc and @raedah on how best to deal with this, but if you can verify that adding the lines below solves the  simnet problem then this is what I'll recommend:

```go
func (exp *explorerUI) simulateASR(StartingDCRBalance float64, IntegerTicketQty bool,
	CurrentStakePercent float64, ActualCoinbase float64, CurrentBlockNum float64,
	ActualTicketPrice float64) (ASR float64, ReturnTable string) {

	// Calculations are only useful on mainnet.  Short circuit calculations if
	// on any other version of chain params.
	if exp.ChainParams.Name != "mainnet" {
		return 0, ""
	}
	
```

freethinkingaway:
Problem solved.

chappjc:
Agreed.  Don't waste time with it.

Issue #297, opened 2017-12-11
Blocks page: Older and Newer links require javascript


xaur:
At the [blocks page](https://explorer.dcrdata.org/explorer) the "Older" and "Newer" navigation links do not work if javascript is disabled. If enabled, they work and send to URL with proper parameters e.g. https://explorer.dcrdata.org/explorer?height=193934&rows=20 . Such URLs work with javascript disabled, so it seems possible to serve Older and Newer links in plain HTML.

chappjc:
resolved by PR https://github.com/decred/dcrdata/pull/476

PR #476, opened 2018-05-18
refine keyboard nav and make it optional via menu toggle

gozart1:
- Left and right arrow only
- '=' toggles menu
- '/' toggles search
- Menu option to enable/disable ( off by default, but stays enabled via cookie just like night mode )
- Subtler highlights with animations for selection and activation ( pulses while loading )

screen shot 2018-05-18 at 2 55 34 am


chappjc:
Just put this up on https://beta.dcrdata.org/.

I really like the subtler highlights.  A couple of observations:

- The "/" key also opens the Firefox ["quick find"](https://stackoverflow.com/q/19650894/2778484) menu.  Maybe "="?  It sorta looks like a menu.
- Enabling hotkeys requires a page reload.  I was confused at first because they didn't work.  Can we make it reload or do something to indicate it's required?

gozart1:
Alright, lets stick with = for menu toggle, the resemblance was my original reasoning for choosing it as well. Gave / a shot since that puts it all in the same location on keyboard, but actually found myself missing liking = better as well. 

Are you sure you actually have to reload? For me the top menu item highlights immediately upon activation. Maybe the highlight is too subtle now and you where expecting it to highlight something in the main page instead of the menu and therefore assumed it wasn't working?

I've been tweaking animation of the outline color as you switch to an item and feel like it has a good balance now. But maybe we need to intensify it in case you're simply overlooking the starting item.

Ah, what you probably did was activate it and the close the menu via mouse. We need to make sure to auto highlight the first item **on the page** at that point. This already happens when closing menu via hotkey, but not via mouse.

gozart1:
Ok, made a few tweaks:

- Give a bit more visual hint on the location of the active selection after a page load and when hot keys are first activated, via a slightly more pronounced pulsate animation of outline color.
- "Refocus" the browser context onto the main page when closing menu via mouse ( focus being applied to the hidden checkbox that governs the js free hamburger toggle was trapping key events )
- hook keynav function into a change handler for the value of the hidden hamburger menu checkbox, so refocusing logic is triggered regardless of how the menu is toggled.
- switch back to = for menu toggle

Thanks for all the feedback, think this is in pretty good shape now. Lmk if anything else.

chappjc:
Oh, there was one thing.  Since this causes us to lose the autofocus on the search box, perhaps we can bind a key jump to the search text box.

chappjc:
Great.  That will address Issue https://github.com/decred/dcrdata/issues/297

chappjc:
Sorry for the conflict just added in extras.tmpl.  It's pretty minor.  Could you rebase too?

gozart1:
- /blocks pagination links are now in go and have a sticky keynav index ( focus stays on "Newer" or "Older" you scroll forwards or backwards )
- added "/" for toggling search, similar to Github ui

I also noticed that /blocks accepted negative heights, so added a fallback to set height to 0 in that scenario https://github.com/decred/dcrdata/pull/476/files#diff-022a8370d807e3c88802b53baa33af77R77

gozart1:
Ah yea you comment about the conflict with Firefox for using / for search went in one ear and right out the other the first time!

Also added [ & ] as alternates keys for going left and right, not sure if using arrow keys is ideal since they may be used to traverse characters while entering values into search input. Right now it doesn't stop you from doing that, but the key nav selector will still hop links as well, so it's just slightly goofy. I don't want to make this feature any more elaborate than it already is by selectively binding and unbinding keys based on various states. So just added [ & ] for now as a hidden feature to see how we like it since that avoids that quirk with no extra code involved.

gozart1:
Added one last little fix to prevent the blocks pagination from trying to request blocks older than height 0. Admittedly it's a bit out of scope for this PR but ¯\_(ツ)_/¯

PR #484, opened 2018-05-27
Err message page clean up

timi-ola:
Simplified the error message page.


chappjc:
Please rebase and resolve, then we will merge.

raedah:
Could use some cleanup on Search and ErrorPage in explorerroutes for the vars being removed that were not serving a purpose.

PR #485, opened 2018-05-28
api changed to view decoded on line 17 on block page

stanleychidi:
block page


raedah:
These are not correct.

Issue #130, opened 2017-09-08
Unconfirmed transactions are showing unconfirmed inputs.


raedah:
Reproduce by taking a txid from getrawmempool and loading in the dcrdata gui.
The inputs of the unconfirmed transaction show N/A for both the Block Height and Value. 
Then click on the input itself, and see that it included in a block and has a value.

raedah:
This probably has the same cause as https://github.com/dcrdata/dcrdata/issues/142

ukane-philemon:
@chappjc, I think this has been fixed by the merged PRs above, no? Maybe should close if yes.

PR #489, opened 2018-06-01
Unconfirmed tx address list

### LINKED ISSUE #142: Recently confirmed transactions do not show recent data
On the transaction view, 
Time | Confirmations | Included in block
N/A | Unconfirmed | N/A

### LINKED ISSUE #277: address view does not show unconfirmed transactions

papacarp:
This PR addresses the unconfirmed transactions not visible in address view and input amounts not visible in transaction view.

This should close #277 , #130 
This could close #142 with a few more small changes - it should be easy to add those from mempool. 

There were two major areas of work:
1.  Blending unconfirmed transactions with confirmed transactions in the current address view.  Mostly extending what @chappjc  already started.
2.  Compensating for AmountIn values not set in dcrd calls to mempool transactions.

Known Issues
1.  Unconfirmed transactions are added at the start of the address view list regardless of indexing (start/count) settings.  I left some commented code in there for now that does the final slice to illustrate where we would do it, but we can remove before merge.  Fixing the indexing issue is a broader issue than this PR and is likely going to happen with PR#427 or I can fix it separately with a variation on the experimental PR I proposed (PR #433 but I do not think we need to add the column to the addresses table anymore).
2.  DCRD is fixing the AmountIn issue.  Depending on the status of that change the repository owners can make a determination if we should leave out those compensating changes or include in this PR.  

**Transaction View**

*Before*
image

*After*
image

**Address View**

*Before*
image

*After*
image

*Before*
image

*After*
image
Note:  age is updated

*Before*
image

*After*
image
Note:  order of transactions is corrected.  Also note that unconfirmed tx is 1, since the transaction only has one confirmation.

*Before*
image

*After*
image


raedah:
I would not work around the dcrd issue unless important functionality is broken. Dcrdata is meant to accurately display what dcrd reports. If the work around does need to be used, https://github.com/decred/dcrdata/issues/130  and https://github.com/decred/dcrdata/issues/142 should be kept open, or a new issue created as a reminder to remove the temporary fix after the dcrd issue is resolved.

raedah:
Looks like tx could be subsorted by in[] or out[]

papacarp:
Rebased and updated to sort by in/out if time is same.  Note transaction confirmations are showing as different for the same tx.  Is that expected?

image

This is ready for review @chappjc

chappjc:
Nope, definitely not expected.  Looks like it's showing confirms for the funding transaction rather than the spending transaction.

papacarp:
I can rebase this today or tomorrow.  The confirmation issue was not fixed AFAIK.  Do you want me to go ahead and fix issue #507 as well as part of this PR?

chappjc:
If you see the root cause and it is trivial to fix, yes, it's cool to include the fix here.  Otherwise this can go in and we can fix the confirmations issue in a follow-up.  Thanks!

papacarp:
rebased.  The  issue #507 seems to have fixed itself in the address view updates @dmigwi did.  However, i see two potentially new issues - I have confirmed they are in master right now.  I suspect both are related.

1.  unspent is error if the total unspent balance is less than zero.
2.  total unspent balance is likely less than zero because this address view has missing transactions.  There should be 30 transactions, but tip only pulls 20 transactions.

looks like something changed in "exp.explorerSource.AddressHistory' so it doesn't get the right balance.

current view: 20 transactions
image

compare to beta:  https://beta.dcrdata.org/address/DsiJmwqFbsCEc1H3HF8UeRjoqJtKt9ZFRKw 

This PR can be merged into master once your review is complete.  I recommend we open up new bugs for the issues noted above.

Mike

dmigwi:
>unspent is error if the total unspent balance is less than zero.
This is a concern that we should address ASAP.

>total unspent balance is likely less than zero because this address view has missing transactions. There should be 30 transactions, but tip only pulls 20 transactions.

This concern is addressed by the links source and spent.  Notice the time in txs that appear here https://beta.dcrdata.org/address/DsiJmwqFbsCEc1H3HF8UeRjoqJtKt9ZFRKw. They are not well ordered but what appears in the above screenshot is well order. That was the main reason why we had to alter the addresses table.

chappjc:
@papacarp, @dmigwi please coordinate on those issues and their resolutions.  I'll get this PR in as is.

PR #479, opened 2018-05-22
Insight API UTXO+ [10] [11] [12] [3] [4]

papacarp:
This PR is ready for testing on the UTXO endpoints:
[10] addr_utxo
[11] multi_addr_utxo
[12] POST_multi_addr_utxo; 

In addition the following changes were implemented:
1.  New standard API failure packet.  Return values for failure modes is highly inconsistent across mainnet and bitpay api.  We used the return error format from bitpay for a blank hash on a block call as our model.  Specifically:
```
{
    "status": 404,
    "url": "/api/block/",
    "error": "Not found"
}
```
I hope to align all Insight API failure responses to use this method.

2.  rawblock endpoint was fixed.  so [3] and [4] are also ready for test.

**Valid UTXO Response**
The target UTXO schema is shown below:
image
(latest version can always be found here:  https://docs.google.com/spreadsheets/d/1WGHezqRG5VY-t4O6cGpZGwsq69IKJ3deEnERaVvQOb4/edit?usp=sharing)

**Example Output of unconfirmed**
```
[
    {
        "address": "Dsbb8DHHwWMkxSSgfAj9czC44VVKXZPWAmg",
        "txid": "fe4b0315638d81070917f84312f8e288121922080662ed43016af37edcb5ff25",
        "vout": 0,
        "ts": 1526970773,
        "scriptPubKey": "76a914748ada15e1200593e21e9d3eb15e1e259cabdbeb88ac",
        "amount": 0.0489327,
        "satoshis": 4893270,
        "confirmations": 0
    },
    {
        "address": "Dsbb8DHHwWMkxSSgfAj9czC44VVKXZPWAmg",
        "txid": "4f44f80db15c0da95e89458371561e9a22db4469b271d17a68d218a8421dec05",
        "vout": 1,
        "scriptPubKey": "76a914748ada15e1200593e21e9d3eb15e1e259cabdbeb88ac",
        "height": 240352,
        "amount": 0.01700671,
        "satoshis": 1700671,
        "confirmations": 561
    }
]
```
**Example response for multi addr**
```
[
    {
        "address": "37p9pUugydmoLpQyFLLqGAgjWmUFERa1Pq",
        "txid": "80106ac7ef410be2b99fab552da9fba0e19cf5d0d79fa9eb64d946ece8a5ff0c",
        "vout": 0,
        "scriptPubKey": "a914432a41db83cc1f7e5cc9c48a0808b00ff2992a3a87",
        "amount": 0.01200328,
[...]

Issue #400, opened 2018-03-20
Insight HTTP API verification and validation


chappjc:
The initial framework for Insight API support (minus websocket) was merged in PR #350. Now we need to be verify, fix and/or implement all of the endpoints:

- API reference: https://github.com/bitpay/insight-api#api-http-endpoints
- Compare results with:  https://mainnet.decred.org/

Verify and validate each endpoint listed below:

- [x] 1. Block. `/block/{hash}` PR https://github.com/decred/dcrdata/pull/483
- [x] 2. Block index.  `/block-index/{height}` PR https://github.com/decred/dcrdata/pull/483
- [x] 3. Raw block (hash).  `/rawblock/{hash}` PR https://github.com/decred/dcrdata/pull/402
- [x] 4. Raw block (height).  `/rawblock/{height}` PR https://github.com/decred/dcrdata/pull/402
- [x] 5. Raw summaries.  `/blocks` (URL query params: `limit`, `blockDate`, ...) PR https://github.com/decred/dcrdata/pull/483
- [x] 6. Transaction.  `/tx/{hash}`  PR https://github.com/decred/dcrdata/pull/493
- [x] 7. Raw transaction. `/rawtx/{hash}`
- [x] 8. Address.  `/addr/{address}` (URL query params: `noTxList`, `from`, `to`, ...) PR https://github.com/decred/dcrdata/pull/483
- [x] 9. Address Props.  `/balance`, `totalReceived`, `totalSent`, `/totalSent`, `unconfirmedBalance` PR https://github.com/decred/dcrdata/pull/483
- [x] 10. UTXOs.  `/addr/{address}/utxo` PR https://github.com/decred/dcrdata/pull/479
- [x] 11. Multi-address UTXOs.  `/addrs/{address0, address1, etc}/utxo` PR https://github.com/decred/dcrdata/pull/479
- [x] 12. POST-variant of multi-address UTXOs. PR https://github.com/decred/dcrdata/pull/479
- [x] 13. Transactions by block.  `/txs/?block={hash}` PR https://github.com/decred/dcrdata/pull/506
- [x] 14. Transactions by address.  `/txs/?address={address}` PR https://github.com/decred/dcrdata/pull/506
- [x] 15. Transactions for multiple addresses.  `/addrs/{addr0, addr1, ...}/txs`  (params: `from`, `to`) PR https://github.com/decred/dcrdata/pull/493
[...]

Issue #260, opened 2017-10-29
Implement the Insight API


chappjc:
For compatibility with many platforms that use it.

alexlyp:
After discussing with @behindtext, we've decided to abandon insight/insight-api altogether in favor of dcrdata along with an the insight-api endpoints implemented.  I will begin discussion of this transition here.  

First step to the process will be do audit all the insight-api endpoints that our various services use (copay/exodus).  I've reached out to the Exodus team and will post their required endpoints here.  I'll research bitcore-wallet-service for it's needed endpoints so copay keeps working.

alexlyp:
From Exodus:
- broadcast tx
- `/tx/${txId}` -- for transaction confirmation
- `/addrs/${addresses}/utxo?noCache=1` -- fetch UTXOs for importing private key and wallet startup
- `/addrs/txs` -- fetch tx data on wallet refreshing
- also we use socket.io for incoming transactions

chappjc:
Completion and testing of HTTP endpoints is being tracked in Issue #400

chappjc:
Resolved by https://github.com/decred/dcrdata/pull/644

PR #493, opened 2018-06-05
Insight API [6], [6.1], [15], [15.1], [16], [16.1], [16.2], [16.3], [16.4]

papacarp:
This PR replaces PR #450 and was developed by both @rmukhamet and I.  This PR addresses issues #260 and #400 but does not yet close them. 

The following Insight API endpoints are completed with exceptions as noted below.
[6]  tx/{txid}
[6.1] tx/{unknown txid}
[15] addrs/{addr1},{addr2}/txs
[15.1] addrs/{addr1},{addr2}/txs?from=X&to=Y
[16]  addrs/txs (POST)
[16.1] addrs/txs (POST with from/to)
[16.2] addrs/txs (POST with noAsm)
[16.3] addrs/txs (POST with noScriptSig)
[16.4] addrs/txs (POST with noSpent)

All endpoint comparisons are visible here:  https://docs.google.com/spreadsheets/d/1WGHezqRG5VY-t4O6cGpZGwsq69IKJ3deEnERaVvQOb4/edit?usp=sharing

**JSON output comparison for [6]**
image

**JSON output comparison for [15],[16]**
image

Notable issues
1.  Vin addresses for transactions that are spending other mempool outputs are not accurate
2.  Vout spend fields for transactions that are spent in mempool are not accurate

These issues affect only transactions in mempool.  I'm recommending we fix these with a broader update of mempool caching sorting/filtering in dcrdata and merge and likely open an issue for these small issues so we can get exodus testing started.

To start Exodus testing all we need is this PR and #479 to get merged.

### DIFF
--- a/api/insight/apimiddleware.go
+++ b/api/insight/apimiddleware.go
@@ -10,15 +10,23 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"strconv"
 
 	apitypes "github.com/decred/dcrdata/api/types"
 	m "github.com/decred/dcrdata/middleware"
+	"github.com/go-chi/chi"
 )
 
 type contextKey int
 
 const (
 	ctxRawHexTx contextKey = iota
+	ctxFrom
+	ctxTo
+	ctxNoAsm
+	ctxNoScriptSig
+	ctxNoSpent
+	ctxAddress
 )
 
 // BlockHashPathAndIndexCtx is a middleware that embeds the value at the url
@@ -50,6 +58,16 @@ func (c *insightApiContext) getBlockHashCtx(r *http.Request) string {
 		}
 	}
 	return hash
+}
+
+// AddressPathCtx returns a http.HandlerFunc that embeds the value at the url
+// part {address} into the request context.
+func (c *insightApiContext) AddressPathCtx(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		address := chi.URLParam(r, "address")
+		ctx := context.WithValue(r.Context(), ctxAddress, address)
+		next.ServeHTTP(w, r.WithContext(ctx))
+	})
 }
 
 // GetRawHexTx retrieves the ctxRawHexTx data from the request context. If not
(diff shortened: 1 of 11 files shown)


api/insight/apimiddleware.go
```diff
@@ -91,3 +109,161 @@ func (c *insightApiContext) PostBroadcastTxCtx(next http.Handler) http.Handler {
...
+		next.ServeHTTP(w, r.WithContext(ctx))
+	})
+}
+
+// Process params given in post body for an addrs endpoint
+func (c *insightApiContext) PostAddrsTxsCtx(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		var req apitypes.InsightMultiAddrsTx
+		var from, to, noAsm, noScriptSig, noSpent int64
+		body, err := ioutil.ReadAll(r.Body)
```
chappjc: Usually we'd first check the request's method (ensure POST), the Content-Type, and any other headers we expect to get.  For example, here's a little file upload/download web app I put together recently:

https://github.com/chappjc/webfiles/blob/c5e702e2b6311a340722f637b8f98933fc63c8ba/server/server.go#L206

api/insight/apimiddleware.go
```diff
@@ -91,3 +109,161 @@ func (c *insightApiContext) PostBroadcastTxCtx(next http.Handler) http.Handler {
...
+		r.Body.Close()
+		if err != nil {
+			apiLog.Errorf("error reading JSON message: %v", err)
+			writeInsightError(w, fmt.Sprintf("error reading JSON message: %v", err))
+			return
+		}
+		err = json.Unmarshal(body, &req)
```
chappjc: Do you have an example of the type of request (header and body) that we should expect for this?

api/insight/apimiddleware.go
```diff
@@ -91,3 +109,161 @@ func (c *insightApiContext) PostBroadcastTxCtx(next http.Handler) http.Handler {
...
+			writeInsightError(w, fmt.Sprintf("Failed to parse request: %v", err))
+			return
+		}
+		// Successful extraction of Body JSON
+		ctx := context.WithValue(r.Context(), ctxAddress, req.Addresses)
+
+		if req.From != "" {
+			from, err = req.From.Int64()
+			if err != nil {
+				apiLog.Errorf("Error converting 'from' to Int.  Expecting Integer.")
```
chappjc: Since these depend on input from the client, it's trivial to fill up the logs with this.  Please make this Debugf, or perhaps check the error itself to only make an Errorf/Warnf in certain circumstances that indicate a server error rather than a bad request.

api/insight/apimiddleware.go
```diff
@@ -91,3 +109,161 @@ func (c *insightApiContext) PostBroadcastTxCtx(next http.Handler) http.Handler {
...
+			} else {
+				ctx = context.WithValue(ctx, ctxTo, int(to))
+			}
+		}
+		if req.NoAsm != "" {
+			noAsm, err = req.NoAsm.Int64()
+			if err != nil {
+				apiLog.Errorf("Error converting 'noAsm' to Int.  Expecting 0 or 1.")
+			} else {
+				if noAsm == 0 {
```
chappjc: So if `req.NoAsm == ""`, then `ctxNoAsm` should not be set at all?

api/insight/apiroutes.go
```diff
@@ -287,27 +310,136 @@ func (c *insightApiContext) getTransactions(w http.ResponseWriter, r *http.Reque
...
+	address := c.GetAddressCtx(r)         // Required
+	noAsm := c.GetNoAsmCtx(r)             // Optional
+	noScriptSig := c.GetNoScriptSigCtx(r) // Optional
+	noSpent := c.GetNoSpentCtx(r)         // Optional
+	from, ok := c.GetFromCtx(r)           // Optional
 
-	if address == "" {
-		http.Error(w, http.StatusText(422), 422)
-		return
+	if !ok {
```
chappjc: You're setting the default to 0 already in `GetFromCtx`

api/insight/apiroutes.go
```diff
@@ -287,27 +310,136 @@ func (c *insightApiContext) getTransactions(w http.ResponseWriter, r *http.Reque
...
+
+	// If no address present we need to stop.  Allow Addresses to be single or
+	// multiple separated by a comma.
+	if address != "" {
+		if strings.Contains(address, ",") {
+			addresses = strings.Split(address, ",")
+		} else {
+			addresses = []string{address}
+		}
+	} else {
```
chappjc: I suggest to check this right after `GetAddressCtx`, then down here you don't need any checks, just `addresses := strings.Split(address, ",")`

api/insight/apimiddleware.go
```diff
@@ -52,6 +60,16 @@ func (c *insightApiContext) getBlockHashCtx(r *http.Request) string {
 	return hash
 }
 
+// AddressPathCtx returns a http.HandlerFunc that embeds the value at the url
+// part {address} into the request context.
+func (c *insightApiContext) AddressPathCtx(next http.Handler) http.Handler {
```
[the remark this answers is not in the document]
papacarp: Thank you for the review.  Whenever I tweaked or threatened to tweak one of the middleware functions I just duplicated it to avoid possible bugs introduced into the other API.   Not a problem though, I just removed it and switched back to the m. context one.

api/insight/apimiddleware.go
papacarp: I just copied the way I saw dcrd do it which I don't think checks that, but I'll add in content checking as an error response along those lines would have saved me hours when I was trying to figure out how insight post params work.

api/insight/apimiddleware.go
papacarp: should be POST and "application/json"

```go
package main

import (
	"fmt"
	"strings"
	"net/http"
	"io/ioutil"
)

func main() {

	url := "http://127.0.0.1:7777/insight/api/addrs/utxo"

	payload := strings.NewReader("{\"addrs\":\"Dca7Vsv42RAJC6cEdw7dyhLER8QJCHiUYNL,DsTHdW8erZsD1Fqt8uCfapVR1Qsf9KaMwBN\"}")

	req, _ := http.NewRequest("POST", url, payload)

	req.Header.Add("Content-Type", "application/json")
	req.Header.Add("Cache-Control", "no-cache")

	res, _ := http.DefaultClient.Do(req)
}
```

api/insight/apimiddleware.go
chappjc: This example is file upload, so it's almost certainly a different content type, but by checking before reading the body we'll potentially avoid transferring useless data

api/insight/apimiddleware.go
chappjc: Awesome thanks.  How about content-length or content-encoding.  I imagine we want to ensure content-encoding is not chunked, and with content-length set, set to something we threshold

api/insight/apimiddleware.go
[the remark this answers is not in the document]
papacarp: One more issue.  ctxAddress is also set with with:
```go
func (c *insightApiContext) PostAddrsTxsCtx(next http.Handler) http.Handler {
```
So it looks like I have to find a way to export ctxAddress from the other context into the insight one.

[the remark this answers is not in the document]
chappjc: I might be missing the issue, but I think you would use `PostAddrsTxsCtx` from middleware instead of insight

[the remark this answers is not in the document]
chappjc: oh, nvm, middleware has `AddressPathCtx` and `GetAddressCtx`, but not `PostAddrsTxsCtx`.  I see...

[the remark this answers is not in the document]
chappjc: How about changing `PostAddrsTxsCtx` to set `m.CtxAddress` instead of `ctxAddress`.  Then you don't need an insight version of AddressPathCtx and GetAddressCtx

api/insight/apimiddleware.go
papacarp: The get procedure will set it to false if its not set at all.  When we Unmarshall the req not set and set to "" are the same thing.  I will streamline this section accordingly as that set false is not required since its assumed in the get call.

api/insight/apimiddleware.go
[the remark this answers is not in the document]
papacarp: Worked perfectly!   BTW, we could probably get rid of `GetRawHexTx` this same way if you want (different PR?)

api/insight/apimiddleware.go
papacarp: Ok, I checked the content-length.    Transfer encoding?

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Transfer-Encoding

api/insight/apimiddleware.go
papacarp: I removed all logging references that were related to user data.  Agreed, not useful unless something is happening that should never happen.

api/insight/apimiddleware.go
papacarp: chi mux has middleware to check content type so I added that in.

papacarp:
Everything is addressed except content-encoding.  Also when I was poking around in the headers I noticed existing bitpay insight compresses the response, as does mainnet, so I compressed our api responses as well.

chappjc:
Good stuff.  I didn't know about the content type validator middleware in chi.  I'll get this up on beta site soon so we can abuse it

PR #496, opened 2018-06-06
Update configuration file description

lte13:
After you are finished with changing the default, you have to move dcrdata.conf to the ./dcrdata folder to start otherwise it gives you an error when you start dcrdata because ./dcrdata is missing the dcrdata.conf. Maybe change this?

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,7 +112,9 @@ Begin with the sample configuration file:
 cp sample-dcrdata.conf dcrdata.conf
 ```
 
-Then edit dcrdata.conf with your dcrd RPC settings. See the output of
+Then edit dcrdata.conf with your dcrd RPC settings. After you are finished, move dcrdata.conf to ./dcrdata.
+
+See the output of
 `dcrdata --help` for a list of all options and their default values.
 
 ### Indexing the Blockchain


chappjc:
The current version looks for dcrdata in ~/.dcrdata/dcrdata.conf.  Please see the `--appdata` option .

chappjc:
Didn't mean to close.  This is a good addition, but it needs to refer to the right location.

chappjc:
Yeah, that's better thanks.  It's a bit different on Windows (%LOCALAPPDATA%\dcrdata) and Mac too, but this is a good correction.

README.md
```diff
@@ -112,7 +112,9 @@ Begin with the sample configuration file:
 cp sample-dcrdata.conf dcrdata.conf
 ```
 
-Then edit dcrdata.conf with your dcrd RPC settings. See the output of
+Then edit dcrdata.conf with your dcrd RPC settings. After you are finished, move dcrdata.conf to the ~/.dcrdata/dcrdata.conf folder.
```
chappjc: Actually the wording is a bit funny as it refers to `~/.dcrdata/dcrdata.conf` as a folder.  Also re-wrap these 3 sentences to it's a bit more nicely formatted.  Thanks for the PR!

Issue #444, opened 2018-04-26
Consider stimulus.js for terser, more organized client side javascript 


gozart1:
It's likely this explorer will continue to get fancier with client side interactions such as charts or table sorts, but at the same time will most likely continue to be based on server rendered html at its core.

So might be good to add a bit of structure to our javascript with Stimulus, a framework that's designed for the turbolinks architecture.  https://stimulusjs.org/handbook/origin

What do ya'll think?

chappjc:
> Think of it like this: in the same way that class is a bridge connecting HTML to CSS, data-controller is a bridge from HTML to JavaScript.

Sounds like a cool way to create a little order out of our growing mess of js.

Thumbs up.

raedah:
Will this create any changes visible to the user, or will it improve code readability?

chappjc:
resolved by PR https://github.com/decred/dcrdata/pull/498 (for home and mempool).  @gozart1 Think this is worth more refactoring?  Seems like it might be.

PR #498, opened 2018-06-07
refactor home page and mempool javascript with stimulus (WIP)

gozart1:
This PR explores structuring client side javascript with stimulus.js as mentioned in issue https://github.com/decred/dcrdata/issues/444. 

Our current approach to javascript makes it tough to follow the program logic as it forces someone unfamiliar with the codebase to do their own detective work to hopefully associate some id on an html element with some javascript code that lives somewhere else in the codebase. It worked great in the 90's... but today there are better approaches available to us.

Given the server rendered templating backbone of our application, we just need something lightweight that makes the javascript more declarative. So that you can simply look at a given chunk of template markup and know exactly what javascript is interacting with it.

So I'm reworking the homepage and mempool page with stimulus since those are the most JS heavy at the moment and should put stimulus through its paces to see if it makes sense to standardize on.


gozart1:
Ok, this should be good to go. There shouldn't be any visible changes other than:
- Newly added rows now have a brief flash animation of their background color
- Tables now get rendered on pageload even if empty ( eg 0 revocations ). Otherwise when a revocation comes in in via websocket, there is no place to put it and the end users never gets to see the new row. So now there are always tables, with empty messages when no transactions.
screen shot 2018-06-15 at 1 56 51 am

PR #499, opened 2018-06-07
accelerate StakeDatabase.ConnectBlock

chappjc:
This changes how StakeDatabase gets the live ticket pool and the diffs (in/out) for each block in ConnectBlock.

The `liveTicketCache` in `StakeDatabase` is now a faithful representation of the live ticket pool at the current best block.  So when the live ticket list is needed, it is no longer necessary to use the `stake.Node` (`BestNode.LiveTickets()`).

Add `poolValue int64` to `StakeDatabase`, updating it on connect/disconnect via += / -+ for the pool in/out tickets.

A major optimization in `ConnectBlock` is using the new `ExpiredByBlock` function in dcrd/stake instead of the much more costly `ExistsExpiredTicket` and `ExistsRevokedTicket`.  This requires dcrd PR https://github.com/decred/dcrd/pull/1221.

 Improve the process of "Pre-populating live ticket cache..." by prefetching the tickets from the stake `BestNode` (dcrdata's best node) rather than asking dcrd for it's idea of the live ticket pool.  This ensures `liveTicketCache` will always contain an accurate ticket pool.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -125,7 +125,7 @@
     "txscript",
     "wire"
   ]
-  revision = "7a457fc9c32a415604aba12a6aedd96dfc3f470e"
+  revision = "5386a6375874cbbc001ee915c887a8240933751b"
 
 [[projects]]
   branch = "master"
--- a/stakedb/stakedb.go
+++ b/stakedb/stakedb.go
@@ -93,8 +93,9 @@ type StakeDatabase struct {
 	BestNode        *stake.Node
 	blkMtx          sync.RWMutex
 	blockCache      map[int64]*dcrutil.Block
-	liveTicketMtx   sync.Mutex
+	liveTicketMtx   sync.RWMutex
 	liveTicketCache map[chainhash.Hash]int64
+	poolValue       int64
 	poolInfo        *PoolInfoCache
 	PoolDB          *TicketPool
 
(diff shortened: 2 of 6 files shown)


chappjc:
Tested with `rebuilddb` (for building just the sqlite and stakedb) and `dcrdata --pg` for the whole shebang.

Full sync with `dcrdata --pg` completed in 2 hrs, 2 minutes:

    PSQL: Avg. speed: 597 tx/s, 2206 vout/s

Not including indexing, which is unchanged, and takes another 45 minutes.

Typical new block:

    Synchronous handlers of collectionQueue.ProcessBlocks() completed in 37.720347ms

NOTE:  To isolate the computational (non-io) bottlenecks, I came up with a trick that is very helpful for the initial sync:

1. Before starting initial sync, mount a `tmpfs` (memory-backed) file system or similar RAM disk.  Call it /tmp.
2. `mkdir /tmp/mainnet` (for mainnet)
3. `ln -s /tmp/mainnet ~/.dcrdata/data/mainnet`
4. Do the initial sync with dcrdata, allowing it to put all the sqlite, ticket pool, and stakedb files in /tmp/mainnet.
5. After sync, stop dcrdata, then: `unlink ~/.dcrdata/data/mainnet`.
6. `mv /tmp/mainnet ~/.dcrdata/data/`

chappjc:
Initial sync in lite mode is now blazing fast:
- This PR: **28 minutes** from genesis to block 246152.
- master: **52 minutes** from genesis to block 246164.

chappjc:
dcrd PR https://github.com/decred/dcrd/pull/1221 is in, so this can be tested more easily now.

chappjc:
I'm putting this in ahead of other older PRs because this will help speed up everyone's work, especially the work involving DB schema changes.

Issue #504, opened 2018-06-12
Prevent update dev fund balance from launching more than once


chappjc:
The update is expensive.  If it tries to update the balance at a bad time, like a reorg, or tries to update while it's already updating, the process grinds to a halt.

chappjc:
Closed by PR #505 https://github.com/decred/dcrdata/pull/505

PR #505, opened 2018-06-12
dcrpg: prevent stacking multiple calls to updateDevFund

chappjc:
Addressing issue https://github.com/decred/dcrdata/issues/504


chappjc:
Been running ok on beta.

PR #510, opened 2018-06-20
api: add /txs route for batch request

chappjc:
This adds the /txs and /txs/trimmed routes for batch requesting transaction details.  The transaction hashes are transmitted in the http request body as JSON.  The structure is `api/types.Txns`. The request
Content-Type must be "application/json".
The return type for /txs is `[]*types.Tx`.
The return type for /txs/trimmed is `[]*types.TrimmedTx`.

version: bump to 2.0.1


chappjc:
Supports politeia PR https://github.com/decred/politeia/pull/337

PR #520, opened 2018-06-28
html/tx: unconfirmed tickets/votes have no maturity status

chappjc:
(db *wiredDB) GetExplorerTx: set Mature="false" when confirmations==0
tx.tmpl: Unconfirmed tickets and votes read "N/A"  for MATURITY, must be
confirmed to show progress meter.


chappjc:
Tested in lite mode.  Pending test in full mode.

PR #521, opened 2018-06-28
rebuilddb2 batch sync flag

chappjc:
This prevents each connected block from updating the dev fund balance.  The dcrdata app already does this, but rebuilddb2 was not updated.


chappjc:
Tested.

PR #527, opened 2018-07-01
stakedb: recovery and robustness with upserts

chappjc:
This adds some automatic recovery functions.

- Add `stakedb.LoadAndRecover`, which attempts recovery of the `StakeDatabase` and its `TicketPool`.
- `NewStakeDatabase` returns the height where it stopped loading. If there is an error, the height may be passed to `LoadAndRecover`.
- `InsertVin[s]` now has an upsert option (dupcheck=true) to avoid failures requiring recovery procedures.
- Also have dupcheck option with `SetSpendingForFundingOP` / `insertSpendingTxByPrptStmt` and `SetSpendingByVinID`.
- Add `MakeAddressRowInsertStatement` to create the upsert statement if checkup=true;
- `InsertVotes` and `MakeVoteInsertStatement` similarly take checkdups option.
- Fix incorrect parameter numbers in `upsertVoteRow` and `insertVoteRowReturnId`.
- `AdvanceToTip` now returns the last valid diff index (cursor - 1).

### DIFF
--- a/stakedb/stakedb.go
+++ b/stakedb/stakedb.go
@@ -116,20 +116,24 @@ const (
 	DefaultTicketPoolDbName = "ticket_pool.db"
 )
 
-// NewStakeDatabase creates a StakeDatabase instance, opening or creating a new
-// ffldb-backed stake database, and loads all live tickets into a cache.
-func NewStakeDatabase(client *rpcclient.Client, params *chaincfg.Params,
-	dataDir string) (*StakeDatabase, error) {
-	// Create DB folder
-	err := os.MkdirAll(dataDir, 0700)
-	if err != nil {
-		return nil, fmt.Errorf("unable to create DB folder: %v", err)
+// LoadAndRecover attempts to load the StakeDatabase and it's TicketPool,
+// rewinding either TicketPool or StakeDatabase so that they are at the same
+// height, and then further rewinding both to the specified height. Finally, it
+// advances the TicketPool to tip, and if there is an error it rewinds both back
+// to that height - 1.  Normally use NewStakeDatabase.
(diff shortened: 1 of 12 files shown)


chappjc:
@dmigwi Would you scan through this when you have a minute?

dmigwi:
Sure.

stakedb/stakedb.go
```diff
@@ -146,7 +251,8 @@ func NewStakeDatabase(client *rpcclient.Client, params *chaincfg.Params,
 
 	stakeDBPath := filepath.Join(dataDir, DefaultStakeDbName)
 	if err = sDB.Open(stakeDBPath); err != nil {
-		return nil, err
+		_ = poolDB.Close()
```
dmigwi: If you are not capture the error here `_ = poolDB.Close()` looks like something is missing. I think it would look neater just as `poolDB.Close()`

stakedb/stakedb.go
```diff
@@ -155,13 +261,15 @@ func NewStakeDatabase(client *rpcclient.Client, params *chaincfg.Params,
 	if heightStakeDB != heightTicketPool {
 		if heightStakeDB-heightTicketPool > 16 {
 			log.Warnf("The ticket DB format has changed in v2.1. Did you upgrade?")
-			return nil, fmt.Errorf("Remove %s and start dcrdata again.",
+			_ = sDB.Close()
```
dmigwi: same here

stakedb/stakedb.go
chappjc: It sort of says "I've not forgotten the error, I just don't care."  So I've added a comment on each one to document this.  It also shuts up linters that look for unhandled errors.

PR #533, opened 2018-07-02
dcrsqlite: fix RetrievePoolInfo empty result

### LINKED ISSUE #532: RetrievePoolInfo, split call on empty result
https://github.com/decred/dcrdata/blob/master/db/dcrsqlite/sqlite.go#L389
```go
// RetrievePoolInfo returns ticket pool info for block height ind
func (db *DB) RetrievePoolInfo(ind int64) (*apitypes.TicketPoolInfo, error) {
    tpi := &apitypes.TicketPoolInfo{
        Height: uint32(ind),
    }
    var hash, winners string
    err := db.QueryRow(db.getPoolSQL, ind).Scan(&hash, &tpi.Size,
        &tpi.Value, &tpi.ValAvg, &winners)
    tpi.Winners = strings.Split(winners, ";")
    return tpi, err
}
```

``` tpi.Winners = strings.Split(winners, ";")``` returns array that contains one element (empty string)

Should probably return empty array;

### PR DESCRIPTION
**Pull request by JFixby**
Ensures `tpi.Winners` is an empty array.
Fixes #532

### DIFF
--- a/db/dcrsqlite/sqlite.go
+++ b/db/dcrsqlite/sqlite.go
@@ -386,7 +386,11 @@ func (db *DB) RetrievePoolInfo(ind int64) (*apitypes.TicketPoolInfo, error) {
 	var hash, winners string
 	err := db.QueryRow(db.getPoolSQL, ind).Scan(&hash, &tpi.Size,
 		&tpi.Value, &tpi.ValAvg, &winners)
-	tpi.Winners = strings.Split(winners, ";")
+	if winners == "" {
+		tpi.Winners = make([]string, 0)
+	} else {
+		tpi.Winners = strings.Split(winners, ";")
+	}
 	return tpi, err
 }
 

### CODE REVIEW

File: db/dcrsqlite/sqlite.go [hunk 1]
```diff
@@ -386,7 +386,11 @@ func (db *DB) RetrievePoolInfo(ind int64) (*apitypes.TicketPoolInfo, error) {
 	var hash, winners string
 	err := db.QueryRow(db.getPoolSQL, ind).Scan(&hash, &tpi.Size,
 		&tpi.Value, &tpi.ValAvg, &winners)
-	tpi.Winners = strings.Split(winners, ";")
+	if winners == "" {
+		tpi.Winners = make([]string, 0)
+	} else {
```
chappjc: The first part is already happening when `tpi` is initialized.  You just need `if winners != "" { tpi.Winners = ... }

chappjc: oh, nvm, I forgot what we talked about earlier.  nil slice is no good, we want zero length non-nil like you did.

PR #534, opened 2018-07-02
dcrsqlite: remove dead code and swap wrongly positioned args

JFixby:


db/dcrsqlite/sqlite.go
```diff
@@ -703,13 +690,13 @@ func (db *DB) RetrieveBlockSizeRange(ind0, ind1 int64) ([]int32, error) {
 	}
 	if N < 0 {
 		return nil, fmt.Errorf("Cannot retrieve block size range (%d<%d)",
-			ind1, ind0)
+			ind0, ind1)
```
chappjc: This one was saying why, so the message here was right.

JFixby: How about to change it to the `"Cannot retrieve block size range [%d,%d]" `?

JFixby:
Found and fixed 3 more cases.
Also switched the `<` sign

db/dcrsqlite/sqlite.go
chappjc: Sure, that's fine.

PR #535, opened 2018-07-03
[WIP] Graphical Block Display

freethinkingaway:
Work for #453. Runs on localhost:7777/nexthome.

screen shot 2018-07-02 at 10 23 32 pm

For the Non-Mempool Blocks (Recent Blocks):

First Row: Votes.
Green blocks represent Valid votes, Red if Not Valid.

Second Row: Ticket & Revocation transactions.
Tickets are grey. Rev's are yellow.
Fee rate (DCR/kb) determines light/darkness of the buttons as follows:

< 0.001 = "lowest"/lightest
0.001 - 0.01 = "low"/light
0.01 - 0.1 = "moderate"/medium
0.1 - 0.5 = "moderate-high"/medium-dark
0.5 - 1.0 = "high"/dark
1.0 < = "very-high"/darkest

Third Row: Regular non-Coinbase transactions.
Width of the slot represents TxAmount (DCR) as follows:

< 50 = "xs"
50 - 100 = "s"
100 - 200 = "m"
200 - 500 = "l"
500 - 1000 = "xl"
1000 < = "xxl"

Height of the slot represents ByteSize (TxSize) as follows:

< 500 B = "xs-size"
500 - 1KB = "s-size"
1KB - 2KB = "m-size"
2KB - 10KB = "l-size"
10KB - 50KB = "xl-size"
50KB < = "xxl-size"

Note: The websocket is configured to pull in and repopulate the Recent Blocks but the blocks currently overflow. As is, the most recent block gets pulled in, old blocks are not removed, and the recent blocks eventually overflow the page. I tried a number of ways to remedy this, but haven't quite cracked the issue.

The Mempool Block:

The websocket is configured to represent every incoming transaction (including the coinbase) to be placed into the next block, however the websocket processes/displays the transactions as they come in and does not currently display them in order or neatly. I developed this using simnet, which even while generating a block at a time, is difficult to accurately represent the live network. Perhaps it won't be as difficult to reorder and display appropriately using the testnet and/or mainnet.

Work To Be Done:

Properly configuring the websocket to display the Mempool in a better displayed/ordered format.
Properly configuring the websocket to display and repopulate the Recent Blocks, remove old blocks as new ones come in.


raedah:
@freethinkingaway has gotten this component off the ground. He is away currently, and someone else can jump in to move this forward.

denyszayets:
1366x800 home alt1 2x

gozart1:
This latest design doesn't pass the "squint test".
By that I mean if I squint my eyes, the primary element I see is the 5 gray bars lining the top of each block. The purpose of the squinting at the design is to identify if the visual weight of the overall composition, you can squint at a good design and get still information from the broad strokes that remain perceivable.

Squinting also helps you find areas where contrast is too low. For example, in this design, the block height and age become invisible when squinting. I think both of these elements too important 
 to de-prioritize like this because they help you get your bearings when you are comparing your own view of the chain with the rest of the world, which is a common use case for a block explorer. 

The icons have no meaning to me. The 2nd own looks like a ticket being submitted into a ballot box to me, which I don't think is what it's supposed to represent.

Adding tooltips to explain them is a design smell to me as it is not very mobile friendly, and in general indicates a failure to make something intuitive on its own. 

I'm not necessarily saying the current design is better in every way, for example if you squint at the current home page the progress bars have too much visual importance. 

I think that we can do better than both versions.

dmigwi:
Closing this since its Superseded by #660

raedah:
for historical ref, technically superceded by
https://github.com/decred/dcrdata/pull/689

PR #536, opened 2018-07-03
Streaming txs endpoint

chappjc:
This makes the HTTP response to the /txs endpoints streaming (i.e. writes each transaction to the response writer as they arrive rather than all at once after they are all fetched).


chappjc:
not necessary

PR #542, opened 2018-07-06
Current Ticket Pool Visualization

dmigwi:
This PR introduces a page that shows real time data of what is happening to the live, immature and mempool tickets.

screen shot 2018-07-07 at 02 44 24


dmigwi:
Possible ways to view the `Ticket Pool Tickets by Purchase Date` graph
graphs

chappjc:
I really like the adjustable binning

xaur:
Ideas for discussion:

* Zoom bar: "all blocks" is inconsistent with human periods like "day" or "week", I suggest "all time"
* first graph:
  * title: Ticket Purchase History
  * add another Y axis with ticket price and add a line plot
  * add label to X axis like in second graph, note that time is UTC
  * swap colors, use blue for live and green for immature ("dat ticket is still so green, so immature")
* second graph:
  * title: Ticket Price Distribution
  * keep showing live tickets as blue
  * add immature tickets in green stacked on top of blue live tickets, so some bars will have two segments (blue bottom, green top)
* both graphs:
  * reduce resolution to have better visible (more fat) bars?
* idea for another graph:
  * title: Ticket Fee Distribution
  * X axis: ticket fee, Y axis: number of tickets
  * we had that on the now-dead posmaster.info
  * not so relevant since sdiff algo is fixed but still interesting sometimes (when whales start their buys)

dmigwi:
@xaur nice Ideas, I will work on them.  
@raedah The page barely fits 3 graphs do I replace the `Ticket Fee Distribution` chart with the pie chart (doughnut) or how should I go about adding the `Ticket Fee Distribution` chart ?

dmigwi:
>reduce resolution to have better visible (more fat) bars?

For the first graph this is already in place with the adjustable data binning (Zoom and Interval group buttons). For the second chart, the bar width will be hard to alter unless we change the chart itself.

We have another PR  #463 in review and it has a Ticket Price Chart. lts not real time though but it covers all the ticket purchased since block 257

xaur:
On the second pic (green charts), the zoom buttons are better but still inconsistent:

* Zoom and Interval buttons have opposite sorting
* 1w is pretty widespread, no need for 1wk

raedah:
Ticket fee distribution is less interesting with the Live Pool view, and much better correlated with the historical view charts.

xaur:
@raedah agreed, took me a while to realize all charts on this page apply to live tickets only. Although, in some rare situations (when whales start buying tickets), realtime visualization of tickets and ticket fees in the _mempool_ is useful.

dmigwi:
>Zoom and Interval buttons have opposite sorting

Zoom alters the range slide bar only and Interval alters that actual range used to data fetched from the backend. i.e. if zoom is set to `wk` only a time span of the last one week appears on the graph and  if interval is `wk` tickets within one week are plotted in a single bar.
>1w is pretty widespread, no need for 1wk

This is just one of the option/mode that the first graph can be viewed in. There more than 10 possible ways/mode in which the first graph can be made to look different.

xaur:
> if zoom is set to wk only a time span of the last one week appears on the graph and if interval is wk tickets within one week are plotted in a single bar.

Seems the latter is what @chappjc called "adjustable binning". I would call it "bar size" or so, "Zoom" and "Interval" together confused me (or it's just me?). For example, on [Poloniex](https://poloniex.com/exchange#btc_dcr) the first control is also called "Zoom" but second called "Candlesticks" since it controls their size. So I'd call the second control "Bars" or "Bar size".

xaur:
Now that I realized all charts only show live tickets the titles I proposed are confusing and the original titles look better. Perhaps the only thing I'd improve is replace "Ticket Pool" with "Live Ticket". So to outline the options:

* naming scheme: Live Tickets by <parameter> vs Live Ticket <parameter> Distribution
* 1st graph: Live Tickets by Purchase Date vs Live Ticket Purchase Date Distribution
* 2nd graph: Live Tickets by Ticket Price vs Live Ticket Price Distribution

@raedah to reduce confusion I imagine adding an info box on top with e.g. "These charts only show show live tickets in the ticket pool, for historic views go <here>", and adding "Purchase date" label to the X axis of first graph.

Also, per Slack discussion, another way to look at first chart is reverse the X axis and call it "Live Tickets by Age".

xaur:
@dmigwi you were right that if you show live, immature and mempool tickets it is incorrect to put "live" in chart titles. Therefore remove "live" from page description (subheading or subtitle) as well:

> These charts show only tickets in the ticket pool, for historic views go to {link}

dmigwi:
Updated ticketpool page
screen shot 2018-07-10 at 02 43 32

dmigwi:
Updated Tickets purchase distribution graph
new-graphs

dmigwi:
@chappjc When you get the time help me review and test this PR.

dmigwi:
@gozart1 I will appreciate a lot If you could help me review and test this PR at your own time.

chappjc:
Will review and test today.

explorer/websockethandlers.go
```diff
@@ -59,55 +60,100 @@ func (exp *explorerUI) RootWebsocket(w http.ResponseWriter, r *http.Request) {
...
+					exp.MempoolData.Unlock()
+
+					var mp = dbtypes.PoolTicketsData{}
+
+					if len(mpData.Tickets) > 0 {
+						mp.Time = append(mp.Time, uint64(mpData.Tickets[0].Time))
+						mp.Price = append(mp.Price, mpData.Tickets[0].TotalOut)
+						mp.Mempool = append(mp.Mempool, uint64(len(mpData.Tickets)))
+					} else {
+						log.Warn("No tickets exists in the mempool")
```
chappjc: This is not unexpected, so `log.Debug` (or `Trace`).

wording: `"No tickets exist in mempool"`

explorer/websockethandlers.go
```diff
@@ -59,55 +60,100 @@ func (exp *explorerUI) RootWebsocket(w http.ResponseWriter, r *http.Request) {
...
+
+				case "getticketpooldata":
+					// Ensures that multiple requests for the same update are
+					// treated as the single request.
+					if !tpEvent.Active {
+						log.Warn("ticketpool event has not updated data.")
+						break
+					}
+
+					tpEvent.Active = false
```
chappjc: this an previous access are racing with the send loop.

explorer/websockethandlers.go
```diff
@@ -166,6 +214,11 @@ func (exp *explorerUI) RootWebsocket(w http.ResponseWriter, r *http.Request) {
 						Extra: exp.ExtraInfo,
 					})
 					exp.NewBlockDataMtx.Unlock()
+
+					tpEvent.RLock()
+					tpEvent.Active = true
```
chappjc: Would need a read-write lock (`Lock` instead of `RLock`)

explorer/websockethandlers.go
```diff
@@ -59,55 +60,100 @@ func (exp *explorerUI) RootWebsocket(w http.ResponseWriter, r *http.Request) {
 
 				// handle received message according to event ID
 				var webData WebSocketMessage
+				//  If the request sent is past the limit exit the loop
+				if len(msg.Message) > requestLimit {
+					log.Debug("Request size over limit")
+					webData.Message = "Request too large"
+					return
```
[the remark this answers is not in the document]
chappjc: But nothing a redundant check is OK, BTW!  Just wanted to point out how `Receive` should behave in this case

explorer/websockethandlers.go
dmigwi: ok. Noted

explorer/websockethandlers.go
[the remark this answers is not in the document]
dmigwi: ok. Noted

dmigwi:
@chappjc I have added a new commit with the updated ticket purchases distribution graph intervals.

dmigwi:
I am waiting for #580 to go in so that I can make changes on top of it.

chappjc:
Noted.  A mainnet reorg finally hit yesterday and I discovered some bugs.  Got them sorted out, and can continue to test tomorrow.  Looks pretty close

chappjc:
How's it going solving the slow queries we talked about?

dmigwi:
Since your PR is almost ready I can give it test and see if things we be better.

chappjc:
Ok to rebase

chappjc:
Slated for 3.1.  Since 3.0 is stabilizing, I'd like to make a 3.0 branch and have it released soon.

chappjc:
@dmigwi How's it going with the performance issues on the queries?  I can look at this with you again if you like.  We can save the concurrency issues until the performance is tolerable.

dmigwi:
@chappjc I am planning to now focus on this issue since the histogram task is done (pretty much). I will be testing all those performance related issues and will give you a decisive answer maybe tomorrow.

dmigwi:
Just seen that tickets table has no `block_time`field but it has a `block_height` field both maybe used interchangeably but I am not sure replacing the time label on the UI with block height is the right call or just implementing a data migration to maintain the current UI is the way to go.

@chappjc and @raedah Any ideas?

dmigwi:
To respond to the concern above, it turns out that this control button complicate things and therefore we shall still have to use the block time as the x-axis for the first graph.

screen shot 2018-08-18 at 19 49 57

explorer/websockethandlers.go
dmigwi: If we use `Lock` shall we be able to activate the tpEvent?

chappjc: The mutex doesn't actually prevent us or any other goroutine from reading or writing to something.  It's only there to *allow* you to set when the data is being read or written.  In this case you want to write, so you use Lock (a R/W lock).  Where you only want to read, your can use `RLock`, the idea being you can have any number of concurrent readers, but only one writer.

I suggest to go through a tutorial on mutexs, such as https://gobyexample.com/mutexes.

But I can look again at this PR and we can resume this concurrency discussion if you think the queries are in a good enough shape.

dmigwi:
@chappjc I have managed to significantly improve the ticket pool queries response time
```
2018/08/29 18:40:12 "GET http://localhost:7777/api/ticketpool/bydate/day HTTP/1.1" from 127.0.0.1:63233 - 200 4117B in 86.960801ms
2018/08/29 18:40:16 "GET http://localhost:7777/api/ticketpool/bydate/wk HTTP/1.1" from 127.0.0.1:63233 - 200 662B in 80.044091ms
2018/08/29 18:40:17 "GET http://localhost:7777/api/ticketpool/bydate/mo HTTP/1.1" from 127.0.0.1:63233 - 200 196B in 74.008159ms
2018/08/29 18:40:19 "GET http://localhost:7777/api/ticketpool/bydate/wk HTTP/1.1" from 127.0.0.1:63233 - 200 662B in 76.488268ms
2018/08/29 18:40:21 "GET http://localhost:7777/api/ticketpool/bydate/day HTTP/1.1" from 127.0.0.1:63233 - 200 4117B in 79.080094ms
2018/08/29 18:40:33 "GET http://localhost:7777/api/ticketpool/bydate/all HTTP/1.1" from 127.0.0.1:63236 - 200 448085B in 159.404955ms
```

chappjc:
I'd say so.  :)  Let me look again at the problem we're trying to solve with this locking.

explorer/websockethandlers.go
dmigwi: I think they are in good shape. Actually I've just seen that `Lock` makes sense

chappjc:
um, I meant to comment on this PR... I'm hacking on this one.

dmigwi:
I think I should embrace the same I idea of not deleting comments that look misplaced in a given context. And rather explain my earlier thought process.

db/dcrpg/queries.go
```diff
@@ -1354,6 +1354,76 @@ func RetrieveAddressTxnOutputWithTransaction(db *sql.DB, address string, current
...
+		if err != nil {
+			return nil, fmt.Errorf("retrieveTicketsByDate %v", err)
+		}
+
+		tickets.Time = append(tickets.Time, timestamp)
+		tickets.Immature = append(tickets.Immature, immature)
+		tickets.Live = append(tickets.Live, live)
+
+		// Returns the average value of a ticket depending on the grouping mode used
+		price = price * 100000000
```
chappjc: Why convert from coin to atoms, then back to coin?  But in general, when converting either way, use dcrutil.  In this case it would have been dcrutil.NewAmount.

db/dcrpg/queries.go
```diff
@@ -1455,6 +1524,32 @@ func RetrieveBlockSummaryByTimeRange(db *sql.DB, minTime, maxTime int64, limit i
 	return blocks, nil
 }
 
+// retrieveTicketsTxsBlockTime fetches the block time for all the tickets
+// transactions in the transactions table. The Tx struct from this function
+// only contains the block height and the block time.
+func retrieveTicketsTxsBlockTime(db *sql.DB) ([]*dbtypes.Tx, error) {
+	var txsData = make([]*dbtypes.Tx, 0)
```
chappjc: Just `var txsData []*dbtypes.Tx` please, unless the caller really needs a non-nil zero-length slice for some reason.

db/dcrpg/upgrades.go
```diff
@@ -486,6 +516,48 @@ func (pgb *ChainDB) handleVinsTableMainchainupgrade() (int64, error) {
 	return rowsUpdated, nil
 }
 
+func (pgb *ChainDB) handleTicketsBlockTimeUpgrade() (int64, error) {
+	log.Info(" - Retrieving all the ticket transactions with blocktime")
```
chappjc: ` - Retrieving block times for all ticket transactions.`

db/dcrpg/upgrades.go
```diff
@@ -486,6 +516,48 @@ func (pgb *ChainDB) handleVinsTableMainchainupgrade() (int64, error) {
...
+	ticketsBlockTime, err := retrieveTicketsTxsBlockTime(pgb.db)
+	if err != nil {
+		return 0, fmt.Errorf("Unable to retreive all a tickets block time: %v", err)
```
chappjc: `unable to retrieve ticket block times: %v` (The string in an `error` type should start lower case).

db/dcrpg/upgrades.go
```diff
@@ -486,6 +516,48 @@ func (pgb *ChainDB) handleVinsTableMainchainupgrade() (int64, error) {
...
+
+	// drop the created index since its no longer need
+	defer func() {
+		_, err = pgb.db.Exec("DROP INDEX xxxx_tickets;")
+		if err != nil {
+			log.Warnf("failed to drop index xxxx_tickets: %v ", err)
+		}
+	}()
+
+	log.Infof(" - Updating tickets data with its block time value ...")
```
chappjc: `" - Updating tickets table with block time values..."`

explorer/explorerroutes.go
```diff
@@ -235,6 +235,54 @@ func (exp *explorerUI) Mempool(w http.ResponseWriter, r *http.Request) {
...
+	exp.MempoolData.Lock()
+	var mpData = exp.MempoolData
+	exp.MempoolData.Unlock()
+
+	if len(mpData.Tickets) > 0 {
+		mp.Time = append(mp.Time, uint64(mpData.Tickets[0].Time))
+		mp.Price = append(mp.Price, mpData.Tickets[0].TotalOut)
+		mp.Mempool = append(mp.Mempool, uint64(len(mpData.Tickets)))
+	} else {
+		log.Warn("No tickets exists in the mempool")
```
chappjc: `Debug` please, unless this requires attention.

explorer/explorerroutes.go
```diff
@@ -235,6 +235,54 @@ func (exp *explorerUI) Mempool(w http.ResponseWriter, r *http.Request) {
...
+	)
+	if err != nil {
+		log.Errorf("Template execute failure: %v", err)
+		exp.StatusPage(w, defaultErrorCode, defaultErrorMessage, ErrorStatusType)
+		return
+	}
+	var mp = dbtypes.PoolTicketsData{}
+	exp.MempoolData.Lock()
+	var mpData = exp.MempoolData
+	exp.MempoolData.Unlock()
```
chappjc: You just copied the `MempoolData` pointer at this point, so you can't unlock yet.  Down below you actually read the data pointed to by `MempoolData` (`mpData`), so you'd need to unlock after that to prevent a writer concurrently modifying this data.

Also, since you are not modifying the data, you just need `RLock` and `RUnlock`, which will not lock out other readers, only writers.

explorer/explorerroutes.go
```diff
@@ -235,6 +235,54 @@ func (exp *explorerUI) Mempool(w http.ResponseWriter, r *http.Request) {
...
+	} else {
+		log.Warn("No tickets exists in the mempool")
+	}
+
+	str, err := exp.templates.execTemplateToString("ticketpool", struct {
+		Version     string
+		NetName     string
+		ChartData   []*dbtypes.PoolTicketsData
+		GroupedData *dbtypes.PoolTicketsData
+		Mempool     dbtypes.PoolTicketsData
```
chappjc: How come `GroupedData` is a pointer but not `Mempool`.  It's not a small struct.

explorer/websockethandlers.go
```diff
@@ -59,55 +60,107 @@ func (exp *explorerUI) RootWebsocket(w http.ResponseWriter, r *http.Request) {
 
 				// handle received message according to event ID
 				var webData WebSocketMessage
+				//  If the request sent is past the limit exit the loop
```
chappjc: Not right to say "exit" the loop since you are using `continue`.  This goes to the next iteration, or back to the beginning of the loop"

explorer/websockethandlers.go
```diff
@@ -59,55 +60,107 @@ func (exp *explorerUI) RootWebsocket(w http.ResponseWriter, r *http.Request) {
...
 					txid, err := exp.blockData.SendRawTransaction(msg.Message)
 					if err != nil {
 						webData.Message = fmt.Sprintf("Error: %v", err)
 					} else {
 						webData.Message = fmt.Sprintf("Transaction sent: %s", txid)
 					}
-				case "getmempooltxs":
-					webData.EventId = msg.EventId + "Resp"
 
+				case "getmempooltxs":
```
chappjc: Please change the below Lock/Unlock to RLock/RUnlock while you are here.

explorer/websockethandlers.go
```diff
@@ -59,55 +60,107 @@ func (exp *explorerUI) RootWebsocket(w http.ResponseWriter, r *http.Request) {
...
+							log.Debugf("Invalid ticket pool interval provided via getticketpooldata: %s", msg.Message)
+							webData.Message = "Error: " + err.Error()
+							break
+						}
+						log.Errorf("TicketPoolVisualization error: %v", err)
+						webData.Message = "Error: Failed to fetch ticketpool data"
+						break
+					}
+
+					exp.MempoolData.Lock()
```
chappjc: same issues as before (RLock/RUnlock, and unlocking too early)

db/dcrpg/internal/txstmts.go
```diff
@@ -130,8 +130,17 @@ const (
...
 	SelectStakeTxByHash   = `SELECT id, block_hash, block_index FROM transactions WHERE tx_hash = $1 and tree=1;`
 
+	SelectAllTicketsBlockTime = `SELECT block_height, block_time FROM transactions WHERE tx_type = 1;`
+
+	SelectTicketsByType = `SELECT
+		SUM(CASE WHEN num_vout = 3 THEN 1 ELSE 0 END) as solo,
+		SUM(CASE WHEN num_vout = 5 THEN 1 ELSE 0 END) as pooled,
+		SUM(CASE WHEN num_vout > 5 THEN 1 ELSE 0 END) as tixsplit
+		FROM transactions WHERE tx_hash = ANY(SELECT tx_hash FROM tickets
+		WHERE pool_status=0 AND is_mainchain = TRUE) AND tx_type=1;`
```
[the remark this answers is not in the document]
chappjc: Of course we might pay a non-trivial penalty on transaction inserts.

explorer/explorerroutes.go
chappjc: And `"No tickets exist in the mempool"`

db/dcrpg/internal/txstmts.go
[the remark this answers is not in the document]
chappjc: Some more thoughts on this query:

- The `AND tx_type=1` query is not necessary since the tx hash is coming from the tickets table.  Removing that cuts the query time more than in half.
- The tickets table has the corresponding row id (the primary key) in the transactions table, so you can use that instead of tx_hash.  That cuts the query down to about 95 ms.
- A `JOIN` instead of the `WHERE tx_hash = ANY(SELECT` might be better.
- The `SUM(CASE` stuff can be avoided with `GROUP BY num_vout` with just a little aggregation for the >5 cases in Go.

This gives:

```
dcrdata=# SELECT num_vout, count(*) FROM transactions JOIN tickets ON transactions.id=purchase_tx_db_id WHERE pool_status=0 AND tickets.is_mainchain = TRUE GROUP BY num_vout;
 num_vout | count 
----------+-------
        3 | 21469
        5 | 20765
        7 |     1
        9 |     3
       11 |     2
       13 |     2
       15 |     2
       17 |     1
       19 |     2
       25 |     1
(10 rows)
```

With the first two changes, it's down to <100ms, but the other two change seem to help readability to me.

Prettied-up, the query is:

```sql
SELECT num_vout,
       count(*)
FROM transactions
JOIN tickets ON transactions.id=purchase_tx_db_id
WHERE pool_status=0
  AND tickets.is_mainchain = TRUE
GROUP BY num_vout;
```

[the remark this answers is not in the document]
chappjc: Or sticking with the `SUM(CASE` approach, and applying the first 3 changes:

```
dcrdata=# SELECT SUM(CASE WHEN num_vout = 3 THEN 1 ELSE 0 END) as solo, SUM(CASE WHEN num_vout = 5 THEN 1 ELSE 0 END) as pooled, SUM(CASE WHEN num_vout > 5 THEN 1 ELSE 0 END) as tixsplit FROM transactions JOIN tickets ON transactions.id=purchase_tx_db_id WHERE pool_status=0 AND tickets.is_mainchain = TRUE;
 solo  | pooled | tixsplit 
-------+--------+----------
 21469 |  20765 |       14
(1 row)
```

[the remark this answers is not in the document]
chappjc: More naturally, using the `width_bucket` aggregate and keeping the table output format:

```sql
SELECT width_bucket(num_vout, array[3, 5, 6]) as ticket_bucket,
       count(*)
FROM transactions
JOIN tickets ON transactions.id=purchase_tx_db_id
WHERE pool_status=0                         -- live tickets
  AND tickets.is_mainchain = TRUE
GROUP BY ticket_bucket;
```
 Output:

```
 ticket_bucket | count 
---------------+-------
             1 | 21469
             2 | 20765
             3 |    14
(3 rows)
```

api/apirouter.go
```diff
@@ -189,6 +189,11 @@ func NewAPIRouter(app *appContext, userRealIP bool) apiMux {
 		r.With(m.ChartTypeCtx).Get("/{charttype}", app.getChartTypeData)
 	})
 
+	mux.Route("/ticketpool", func(r chi.Router) {
+		r.Get("/", http.NotFound)
```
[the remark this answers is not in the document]
dmigwi: ok. Noted

api/apiroutes.go
```diff
@@ -722,6 +724,21 @@ func (c *appContext) getSSTxDetails(w http.ResponseWriter, r *http.Request) {
 	writeJSON(w, sstxDetails, c.getIndentQuery(r))
 }
 
+func (c *appContext) getTicketPoolByDate(w http.ResponseWriter, r *http.Request) {
+	var tp = m.GetTpCtx(r)
+	var maturityBlock = c.AuxDataSource.GetTicketPoolBlockMaturity()
```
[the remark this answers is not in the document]
dmigwi: ok. Noted

api/apiroutes.go
```diff
@@ -722,6 +724,21 @@ func (c *appContext) getSSTxDetails(w http.ResponseWriter, r *http.Request) {
...
+
+	var tpData, err = c.AuxDataSource.GetTicketPoolByDateAndInterval(maturityBlock,
```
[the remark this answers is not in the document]
dmigwi: ok. Noted

db/dcrpg/internal/stakestmts.go
```diff
@@ -60,6 +61,18 @@ const (
 	SelectUnspentTickets         = `SELECT id, tx_hash FROM tickets WHERE spend_type = 0 OR spend_type = -1
 		AND is_mainchain = true;`
 
+	SelectTicketsByPrice = `SELECT price,
+		SUM(CASE WHEN block_height >= $1 THEN 1 ELSE 0 END) as immature,
+		SUM(CASE WHEN block_height < $1 THEN 1 ELSE 0 END) as live FROM tickets
+		WHERE pool_status=0 AND is_mainchain=TRUE GROUP BY price ORDER BY price;`
+
+	SelectTicketsByPurchaseDate = `SELECT (block_time/$1)*$1 as timestamp,
```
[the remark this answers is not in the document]
dmigwi: ok. My end the queries we so slow in transactions table. Too 3 secs at best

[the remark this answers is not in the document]
dmigwi: Do we keep the `pool_status` index? My laptop is super slow without the index. I initially tried out the join on transactions and tickets table but that brought no significant performance increase.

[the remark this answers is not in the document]
dmigwi: This query should be our worst case scenario since it returns the biggest payload and its generally been the slowest 

```sql
SELECT (block_time/1)*1 as timestamp, 
   SUM(price) as price, 
   SUM(CASE WHEN block_height >= 267622 THEN 1 ELSE 0 END) as immature,
   SUM(CASE WHEN block_height < 267622 THEN 1 ELSE 0 END) as live 
FROM tickets 
WHERE pool_status = 0 AND is_mainchain = TRUE  
GROUP BY timestamp ORDER BY timestamp;
```

Without the Index 
```
dcrdata=# explain analyze SELECT (block_time/1)*1 as timestamp, SUM(price) as price, SUM(CASE WHEN block_height >= 267622 THEN 1 ELSE 0 END) as immature, SUM(CASE WHEN block_height < 267622 THEN 1 ELSE 0 END) as live FROM tickets WHERE pool_status = 0 AND is_mainchain = TRUE  GROUP BY timestamp ORDER BY timestamp;
                                                                    QUERY PLAN
---------------------------------------------------------------------------------------------------------------------------------------------------
 Finalize GroupAggregate  (cost=188362.39..193490.83 rows=30268 width=32) (actual time=5470.357..5494.540 rows=16635 loops=1)
   Group Key: (((block_time / 1) * 1))
   ->  Gather Merge  (cost=188362.39..192706.87 rows=32994 width=32) (actual time=5470.354..5486.915 rows=19750 loops=1)
         Workers Planned: 2
         Workers Launched: 2
         ->  Partial GroupAggregate  (cost=187362.36..187898.52 rows=16497 width=32) (actual time=5444.093..5451.107 rows=6583 loops=3)
               Group Key: (((block_time / 1) * 1))
               ->  Sort  (cost=187362.36..187403.61 rows=16497 width=20) (actual time=5444.078..5445.483 rows=14202 loops=3)
                     Sort Key: (((block_time / 1) * 1))
                     Sort Method: quicksort  Memory: 2749kB
                     ->  Parallel Seq Scan on tickets  (cost=0.00..186206.76 rows=16497 width=20) (actual time=0.146..5435.731 rows=14202 loops=3)
                           Filter: (is_mainchain AND (pool_status = 0))
                           Rows Removed by Filter: 446455
 Planning time: 0.574 ms
 Execution time: 5495.744 ms
(15 rows)
```

With the index
```
dcrdata=# explain analyze SELECT (block_time/1)*1 as timestamp, SUM(price) as price, SUM(CASE WHEN block_height >= 267622 THEN 1 ELSE 0 END) as immature, SUM(CASE WHEN block_height < 267622 THEN 1 ELSE 0 END) as live FROM tickets WHERE pool_status = 0 AND is_mainchain = TRUE  GROUP BY timestamp ORDER BY timestamp;
                                                                     QUERY PLAN
-----------------------------------------------------------------------------------------------------------------------------------------------------
 Sort  (cost=32566.09..32642.24 rows=30462 width=32) (actual time=112.169..114.041 rows=16635 loops=1)
   Sort Key: (((block_time / 1) * 1))
   Sort Method: quicksort  Memory: 2068kB
   ->  HashAggregate  (cost=29840.54..30297.47 rows=30462 width=32) (actual time=103.263..107.209 rows=16635 loops=1)
         Group Key: ((block_time / 1) * 1)
         ->  Index Scan using pool_status_index on tickets  (cost=0.43..29241.46 rows=39939 width=20) (actual time=0.051..86.166 rows=42607 loops=1)
               Index Cond: (pool_status = 0)
               Filter: is_mainchain
 Planning time: 0.356 ms
 Execution time: 118.356 ms
(10 rows)
```

Without the index you can see the query takes 5 sec compared to 120 ms with the index.

db/dcrpg/internal/txstmts.go
```diff
@@ -130,8 +130,17 @@ const (
 	SelectRegularTxByHash = `SELECT id, block_hash, block_index FROM transactions WHERE tx_hash = $1 and tree=0;`
 	SelectStakeTxByHash   = `SELECT id, block_hash, block_index FROM transactions WHERE tx_hash = $1 and tree=1;`
 
+	SelectAllTicketsBlockTime = `SELECT block_height, block_time FROM transactions WHERE tx_type = 1;`
+
+	SelectTicketsByType = `SELECT
```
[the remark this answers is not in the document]
dmigwi: What should I use here instead?

db/dcrpg/internal/txstmts.go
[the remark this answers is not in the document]

db/dcrpg/queries.go
```diff
@@ -1354,6 +1354,76 @@ func RetrieveAddressTxnOutputWithTransaction(db *sql.DB, address string, current
 	return outputs, nil
 }
 
+// retrieveTicketsByDate fetches the tickets in the current ticketpool by the
```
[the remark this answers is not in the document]
dmigwi: ok. Noted

db/dcrpg/queries.go
```diff
@@ -1354,6 +1354,76 @@ func RetrieveAddressTxnOutputWithTransaction(db *sql.DB, address string, current
...
+
+// retrieveTicketByPrice fetches the tickets in the current ticketpool by the
+// purchase price.
+func retrieveTicketByPrice(db *sql.DB, maturityBlock int64) (*dbtypes.PoolTicketsData, error) {
+	// Create the query statement and retrieve rows
+	rows, err := db.Query(internal.SelectTicketsByPrice, maturityBlock)
+	if err != nil {
+		return nil, err
+	}
+
```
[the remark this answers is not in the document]
dmigwi: Will add it.

db/dcrpg/queries.go
dmigwi: We need to get the average price and still maintain the 8 decimal places for our average price.
Converting it to atoms, get the average price then format to have a max of 8 decimal places did seem like the easiest way.

db/dcrpg/queries.go
dmigwi: ok. Noted

db/dcrpg/upgrades.go
dmigwi: ok. Noted

db/dcrpg/upgrades.go
dmigwi: ok. Noted

db/dcrpg/upgrades.go
dmigwi: ok. Noted

explorer/explorerroutes.go
dmigwi: ok. Noted

explorer/explorerroutes.go
dmigwi: ok. Noted

explorer/explorerroutes.go
dmigwi: ok. I will fix that.

explorer/websockethandlers.go
dmigwi: ok. Noted

explorer/websockethandlers.go
dmigwi: ok. Noted

explorer/websockethandlers.go
dmigwi: ok. Noted

db/dcrpg/internal/stakestmts.go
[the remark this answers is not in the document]
chappjc: Sure, the pool_status index is fine, although it's surprising the difference is so dramatic in your machine.

The main point in my comments here are that `JOIN transactions ON purchase_tx_db_id=transactions.id` will remove the need to duplicate block_time.

db/dcrpg/internal/txstmts.go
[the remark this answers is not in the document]
chappjc: There is no way to make this determination.

This is why I asked gozart to soften the language on the frontend recently.

We just need a comment for this code... "this is just guessing based on commonly structured ticket purchases"

db/dcrpg/queries.go
chappjc: Hmm, it seems like the max of 8 decimal places is a display/string processing issue.  The average quantity is just a statistic, not an actual single quantity accounted directly in the block chain.  So there's no need to keep the decimals to 8 except for display consistency

db/dcrpg/internal/stakestmts.go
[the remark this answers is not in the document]
dmigwi: ok.  I haven't tried that out.

db/dcrpg/internal/txstmts.go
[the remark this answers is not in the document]

db/dcrpg/queries.go
dmigwi: Yes, display consistency is why I had to do all that.

db/dcrpg/internal/stakestmts.go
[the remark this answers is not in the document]
dmigwi: Its still abit slow but not that much..
```
dcrdata=# explain analyze SELECT (tickets.block_time/86400)*86400 AS TIMESTAMP,  SUM(tickets.price) AS price,  SUM(CASE   WHEN tickets.block_height >= 269601 THEN 1 ELSE 0 END) AS immature, SUM(CASE  WHEN tickets.block_height < 269601 THEN 1  ELSE 0  END) AS live FROM tickets JOIN transactions ON purchase_tx_db_id=transactions.id WHERE pool_status = 0 AND tickets.is_mainchain = TRUE GROUP BY TIMESTAMP ORDER BY TIMESTAMP;
                                                                                   QUERY PLAN
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Finalize GroupAggregate  (cost=159173.92..163272.42 rows=30462 width=32) (actual time=216.365..223.205 rows=142 loops=1)
   Group Key: (((tickets.block_time / 86400) * 86400))
   ->  Gather Merge  (cost=159173.92..162580.55 rows=23494 width=32) (actual time=216.356..227.112 rows=284 loops=1)
         Workers Planned: 1
         Workers Launched: 1
         ->  Partial GroupAggregate  (cost=158173.91..158937.46 rows=23494 width=32) (actual time=212.026..217.625 rows=142 loops=2)
               Group Key: (((tickets.block_time / 86400) * 86400))
               ->  Sort  (cost=158173.91..158232.64 rows=23494 width=20) (actual time=212.012..213.536 rows=21304 loops=2)
                     Sort Key: (((tickets.block_time / 86400) * 86400))
                     Sort Method: quicksort  Memory: 2741kB
                     ->  Nested Loop  (cost=0.86..156468.24 rows=23494 width=20) (actual time=0.141..204.344 rows=21304 loops=2)
                           ->  Parallel Index Scan using pool_status_index on tickets  (cost=0.43..28877.31 rows=23494 width=28) (actual time=0.081..66.663 rows=21304 loops=2)
                                 Index Cond: (pool_status = 0)
                                 Filter: is_mainchain
                           ->  Index Only Scan using transactions_pkey on transactions  (cost=0.43..5.43 rows=1 width=8) (actual time=0.006..0.006 rows=1 loops=42607)
                                 Index Cond: (id = tickets.purchase_tx_db_id)
                                 Heap Fetches: 25254
 Planning time: 0.752 ms
 Execution time: 227.584 ms
(19 rows)
```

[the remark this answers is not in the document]
dmigwi: - We still need the `pool_status_index` as without it the Join query takes 5 sec and with it takes 200ms
- Even though the join query is slower by some milliseconds, is it fast enough for our use case?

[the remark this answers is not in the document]
dmigwi: @chappjc Whats do we de here?

db/dcrpg/queries.go
dmigwi: Display consistency was the reason.

chappjc:
I'm AFK for the day. Try out the queries

db/dcrpg/internal/txstmts.go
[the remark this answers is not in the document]
dmigwi: For sure this new query version is faster,
```
dcrdata=# explain analyze SELECT width_bucket(num_vout, array[3, 5, 6]) as ticket_bucket,  count(*) FROM transactions JOIN tickets ON transactions.id=purchase_tx_db_id WHERE pool_status=0   AND tickets.is_mainchain = TRUE GROUP BY ticket_bucket;
                                                                                   QUERY PLAN
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Finalize GroupAggregate  (cost=157614.16..157615.90 rows=87 width=12) (actual time=357.169..357.171 rows=3 loops=1)
   Group Key: (width_bucket(transactions.num_vout, '{3,5,6}'::integer[]))
   ->  Sort  (cost=157614.16..157614.37 rows=87 width=12) (actual time=357.166..357.167 rows=6 loops=1)
         Sort Key: (width_bucket(transactions.num_vout, '{3,5,6}'::integer[]))
         Sort Method: quicksort  Memory: 25kB
         ->  Gather  (cost=157601.57..157611.35 rows=87 width=12) (actual time=357.143..362.777 rows=6 loops=1)
               Workers Planned: 1
               Workers Launched: 1
               ->  Partial HashAggregate  (cost=156601.57..156602.65 rows=87 width=12) (actual time=348.596..348.597 rows=3 loops=2)
                     Group Key: width_bucket(transactions.num_vout, '{3,5,6}'::integer[])
                     ->  Nested Loop  (cost=0.86..156484.10 rows=23494 width=4) (actual time=0.986..342.367 rows=21182 loops=2)
                           ->  Parallel Index Scan using pool_status_index on tickets  (cost=0.43..28881.31 rows=23494 width=8) (actual time=0.461..197.584 rows=21182 loops=2)
                                 Index Cond: (pool_status = 0)
                                 Filter: is_mainchain
                           ->  Index Scan using transactions_pkey on transactions  (cost=0.43..5.43 rows=1 width=12) (actual time=0.006..0.006 rows=1 loops=42364)
                                 Index Cond: (id = tickets.purchase_tx_db_id)
 Planning time: 2.520 ms
 Execution time: 362.918 ms
(18 rows)
```

Compared To
```
dcrdata=# explain analyze SELECT SUM(CASE WHEN num_vout = 3 THEN 1 ELSE 0 END) as solo, SUM(CASE WHEN num_vout = 5 THEN 1 ELSE 0 END) as pooled, SUM(CASE WHEN num_vout > 5 THEN 1 ELSE 0 END) as tixsplit FROM transactions WHERE tx_hash = ANY(SELECT tx_hash FROM tickets WHERE pool_status=0 AND is_mainchain = TRUE) AND tx_type=1;
                                                                               QUERY PLAN
------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Finalize Aggregate  (cost=285825.65..285825.66 rows=1 width=24) (actual time=7391.852..7391.852 rows=1 loops=1)
   ->  Gather  (cost=285825.43..285825.64 rows=2 width=24) (actual time=7374.604..7399.996 rows=3 loops=1)
         Workers Planned: 2
         Workers Launched: 2
         ->  Partial Aggregate  (cost=284825.43..284825.44 rows=1 width=24) (actual time=7375.099..7375.099 rows=1 loops=3)
               ->  Hash Semi Join  (cost=30014.00..284753.26 rows=4811 width=4) (actual time=6655.908..7369.732 rows=14121 loops=3)
                     Hash Cond: (transactions.tx_hash = tickets.tx_hash)
                     ->  Parallel Seq Scan on transactions  (cost=0.00..239126.91 rows=578984 width=69) (actual time=0.338..6659.978 rows=461408 loops=3)
                           Filter: (tx_type = 1)
                           Rows Removed by Filter: 1140639
                     ->  Hash  (cost=29045.76..29045.76 rows=39939 width=65) (actual time=350.220..350.220 rows=42364 loops=3)
                           Buckets: 65536  Batches: 2  Memory Usage: 2514kB
                           ->  Index Scan using pool_status_index on tickets  (cost=0.43..29045.76 rows=39939 width=65) (actual time=0.353..325.339 rows=42364 loops=3)
                                 Index Cond: (pool_status = 0)
                                 Filter: is_mainchain
 Planning time: 7.588 ms
 Execution time: 7400.215 ms
(17 rows)
```

db/dcrpg/internal/stakestmts.go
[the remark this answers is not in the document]
chappjc: I see in the query in your comment above it's using `tickets.block_time`, but you've done the join so you don't need that right?

chappjc:
Good work on the last edits.  I'm still not convinced we need a new column `block_time` in the tickets table since the join with the transactions table via the primary key (`purchase_tx_db_id`) seems to be super fast.  In all of my test queries I showed in the review, I had not run the upgrade to get block_time into tickets.

But it looks like it's ~220ms with the transactions join and ~120 without.  Do avoid a redundant call, this may be worth it.

Trying these queries (with and without various indexes and extra columns) on a "production equivalent" cloud VM (2 core, 6-8 GB RAM) would help us decide how to go.

db/dcrpg/queries.go
chappjc: In that case, the float64AsDecimalParts and/or amountAsDecimalParts template functions take care of the formatting.  There's really no reason to do the `price = price * 100000000`, then `dcrutil.Amount(price/total).ToCoin()` when `price` starts in units of coin. We just need `price/total` and the template function handles formatting.

explorer/websockethandlers.go
```diff
@@ -59,55 +60,108 @@ func (exp *explorerUI) RootWebsocket(w http.ResponseWriter, r *http.Request) {
...
+					var mp = dbtypes.PoolTicketsData{}
 
+					if len(mpData.Tickets) > 0 {
+						mp.Time = append(mp.Time, uint64(mpData.Tickets[0].Time))
+						mp.Price = append(mp.Price, mpData.Tickets[0].TotalOut)
+						mp.Mempool = append(mp.Mempool, uint64(len(mpData.Tickets)))
+					} else {
+						log.Debug("No tickets exists in the mempool")
+					}
+
```
chappjc: You could put `exp.MempoolData.RUnlock()` here since you are done with `mpData`, but it's not a big difference since only a small struct definition follows.

db/dcrpg/internal/stakestmts.go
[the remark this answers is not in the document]
dmigwi: with `transactions.block_time` its even worse
```
histogram_dcrdata=# explain analyze SELECT (transactions.block_time/86400)*86400 AS TIMESTAMP,  SUM(tickets.price) AS price,  SUM(CASE   WHEN tickets.block_height >= 269601 THEN 1 ELSE 0 END) AS immature, SUM(CASE  WHEN tickets.block_height < 269601 THEN 1  ELSE 0  END) AS live FROM tickets JOIN transactions ON purchase_tx_db_id=transactions.id WHERE pool_status = 0 AND tickets.is_mainchain = TRUE GROUP BY TIMESTAMP ORDER BY TIMESTAMP;
                                                                            QUERY PLAN
-------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Finalize GroupAggregate  (cost=244787.11..250085.76 rows=39821 width=32) (actual time=5256.253..5263.711 rows=143 loops=1)
   Group Key: (((transactions.block_time / 86400) * 86400))
   ->  Gather Merge  (cost=244787.11..249156.61 rows=33184 width=32) (actual time=5256.235..5267.032 rows=428 loops=1)
         Workers Planned: 2
         Workers Launched: 2
         ->  Partial GroupAggregate  (cost=243787.09..244326.33 rows=16592 width=32) (actual time=5250.463..5256.348 rows=143 loops=3)
               Group Key: (((transactions.block_time / 86400) * 86400))
               ->  Sort  (cost=243787.09..243828.57 rows=16592 width=20) (actual time=5250.445..5251.930 rows=14152 loops=3)
                     Sort Key: (((transactions.block_time / 86400) * 86400))
                     Sort Method: quicksort  Memory: 1534kB
                     ->  Nested Loop  (cost=0.43..242624.14 rows=16592 width=20) (actual time=0.556..5242.289 rows=14152 loops=3)
                           ->  Parallel Seq Scan on tickets  (cost=0.00..152400.19 rows=16592 width=20) (actual time=0.195..4626.668 rows=14152 loops=3)
                                 Filter: (is_mainchain AND (pool_status = 0))
                                 Rows Removed by Filter: 447000
                           ->  Index Scan using transactions_pkey on transactions  (cost=0.43..5.43 rows=1 width=16) (actual time=0.042..0.042 rows=1 loops=42457)
                                 Index Cond: (id = tickets.purchase_tx_db_id)
 Planning time: 2.890 ms
 Execution time: 5267.554 ms
(18 rows)
```

db/dcrpg/queries.go
dmigwi: But in the query we are getting the `SUM(tickets.price) AS price,` that match a given grouping provided. From the sum thats where we need to get the average of the summed prices.

dmigwi: Doing this on the frontend might slow down the browser loading the ticketpool page everytime new data is available. I think this is cheaper to do it the backend while we fetch the summed prices.

explorer/websockethandlers.go
dmigwi: ok. Noted

db/dcrpg/queries.go
dmigwi: In the worst case scenario we may end up having 40k records I think using float64AsDecimalParts and/or amountAsDecimalParts template to do the formatting may make small devices browser unresponsive when loading data in the ticketpool page.

dmigwi:
Ok. I will remove the `block_time` column and us the `join` instead.

dmigwi:
I have just removed the `block_time` column from tickets and used the `join` query on transactions table instead, on my end there is some small performance decrease comparing https://github.com/decred/dcrdata/pull/542#issuecomment-417000505
with 
```
2018/09/03 04:51:46 "GET http://127.0.0.1:7777/api/ticketpool/bydate/mo HTTP/1.1" from 127.0.0.1:53100 - 200 225B in 596.088475ms
2018/09/03 04:51:48 "GET http://127.0.0.1:7777/api/ticketpool/bydate/wk HTTP/1.1" from 127.0.0.1:53100 - 200 661B in 440.863891ms
2018/09/03 04:51:49 "GET http://127.0.0.1:7777/api/ticketpool/bydate/day HTTP/1.1" from 127.0.0.1:53100 - 200 4116B in 413.580266ms
2018/09/03 04:51:51 "GET http://127.0.0.1:7777/api/ticketpool/bydate/all HTTP/1.1" from 127.0.0.1:53100 - 200 449886B in 520.34401ms
```
But I think we can gracefully handle it.

chappjc:
Trying this on a cloud VM now.  Going from ~90 ms to ~450 ms is not great, so I want to see what is would be like in production.

chappjc:
I have this up on alpha.dcrdata.org now.  Here are the timings (dcrpg format v3.5.2):

```
http://127.0.0.1:7777/api/ticketpool/bydate/mo
 - 34 sec initial (100 - 140 ms after)

http://127.0.0.1:7777/api/ticketpool/bydate/wk
 - 105 ms

http://127.0.0.1:7777/api/ticketpool/bydate/day
 - 125 ms

http://127.0.0.1:7777/api/ticketpool/bydate/all
 - 175 - 190 ms
```

chappjc:
I don't think these query times are problematic, especially with a reasonable nginx config.

image


Just a couple notes after running it live:
- Please rephrase `These charts only show current tickets in the ticket pool, for historic views go to here` to "These charts represent the current ticket pool. For historic views go here."
- For circle chart for "Tickets Grouped by # of Outputs", we need to soften the language here too.  Instead of `1. Solo Tickets: -3 outputs.` use `3 Outputs - typically solo.  Similarly, `5 Outputs - typically Voting Service Provider (VSP)` and `More than 5 Outputs - likely multi-party (split) tickets.`
- Note the VSP abbreviation (not VPS).
- Please rename "Tickets Grouped by # of Outputs" to "Number of Ticket Outputs".

Perhaps a better overall approach is not to report the number of total sstx outputs, but the number of sstxcommitment-tagged outputs, as per consensus rules this will be the number of ways the subsidy is split.  But hold off on that change since we may want feedback first.

chappjc:
The first chart with two y axes is confusing without a legend or additional explanation.  Do the bars go with the left axis (number of tickets)?  This makes sense since this would make it a ticket pool histogram with purchase date bins.  If there were a legend that showed the pink line with "Avg. Ticket Value" and a blue bar with "Number of Ticket Purchases" (as in number of ticket purchases in the time bin), then this would be clear very quickly.

dmigwi:
Sorry, I was AFK.

The feedback about the ticketpool charts is noted. 

Concerning the legend, Let me see how that works out.

dmigwi:
@chappjc I have two designs to fix the legend issue.

**Option 1**
screen shot 2018-09-04 at 20 23 38


**Option 2**
screen shot 2018-09-04 at 20 29 07


Personally I feel option 2 fits the bill.

dmigwi:
Here is the third chart

screen shot 2018-09-04 at 20 36 33

chappjc:
Yeah, I like the second option for Ticket Purchase Distribution.

On the circle chart, We can drop the "1.", "2.", "3." from the descriptions.  If it looks funny as just text, we could use bullets, but I think it's fin without any numbering or bullets.

dmigwi:
Final version

screen shot 2018-09-04 at 20 54 43

PR #546, opened 2018-07-12
[2.0] explorer/main: add no-dev-prefetch option

chappjc:
The `no-dev-prefetch` flag does the following:  Disable automatic dev/project fund balance query on new blocks. When true, the query will still be run on-demand (e.g. explorer address page for dev fund), but not automatically after new blocks are connected.


chappjc:
independently tested by lykke

PR #555, opened 2018-07-19
dcrsqlite: correctly splits given string into array of strings

JFixby:
Fixes number of issues similar to the one described in the #532
`strings.Split(winners, ";")` is not a correct way to split the `winners` string.


JFixby:
Related PR: https://github.com/decred/dcrdata/pull/533

db/dcrsqlite/sqlite.go
```diff
@@ -927,3 +903,13 @@ func logDBResult(res sql.Result) error {
 
 	return nil
 }
+
+// utility function, correctly splits given string into array of strings
```
chappjc: Please start comment with // splitToArray ...

PR #514, opened 2018-06-26
dcrsqlite: +TestInitDB

JFixby:
TestInitDB creates default DB and performs some basic checks.

This is work towards the https://github.com/decred/dcrdata/issues/5

PR #559, opened 2018-07-20
 testutil: refactor test setup

JFixby:
This is tier 1 of the #514 
- refactors the test setup
- also includes minor refactoring of the `dcrsqlite/sqlite.go`

### DIFF
--- a/testutil/testutil.go
+++ b/testutil/testutil.go
@@ -1,0 +1,58 @@
+package testutil
+
+import (
+	"fmt"
+	"testing"
+)
+
+// Tests are written in a way that allows easily
+// to switch to any other test framework
+//
+// Current test setup wraps GO-Lang/testing framework
+var t *testing.T
+
+func CurrentTestSetup() *testing.T {
+	return t
+}
+
+// BindCurrentTestSetup deploys GO-Lang/testing framework
+func BindCurrentTestSetup(set *testing.T) {
+	t = set
+}
+
+// Set this flag "true" to use "github.com/davecgh/go-spew/spew" for arrays pretty print
+var UseSpewToPrettyPrintArrays = true
+
+// Defines what to do on test setup crash
+var PanicOnTestSetupFailure = true
+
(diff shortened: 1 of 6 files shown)


testutil/testutil.go
```diff
@@ -0,0 +1,58 @@
+package testutil
```
chappjc: In this file, can you put a comment above `package testutil` starting with `// package testutil` so that godoc has something to say about the package in general?

JFixby:
@chappjc Smth like:
```
// Copyright (c) 2018+ The Decred developers
// Use of this source code is governed by an ISC license that can be found in the LICENSE file.

/*
Package testutil provides convenience functions and types for testing.

Generally tests should be written in a way that allows easily to switch to any other test framework.

testutil stands as a proxy between user-code and a test framework giving some flexibility on configuring test-setup.

Currently testutil wraps only the GO-Lang/testing framework
*/
```
?

chappjc:
@JFixby yeah, except we need to wrap all comments after 80 columns.  The consistent width is friendly with a variety of tools, but the main reason we do it that way is to facilitate side-by-side comparison such as when performing a diff.

With code, I'm generally OK with having source go to ~90 columns, but with comments we try to be strict about the 80 column thing.  It's a minor thing really.  I use an extension called Rewrap in VS Code that will do it automatically with ALT+Q.  No effort needed.

JFixby:
Ok, time to wrap comments.

chappjc:
Feels like nitpicking that doesn't really affect anything, but consistency is important.  We can focus on technical aspects now...  :)

JFixby:
I would also suggest 120 chars limit. 80 chars feels too tight.

chappjc:
I generally agree that 80 is quite tight for code.  For comments 80 is not an issue.

Anyway, for code, I'm usually good with <100, but again the side-by-side diff issue makes >100 really irritating

JFixby:
Good. I'm preparing the next tier.

Issue #442, opened 2018-04-25
No vote issue list on tx, display msg and hide column headers


raedah:
http://explorer.dcrdata.org/tx/cd658ea50ed9d21b867fb4b8b7a6878aea9a5829bc270d2ce08fac6242b40ca0

Do in the same way that missed votes is shown here, http://explorer.dcrdata.org/block/104699

SeaLightHouse:
I am doing this.

PR #560, opened 2018-07-22
Add message for no votes on a block.

snadorp:
Remove the table and display a nice message if there are no votes in a transaction.

Fixes: https://github.com/decred/dcrdata/issues/442

### DIFF
--- a/views/tx.tmpl
+++ b/views/tx.tmpl
@@ -353,7 +353,9 @@
         <div class="col-md-12">
             <h4>Vote Info</h4>
             <p>Last Block Valid: <span class="mono"><strong>{{.Validation.Validity}}</strong></span><br>
-            Version: <span class="mono">{{.Version}}</span> | Bits: <span class="mono">{{printf "%#04x" .Bits}}</span></p>
+            Version: <span class="mono">{{.Version}}</span> | Bits: <span class="mono">{{printf "%#04x" .Bits}}</span>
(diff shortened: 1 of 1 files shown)


snadorp:
Before:
screen shot 2018-07-22 at 15 54 39
After:
screen shot 2018-07-22 at 15 07 19

JFixby:
Test link: 

http://explorer.dcrdata.org/tx/cd658ea50ed9d21b867fb4b8b7a6878aea9a5829bc270d2ce08fac6242b40ca0

This is what it looks like on my machine:

http://localhost:8080/tx/cd658ea50ed9d21b867fb4b8b7a6878aea9a5829bc270d2ce08fac6242b40ca0

image

Looks fine.

views/tx.tmpl
```diff
@@ -375,6 +377,9 @@
                     {{end}}
                 </tbody>
             </table>
+            {{else}}
+            <br>No votes in this block.</p>
```
chappjc: No votes in this *transaction*?

chappjc: But in general, I found the wording a bit confusing since a vote must always validate the previous block.  Really what this change intends to say is that this vote/ssgen/transaction contains no choices on known agendas

chappjc:
Thanks @snadorp, and thanks for testing @JFixby

PR #561, opened 2018-07-22
explorer: votes filter now flags votes instead of excluding them.

papacarp:
This PR replaces PR #419 .  Progress so far:

- [x]  Modified js to match current websocket stimulus architecture
- [x] minor change to fix metalinter issue with unrelated code (%s to %v)
- [x] TOTAL SENT updates on mempool ws message
- [x] SIZE updates with mempool ws message
- [x] newtx for votes are incrementally updated and sorted in the votes table and colored if they are not for the most recent block.
- [x] last block time ago fixed when new block us updated with ws.
- [x] confirm vote choice X/5 is working properly and does not count duplicates (avoid 7/5 issue)
- [x] "pink bgcolor tweaked for dark mode" what color do you want it in dark mode?  
- [x] Is "Vote Choice" the right name for the column that says the vote is valid or invalid?

Issues:
It looks like all the votes are sent through the WS interface first as newTX before the new block (and thus mempool update) is sent.  The result is that we get a bunch of new vote transactions on the next block before we update the block.  Since the block hashes don't match they are colored pink.

Possible solutions:
1.  Investigate why we are getting newtx votes before a new block message (is this DCRD or pehaps in our websocket controller in go?) and reverse the order so we get a new block before we get any votes for that block.
2.  do not color votes pink if height is greater than current best height (it will still look a little strange but at least the votes will not stand out).

Below I captured how this looks on block updates.

ezgif com-video-to-gif 1

### DIFF
--- a/explorer/explorertypes.go
+++ b/explorer/explorertypes.go
@@ -138,6 +138,7 @@ type VoteInfo struct {
 	Choices            []*txhelpers.VoteChoice `json:"vote_choices"`
 	TicketSpent        string                  `json:"ticket_spent"`
 	MempoolTicketIndex int                     `json:"mempool_ticket_index"`
+	ForLastBlock       bool                    `json:"last_block"`
 }
 
 // BlockValidation models data about a vote's decision on a block
@@ -304,18 +305,31 @@ type MempoolInfo struct {
 
 // MempoolShort represents the mempool data sent as the mempool update
 type MempoolShort struct {
-	LastBlockHeight    int64          `json:"block_height"`
-	LastBlockTime      int64          `json:"block_time"`
-	TotalOut           float64        `json:"total"`
-	TotalSize          int32          `json:"size"`
-	NumTickets         int            `json:"num_tickets"`
-	NumVotes           int            `json:"num_votes"`
-	NumRegular         int            `json:"num_regular"`
-	NumRevokes         int            `json:"num_revokes"`
-	NumAll             int            `json:"num_all"`
-	LatestTransactions []MempoolTx    `json:"latest"`
-	FormattedTotalSize string         `json:"formatted_size"`
-	TicketIndexes      map[string]int `json:"ticket_indexes"`
+	LastBlockHeight    int64                    `json:"block_height"`
+	LastBlockHash      string                   `json:"block_hash"`
+	LastBlockTime      int64                    `json:"block_time"`
+	TotalOut           float64                  `json:"total"`
+	TotalSize          int32                    `json:"size"`
+	NumTickets         int                      `json:"num_tickets"`
+	NumVotes           int                      `json:"num_votes"`
+	NumRegular         int                      `json:"num_regular"`
+	NumRevokes         int                      `json:"num_revokes"`
+	NumAll             int                      `json:"num_all"`
+	LatestTransactions []MempoolTx              `json:"latest"`
+	FormattedTotalSize string                   `json:"formatted_size"`
+	TicketIndexes      map[int64]map[string]int `json:"ticket_indexes"`
+	VotingInfo         VotingInfo               `json:"voting_info"`
+}
+
+// VotingInfo models data about the validity of the next block from mempool
+type VotingInfo struct {
+	Valids         uint16 `json:"choice_valid"`
+	Invalids       uint16 `json:"choice_invalid"`
+	TotalCollected uint16 `json:"total_votes_collected"`
+	TotalNeeded    uint16 `json:"total_votes_required"`
+	Required       uint16 `json:"total_choices_required"`
+	BlockValid     bool   `json:"block_valid"`
+	voted          map[string]bool
 }
 
 // ChainParams models simple data about the chain server's parameters used for some
--- a/explorer/explorertypes.go
+++ b/explorer/explorertypes.go
@@ -317,7 +317,7 @@ type MempoolShort struct {
 	NumAll             int                      `json:"num_all"`
 	LatestTransactions []MempoolTx              `json:"latest"`
 	FormattedTotalSize string                   `json:"formatted_size"`
-	TicketIndexes      map[int64]map[string]int `json:"ticket_indexes"`
+	TicketIndexes      map[int64]map[string]int `json:"-"`
 	VotingInfo         VotingInfo               `json:"voting_info"`
 }
 
--- a/explorer/mempool.go
+++ b/explorer/mempool.go
@@ -284,7 +284,6 @@ func (exp *explorerUI) getLastBlock() (lastBlockHash string, lastBlock int64, la
 	if err != nil {
 		log.Warnf("Could not get bloch hash for last block")
 	}
-	log.Debug("Last Block Hash: ", lastBlockHash)
 	return
 }
 
--- a/explorer/mempool.go
+++ b/explorer/mempool.go
@@ -96,6 +96,8 @@ func (exp *explorerUI) mempoolMonitor(txChan chan *NewMempoolTx) {
 			exp.MempoolData.Votes = append([]MempoolTx{tx}, exp.MempoolData.Votes...)
 			sort.Sort(byHeight(exp.MempoolData.Votes))
 			exp.MempoolData.NumVotes++
+			// Multiple transactions can be in mempool from the same ticket.  We
+			// need to insure we do not double count these votes.
 			if tx.VoteInfo.ForLastBlock && !exp.MempoolData.VotingInfo.voted[tx.VoteInfo.TicketSpent] {
 				exp.MempoolData.VotingInfo.voted[tx.VoteInfo.TicketSpent] = true
 				exp.MempoolData.VotingInfo.TotalCollected++
@@ -199,7 +201,8 @@ func (exp *explorerUI) storeMempoolInfo() (lastBlockHash string, lastBlock int64
 			tickets = append(tickets, tx)
 		case "Vote":
 			addVoteIndex(tx.VoteInfo, txindexes)
-			if tx.VoteInfo.ForLastBlock = voteForLastBlock(lastBlockHash, tx.VoteInfo.Validation.Hash); tx.VoteInfo.ForLastBlock && !votingInfo.voted[tx.VoteInfo.TicketSpent] {
+			tx.VoteInfo.ForLastBlock = voteForLastBlock(lastBlockHash, tx.VoteInfo.Validation.Hash)
+			if tx.VoteInfo.ForLastBlock && !votingInfo.voted[tx.VoteInfo.TicketSpent] {
 				votingInfo.voted[tx.VoteInfo.TicketSpent] = true
 				votingInfo.TotalCollected++
 				if tx.VoteInfo.Validation.Validity {
@@ -247,8 +250,9 @@ func (exp *explorerUI) storeMempoolInfo() (lastBlockHash string, lastBlock int64
 	return
 }
 
-// addVoteIndex sets v.MempoolTicketIndex to the corresponding map value if it exits in
-// txindexes and creates a new index map if it doesn't
+// addVoteIndex assigns a unique index for a vote in a block if an index has not
+// already been assigned.  This index is used for sorting in views and counting
+// total unique votes for a block.
 func addVoteIndex(v *VoteInfo, txindexes map[int64]map[string]int) {
 	if idxs, ok := txindexes[v.Validation.Height]; ok {
 		if idx, ok := idxs[v.TicketSpent]; ok {
--- a/explorer/mempool.go
+++ b/explorer/mempool.go
@@ -57,9 +57,6 @@ func (exp *explorerUI) mempoolMonitor(txChan chan *NewMempoolTx) {
 		var voteInfo *VoteInfo
 		if ok := stake.IsSSGen(msgTx); ok {
 			validation, version, bits, choices, err := txhelpers.SSGenVoteChoices(msgTx, exp.ChainParams)
-			if !voteForLastBlock(lastBlockHash, validation.Hash.String()) {
-				continue
-			}
 			if err != nil {
 				log.Debugf("Cannot get vote choices for %s", hash)
 			} else {
@@ -69,10 +66,11 @@ func (exp *explorerUI) mempoolMonitor(txChan chan *NewMempoolTx) {
 						Height:   validation.Height,
 						Validity: validation.Validity,
 					},
-					Version:     version,
-					Bits:        bits,
-					Choices:     choices,
-					TicketSpent: msgTx.TxIn[1].PreviousOutPoint.Hash.String(),
+					Version:      version,
+					Bits:         bits,
+					Choices:      choices,
+					TicketSpent:  msgTx.TxIn[1].PreviousOutPoint.Hash.String(),
+					ForLastBlock: voteForLastBlock(lastBlockHash, validation.Hash.String()),
 				}
 			}
 		}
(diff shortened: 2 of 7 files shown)


papacarp:
1.  confirmed 7/5 working properly - it will show a max of 5/5.
2.  new block tx are not painted pink if the block height is greater than current block height.
3.  Column headers changed to "last block" and function is valid or invalid for the last block.  updates dynamically.

This PR is ready for review and merge.

@gozart1 since you did the original Stimulus implementation I would appreciate quick look at this for the js changes if you have a moment.  Also if you have any thoughts on the right pink to use for "dark mode" as @raedah asked I would appreciate it.  Otherwise we can let that go and leave it for a future update as well when we are doing CSS work.

gozart1:
sure thing, taking a look now

papacarp:
updated upcoming votes to light grey background as per dialog with @raedah

public/js/controllers/mempool.js
```diff
@@ -61,6 +61,47 @@
         $(newRowHtml).insertBefore(rows.first())
     }
 
+    function filterVotesLastBlock(target) {
+        var best_block_hash=$(target).data("hash");
```
gozart1: Please use camelCase for client side variables, so `best_block_hash` should be `bestBlockHash`, it's simply the convention amongst the majority of open source javascript codebases.

public/js/controllers/mempool.js
```diff
@@ -61,6 +61,47 @@
...
+        var best_block_height=$(target).text();
+        console.log(best_block_height);
+        $('*[data-target="mempool.voteTransactions"] tr').each(function(i,el) {
```
gozart1: If these functions are moved inside this controller class, then you will have the targets available via `this.voteTransactionsTargets`, that would be the stimulus way of accessing the DOM.

public/js/controllers/mempool.js
```diff
@@ -138,8 +184,10 @@
             });
             ws.registerEvtHandler("getmempooltxsResp", (evt) => {
                 this.handleTxsResp(evt)
+                filterVotesLastBlock(this.bestBlockTarget);
+                sortVotesTable();
                 keyNav(evt, false, true)
-            })
+            })
```
gozart1: No trailing whitespace please. The reason for this is that some contributes have their editors configured to auto delete trailing whitespace, and this muddies up the git history with lines being "changed" with no meaning other than whitespace deletion.

views/mempool.tmpl
```diff
@@ -35,6 +35,13 @@
                             <td class="text-right pr-2 lh1rem nowrap p03rem0">TICKETS</td>
                             <td data-target="mempool.numTicket" class="lh1rem">{{.NumTickets}}</td>
                         </tr>
+                        <tr>
+                            <td class="text-right pr-2 lh1rem nowrap p03rem0">VOTE CHOICES</td>
+                            <td id="mempool_vote_choices" class="lh1rem"><span data-target="mempool.totalCollected">
```
gozart1: this id looks unused

views/mempool.tmpl
```diff
@@ -35,6 +35,13 @@
...
+                            {{.VotingInfo.TotalCollected}}</span>/<span data-target="mempool.totalNeeded">{{.VotingInfo.TotalNeeded}}</span>
+                            <span id="mempool_block_validity" class="lh1rem {{if lt .VotingInfo.TotalCollected .VotingInfo.TotalNeeded}}hidden{{end}}">({{if .VotingInfo.BlockValid}}Valid{{else}}Invalid{{end}})</span>
```
gozart1: this id also looks unused

public/js/controllers/mempool.js
```diff
@@ -61,6 +61,47 @@
...
+            var heightB = parseInt($('td:nth-child(2)',b).text());
+            if(heightA == heightB) {
+                 var indexA = parseInt($('td:nth-child(3)',a).text());
+                 var indexB = parseInt($('td:nth-child(3)',b).text());
+                 return (indexA - indexB);
+            }else{
+                return (heightB - heightA);
+            }
+        });
+        $.each($rows, function(index, row){
```
gozart1: Instead of looping, it's more efficient to update the html in a single operation via `$table.html($rows);`

gozart1:
Since the background is already kind of gray, how about blue for upcoming?
Here's some semi-transparent colors that could work for both dark & light mode. 
```
.old-vote {
  background-color: #ff3d001f;
}
.upcoming-vote {
  background: #0078ff1c;
}
```
screen shot 2018-07-24 at 12 39 17 pm
screen shot 2018-07-24 at 12 39 09 pm

papacarp:
@gozart1 thank you for the thorough comments!  I learned a lot from your review and I really appreciate it.  I think it hit all of the issues you brought up with the latest push.

public/js/controllers/mempool.js
```diff
@@ -232,5 +184,42 @@
                 addTxRow(tx, this[tx.Type.toLowerCase() + "TransactionsTarget"], rowFn)
             })
         }
+
+        filterVotesLastBlock() {
```
gozart1: One last nit about code formatting. 
I'm not into bike-shedding about semi-colons or no semi-colons, and don't think we need a js linter to enforce style. But I do think we it is beneficial to have a baseline of standards.

I think this is the smallest set of rules that are worth aiming for as a team for the sake of having a relatively consistent js codebase. https://crockford.com/javascript/code.html 

So in terms of this function, there should be whitespace between control flow characters, comparators, and variable assignments.
eg `var bestBlockHash = $(this.bestBlockTarget).data("hash");`
and `if (voteBlockHeight > bestBlockHeight) {`

papacarp: I actually was curious what my protocol should be on semi-colons.  I just pushed those changes.  I'll try to assimilate the code.html strategies you sent over as I go.  I appreciate the feedback - the more aligned we all are to a common theme the more professional we all look.

public/js/controllers/mempool.js
```diff
@@ -176,6 +184,42 @@
                 addTxRow(tx, this[tx.Type.toLowerCase() + "TransactionsTarget"], rowFn)
             })
         }
-    })
 
+        filterVotesLastBlock() {
```
gozart1: How about renaming this to something more semantic like `labelVotes`? 
The word "filter" implies creating a subset to me.

gozart1: or `categorizeVotes` or ...`classifyVotes`, too many good names to chose!

public/js/controllers/mempool.js
gozart1: I tend to leave them out.
If I had to chose a linter it would probably be https://standardjs.com/ which is the main one people that who don't use semi-colons use. I point people to the crockford site since that's like 20 years old and still pretty much looks like most of the well written js you see out there today, so I figure that's evidence that the recommendations are reasonable.  Not even TC39 can decide on the semi-colons though lol https://github.com/tc39/agendas/commit/c0b54452ff2e2366e6e7847559ef237d5b2f6783#diff-5a8c1d412a7069f29901ea548d76ac26

gozart1:
PR looks pretty good to me. I noticed the same thing with regards to transactions related to new blocks being received by multiple handlers. Your solution of using an idempotent labeling function seems like a good way to handle it without having to get into edgecases.

public/js/controllers/mempool.js
papacarp: That sounds like a good idea.  i'll update that later today.

chappjc:
I updated the PR comment as all the TODO items have been hit.

Regarding the upcoming votes showing up before the block, I believe what you are referring to is a consequence of when the votes are broadcast vs. mined:
1. Block N is mined including votes on block N-1.
2. Within seconds of (1), the votes on the validity of block N are broadcast.
3. The miners start working on block N+1, which includes the votes from (2).
4. A solution to block N+1 is found and the block is broadcast.

If I understand your observation, the two events that show duplicate votes are (2) and (4).  But those are ~5 minutes apart, whereas (1) and (2) are seconds apart.

explorer/explorertypes.go
```diff
@@ -304,18 +305,31 @@ type MempoolInfo struct {
...
+	TotalOut           float64                  `json:"total"`
+	TotalSize          int32                    `json:"size"`
+	NumTickets         int                      `json:"num_tickets"`
+	NumVotes           int                      `json:"num_votes"`
+	NumRegular         int                      `json:"num_regular"`
+	NumRevokes         int                      `json:"num_revokes"`
+	NumAll             int                      `json:"num_all"`
+	LatestTransactions []MempoolTx              `json:"latest"`
+	FormattedTotalSize string                   `json:"formatted_size"`
+	TicketIndexes      map[int64]map[string]int `json:"ticket_indexes"`
```
chappjc: I do not believe this will marshal to JSON.  Probably will be something like `json: unsupported type: map[int64]...`.

Since this is only being used for the go html template processing, not JSON marshaling (presently), I suggest `json:"-"` for this

chappjc:
not forgotten, review coming in a.m.

explorer/mempool.go
```diff
@@ -180,23 +183,30 @@ func (exp *explorerUI) storeMempoolInfo() (lastBlockHash string, lastBlock int64
...
-				continue
-			}
-			if idx, ok := txindexes[tx.VoteInfo.TicketSpent]; ok {
-				tx.VoteInfo.MempoolTicketIndex = idx
-			} else {
-				idx := len(txindexes) + 1
-				txindexes[tx.VoteInfo.TicketSpent] = idx
-				tx.VoteInfo.MempoolTicketIndex = idx
+			addVoteIndex(tx.VoteInfo, txindexes)
+			if tx.VoteInfo.ForLastBlock = voteForLastBlock(lastBlockHash, tx.VoteInfo.Validation.Hash); tx.VoteInfo.ForLastBlock && !votingInfo.voted[tx.VoteInfo.TicketSpent] {
```
chappjc: please put the assignment on a separate line

explorer/mempool.go
```diff
@@ -230,14 +242,38 @@ func (exp *explorerUI) storeMempoolInfo() (lastBlockHash string, lastBlock int64
 		LatestTransactions: latest,
 		FormattedTotalSize: humanize.Bytes(uint64(totalSize)),
 		TicketIndexes:      txindexes,
+		VotingInfo:         votingInfo,
 	}
 	return
 }
 
+// addVoteIndex sets v.MempoolTicketIndex to the corresponding map value if it exits in
+// txindexes and creates a new index map if it doesn't
```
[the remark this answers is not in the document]
papacarp: Thank you.  Ok, I didn't touch this section of the PR at all.  I'll try to sort out what @RogueElement was aiming for.  Same with the rest of your comments.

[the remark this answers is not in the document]
chappjc: Oh, I see.  I forgot this work was passed to you.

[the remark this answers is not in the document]
papacarp: no problem.  It still has to be right!

explorer/mempool.go
```diff
@@ -180,23 +183,30 @@ func (exp *explorerUI) storeMempoolInfo() (lastBlockHash string, lastBlock int64
 
 	var totalOut float64
 	var totalSize int32
+	var votingInfo VotingInfo
+
+	lastBlockHash, lastBlock, lastBlockTime = exp.getLastBlock()
+
+	votingInfo.TotalNeeded = exp.ChainParams.TicketsPerBlock
```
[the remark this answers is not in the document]
papacarp: perhaps the right terminology here would be "TargetTotalVotes" ?  I think he was just trying to get to the number 5 here. (for mainnet)

explorer/mempool.go
```diff
@@ -230,14 +242,38 @@ func (exp *explorerUI) storeMempoolInfo() (lastBlockHash string, lastBlock int64
...
+		v.MempoolTicketIndex = 1
+	}
+}
+
 func voteForLastBlock(blockHash, validationHash string) bool {
 	return blockHash == validationHash && blockHash != ""
 }
 
+func updateBlockValidity(votingInfo *VotingInfo) {
+	votingInfo.BlockValid = votingInfo.Valids+votingInfo.Invalids >= votingInfo.TotalNeeded && votingInfo.Valids >= votingInfo.Required
```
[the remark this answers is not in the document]
papacarp: I see your point.  I reviewed DCRD briefly and it checks a lot of things that would likely require quite a few rpc calls to replicate if we really want to say a block is valid.  I recommend one of three paths here:

1.  Remove "Vote Choices" completely from the mempool page.  Since the votes seem to come in via ws all at once anyway after a block is sent, we don't really get useful information out of this (IE, we don't actually see votes come in one at a time allowing us to watch the block become "valid")
2.  Remove the assertion of validity here and change it to be something like:
VOTES:  3 found / 5 targeted
3.  tone down the assertion of validity:
VOTES:  3 found / 5 targeted (vote quantity valid)

explorer/mempool.go
```diff
@@ -94,15 +92,20 @@ func (exp *explorerUI) mempoolMonitor(txChan chan *NewMempoolTx) {
...
-			} else {
-				idx := len(exp.MempoolData.TicketIndexes) + 1
-				exp.MempoolData.TicketIndexes[tx.VoteInfo.TicketSpent] = idx
-				tx.VoteInfo.MempoolTicketIndex = idx
-			}
+			addVoteIndex(tx.VoteInfo, exp.MempoolData.TicketIndexes)
 			exp.MempoolData.Votes = append([]MempoolTx{tx}, exp.MempoolData.Votes...)
+			sort.Sort(byHeight(exp.MempoolData.Votes))
 			exp.MempoolData.NumVotes++
+			if tx.VoteInfo.ForLastBlock && !exp.MempoolData.VotingInfo.voted[tx.VoteInfo.TicketSpent] {
```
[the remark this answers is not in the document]
papacarp: For whatever reason sometimes multiple transactions are generated and submitted to the mempool for a single ticket that is voted.  Those duplicate transactions may be part of our mempool pruning issues but either way this section helps us insure we never have a 7/5 type vote count.   I added comments accordingly.

// Multiple transactions can be in mempool from the same ticket.  We
// need to insure we do not double count these votes.

explorer/mempool.go
```diff
@@ -180,23 +183,30 @@ func (exp *explorerUI) storeMempoolInfo() (lastBlockHash string, lastBlock int64
 
 	var totalOut float64
 	var totalSize int32
+	var votingInfo VotingInfo
+
+	lastBlockHash, lastBlock, lastBlockTime = exp.getLastBlock()
+
+	votingInfo.TotalNeeded = exp.ChainParams.TicketsPerBlock
+	votingInfo.Required = exp.ChainParams.StakeRewardProportion
```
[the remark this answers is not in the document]
papacarp: That makes sense.  Actually I'm not seeing that "Required" (aka "total_choices_required" in json) is used anywhere in the code except for block validity which (as you point out) should really be done by dcrd.  I'm tempted to remove this altogether - see my other comments.

either way, exp.ChainParams.StakeRewardProportion is absolutely not correct.

explorer/mempool.go
papacarp: done

chappjc:
I slated this for 3.0 because it is addressing some usability issues, but it can be deferred to 3.1 if needed.

explorer/mempool.go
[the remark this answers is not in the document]
papacarp: @chappjc or @raedah any feedback on which option to take here?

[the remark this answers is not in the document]
chappjc: I generally agree with 1.  Votes come all at once (within a window of a couple seconds at most), so it's not so useful in the new tx loop.  However, if done after a few second delay, it perhaps makes sense.

But say we wanted to write the validity check correctly, I would not favor RPC calls (which ones, by the way?).  But thinking about options 2. and 3., I do think it is wise not to jump to conclusions about validity until at least 3 up/down votes are received.  We know that 2 up + 2 down = invalid/disapprove, but until a third comes in or the next block is mined, it's technically up in the air.

So I see this not as a boolean but

```go
type BlockValidState int8
const (
	BlockValidYes BlockValidState = iota
	BlockValidNo
	BlockValidUncertain
)
```

explorer/mempool.go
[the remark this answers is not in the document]
chappjc: Thanks.  I understood the reasoning, but it's not clear to someone less familiar just reading the code.  :)

explorer/mempool.go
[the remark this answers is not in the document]
papacarp: Throwing this out there as I'm not familiar with the details:
A.  miners create what they think is a valid block
B.  voters validate the block by running that through a bunch of checks (dcrd) in maybeacceptblock and then even more in checkConnectBlock :

https://github.com/decred/dcrd/blob/9536f0c88fa85180c348043766f6c1d4d42107da/blockchain/accept.go#L104-L116

https://github.com/decred/dcrd/blob/9536f0c88fa85180c348043766f6c1d4d42107da/blockchain/validate.go#L2305-L2323

Items confirmed would be things like:
1.  previous block is valid
2.  check block header
3.  check block max size
4.  check that all transactions are finalized in block
5.  confirm coinbase contains block height
6.  confirm votes are for winning tickets
7.  check for duplicate transactions
8.  check block scripts
9.  check that there is at least 1 regular transaction
10.  check that there is only one coinbase transaction
11.  block header commits to votes and tickets and revocations

I'm sure there is more, and I'm sure this is a partial list, but thats what I saw when I was looking at replicating block validity from dcrd into dcrdata.  

Shall I implement #1 here and remove the reference to validity so we can close out this PR?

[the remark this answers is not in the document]
chappjc: We're only concerned with B.  This is actually why some people, @davecgh for one if I recall correctly, have have emphasized that the voters *approve* or *disapprove*.  Strictly speaking, validation is what you've described primarily in your last comment.  We can assume that dcrd has done its job, and that it is a valid block.

This is mempool information, so all we want to do it consider the votes that are likely to be included in the next block, which approve the current best block, and indicate how the block would be treated (approved or disapproved).

Anyway, yes, I say we not provide an indication of the collective vote choice on the mempool page.  Besides the implementation and timing issues, it's really not certain until a miner includes the votes in a block, which they really don't have to do (although they are incentivized to include all votes).

chappjc:
@papacarp Assigning the review edits to me.  Will provide your branch with a PR for adding to this.

chappjc:
I am just now getting to this, so if you want it back @papacarp, you may.  I'm not sure what else is on your plate. Either way is fine with me.  Just let me know.

chappjc:
I have a PR for your branch @papacarp .  Could you rebase this one on master so the diff is sensible? (I started my work rebased on master for some reason).

chappjc:
Superseded (possibly) by PR https://github.com/decred/dcrdata/pull/634, a combination of work by @RogueElement , @papacarp , and @chappjc.

chappjc:
closed by https://github.com/decred/dcrdata/pull/634

PR #563, opened 2018-07-23
dcrsqlite: test RetrieveAllPoolValAndSize and RetrieveBlockFeeInfo, tier 3

JFixby:
This is tier 3 of the #514 
Testing:
- `dcrsqlite/sqlite.go` `DB.RetrieveAllPoolValAndSize()`
- `dcrsqlite/sqlite.go` `DB.RetrieveBlockFeeInfo()`

### DIFF
--- a/db/dcrsqlite/dcrsqlite_test.go
+++ b/db/dcrsqlite/dcrsqlite_test.go
@@ -30,3 +30,19 @@ func InitTestDB(targetDBFile string) *DB {
 	}
 	return db //is not nil
 }
+
+var reusableEmptyDB *DB = nil
+
+// ObtainReusableEmptyDB returns a single reusable instance of an empty DB. The instance
+// is created once during the first call. All the subsequent calls will return
+// result cached in the reusableEmptyDB variable above.
+func ObtainReusableEmptyDB() *DB {
+	if reusableEmptyDB == nil {
+		testName := "reusableEmptyDB"
+		testutil.ResetTempFolder(&testName)
+		target := filepath.Join(testName, testutil.DefaultDBFileName)
+		targetDBFile := testutil.FilePathInsideTempDir(target)
+		reusableEmptyDB = InitTestDB(targetDBFile)
+	}
+	return reusableEmptyDB
+}
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,6 @@ testnet/
 testdata/
 \.project
 test-data-repo
+dcrdata-testdata-gitlfs
+dcrdata-testdata
+.vendor-new
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -273,6 +273,17 @@
 
 [[projects]]
   digest = "1:e22469bf84466c427efe488bdb825101c527ecfd79f7829f72505c5ec64a7c06"
+  name = "github.com/google/go-cmp"
+  packages = [
+    "cmp",
+    "cmp/internal/diff",
+    "cmp/internal/function",
+    "cmp/internal/value"
+    ]
+  revision = "3af367b6b30c263d47e8895973edcca9a49cf029"
+  version = "v0.2.0"
+
+[[projects]]
   name = "github.com/google/gops"
   packages = [
     "agent",
--- a/db/dcrsqlite/blocksretrieval_test.go
+++ b/db/dcrsqlite/blocksretrieval_test.go
@@ -1,0 +1,50 @@
+package dcrsqlite
+
+import (
+	"testing"
+
+	"github.com/decred/dcrdata/db/dbtypes"
+	"github.com/decred/dcrdata/testutil"
+	"github.com/google/go-cmp/cmp"
+)
+
+func TestEmptyDBRetrieveAllPoolValAndSize(t *testing.T) {
+	testutil.BindCurrentTestSetup(t)
+	db := ObtainReusableEmptyDB()
+
+	result, err := db.RetrieveAllPoolValAndSize()
+	if err != nil {
+		testutil.ReportTestFailed(
+			"RetrieveAllPoolValAndSize() failed: default result expected: %v",
+			err)
+	}
+
+	// Expected value:
+	var defaultChartsData dbtypes.ChartsData
+
+	if !cmp.Equal(*result, defaultChartsData) {
+		testutil.ReportTestFailed(
+			"RetrieveAllPoolValAndSize() failed: default result expected:\n%v",
+			cmp.Diff(*result, defaultChartsData))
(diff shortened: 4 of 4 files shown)


db/dcrsqlite/dcrsqlite_test.go
```diff
@@ -30,3 +32,42 @@ func InitTestDB(targetDBFile string) *DB {
...
+	ExpectEmptyArray(testedFunctionName, "chartsData.SizeF", chartsData.SizeF)
+	ExpectEmptyArray(testedFunctionName, "chartsData.ValueF", chartsData.ValueF)
+	ExpectEmptyArray(testedFunctionName, "chartsData.Unspent", chartsData.Unspent)
+	ExpectEmptyArray(testedFunctionName, "chartsData.Revoked", chartsData.Revoked)
+	ExpectEmptyArray(testedFunctionName, "chartsData.Height", chartsData.Height)
+	ExpectEmptyArray(testedFunctionName, "chartsData.Pooled", chartsData.Pooled)
+	ExpectEmptyArray(testedFunctionName, "chartsData.Solo", chartsData.Solo)
+}
+
+func ExpectEmptyArray(
```
chappjc: Why does this have to take an interface and use reflect / .Len when you could simply test with `len`?  e.g. 

```if len(chartsData.TimeStr) != 0```

File: db/dcrsqlite/dcrsqlite_test.go [hunk 2]
```diff
@@ -30,3 +32,42 @@ func InitTestDB(targetDBFile string) *DB {
...
+//  - all the returned arrays expected to be empty
+//  - chartsData is not null (nil)
+//
+// These checks are sorted according to the ChartsData struct fields order
+func checkChartsDataIsDefault(testedFunctionName string, chartsData *dbtypes.ChartsData) {
+	if chartsData == nil {
+		testutil.ReportTestFailed(
+			testedFunctionName + "() failed: value is nil")
+	}
+	// Array checks:
```
chappjc: I don't see the necessity of all these lines.  If you want to make sure that `chartsData` is initialized to the zero values, just do:

```go
var emptyChartsData dbtypes.ChartsData
if !reflect.DeepEqual(emptyChartsData, *chartsData) {
```

db/dcrsqlite/retrieveallpoolvalandsize_test.go
```diff
@@ -0,0 +1,23 @@
...
+}
+
+func testEmptyDBRetrieveAllPoolValAndSize(db *DB) {
+	result, err := db.RetrieveAllPoolValAndSize()
+	if err != nil {
+		testutil.ReportTestFailed(
+			"RetrieveAllPoolValAndSize() failed: default result expected:",
+			err)
+	}
+	checkChartsDataIsDefault("RetrieveAllPoolValAndSize", result)
```
chappjc: So instead of all that other code, you can just check that the pointer is non-nil, and the data is default with `DeepEqual` as I said.  Or you could use something like https://github.com/google/go-cmp

db/dcrsqlite/retrieveblockfeeinfo_test.go
```diff
@@ -0,0 +1,23 @@
...
+import (
+	"testing"
+
+	"github.com/decred/dcrdata/testutil"
+)
+
+func TestEmptyDBRetrieveBlockFeeInfo(t *testing.T) {
+	testutil.BindCurrentTestSetup(t)
+	db := InitTestDB(DBPathForTest())
+	testEmptyDBRetrieveBlockFeeInfo(db)
```
chappjc: Is it necessary to have `InitTestDB` run repeatedly.  It would be much more concise to have lots of these "empty" tests just use the one empty db you init.  This is one reason I referred you to golang's established test harnesss setup: https://golang.org/pkg/testing/#hdr-Main.

db/dcrsqlite/dcrsqlite_test.go
JFixby: All the arrays above has different types: `uint[]`, `float[]`, `string[]` and so on. Reflect helps here to avoid creating a dedicated function for each array type.

db/dcrsqlite/dcrsqlite_test.go
JFixby: DeepEqual wouldn't provide detailed info on what went wrong (which field is incorrect).

db/dcrsqlite/retrieveblockfeeinfo_test.go
JFixby: Thank you.  I'll look into it.

JFixby: I found the following solution to address the issue: 
```
var ReusableEmptyDB *DB = nil

// ObtainReusableEmptyDB returns a single reusable instance of an empty DB. The instance
// is created once during the first call. All the subsequent calls will return
// result cached in the ReusableEmptyDB variable above.
func ObtainReusableEmptyDB() *DB {
	if ReusableEmptyDB == nil {
		testName := "ReusableEmptyDB"
		testutil.ResetTempFolder(&testName)
		target := filepath.Join(testName, testutil.DefaultDBFileName)
		targetDBFile := testutil.FilePathInsideTempDir(target)
		ReusableEmptyDB = InitTestDB(targetDBFile)
	}
	return ReusableEmptyDB
}
```

db/dcrsqlite/dcrsqlite_test.go
chappjc: As I mentioned in my other comment, you don't need a dedicated function at all. `len(uintArray)`, `len(floatArray)`, all fine.

db/dcrsqlite/retrieveblockfeeinfo_test.go
```diff
@@ -0,0 +1,23 @@
+package dcrsqlite
+
+import (
+	"testing"
+
+	"github.com/decred/dcrdata/testutil"
+)
+
+func TestEmptyDBRetrieveBlockFeeInfo(t *testing.T) {
+	testutil.BindCurrentTestSetup(t)
```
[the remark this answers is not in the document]
JFixby: I really did.

chappjc:
Thank you.  The final request I have is to combine `retrieveallpoolvalandsize_test.go` and `retrieveblockfeeinfo_test.go`.  There's no compelling reason to have the tests in different files.  They  both involve dcrsqlite, the blocks table, and retrieval (even empty DB tests).  The bar for keeping tests in the same file is even lower than this, so please do merge.

JFixby:
Should be good now.

PR #565, opened 2018-07-24
dcrsqlite: test RetrieveBlockSummary*, tier 5

JFixby:
This is tier 5 of the #514 

Testing:
- `dcrsqlite/sqlite.go` `DB.RetrieveBlockSummary()`
- `dcrsqlite/sqlite.go` `DB.RetrieveBlockSummaryByHash()`
- `dcrsqlite/sqlite.go` `DB.RetrieveBlockSummaryByTimeRange()`
- `dcrsqlite/sqlite.go` `DB.RetrieveLatestBlockSummary()`

### DIFF
--- a/db/dcrsqlite/blocksummary_test.go
+++ b/db/dcrsqlite/blocksummary_test.go
@@ -1,0 +1,73 @@
+package dcrsqlite
+
+import (
+	"testing"
+
+	"github.com/decred/dcrdata/testutil"
+)
+
+func TestEmptyDBRetrieveLatestBlockSummary(t *testing.T) {
+	testutil.BindCurrentTestSetup(t)
+	db := ObtainReusableEmptyDB()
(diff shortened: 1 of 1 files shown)


chappjc:
So I'm still good with having some useful utilities and test setup functions, but could you comment on the need for testing all the functions on an empty DB?

chappjc:
What I have been doing is checking in a reference db to the appropriate folder in https://github.com/dcrlabs/bug-free-happiness and using that during testing.  I suggest a DB that is syncd up to a modest height, so that the file is still small, and using that to verify expected output.

FYI, the test data gets accessed by this:

https://github.com/decred/dcrdata/blob/c4864424c5156c8c40983867a8d1a623cbed60e0/run_tests.sh#L50-L52

JFixby:
Sure. As I said, I want to finish with empty DB first.
Once it is merged, non empty-db is the next step.

PR #570, opened 2018-07-26
Fix bugs in addresses page view

dmigwi:
Before :
screen shot 2018-07-26 at 05 23 38

 After:
screen shot 2018-07-26 at 05 19 59

### DIFF
--- a/db/dcrpg/pgblockchain.go
+++ b/db/dcrpg/pgblockchain.go
@@ -610,12 +610,9 @@ func (pgb *ChainDB) AddressHistory(address string, N, offset int64,
 		return addressRows, nil, fmt.Errorf("ReduceAddressHistory failed. len(addressRows) = %d", len(addressRows))
 	}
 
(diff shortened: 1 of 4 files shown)


dmigwi:
Ref https://github.com/decred/dcrdata/pull/489

dmigwi:
**Bug 1: Incorrect SPENT, RECEIVED and UNSPENT values**
- Before the address table format was changed each row could be either 1 funding and 0 spending, or 1 funding and 1 spending but now each row is either a funding or spending tx. Determining if you've fetched all transactions was only done with funding transactions but now both funding and spending transactions are needed to determine if all transactions were fetched.

**Bug 2: Wrong Transactions count on the consolidated debit view.**
- The consolidated debit view transactions count wasn't set thus the transaction count was always zero.

Before:
screen shot 2018-07-26 at 07 58 16

After:
screen shot 2018-07-26 at 07 48 31

Issue #314, opened 2018-01-14
Ticket Pool Size from the explorer block not the same as that of the api block summary


oluwandabira:
But it is the same for the api block verbose endpoint. The api summary gets its info from the sqlitedb while both the explorer and api verbose endpoint get theirs from the rpc. The api summary seems the give the ticket pool value of the next block sometimes and that suggests an bug with the db query.

chappjc:
I believe the data from sqlite (originally from stakedb accounting) is correct... but so is the RPC data.  The issue is that the ticket pool stats from RPC do not reflect the tip block.  I recall discussing this with @davecgh at some point.  I'll have to look into that.

As for what to do...

chappjc:
Yeah, `GetBlockVerboseByHash` returns block data that includes the ticket pool size *before* that block got mined.  But if you did `livetickets` after that block was mined, you would find the pool size could be different.

So, the verbose endpoint (/api/block/best/verbose) that calls that RPC is fine given the interpretation of that data structure I've described.

The explorer on the other hand, this could be confusing to the user.  The pool size there has the same meaning as the RPC result but it's presented in such a way that you would expect the pool size *given the transactions mined in that block*.  The explorer page should query sqlite for this information instead.

davecgh:
The value in the header, which is what `GeBlockVerboseByHash` returns, can't be changed as that would invalidate the block.

chappjc:
Perhaps we accept the differences and address this with different terminology.  For instance, "confirmed live ticket pool" vs. "tip live ticket pool".

ukane-philemon:
@chappjc, is this still an issue?

chappjc:
Non-issue.  The JSON data for a block header is what it is.  The home page (and the /api/stake/pool response) will show ticket pool size that reflects the best block even if the best block's header doesn't reflect itself.

PR #572, opened 2018-07-26
Insight API fixes for upgraded tables

papacarp:
Looks like a few things broke when we upgraded the tables with PR #427.  I'll go through all of insight and retest.

[1] ok
[1.1] ok
[2] ok
[2.1] ok
[3] ok
[4] ok
[5] ok
[5.1] ok
[6] missing addr field - fixed, missing spent fields - fixed
[6.1] ok
[7] ok
[7.1] ok
[8],[8.2] missing transaction list - fixed
[8.1] ok
[9.1] ok
[9.2] ok
[9.3] ok
[9.4] ok
[10] incorrect UTXO - fixed
[11] ok
[12] ok
[13] duplicate transactions - fixed
[14] ok
[15] ok
[16] ok
[17] not tested (send raw transaction)
[19] ok
[20]  ok
[21] ok

### DIFF
--- a/db/dcrpg/internal/addrstmts.go
+++ b/db/dcrpg/internal/addrstmts.go
@@ -7,7 +7,7 @@ const (
 
 	InsertAddressRow = insertAddressRow0 + `RETURNING id;`
 
-	UpsertAddressRow = insertAddressRow0 + `ON CONFLICT (tx_vin_vout_row_id, address, is_funding) DO UPDATE 
+	UpsertAddressRow = insertAddressRow0 + `ON CONFLICT (tx_vin_vout_row_id, address, is_funding) DO UPDATE
 		SET block_time = $7, valid_mainchain = $9 RETURNING id;`
 	InsertAddressRowReturnID = `WITH inserting AS (` +
 		insertAddressRow0 +
@@ -18,7 +18,7 @@ const (
 		SELECT id FROM inserting
 		UNION  ALL
 		SELECT id FROM addresses
-		WHERE  address = $1, is_funding = TRUE 
+		WHERE  address = $1, is_funding = TRUE
 		AND tx_vin_vout_row_id = $5
 		LIMIT  1;`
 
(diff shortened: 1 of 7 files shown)


chappjc:
all ready, except for rebase?

papacarp:
Not yet, I have not finished testing all endpoints.  I saw some potential issues related to UTXO but won’t have time to investigate those until Monday.    It was just some differences between results from existing insight api and dcrdata api.

chappjc:
OK, thanks.  When looking at insight api vs. dcrdata api, there are some known differences such as https://github.com/decred/dcrdata/issues/314
In that case it was a matter of semantics that could be resolved by variable naming, but there may certainly be straight-up bugs.

papacarp:
The issue I saw was just a utxo that showed up on existing insight api but not dcrdata insight api.  I had those pretty well correlated before so either something changed with address tables, I had a dB sync inconsistency, or ...

db/dcrpg/internal/addrstmts.go
```diff
@@ -41,32 +41,43 @@ const (
...
-		where address = $1 order by tx_time desc;`
-
-	SelectAddressesTxnByFundingTx = `SELECT tx_vin_vout_index, tx_hash, tx_vin_vout_index, 
-		block_height FROM addresses LEFT JOIN 
-		transactions on transactions.tx_hash=tx_hash and is_funding = FALSE WHERE 
-		address = $1 and tx_hash=$2;`
-
-	SelectAddressUnspentCountAndValue = `SELECT COUNT(*), SUM(value) FROM addresses 
+	SelectAddressesAllTxn     = `WITH these as (SELECT addresses.tx_hash, ftxd.time as tx_time, ftxd.block_height as height
+		from addresses left join transactions as ftxd on addresses.tx_hash=ftxd.tx_hash
```
dmigwi: On the transactions table a `tx_hash` is only unique within a given block. This means we may have similar tx_hashes in multiple blocks. Therefore doing a left join on the `tx_hash` field might not be as effective as needed.
The only way I think you can fetch the information in the query above effectively is as below:
- Do a left join with vins table id & vouts table id  since they exist in both addresses as `tx_vin_vout_row_id` column and transactions table as `vin_db_ids` or `vout_db_ids` columns.
- Do a data migration to add `block_height` to  the addresses table.

dmigwi: Actually doing a data migration to add the `block_height` back might help a lot with the left joins in the long run.

dmigwi: `block_time` field already exists in the addresses table.

chappjc:
Be aware of the new columns.  valid_mainchain in addresses table

papacarp:
rebased but cannot re-test yet due to an issue with upgrade:

```
2018-08-10 13:04:37.953 [INF] PSQL: Starting blocks table mainchain upgrade at block 00000000000000007c575ef20913a92ebad688333040e33f0c6a2ed7fca28721 (working back to genesis)
2018-08-10 13:04:44.119 [ERR] PSQL: Failed to update next_hash in block_chain for block 00000000000000007ba33357b86af74c3fca7b63f7ad7a2827246f335c237c68
2018-08-10 13:04:44.219 [INF] DATD: Closing connection to dcrd.
2018-08-10 13:04:44.220 [INF] DATD: Bye!
2018-08-10 13:04:44.497 [ERR] DATD: failed to upgrade blocks table to version 3.3.0. Error: upgrade of blocks table ended prematurely after 1408 blocks. Error: SetMainchainByBlockHash for block 00000000000000007ba33357b86af74c3fca7b63f7ad7a2827246f335c237c68 failed: sql: no rows in result set
```

chappjc:
There's somehow missing data in the block_chain table.  That could happen (given a bug that was fixed) if dcrdata was started and stopped twice on consecutive blocks, but without receiving new blocks.  It's not an issue any more hopefully.  Sorry for the inconvenience.  There are DBs for bootstrapping shared in Slack.

db/dcrpg/internal/addrstmts.go
papacarp: If there is a realistic possibility of collisions for TxIDs then I will need to rebuild this since the strategy used is to get all the TxIDs ordered by block/time and then fill in the details for only the transactions that will be displayed.  The premise is that txids are unique.   

I just searched the db and we don't have any duplicate txids for is_valid=true transactions.  We do have 10 duplicates if we include is_valid=false transactions.

I'm not thrilled with this query for performance reasons, but given our desire to get testing underway with insight I think getting it merged in and learning what we don't know might be best.  I'm just pushing an update to only pull is_valid transactions.

papacarp:
this is rebased and retested - should be ready for insight testing.  Please let me know if you want me to deal with the possibility of duplicate transaction ids as it will require a new strategy with some of the queries.

chappjc:
I'm not thrilled with that query either, but your work here is helping shed light on the details that were overlooked.  I'm going to play around with psql.

To clarify, there will not be duplicate row ids of course since they are serials, but there will be dup's of the tx_hash.  This is why the unique index had to be (tx_hash, block_hash).  But as you observer, only one will be mainchain && valid.

Please have a look at the last several comments in https://github.com/decred/dcrdata/pull/580 to see how I've demonstrated it.

papacarp:
Yea, that query got a bit out of hand.  block height is used to filter the return result to a subset that can then be used to compare against mempool - this was done to de-duplicate the transaction list as sometimes transactions are in both places.  We could eliminate that if we tighten up mempool to always trim transactions that have been mined.

The other reason for the joins is the the insight API wants transactions displayed and ordered with transaction time, not block time.  That might be a nuance we can just overlook, but I felt it was important to use the actual transaction time wherever possible.  Furthermore debit/credit transactions need to be interleaved as they occur and not displayed separately or will all vin/vout details as we are currently doing in the address view.  

Having a built in indexed method of correlating the address table to the transaction table seems like an important thing to include - we could do that with just an addition to the current addresses table like tx_row_id - we just might need two of them though - one for funding tx and one for spending tx.  

As you know, I'm also a strong proponent of having another (or replacement) table with transactions ordered by transaction time that we can easily offset and limit into using queries to pull up the full debit/credit ordered transaction list for a particular address or set of addresses.

Something like:
address | transaction_time | block_height | tx_row_id

Of if we want to minimize joins (not sure why we would at the expense of even more table data) we could add a little more detail like "tx_hash | amount"

Just some thoughts.  I will be back on it tomorrow afternoon.

papacarp:
@chappjc I think this is ready to go.  Please let me know if there is something I need to do about the travis issue on 1.10.   I ran dep ensure on my local but did not checkin the lock file as its derived.  anything I need to do differently?

chappjc:
No, you're good.  I'm just trying a few things out.

chappjc:
We chatted a bit on slack and the following queries are being considered.

For `SelectAddressesTxnByFundingTx`, something similar to the following, except with a transactions join to get `block_hash` instead of `block_time`:

```sql
SELECT
    s.block_time,
    s.tx_hash AS spending_hash,
    s.tx_vin_vout_index AS vin_index,
    s.matching_tx_hash AS funding_hash,
    f.tx_vin_vout_index AS vout_index 
FROM
    addresses s 
    JOIN
        addresses f 
        ON f.matching_tx_hash = '51da0cc56ca8886efdbf783cb627f5793e88e7b3bc6e84857fed1ab6b89485b9' 
        AND f.is_funding = TRUE 
WHERE
    s.address = ANY('{"TsonmCnrgEd5ptsF8VYtnLyUE6uFJy7x892"}') 
    AND 
    (
        s.tx_hash = '51da0cc56ca8886efdbf783cb627f5793e88e7b3bc6e84857fed1ab6b89485b9' 
        AND s.is_funding = FALSE
    )
;
```

For `SelectAddressesAllTxn`:

```sql
SELECT DISTINCT
    ON(transactions.time, transactions.tx_hash) transactions.tx_hash,
    transactions.time,
    block_height,
    is_funding,
    is_valid,
    is_mainchain,
    valid_mainchain 
FROM
    addresses 
    INNER JOIN
        transactions 
        ON addresses.tx_hash = transactions.tx_hash 
        AND is_mainchain = TRUE
WHERE
    address = ANY('{"TsUgA1cEKDeUxcytP7FSuC8N7B64ePk3bGN", "TcrypGAcGCRVXrES7hWqVZb5oLJKCZEtoL1"}') AND valid_mainchain=true
ORDER BY
    time DESC,
    transactions.tx_hash ASC;
```

papacarp:
@chappjc this should be ready to go.

papacarp:
@chappjc please let me know your thoughts on this one.  I left it as a separate function rather than try to add on to `SpendingTransaction` or `SpendingTransactions` due to the extrain join required to get block height.

PR #583, opened 2018-07-31
Display the tiny qr code thumbnail by default

dmigwi:
Issue Fixed.
- [x] Display the Tiny QR code thumbnail by default.

**Before:**

screen shot 2018-08-01 at 01 23 09

**After:**

screen shot 2018-08-01 at 01 19 47

### DIFF
--- a/views/address.tmpl
+++ b/views/address.tmpl
@@ -17,7 +17,7 @@
                     {{.Address}}<a
                         id="qrcode-init"
                         href="javascript:showAddressQRCode('{{.Address}}');"
-                        class="dcricon-qrcode jsonly no-underline color-inherit p10"
+                        class="dcricon-qrcode no-underline color-inherit p10"
                     ></a>
                 </div>
                 <div class="row">
--- a/views/address.tmpl
+++ b/views/address.tmpl
@@ -17,7 +17,7 @@
                     {{.Address}}<a
                         id="qrcode-init"
                         href="javascript:showAddressQRCode('{{.Address}}');"
-                        class="dcricon-qrcode no-underline color-inherit p10"
+                        class="dcricon-qrcode jsonly no-underline color-inherit p10"
                     ></a>
                 </div>
                 <div class="row">
@@ -259,6 +259,8 @@
     </div>
 
     <script>
+        $(".jsonly").show()
+        
         $("#txntype").change(function(ev) {
             Turbolinks.visit(
                 window.location.pathname


chappjc:
but it's not clickable without JS is it?

dmigwi:
It needs Js.

chappjc:
So just to be clear, the idea is to show it anyway?  I'm OK with that, but it would be best if it showed "enable JavaScript to view full QR code" when hovering on the disabled QR icon.

gozart1:
`$(".jsonly").show()` used to run when the app boots up and on page transitions. 
I looks like that get removed by accident, so could just add that back in.
But there might be a better way, such as using javascript in the initial load to add a bit of css into the global header which is persisted between turbolink page transitions. 
```
body .jsonly {
  display: block
}
```
That way we don't need to worry about page changes since the css would be overridden permanently.

chappjc:
Oh, I didn't realize `jsonly` just wasn't working anymore.  That CSS fix sounds like the right approach.

views/address.tmpl
```diff
@@ -17,7 +17,7 @@
                     {{.Address}}<a
                         id="qrcode-init"
                         href="javascript:showAddressQRCode('{{.Address}}');"
-                        class="dcricon-qrcode jsonly no-underline color-inherit p10"
+                        class="dcricon-qrcode no-underline color-inherit p10"
```
chappjc: @gozart1 suggested having this CSS loaded into the global header by javascript in the initial page load:

```css
body .jsonly {
  display: block
}
```

Could you give that a shot (while retaining the `jsonly` class for this link)?

dmigwi: ok. Noted.

chappjc: I'm not sure how @gozart1 would do this, but I thought he meant to inject the CSS

```css
.jsonly {
  display: block;
}
```

into the html header so it overrides the default in main.css of:

```css
.jsonly {
  display: none;
}
```

https://www.w3schools.com/cssref/pr_class_display.asp

Note that `block` does not mean to block it from displaying, it means to show it as a block element.

chappjc: So maybe in `<head>`:

```html
<script type="text/javascript">
  document.write("<style>body .jsonly {display :block }</style>");
</script>
```

dmigwi: I made the page work by just adding `$(".jsonly").show()` to the page's css but I can still inject the  css.

dmigwi: Its also been used here https://github.com/decred/dcrdata/blob/master/views/extras.tmpl#L99

chappjc: Gotcha.

PR #578, opened 2018-07-27
[wip] dcrsqlite: test BestBlockHash, BestBlockHeight, StakeInfoHeight

JFixby:
Testing:
- `dcrsqlite/sqlite.go` `DB.GetBestBlockHash()`
- `dcrsqlite/sqlite.go` `DB.GetBestBlockHeight()`
- `dcrsqlite/sqlite.go` `DB.GetStakeInfoHeight()`

This PR awaits for (blocked by):
 - #584 provides required functionality for testing (loads db and expected outputs)

PR #584, opened 2018-08-01
[wip] testutil, dcrsqlite: add tooling for non-empty DB test, verifying expected outputs

JFixby:
The goal of this PR is to provide some functions for non-empty DB tests.

The idea is to store pre-filled DB together with expected outputs (test descriptor file) on a test-data repo.

The test setup downloads test-data, loads db, runs db-calls and compares results with expected outputs stored in related test descriptor.

Proposed changes:

1) In `testutil` there were added some basic I/O operations for testing and debugging: simplified calls `ToJson/FromJson()`, `Read/Write StringToFile()` and `FullPathToFile()`

2) `dcrsqlite` uses the functions above to load pre-filled DB and expected outputs to run DB tests.
See usage example in the [#578/getbestblockhash_test.go](https://github.com/decred/dcrdata/pull/578/commits/3f5ef98fc4834867b5dc44d73b2b39b7885a329c#diff-e8ada86f0b74b8bbdcaab02f9316862bR26) 

Test data structure:
 - Each test has a dedicated folder. The folder name is also the test name. Example: the `synced_up_to_260241`.
 - Inside each test folder, there is compressed db file `test-data.tar.xz`. Unpacking it gives `test-data.db` file.
 - Also there is the `test-description.json` containing information related to the test: name, expected values and so on.

image

`synced_up_to_260241` - is a fully synced db up to the block `# 260241`

3) In this PR testdata are stored at the https://github.com/JFixby/dcrdata-testdata-gitlfs repo.
`run_tests.sh` is modified to retrieve it. I need an opinion here about that.

### DIFF
--- a/testutil/blockio.go
+++ b/testutil/blockio.go
@@ -1,0 +1,99 @@
+package testutil
+
+import (
+	"bytes"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strconv"
+
+	"github.com/decred/dcrd/dcrutil"
+	"github.com/decred/dcrd/wire"
+)
+
+// BlockFilename generates file name for a Block
+func BlockFilename(index uint64) string {
+	return "block-" + padNumberWithZero(index, 10) + ".bin"
+}
+
+func padNumberWithZero(value uint64, zeroes uint64) string {
+	return fmt.Sprintf("%0"+strconv.FormatUint(zeroes, 10)+"d", value)
+}
+
+// SaveBlockToFile writes Block to a file in the target folder
+func SaveBlockToFile(block *dcrutil.Block, targetFolder string) (*string, error) {
+	bytes, err := block.Bytes()
+	if err != nil {
+		Log(" failed", err)
+		return nil, err
+	}
+
+	index := block.MsgBlock().Header.Height
+	filename := filepath.Join(targetFolder, BlockFilename(uint64(index)))
+	filename, err = filepath.Abs(filename)
+
+	if err != nil {
+		return nil, err
+	}
+
+	Log("writing", filename)
+	parent := filepath.Dir(filename)
+	err = os.MkdirAll(parent, 0755)
+	if err != nil {
+		Log(" failed", err)
+		return nil, err
+	}
+	err = ioutil.WriteFile(filename, bytes, 0777)
+
+	if err != nil {
+		Log(" failed", err)
+		return nil, err
+	}
+
+	return &filename, nil
+}
+
+// ReadBlock reads Block from file
+func ReadBlock(file string) (*dcrutil.Block, error) {
+	//Log("reading", file)
+	file, err := filepath.Abs(file)
+
+	if err != nil {
+		return nil, err
+	}
+
+	data, err := ioutil.ReadFile(file)
+
+	if err != nil {
+		return nil, err
+	}
+
+	var msgBlock wire.MsgBlock
+	err = msgBlock.Deserialize(bytes.NewReader(data))
+	if err != nil {
+		return nil, err
+	}
+	block := dcrutil.NewBlock(&msgBlock)
+
+	return block, nil
+}
+
+func PreLoadBlocks(tag string, fromIndex uint64, toIndex uint64) map[int64]*dcrutil.Block {
+	testBlockchain := make(map[int64]*dcrutil.Block)
+	for i := fromIndex; i <= toIndex; i++ {
+		index := int64(i)
+		blockFileName := BlockFilename(i)
+		target := PathToTestDataFile(tag, blockFileName)
+		block, err := ReadBlock(target)
+		if err != nil {
+			ReportTestIsNotAbleToTest("error reading", target)
+		}
+
+		testBlockchain[index] = block
+		if block == nil {
+			ReportTestIsNotAbleToTest("block not found", i)
+		}
+	}
+	return testBlockchain
+}
--- a/testutil/io.go
+++ b/testutil/io.go
@@ -1,0 +1,81 @@
+package testutil
+
+import (
+	"crypto/md5"
+	"encoding/hex"
+	"encoding/json"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+)
+
+func ToJson(object interface{}) string {
+	bytes, err := json.MarshalIndent(object, "", "    ")
+	if err != nil {
+		ReportTestIsNotAbleToTest(
+			"Failed to produce json for %v: %v",
+			object,
+			err)
+	}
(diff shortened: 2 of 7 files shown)


testutil/testdatahandler.go
```diff
@@ -0,0 +1,105 @@
...
+		ExpectedStakeInfoHeight:                 -1,
+		ExpectedBestBlockHash:                   "",
+		ExpectedMD5OfRetrieveBlockFeeInfoString: "d41d8cd98f00b204e9800998ecf8427e",
+	}
+
+	jsonString := ToJson(exampleDescriptor)
+
+	Log("writing", targetFile)
+	Log(jsonString)
+	WriteStringToFile(targetFile, jsonString)
```
chappjc: why marshal to a string and then write string when you can Encode to the .json file?

testutil/testdatahandler.go
```diff
@@ -0,0 +1,105 @@
...
+	return PathToTestDataFile(tag, TestDescriptorFileName)
+}
+
+func ProduceExampleTestDescriptor(targetFile string) {
+	exampleDescriptor := TestDescriptorJson{
+		TestName:                                "exampleTest",
+		ExpectedBestBlockHeight:                 -1,
+		ExpectedStakeInfoHeight:                 -1,
+		ExpectedBestBlockHash:                   "",
+		ExpectedMD5OfRetrieveBlockFeeInfoString: "d41d8cd98f00b204e9800998ecf8427e",
```
chappjc: Why are you planning to compare MD5 values of the result instead of the actual result?  It's not a great byte savings.  It would be more helpful, especially when dealing with non-zero value results, to have the expected result right here.

testutil/testutil.go
```diff
@@ -41,7 +41,7 @@ var UseSpewToPrettyPrintArrays = true
 var PanicOnTestSetupFailure = true
 
 // Set this flag "true" to print a stack stace when a test failed
-var PanicOnTestFailure = false
+var PanicOnTestFailure = !false
```
chappjc: I guess you used `!` to switch over quickly on your end, but please change back to false.

testutil/log.go
```diff
@@ -0,0 +1,29 @@
...
+func log(tag string, a interface{}) {
+	if a == nil {
+		fmt.Println(tag)
+		return
+	}
+
+	// a!=nil
+	array := reflect.ValueOf(a)
+	if array.Kind() == reflect.Array || array.Kind() == reflect.Slice {
+		fmt.Println(ArrayToString(tag, a))
```
chappjc: Using `fmt.Println` while running tests is no good.  This is why I complained last time about not being able to have access to `testing.T` in these functions -- you need to be doing `t.Logf` so that you get it logged to the console with `go test -v`.

testutil/testdatahandler.go
```diff
@@ -0,0 +1,105 @@
...
+	"path/filepath"
+)
+
+const (
+	TestDataFolderName     = "dcrdata-testdata"
+	TestDescriptorFileName = "test-description.json"
+)
+
+type TestDataHandler struct {
+	expected TestDescriptorJson
```
chappjc: I don't see the need for `TestDataHandler`.  The methods can have `(*TestDescriptorJson)` receivers.

testutil/testdatahandler.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+func (db *TestDataHandler) GetExpectedStakeInfoHeight() int64 {
+	return db.expected.ExpectedStakeInfoHeight
+}
+
+func (db *TestDataHandler) GetExpectedBestBlockHash() string {
+	return db.expected.ExpectedBestBlockHash
+}
+func (db *TestDataHandler) GetExpectedMD5OfRetrieveBlockFeeInfoString() string {
```
chappjc: None of these `Get` functions are needed either.  The fields of `TestDescriptorJson` are all exported.

testutil/testdatahandler.go
```diff
@@ -0,0 +1,105 @@
...
+	return db.expected.ExpectedMD5OfRetrieveBlockFeeInfoString
+}
+
+func LoadTestDataHandler(tag string) *TestDataHandler {
+	var tdb TestDataHandler
+	testDescriptorFile := PathToTestDescriptorFile(tag)
+
+	Log("reading:")
```
chappjc: Why is this in a separate log statement?

testutil/testdatahandler.go
```diff
@@ -0,0 +1,105 @@
...
+	Log("testDescriptorFile", testDescriptorFile)
+	// uncomment this to produce example descriptor file
+	//ProduceExampleTestDescriptor(PathToTestDescriptorFile(tag))
+
+	testDescriptor := ReadTestDescriptorJson(testDescriptorFile)
+	tdb = TestDataHandler{
+		expected: testDescriptor,
+	}
+
+	return &tdb
```
chappjc: consider the simplification here too if you get rid of the unneeded struct type: `return ReadTestDescriptorJson(testDescriptorFile)`

testutil/log.go
```diff
@@ -0,0 +1,29 @@
...
+
+func log(tag string, a interface{}) {
+	if a == nil {
+		fmt.Println(tag)
+		return
+	}
+
+	// a!=nil
+	array := reflect.ValueOf(a)
+	if array.Kind() == reflect.Array || array.Kind() == reflect.Slice {
```
chappjc: More avoidable complexity.  What about the spew stuff you wired up?  I really hate seeing `reflect` used so liberally like this.  An exception I make is `reflect.DeepEqual`, when appropriate.

testutil/io.go
```diff
@@ -0,0 +1,81 @@
...
+	if err != nil {
+		ReportTestIsNotAbleToTest(
+			"Failed to build abs path for %v: %v",
+			relative,
+			err)
+	}
+	return abs
+}
+
+func MD5ofString(str string) string {
```
chappjc: Why choose a non-cryptographic (broken) hash function?

testutil/blockio.go
```diff
@@ -0,0 +1,99 @@
...
+	"path/filepath"
+	"strconv"
+
+	"github.com/decred/dcrd/dcrutil"
+	"github.com/decred/dcrd/wire"
+)
+
+// BlockFilename generates file name for a Block
+func BlockFilename(index uint64) string {
+	return "block-" + padNumberWithZero(index, 10) + ".bin"
```
chappjc: are you planning to use `padNumberWithZero` often?  I would just have said:

`return fmt.Sprintf("block-%010d.bin", index)`

Simpler, shorter, more readable.

Issue #587, opened 2018-08-02
agendadb: location and permissions of agenda db file


chappjc:
Some issues with the agendas db file that need adjustment:
* The agenda db file is hard coded
* The location of the agenda db file should be same as `datadir`.
* `GetAllAgendas` has unhandled errors.

chappjc:
`GetAllAgendas` should handle the `Open` error instead of just logging it:

Change to something like:
```go
// GetAllAgendas returns all agendas and their info in the db.
func GetAllAgendas() (agendas []*AgendaTagged, err error) {
	var adb *AgendaDB
	adb, err = Open(dbName)
	if err != nil {
		log.Errorf("Failed to open new DB: %v", err)
		return
	}
	err = adb.sdb.All(&agendas)
	if err != nil {
		log.Errorf("Failed to get data from DB: %v", err)
	}

	err = adb.Close()

	return
}```

Not that `err` was also shadowed previously

dmigwi:
ok. Noted

chappjc:
Partially resolved by PR https://github.com/decred/dcrdata/pull/594, but still need agendas db in datadir folder.

PR #594, opened 2018-08-03
agendadb: Fix unhandled errors while fetching all agendas

dmigwi:
When fetching all agendas if opening the db failed, the application still proceeded to use an invalid agendas db connection failed.

- [x] Fixed unhandled errors

https://github.com/decred/dcrdata/issues/587

### DIFF
--- a/db/agendadb/db.go
+++ b/db/agendadb/db.go
@@ -217,22 +217,23 @@ func GetAgendaInfo(agendaId string) (*AgendaTagged, error) {
 
 // GetAllAgendas returns all agendas and their info in the db.
 func GetAllAgendas() (agendas []*AgendaTagged, err error) {
-	adb, err := Open(dbName)
+	var adb *AgendaDB
+	adb, err = Open(dbName)
 	if err != nil {
-		log.Errorf("Failed to open new DB: %v", err)
+		log.Errorf("Failed to open new Agendas DB: %v", err)
 		return
 	}
 
 	defer func() {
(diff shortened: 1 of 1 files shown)


dmigwi:
@chappjc Please have a look at this when you get the time.

db/agendadb/db.go
```diff
@@ -220,13 +220,20 @@ func GetAllAgendas() (agendas []*AgendaTagged, err error) {
 	adb, err := Open(dbName)
 	if err != nil {
 		log.Errorf("Failed to open new DB: %v", err)
+		return
 	}
+
+	defer func() {
+		err = adb.Close()
+		if err != nil {
+			log.Errorf("Failed to close the DB: %v", err)
```
chappjc: Please mention agenda DB, otherwise it could be confusing what threw the error.

db/agendadb/db.go
```diff
@@ -220,13 +220,20 @@ func GetAllAgendas() (agendas []*AgendaTagged, err error) {
 	adb, err := Open(dbName)
```
chappjc: I believe err here shadows the err declared in the function signature

db/agendadb/db.go
dmigwi: Ok. Noted

db/agendadb/db.go
dmigwi: Ok. Noted

Issue #451, opened 2018-04-27
New Design Implementation: Phase 1


denyszayets:
Hello to all,

Recently design team completed the redesign of explorer.dcrdata.org. We plan to divide this restyling process into a few milestones/phases that will help us all make this project successful. 

Based on my conversations with @raedah, his team is happy to take on this project in a few phases.

A subdomain `redesign.dcrdata.org` should work.

**Phase 1**
Application of visual and style changes (HTML, CSS) to the **homepage**, utilizing features available at the moment. 

- Header
- Search Box
- Menu (button only phase 1)
- Footer
- Network Statistics
- Decred Blockchain

All links from the homepage can be pointed to non-styled pages with a default template.

Items highlighted in **red** will be developed in **phase 2** but will require developers to keep in mind these changes during the development of phase 1.

Attached below is the screenshot highlighting **items to exclude** in phase 1.

Prototype Link: https://xd.adobe.com/view/b0dcf31d-cc18-450c-acfa-2b5d223ffcad/
Design Specifications: https://xd.adobe.com/spec/e9f3f22a-7eb2-4bef-8921-9fc4b04a8b6d/ 
_(this page will have all the details on the colors, size, spacings, fonts, etc.)_

Looking forward to getting our explorer more average user-friendly. Please let me know if I missed anything. Thank you all.

![dcrdata phase 1 2x](https://user-images.githubusercontent.com/33739891/39410128-0b3a0bb8-4bc1-11e8-9c89-d7d6b0fa24a0.png)

Mampool Design
![39456446-80b90dfa-4cb4-11e8-89e9-89ad6d343b01](https://user-images.githubusercontent.com/33739891/39491479-8fbb9a38-4d5a-11e8-9860-12b2a65ef5cc.png)

denyszayets:
Updated mockup and prototype with Phase 1 elements only.

denyszayets:
1. Removing General Block for now.
2. Staking block stays the same
3. Mining block: replacing Hashrate with Available Supply.

updated layout coming soon.

xaur:
Posting my feedback per @gozart1 's request

Overview:

* Total DCR Locked, Available Supply: no need for 4-digit precision, even 0 would work.
* Avg Locked 24.4 days -- is this number correct? I thought it's 28 days + immaturity periods.
* Ticket Price, PoS Reward, PoW Reward: the smaller numbers (3-4th decimal places) don't add much
[...]

PR #598, opened 2018-08-08
chart theming, url backed state & universal zoom

gozart1:
Changes include:

- Use [History pushState](https://developer.mozilla.org/en-US/docs/Web/API/History_API) to add sharable urls for each chart
- Night mode chart theme
- Use color palette from https://github.com/decred/dcrdata/issues/451 for series colors
- Utilize all available horizontal screen real estate
- Enable client side zoom helper for all charts
- Rename "Tickets By Output" to "Ticket Purchase"
- Remove grouping helper for "Ticket Purchase" chart and instead show them as separate options: "Ticket Purchases By Block" and "Ticket Purchases By Price Window" ( simplifies ui & url scheme )
- Move hover state into external div instead of overlaying it on the chart ( this way we get a permanently visible legend, which is useful for screenshots and touch interfaces )

I'd like to further improve the url backing by including the selected range or zoom. But I think we should consider hooking that into the backend. That way we can set a default zoom such as one month and only load the full range if a users actually wants that. This will make the charts load a whole lot faster. In any case, url backed state for zoom/selection should be separate PR since it probably warrants further discussion and there is no need for it to be a blocker to this PR.

Day
Night


linnutee:
Can consider aligning "Zoom: Day, Week, Month" with the rest of the controls UI style/colors, so it wouldn't be competing with the chart elements.

chappjc:
For the Gopkg.lock conflicts, there should be 5 `revision` changes that you'll want to take, and you can ignore all the `digest =` lines.  But after the resolve, and before commit of the rebase, I'd update dep (go get **-u** -v github.com/golang/dep/cmd/dep) and run a `dep ensure` to get it all spiffy.

chappjc:
I think this is pretty much needed for 3.0, so slated as such.

xaur:
Nice colors!

Ideas:

* Lay out "Window/Solo/Pooled" horizontal, in same lane as other controls, when there's enough width.
* When not enough width or if you decide to keep this block vertical, at least right-align the numbers.
* Add ticket price data points and second Y axis for that on the right.

gozart1:
Thanks for all the feedback, very helpful!
Here's another rev on the chart controls and legend that seems well to work under a wide variety of circumstances ( labels + values + screen sizes ).

Added a control for applying a moving average to the dataset. The rollups are more interesting for some charts vs others. So, I'm not sure if we should always show it, but I don't see why not at this point since there aren't that many controls competing for attention. Also, since the computation is done client side, it can get a bit heavy when computing a large moving average for large datasets ( I've set a max moving average period to 10,000, for the sake of putting some sort of upper limit to prevent completely freezing up the browser tab ).

tickets-month
tickets-raw

xaur:
Is there a reason why you changed the zoom buttons sorting from ascending to descending scale?

gozart1:
@xaur Since 'All' is the default it just seems more natural to me for the default to be the first option. 
The reason 'All' is the default is because the current implementation loads all the entire dataset, so we might as well show it after putting the user through the hard work of loading and parsing it.

As mentioned above, I think we should probably switch to server-side zoom in the future, and default to only loading a recent portion of the dataset (or some sort of aggregate) for the sake of performance. 

But the main reason to go from bigger to smaller is because that lines up with how the range slider will update. The further right an option is on the zoom helper, the further to the right the start of the selected range will be. Again, subtle, but it just feels more natural to me for those controls to be aligned vs inverted.

xaur:
@gozart1 thanks for sharing. I noticed it because it stands out compared to most charts I used to see (coinmarketcap, onchainfx, etc) that all have their buttons go from smaller to bigger scale left to right. But your considerations are reasonable and interesting. No hard opinion here, curious to see how it goes!

gozart1:
Good point. Now that I look around they really do all go from smaller to bigger. But those are all trading sites, it must be a conspiracy to bias people towards thinking along short term time horizons! j/k

But in all seriousness, there likely is a very good reason it's so ubiquitous. Maybe it is something along the lines of this: If a user is reaching for the zoomer, then prioritizing the option which is _not_ zooming ('All') is less ergonomic. Genuinely curious now how this came about.

I'm happy to flip it back to ascending. I didn't realize this was such a convention and think flouting UX convention should be avoided unless there is a strong case.

xaur:
Haha, long term by default. I like it.

chappjc:
Been testing and I love it.  But after the mega reorg, we clearly need to limit the queries to mainchain only (TODO in future backend PR).  From the transactions per block chart:

image

chappjc:
The ticket purchases by price window could use some date or block number info on the x axis too, but if that requires API changes it can wait too.

chappjc:
To reiterate my discussion with Migwi, we'll have to reconsider how we present the Tickets By Output Count as it's really impossible to say pool or solo based on number of outputs.  The best I think we can do without being misleading is to simply show 3 output / 5 output / other.

chappjc:
BTW, I'm really happy with the response times.  The backend served every API endpoint in <1ms each.

```"GET http://127.0.0.1:8777/api/chart/tx-per-day HTTP/1.1" from 127.0.0.1:52600 - 200 275B in 35.687µs
"GET http://127.0.0.1:8777/api/chart/tx-per-block HTTP/1.1" from 127.0.0.1:52622 - 200 43393B in 367.822µs
"GET http://127.0.0.1:8777/api/chart/pow-difficulty HTTP/1.1" from 127.0.0.1:52622 - 200 1035B in 35.849µs
"GET http://127.0.0.1:8777/api/chart/coin-supply HTTP/1.1" from 127.0.0.1:52654 - 200 131507B in 954.794µs
"GET http://127.0.0.1:8777/api/chart/fee-per-block HTTP/1.1" from 127.0.0.1:52686 - 200 95170B in 818.505µs
"GET http://127.0.0.1:8777/api/chart/coin-supply HTTP/1.1" from 127.0.0.1:52708 - 200 131507B in 909.299µs
"GET http://127.0.0.1:8777/api/chart/fee-per-block HTTP/1.1" from 127.0.0.1:52730 - 200 95170B in 875.099µs
"GET http://127.0.0.1:8777/api/chart/duration-btw-blocks HTTP/1.1" from 127.0.0.1:52730 - 200 47602B in 598.113µs
"GET http://127.0.0.1:8777/api/chart/ticket-spend-type HTTP/1.1" from 127.0.0.1:52764 - 200 27355B in 312.142µs
"GET http://127.0.0.1:8777/api/chart/ticket-by-outputs-windows HTTP/1.1" from 127.0.0.1:52854 - 200 307B in 35.118µs
"GET http://127.0.0.1:8777/api/chart/ticket-by-outputs-blocks HTTP/1.1" from 127.0.0.1:53068 - 200 28313B in 446.388µs
```

This should be reasonable for the public server to handle.

public/js/util.js
```diff
@@ -0,0 +1,8 @@
+// a place for handy functions utility that can be shared accross the client side application
```
chappjc: across

views/charts.tmpl
```diff
@@ -5,69 +5,142 @@
...
+
+        <div data-target="charts.chartWrapper" class="chart-wrapper pl-2 pr-2">
+            <div
+                data-target="charts.chartsView"
+                style="width:100%; height:76vh; margin:0 auto;">
+            </div>
+            <div class="spinner-wrapper">
+                <div class="spinner-centerer d-flex align-items-center justify-content-center">
+                    <div class="spinner">
+                        <i></i><i></i><i></i><i></i><i></i><i></i><i></i><i></i><i></i><i></i><i></i><i></i>
```
chappjc: What do all these `<i>`s do?

gozart1: They are the spokes of the spinner wheel
spinner

gozart1:
I've added the block range of price window since we can just compute it on the client.
Changed the Ticket Purchase charts back to Ticket Output Counts. What do you think of the labels suggesting the inference "(likely Solo)" ?
likely

gozart1:
Other than settling on a nomenclature for the Ticket Outputs charts, I think this PR is almost ready to go. 

One last item is either to remove the query param parser lib I've added here https://github.com/decred/dcrdata/pull/598/files#diff-5bab677d36ae3694a8a856712a6b0e81, or put it to use. What I had in my is to represent all the chart controls via history pushState in the url, so that an exact view is sharable via a link.
 
The proposed structure is:
 ```/charts#?type=${chart_type}&start=${start}&end=${end}&rollPeriod=${rollPeriod}```

So if blocks 1 - 12,356 of Coin Supply is selected the url would be:
```/charts#?type=coin-supply&start=1&end=12356```

When a chart initially loads, or if a specific zoom is clicked the url would look like this
```/charts#?type=coin-supply&zoom=0```
So a url would either have zoom or start/end, never both.

Another thing to note is in this design, everything before the `#` is considered server-side param, everything after it is a purely the concern of the client side.

It's a bit complicated, so I backed out on implementing it for now. 
If we implement this here, the end user would expect it to work like that in other places as well. So I want to make sure everyone is on board before getting too deep.

I'm leaning towards omitting url backed state for now, and revisit this one we have more views with advanced client side state as that may surface some additional considerations that might help us decide on things like wether to bring in a dedicated library for client side routing.
 
@dmigwi Any thoughts on this since you are in the thick of it with your charts work?

gozart1:
OK, have removed the unused query param parsing lib. Let's merge this and hash out an appropriate client side routing architecture with a separate Issue/PR.

PR #604, opened 2018-08-10
go fmt master

JFixby:


chappjc:
go fmt changed in go 1.11.  we're just ignoring the differences for now

PR #605, opened 2018-08-11
Add color themes for different networks

### LINKED ISSUE #494: Change color theme when viewing testnet
As a visual indicator, use a different color theme when on a different network.

codemaestro64:
This applies different color themes depending on the depending on the current network. 
This fixes #494


chappjc:
I'll test and comment on certain aspects, but I'd love a second set of eyes for the html/css bits.  @gozart1, any comments on the theme implementation here?

views/agenda.tmpl
```diff
@@ -4,7 +4,7 @@
     <link href="/css/charts.css" rel="stylesheet">
     <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/dygraph/2.1.0/dygraph.min.css" />
     {{template "html-head" printf "Decred Vote Agenda Pages"}}
-    <body>
+    <body class="{{ .Theme }}">
```
chappjc: OK, but consider that you could avoid the `Theme` struct field in the page's data by doing something like `{{print "theme-" .NetName}}`

chappjc: There could even be `{{$Theme := print "theme-" .NetName}}` in the header I think.

chappjc: Well, I've neglected `ToLower`, but there are several ways to do this anyway.   You can add to the `template.FuncMap`:

```go
"lower": func(s string) string {
	return strings.ToLower(s)
},
```

or cut to the chase and edit `makeTemplateFuncMap`:

```go
func makeTemplateFuncMap(params *chaincfg.Params) template.FuncMap {
	netTheme := "theme-" + strings.ToLower(netName(params))
	return template.FuncMap{
...
		"theme": func() string {
			return netTheme;
		}
...
```

That would capture the netTheme for the current network in the closure so that it is only generated once, and it only needs to be in this one place instead of several places in each page handler.

codemaestro64:
@chappjc requested changes have been made.

codemaestro64:
Screen grabs of the different themes 
Mainnet
screenshot from 2018-08-15 22-59-12

Testnet 
screenshot from 2018-08-15 22-56-59

Simnet
screenshot from 2018-08-15 23-00-24

gozart1:
I think the concept of color coding the networks is a great idea, but I think it will be far more maintainable to simply color the header and footer with test net and sim net colors, and leave the main body the same for all networks. 

Otherwise, we effectively end up creating and having to maintain 6 themes, one dark and one light for each network. 

The problem with coloring the whole body background is it creates subtle contrast issues, especially when we get more into data visualization which involves the use of multiple colors against the backdrop of the main body background. It's tough enough to find 5 colors that look good together in a chart. We've already compounded the challenge with dark/light mode ( guilty as charged ) and we've already had several design bugs where the night theme colors where forgotten about. So I'm against compounding this challenge further, especially if we can achieve the same end result in a way that does not increase the styling complexity.

chappjc:
I'm still seeing different background colors for the different networks.  If I understood @gozart1's comment, the would be best to keep the body background the same for each network, just change the header and footer colors.

PR #606, opened 2018-08-12
added TargetTimePerBlock in views/parameters.tmpl

SeaLightHouse:
before
targettimeperblock


SeaLightHouse:
OK in this page https://github.com/decred/dcrd/blob/master/chaincfg/params.go#L269, the description is as follows : Desired amount of time to generate each block . 
I added the word desired . Hope that works .

chappjc:
I won't argue with the `chaincfg` docs, but the description should still be better for a user-facing website, I would argue.  This is good to have in now anyway.

PR #612, opened 2018-08-13
simplify the dummy address page

### LINKED ISSUE #358: Deny search for DsQxuVRvS4eaJ42dhQEsCXauMWjvopWgrVg

JFixby:
Handles DsQxuVRvS4eaJ42dhQEsCXauMWjvopWgrVg

Closes #358

### DIFF
--- a/blockdata/blockdata.go
+++ b/blockdata/blockdata.go
@@ -91,7 +91,7 @@ func (b *BlockData) ToBlockExplorerSummary() apitypes.BlockExplorerBasic {
 		FreshStake:             b.Header.FreshStake,
 		StakeDiff:              b.Header.SBits,
 		BlockExplorerExtraInfo: extra,
-		Time: b.Header.Time,
+		Time:                   b.Header.Time,
 	}
 }
 
--- a/blockdata/blockdata.go
+++ b/blockdata/blockdata.go
@@ -91,7 +91,7 @@ func (b *BlockData) ToBlockExplorerSummary() apitypes.BlockExplorerBasic {
 		FreshStake:             b.Header.FreshStake,
 		StakeDiff:              b.Header.SBits,
 		BlockExplorerExtraInfo: extra,
-		Time:                   b.Header.Time,
+		Time: b.Header.Time,
 	}
 }
 
--- a/db/dcrsqlite/apisource.go
+++ b/db/dcrsqlite/apisource.go
@@ -1351,9 +1351,11 @@ func (db *wiredDB) GetExplorerAddress(address string, count, offset int64) *expl
 // hash) address. See https://github.com/decred/dcrdata/issues/358 for details.
 func IsZeroHashP2PHKAddress(checkAddressString string, params *chaincfg.Params) bool {
 	zeroed := [20]byte{}
+	// expecting DsQxuVRvS4eaJ42dhQEsCXauMWjvopWgrVg address for mainnet
 	address, err := dcrutil.NewAddressPubKeyHash(zeroed[:], params, 0)
 	if err != nil {
-		return false //invalid state, should panic
+		log.Errorf("Incorrect pub key hash or invalid network params %v", params)
+		return false
 	}
 	zeroAddress := address.String()
 	return checkAddressString == zeroAddress
--- a/db/dcrsqlite/apisource.go
+++ b/db/dcrsqlite/apisource.go
@@ -28,7 +28,7 @@ import (
 	"github.com/decred/dcrdata/rpcutils"
 	"github.com/decred/dcrdata/stakedb"
 	"github.com/decred/dcrdata/txhelpers"
-	humanize "github.com/dustin/go-humanize"
+	"github.com/dustin/go-humanize"
 )
 
 // wiredDB is intended to satisfy DataSourceLite interface. The block header is
@@ -1241,7 +1241,16 @@ func (db *wiredDB) GetExplorerAddress(address string, count, offset int64) *expl
 		log.Infof("Invalid address %s: %v", address, err)
 		return nil
 	}
+
 	isTechnicalAddress := IsZeroHashP2PHKAddress(address, db.params)
+	if isTechnicalAddress {
+		return &explorer.AddressInfo{
+			Address:            address,
+			IsTechnicalAddress: isTechnicalAddress,
+			Fullmode:           true,
+		}
+	}
+
 	maxcount := explorer.MaxAddressRows
 	txs, err := db.client.SearchRawTransactionsVerbose(addr,
 		int(offset), int(maxcount), true, true, nil)
@@ -1328,7 +1337,7 @@ func (db *wiredDB) GetExplorerAddress(address string, count, offset int64) *expl
 		KnownTransactions:  numberMaxOfTx,
 		KnownFundingTxns:   numReceiving,
 		KnownSpendingTxns:  numSpending,
-		IsTechnicalAddress: isTechnicalAddress,
+		IsTechnicalAddress: false,
 	}
 }
 
--- a/db/dcrsqlite/apisource.go
+++ b/db/dcrsqlite/apisource.go
@@ -1242,12 +1242,19 @@ func (db *wiredDB) GetExplorerAddress(address string, count, offset int64) *expl
 		return nil
 	}
 
-	isDummyAddress := IsZeroHashP2PHKAddress(address, db.params)
-	if isDummyAddress {
-		return &explorer.AddressInfo{
-			Address:        address,
-			IsDummyAddress: isDummyAddress,
-			Fullmode:       true,
+	{
+		// Short circuit the transaction and balance queries if the provided address
+		// is the zero pubkey hash address commonly used for zero value
+		// sstxchange-tagged outputs.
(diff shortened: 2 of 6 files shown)


blockdata/blockdata.go
```diff
@@ -91,7 +91,7 @@ func (b *BlockData) ToBlockExplorerSummary() apitypes.BlockExplorerBasic {
 		FreshStake:             b.Header.FreshStake,
 		StakeDiff:              b.Header.SBits,
 		BlockExplorerExtraInfo: extra,
-		Time:                   b.Header.Time,
+		Time: b.Header.Time,
```
chappjc: please undo the go fmt changes as we'll just keep going back and forth on this.  We're ignoring gofmt warnings in CI for now, so it's fine.

db/dcrsqlite/apisource.go
```diff
@@ -28,7 +28,7 @@ import (
 	"github.com/decred/dcrdata/rpcutils"
 	"github.com/decred/dcrdata/stakedb"
 	"github.com/decred/dcrdata/txhelpers"
-	humanize "github.com/dustin/go-humanize"
+	"github.com/dustin/go-humanize"
```
chappjc: the package name was repeated, seemingly unnecessarily, because I ran into linters that removed the import claiming it was unused.  One of the reasons I can't stand repos that are not named as the package.

views/address.tmpl
```diff
@@ -23,6 +23,10 @@
                 <div class="row">
                     <a href="javascript:hideAddressQRCode();" id="address-qrcode" class="col opacity-transition d-block" style="opacity: 0"></a>
                 </div>
+                {{if .IsTechnicalAddress}}
+                  <br/>
+                  This address is special. Please update this message once you know why it is special.
```
chappjc: A message here is OK (but not this message ofc), but the main point of issue #358 is to avoid the pointless and expensive DB query.

db/dcrsqlite/apisource.go
```diff
@@ -1312,23 +1312,46 @@ func (db *wiredDB) GetExplorerAddress(address string, count, offset int64) *expl
...
+		IsTechnicalAddress: isTechnicalAddress,
+	}
+}
+
+// IsZeroHashP2PHKAddress checks if given address is special
+// see https://github.com/decred/dcrdata/issues/358 for details
+func IsZeroHashP2PHKAddress(checkAddressString string, params *chaincfg.Params) bool {
+
+	//sanity check
+	checkAddress, err := dcrutil.DecodeAddress(checkAddressString)
```
chappjc: why sanity check at all?  You just need to see if the input string is equal to the zero-hash p2pkh.

db/dcrsqlite/apisource.go
```diff
@@ -1312,23 +1312,46 @@ func (db *wiredDB) GetExplorerAddress(address string, count, offset int64) *expl
...
+	zeroed := [20]byte{}
+	address, err := dcrutil.NewAddressPubKeyHash(zeroed[:], params, 0)
+	if err != nil {
+		return false //invalid state, should panic
 	}
+
+	a := checkAddress.String()
+	b := address.String()
+	equal := a == b
+	return equal
```
chappjc: `return checkAddressString == zeroAddress`

JFixby:
I updated the call to ignore DB calls for the address.

image

JFixby:
I updated the call to ignore DB calls for the address.

image

db/dcrsqlite/apisource.go
```diff
@@ -1328,7 +1337,20 @@ func (db *wiredDB) GetExplorerAddress(address string, count, offset int64) *expl
 		KnownTransactions: numberMaxOfTx,
 		KnownFundingTxns:  numReceiving,
 		KnownSpendingTxns: numSpending,
+		IsDummyAddress:    false,
+	}
+}
+
+// IsZeroHashP2PHKAddress checks if given address is dummy
```
chappjc:
```go
// IsZeroHashP2PHKAddress checks if the given address is the dummy (zero pubkey
// hash) address. See https://github.com/decred/dcrdata/issues/358 for details.
```

db/dcrsqlite/apisource.go
```diff
@@ -1328,7 +1337,20 @@ func (db *wiredDB) GetExplorerAddress(address string, count, offset int64) *expl
 		KnownTransactions: numberMaxOfTx,
 		KnownFundingTxns:  numReceiving,
 		KnownSpendingTxns: numSpending,
+		IsDummyAddress:    false,
```
chappjc: Please remove.  false is the zero value for booleans used by default during struct init.

db/dcrsqlite/apisource.go
```diff
@@ -1242,6 +1242,15 @@ func (db *wiredDB) GetExplorerAddress(address string, count, offset int64) *expl
 		return nil
 	}
 
+	isDummyAddress := IsZeroHashP2PHKAddress(address, db.params)
```
chappjc:
```go
// Short circuit the transaction and balance queries if the provided address
// is the zero pubkey hash address commonly used for zero value
// sstxchange-tagged outputs.
```

db/dcrsqlite/apisource.go
```diff
@@ -1242,6 +1242,15 @@ func (db *wiredDB) GetExplorerAddress(address string, count, offset int64) *expl
...
+	if isDummyAddress {
+		return &explorer.AddressInfo{
+			Address:        address,
+			IsDummyAddress: isDummyAddress,
+			Fullmode:       true,
```
chappjc: also, `Balance:        new(explorer.AddressBalance),` to avoid a potential nil pointer dereference panic in template execution

chappjc: and `UnconfirmedTxns: new(AddressTransactions),` please

db/dcrsqlite/apisource.go
```diff
@@ -1328,7 +1337,20 @@ func (db *wiredDB) GetExplorerAddress(address string, count, offset int64) *expl
...
+	}
+}
+
+// IsZeroHashP2PHKAddress checks if given address is dummy
+// see https://github.com/decred/dcrdata/issues/358 for details
+func IsZeroHashP2PHKAddress(checkAddressString string, params *chaincfg.Params) bool {
+	zeroed := [20]byte{}
+	address, err := dcrutil.NewAddressPubKeyHash(zeroed[:], params, 0)
+	if err != nil {
+		return false //invalid state, should panic
```
chappjc: So log the error, or return the error as well.  If `err != nil`, it's because `params` wasn't valid, or the length of a pubkey hash is no longer 20.  In any case, the error would be informative.

explorer/explorertypes.go
```diff
@@ -257,6 +257,13 @@ type AddressInfo struct {
 	// KnownMergedSpendingTxns refers to the total count of unique debit transactions
 	// that appear in the merged debit view.
 	KnownMergedSpendingTxns int64
+
+	// IsDummyAddress flag is true when the address has a special status,
+	// see https://github.com/decred/dcrdata/issues/358 for details
+	//
+	// The flag signals this is dummy address, typically used for
+	// unspendable ticket change outputs.
```
chappjc: Please take more care to use proper grammar and formatting.

```go
// IsDummyAddress is true when the address is the dummy address typically
// used for unspendable ticket change outputs. See
// https://github.com/decred/dcrdata/issues/358 for details.
```

views/address.tmpl
```diff
@@ -23,6 +23,12 @@
                 <div class="row">
                     <a href="javascript:hideAddressQRCode();" id="address-qrcode" class="col opacity-transition d-block" style="opacity: 0"></a>
                 </div>
+                {{if .IsDummyAddress}}
+                  <br/>
+                  This is dummy address, typically used for unspendable ticket change outputs.
```
chappjc: This is *a* dummy address...

JFixby:
image

JFixby:
Should we also remove the address from ticket outputs?

image

chappjc:
@JFixby We could make the link in the ticket outputs unclickable, but we need to show the address.  However, I'd prefer it be clickable otherwise the user would not be able to get to the page with the message you added, and thus not be able to figure out why it's not clickable.

db/dcrsqlite/apisource.go
```diff
@@ -1331,6 +1347,20 @@ func (db *wiredDB) GetExplorerAddress(address string, count, offset int64) *expl
 	}
 }
 
+// IsZeroHashP2PHKAddress checks if the given address is the dummy (zero pubkey
+// hash) address. See https://github.com/decred/dcrdata/issues/358 for details.
+func IsZeroHashP2PHKAddress(checkAddressString string, params *chaincfg.Params) bool {
```
chappjc: Please make a `IsZeroHashP2PHKAddress` in apisource_test.go to verify that it returns the expected result for various strings and all three networks.

JFixby:
Test: https://github.com/decred/dcrdata/pull/612/files#diff-4d4fab4a8bbb8d9ce3c87ceb33d90ce5R10

PR #617, opened 2018-08-15
Add utility that displays next expeted ticket price

### LINKED ISSUE #540: Show next expected ticket price
from rpc estimatestakediff
"expected": 85.67748345

TICKET PRICE   85.15854659 DCR (next expected 85.67748345 DCR)

codemaestro64:
This displays the next expected ticket price alongside the current ticket price 
Fixes #540

### DIFF
--- a/blockdata/blockdata.go
+++ b/blockdata/blockdata.go
@@ -91,7 +91,7 @@ func (b *BlockData) ToBlockExplorerSummary() apitypes.BlockExplorerBasic {
 		FreshStake:             b.Header.FreshStake,
 		StakeDiff:              b.Header.SBits,
 		BlockExplorerExtraInfo: extra,
-		Time:                   b.Header.Time,
+		Time: b.Header.Time,
 	}
 }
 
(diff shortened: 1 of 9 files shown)


chappjc:
There are a couple of important aspects of this issue to address:
1. The estimate of next stake diff is based on what assumptions?  We need to have them made clear via some hover title or whatever.
2. The accuracy of the estimate is generally related to where in the current sdiff window the chain is?  In other words, when the sdiff has just adjusted, the estimate is completely useless, while it is quite accurate near the end of the window.
3. There is a min and max that goes with the estimate that relates to (2).  You can think of these as the bounds of an uncertainty window.

So what I suggest for implementing this properly on the UI would include:
1. An indication of window progress and/or uncertainty of the estimate
2. The upper and lower bounds to indicate quantitatively how much deviation from the estimate can be expected.

codemaestro64:
@chappjc i just got to understand what you mean. will be pushing commits soon to address this.

chappjc:
That's exactly what I had in mind.  Here's the current display with your latest edits:

image

My only critique now is purely cosmetic:
1. I don't think we need so many decimal places for the bounds.  Maybe just 2 decimals.
2. There's a slightly awkward vertical space between the estimate and the bounds.  This might look better on the same line, which would be possible with fewer decimals for the bounds.
3. Some of the newlines added in the following code introduced extra space between the numbers and "DCR":

```html
<span id="blocksdiff">
    {{template "decimalParts" (float64AsDecimalParts .StakeDiff false)}}
</span>
<span class="pl-1 unit">DCR</span>
```

On both the price and estimate lines.

chappjc:
Maybe something like this:

image

chappjc:
Works for me.  The bounds text is a bit large IMO, but we can tweak later.

PR #624, opened 2018-08-17
dcrpg: fix the 3.2.0 -> 3.3.0 upgrade

chappjc:
The `HashDB` function required the `is_mainchain` column, but it did not
yet exist at the time of upgrade.  This was added to the query after
the upgrade was initially tested.


chappjc:
This addresses an issue reported by @papacarp with the upgrade to 3.3.0.  Migwi and I did not hit this issue, presumably because we performed the upgrade before the `is_mainchain = TRUE` constraint was added to this query.

PR #625, opened 2018-08-17
README: update for 3.0

chappjc:
Remove github releaes badge (only tags matter).
Update repository overview with new folders and better descriptions.
Update required dcrd to 1.3 (testnet3).
Clean up and clarify Upgrading Instructions.
Add more about PostgreSQL tuning.
Add System hardware requirements section.
Update JSON API tables with new endpoints (block subsidy, tx vinfo,
tx hex, tx decoded, address totals).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,6 @@
 # dcrdata
 
 [![Build Status](https://img.shields.io/travis/decred/dcrdata.svg)](https://travis-ci.org/decred/dcrdata)
-[![GitHub release](https://img.shields.io/github/release/decred/dcrdata.svg)](https://github.com/decred/dcrdata/releases)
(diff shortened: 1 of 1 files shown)


dmigwi:
Hi @chappjc You might want to recommend people to use postgres version 10 and above instead of 9.6, since `pg_restore` has a limitation in terms of the file types it supports.
To be specific, the compressed pg dump that you create only work with version 10 and above.

On the repository overview the `required` has been mispelled to `requried`.

You haven't mentioned about the datadir and the possible permission issues that may arise especially on a linux based O.S.

dmigwi:
Also 
```
txindex=1
addrindex=1
```
This values are just mentioned as not set by default but its not listed why one should consider setting them or not.  i.e. their benefits.

PR #626, opened 2018-08-21
go mod: Use go modules

chappjc:
Specify dcrd and dcrwallet module versions.
Remove dcrwallet/netparams import.
Update all other deps.


chappjc:
Testers, please test with go1.11rc1.

```
go get golang.org/dl/go1.11rc1
go1.11rc1 download
export GO111MODULE=on
go build/install/whatever
```

chappjc:
This seems to work, and the PR does not remove the dep files, so merging.

jrick:
This wasn't done entirely correctly:

```
$ go get github.com/decred/dcrdata/v3@master
go: finding github.com/decred/dcrdata/v3 master
go: github.com/decred/dcrdata/v3@v3.0.0-20180821200455-2f683513e4a6: go.mod has non-.../v3 module path "github.com/decred/dcrdata" (and .../v3/go.mod does not exist) at revision 2f683513e4a6
go: error loading module requirements
```

With a go.mod present, go assumes that you are following [semantic import versioning](https://research.swtch.com/vgo-import).  That is, for major version 3, the go.mod must have a module name ending in `/v3`, or a v3 directory exists in the repo with the source code of major version 3 (and the go.mod must again end with a `/v3` module name).  The first way allows branches to switch between major versions, while the second way allows multiple major versions to coexist in the same checkout.

chappjc:
We've not tagged v3.0.0 yet.  And we're definitely not following semantic import versioning as it's missing the /v3 in the module name.  I did see this bit from the golang wiki:

> If the module is version v2 or higher, the major version of the module must be included in both the module path in the go.mod file (e.g., module example.com/my/mod/v2) and the package import path (e.g., import "example.com/my/mod/v2/foo").

At this point I think it's just setup to allow building with go build.  I've not defined any modules inside dcrdata.  The build should at least work I think.

jrick:
Ah, sorry I thought v3 was already out.  The same issue occurs with v2 though.

The existing tags make this transition difficult.  We didn't experience this in dcrd or dcrwallet because v2.x.x was never released.

It would be possible to start off at version 2, but this requires some code shuffling.  Either drop everything under v2 (and give an appropriate github.com/decred/dcrdata/v2 module declaration), or work off a v2 branch, keep a v2 go.mod in the root, and rewrite all import paths to reference v2 packages.  The first way is backwards compatible to everybody, but the second way will (I think) break dcrdata importers who use dep to vendor dcrdata, since it won't add the major version to the vendor directory structure.

jrick:
After further thought, this approach is fine.  I tested trying to import v2 not as module /v2, but @2.1.2, and `require github.com/decred/dcrdata v2.1.2+incompatible` was added to my go.mod.  This shows that go understands the project does not follow semantic import versioning, and there is no problem referencing packages in the current release by their existing paths.

If/when a complete switch to module is made, the major versions must appear in the module name, and one of the two approaches I mentioned above should be taken (for dcrd and dcrwallet, because we also want to define submodules, we keep v1 in the root and v2, v3... directories for major API breaks).

Go modules still expect you to follow semantic versioning though, even if it's not semantic import versioning.  If v2.2.0 includes a breaking change, `go get -u github.com/decred/dcrdata@latest` won't hesitate to pull it in and break the build.  Due to the design of modules, it is only possible to specify minimum compatibility, so maximum (minor) versions, (minor) version ranges, and listed incompatible versions are all impossible to enforce on your importers.  I only stress this since the success of modules relies on Go developers making commitments to compatibility once projects have reached major version 1.

chappjc:
That's why major version is already 2 going on 3.  We love to break stuff.  Pretty soon we'll be on v69 like Chrome.  :)  Seriously, thanks for the guidance.  This is slowly becoming clear to me.

Regarding the vN directories for major API breaks, that's an approach that doesn't sit too well with me.  I don't want irrelevant files checked out and angering my dumb IDE.  I'm OK requiring 1.9.7+ and 1.10.3+.

So for now my approach will be to wait until 3.0.0 is ready and start at module /v3.  I don't know yet what submodules we want to make however.

PR #628, opened 2018-08-22
Fix bugs on the charts

dmigwi:
**Summary of Tasks done**
- [x] Update agenda voting milestones (Activated, hardCoded and lockedIn) when data migration is running from scratch as a temporary solution.
- [x] Delete unnecessary/duplicated code.
- [x] Rename the agenda `Id`  variable to `ID`.
- [x] Fix the deformed legend on both addresses and agendas charts.
- [x] Make charts axises with a value more than 999 gets formatted with a `k` appended to it.
- [x] Make agenda page not throw errors if agenda voting data is missing.


dmigwi:
Please do confirm if all charts are still working properly with this PR merged.

PR #631, opened 2018-08-23
Make agendadb filename configurable and house it in datadir folder

codemaestro64:
This adds a configuration option that make the agendadb filename configurable. 
It also ensures that the file location is housed in the datadir. 

This is a final fix for #587


chappjc:
For future reference, you do not need to push a commit to trigger a build.  Just click through (the x or checkmark) to Travis, and click the "Restart Build" button on the top right.  You must be logged into Travis via github.

PR #633, opened 2018-08-24
Fix performance issues on address page

dmigwi:
**Summary of Tasks Done**
- [x] Add the Page size controls back.
- [x] Implemented to a way to determine the best group by interval to use by default when accessing the chart view from the list view.
        -  Either `month`, `week`, `day` or `all blocks` interval can be selected. i.e. `month` is only select if the oldest transaction belonging to a given address was processed more than a month ago from now.  The same also happens to the rest intervals.
        - Also if the total transactions count is less than 20, interval selected defaults to `all`.
        - This helps in dealing with the network latency issues since the payload now requested by default will be smaller and more meaningful to the user.

__for Project Fund Address, it will always default to `month` interval__

screen shot 2018-08-24 at 04 28 00

- [x] Add the programmer names for the transactions types to the Transaction Types legend i.e. `Tickets` becomes  `Tickets (sstx)`

- [x] Split the Regular Txs into Sent and Received Regular Txs.
screen shot 2018-08-24 at 04 31 04

- [x] Deactivate the chart view button when running lite mode

### DIFF
--- a/db/dcrpg/internal/txstmts.go
+++ b/db/dcrpg/internal/txstmts.go
@@ -131,7 +131,7 @@ const (
 	SelectStakeTxByHash   = `SELECT id, block_hash, block_index FROM transactions WHERE tx_hash = $1 and tree=1;`
 
 	IndexTransactionTableOnBlockIn = `CREATE UNIQUE INDEX uix_tx_block_in
-		ON transactions(block_hash, block_index, tree);`  // with cockroach: STORING (tx_hash, block_hash)
+		ON transactions(block_hash, block_index, tree);` // with cockroach: STORING (tx_hash, block_hash)
 	DeindexTransactionTableOnBlockIn = `DROP INDEX uix_tx_block_in;`
 
 	IndexTransactionTableOnHashes = `CREATE UNIQUE INDEX uix_tx_hashes
--- a/db/dcrpg/internal/txstmts.go
+++ b/db/dcrpg/internal/txstmts.go
@@ -131,7 +131,7 @@ const (
 	SelectStakeTxByHash   = `SELECT id, block_hash, block_index FROM transactions WHERE tx_hash = $1 and tree=1;`
 
 	IndexTransactionTableOnBlockIn = `CREATE UNIQUE INDEX uix_tx_block_in
-		ON transactions(block_hash, block_index, tree);` // with cockroach: STORING (tx_hash, block_hash)
+		ON transactions(block_hash, block_index, tree);`
 	DeindexTransactionTableOnBlockIn = `DROP INDEX uix_tx_block_in;`
 
 	IndexTransactionTableOnHashes = `CREATE UNIQUE INDEX uix_tx_hashes
--- a/public/js/controllers/address.js
+++ b/public/js/controllers/address.js
@@ -81,7 +81,7 @@
             var _this = this
             $.getScript('/js/dygraphs.min.js', () => {
                 _this.typesGraphOptions = {
-                    labels: ['Date', 'SentTx (tx)', 'ReceivedTx (tx)', 'Tickets (sstx)', 'Votes (ssgen)', 'RevokedTx (ssrtx)'],
+                    labels: ['Date', 'Sending (regular)', 'Receiving (regular)', 'Tickets', 'Votes', 'Revocations'],
                     colors: ['#69D3F5', '#2971FF', '#41BF53', 'darkorange', '#FF0090'],
                     ylabel: 'Number of Transactions by Type',
                     title: 'Transactions Types',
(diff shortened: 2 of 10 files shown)


dmigwi:
@gozart1 I tried to make the legend always visible but my implementation end up not being that good. Please help our here when you get the time.
screen shot 2018-08-24 at 04 56 39

xaur:
Can the date format be `YYYY-MM-DD HH:mm` please?

db/dcrpg/internal/txstmts.go
```diff
@@ -131,7 +131,7 @@ const (
 	SelectStakeTxByHash   = `SELECT id, block_hash, block_index FROM transactions WHERE tx_hash = $1 and tree=1;`
 
 	IndexTransactionTableOnBlockIn = `CREATE UNIQUE INDEX uix_tx_block_in
-		ON transactions(block_hash, block_index, tree);`  // with cockroach: STORING (tx_hash, block_hash)
+		ON transactions(block_hash, block_index, tree);` // with cockroach: STORING (tx_hash, block_hash)
```
chappjc: Let's remove the comment.  `gofmt` keeps switching it around and we're not working on cockroachdb yet.

public/js/controllers/address.js
```diff
@@ -83,11 +81,11 @@
             var _this = this
             $.getScript('/js/dygraphs.min.js', () => {
                 _this.typesGraphOptions = {
-                    labels: ['Date', 'RegularTx', 'Tickets', 'Votes', 'RevokeTx'],
-                    colors: ['#2971FF', '#41BF53', 'darkorange', '#FF0090'],
-                    ylabel: '# of Tx Types',
+                    labels: ['Date', 'SentTx (tx)', 'ReceivedTx (tx)', 'Tickets (sstx)', 'Votes (ssgen)', 'RevokedTx (ssrtx)'],
```
chappjc: I suggest the following wording:

'Sending (regular)', 'Receiving (regular)', 'Tickets', 'Votes', 'Revocations'

views/address.tmpl
```diff
@@ -13,7 +15,8 @@
         <div class="row">
             <div class="col-md-8 col-sm-6">
                 <h4>Address</h4>
-                <div class="mono" data-target="address.addr" style="margin-bottom: 12px;">
+                <div class="mono" data-target="address.addr"
+                    id="{{$TxTime}}" style="margin-bottom: 12px;">
```
chappjc: I don't follow why this is tagged with `id="{{$TxTime}}"`.  Could you explain the idea?

views/address.tmpl
```diff
@@ -5,6 +5,8 @@
 {{template "html-head" printf "Decred Address %s" .Data.Address}}
 <body class="{{ theme }}">
     {{template "navbar" . }}
+    {{$TxTime := .OldestTxTime}}
+    {{$IsLite := .IsLiteMode}}
```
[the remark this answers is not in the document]
dmigwi: `{{with .Data}}` (line 10) changes our scope and thats why this variables have to be declared outside so that they can be used in the scope of `{{with .Data}}` without its effect.

public/js/controllers/address.js
dmigwi: ok. Noted

views/address.tmpl
dmigwi: This variable helps in choosing the most appropriate `group by` interval that should have lesser payload size and also provide more meaningful chart information to the user by default (when user accesses the chart view for the first time).
Its also helps deactivate by default the `zoom` and `group by` buttons that are not applicable in relation to the current address's data set.
Used here https://github.com/decred/dcrdata/pull/633/files#diff-90ad06f5b024ce10091b98d9a3be7766R230

db/dcrpg/internal/txstmts.go

views/address.tmpl
dmigwi: I also wanted a way to tag that time with an existing data-target rather that create a new data-target for it.

views/address.tmpl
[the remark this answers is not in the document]
chappjc: Yeah, that's OK.  Just note that we can do `$.IsLiteMode` to get back to the root scope regardless of the current scope.

views/address.tmpl
chappjc: Ah, the `var val = parseInt(this.addrTarget.id)` was what I was missing.  Thanks!

chappjc:
@dmigwi I'm hacking on this a bit today.  Will get back to you shortly.

chappjc:
Ah, well I'll wait for @gozart1's review.

dmigwi:
ok. I will look into it.

PR #634, opened 2018-08-24
explorer: flag non-mainchain votes

chappjc:
Votes for the current best block have the usual background, while older or side chain votes have a reddish background.

The header table with a summary of the mempool is updated so that it show the vote and ticket counts, and "TICKETS VOTED ON LAST BLOCK  x/5".  There is no longer any attempt to compute the expected validity of the previous block based on the votes in mempool.  There are several reasons, but the most important is that it is up to the miners what to include in the mined block.  Further, duplicating the consensus rules to decide what's valid (e.g. 2 yes / 2 no == ?, it is no, but we're not coding up these rules) is not something we should pursue in dcrdata.

The columns are also renamed and reordered.  Best Block is yes/no to indicate if the vote is for the main chain tip.  So old votes or votes on side chains at the same height would say "no".  The Block column is now "Voting On" to be clear that the vote is "voting on" the validity of the block listed.  The column text is block height, but the link is via hash to support side-chains.  The ticket index column is now "Validator ID" to represent that it is an identifier (arbitrary) for the ticket hash chosen to validate the block, and as such it links to the ticket hash that won the lottery and is being spent in the vote.

image

There are a number of fixes to the mempool transactions handling:
* Duplicates are screened out in both the fresh mempool update and the mempool monitor by tracking known mempool txns at the current best block.
* A javascript race is also worked around.  It is possible for a new txn notification to be sent for a vote on a block that has not yet been seen, and when the full mempool update is handled, that transaction is included again.  If JS handled it in the same order, it would be fine (removing the transaction that came early and starting with the full update), however, if JS processes the new block update of mempool and *then* the early vote's notification, it will end up on the table twice.
* The unique ticket indexes (now a.k.a. Validator ID) were based on block height and ticket hash, but this did not support side chains where multiple blocks would have the same height.  The index map (`BlockValidatorIndex`) is changed to use block hash.
* `explorerUI.NewBlockData` now contains `Hash` to that and RPC call may be skipped, and to support this, `explorer.BlockBasic` now includes `Hash`.

### DIFF
--- a/api/apiroutes.go
+++ b/api/apiroutes.go
@@ -489,7 +489,7 @@ func (c *appContext) getDecodedTransactions(w http.ResponseWriter, r *http.Reque
 	for i := range txids {
 		tx := c.BlockData.GetTrimmedTransaction(txids[i])
 		if tx == nil {
-			apiLog.Errorf("Unable to get transaction %s", tx)
+			apiLog.Errorf("Unable to get transaction %v", tx)
 			http.Error(w, http.StatusText(422), 422)
 			return
 		}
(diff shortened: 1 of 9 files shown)


chappjc:
The intent is to rebase all 3 commits w/o merge onto master so that each of the different contributors shows up in the log.

PR #641, opened 2018-08-28
Stats page

McEdward:
New Statistics page, as stated in issue #451
screenshot 87
 issue #451.


chappjc:
I suspect  we'll get lots of input on the design, so let's wait for some discussion.

My comments:

* In general I like the material design-like cards/boxes.
* The text in the boxes is too small for the size of the boxes.
* The DCR units do not belong in the "VOTES IN MEMPOOL" and "TICKETS IN MEMPOOL" boxes.
* "TICKETROI" should be two works, and it needs more than just the percent (time period, and equiv. return per year).
* "TICKET POOL VALUE" has no DCR units, and it says 0.00% of total supply.
* "PROOF-OF-STAKE REWARD" says 6.100967 DCR *per ticket*.  This is incorrect.
* The term DEVELOPMENT FUND, is no longer used, in favor of either PROJECT FUND or PROJECT treasury.  Also, the term "reward" in this context is awkward, and I prefer "subsidy" for the project fund.
* For the BLOCK REWARD ADJUSTMENT and TICKET WINDOW PROGRESS boxes, the current and next (estimated in the case of sdiff) should be shown.
* BLOCK REWARD ADJUSTMENT does not belong under Proof-of-Stake Statistics.  It should be at the top of the Block Rewards section.
* TOTAL SUPPLY should also give the "ultimate supply" to give an indication of what portion of DCR have been generated so far.
* DEVELOPMENT FUND missing DCR units, and also needs renaming as described above.
* I'm not sure PThash is right.  I think this means petahash, so that's probably PHash.
* More generally, I think there needs to be more eye candy.  For example, if the ticket pool size is high (some percentage above target), maybe make a orange arrow pointing up with the text "over target", etc.    More visual indicators like that to draw attention to anything of interest.

chappjc:
Let's broaden the review audience for the next iteration.

raedah:
Once we get this page up I expect it will also continue to evolve. Design elements such as font size and number of decimal places should be following the original mockup, https://user-images.githubusercontent.com/33739891/43277886-6f8c92ec-90d7-11e8-962d-3326f0db4e0b.png

McEdward:
screenshot 109

explorer/explorerroutes.go
```diff
@@ -1000,20 +1000,28 @@ func (exp *explorerUI) StatsPage(w http.ResponseWriter, r *http.Request) {
 	var stats StatsInfo
 	exp.MempoolData.RLock()
 	stats.TotalSupply = exp.ExtraInfo.CoinSupply
+	if exp.NetName == "Mainnet" {
+		stats.UltimateSupply = 15000000 * 100000000
```
chappjc: What's this from?  There is ultimate supply calculation else where in the dcrdata code you can reference?

McEdward: I couldn't find a reference to the ultimate supply in the code so zaephy gave this option, or could you help me with finding it?

chappjc: Perhaps the naming is not the best: https://github.com/decred/dcrdata/blob/master/txhelpers/subsidy.go#L15

McEdward: Alright, i'll get on that.

db/dcrsqlite/apisource.go
```diff
@@ -1501,3 +1501,13 @@ func (db *wiredDB) TxHeight(txid string) (height int64) {
 	height = txraw.BlockHeight
 	return
 }
+
+//GetDifficulty returns the difficulty
```
chappjc: Two notes:
1. We are trying to avoid functions starting with `Get`, so just `Difficulty` would be good.
2. Comment formatting needs to be consistent, with a space between `//` and the function name. The docs should be full sentences ending in a period.

explorer/explorerroutes.go
```diff
@@ -994,3 +994,59 @@ func (exp *explorerUI) AgendasPage(w http.ResponseWriter, r *http.Request) {
 	w.WriteHeader(http.StatusOK)
 	io.WriteString(w, str)
 }
+
+// StatsPage is the page handler for the "/stats" path
+func (exp *explorerUI) StatsPage(w http.ResponseWriter, r *http.Request) {
+	var stats StatsInfo
+	exp.MempoolData.RLock()
```
chappjc: Also need to RLock the NewBlockDataMtx, which guards both `ExtraInfo` and `NewBlockData`.

explorer/explorerroutes.go
```diff
@@ -994,3 +994,59 @@ func (exp *explorerUI) AgendasPage(w http.ResponseWriter, r *http.Request) {
...
+	stats.TotalSupply = exp.ExtraInfo.CoinSupply
+	stats.UltimateSupply = txhelpers.UltimateSubsidy(exp.ChainParams)
+	stats.TotalSupplyPercentage = float64(stats.TotalSupply) / float64(stats.UltimateSupply) * 100
+	stats.DevFunds = exp.ExtraInfo.DevFund
+	stats.DevAddress = exp.ExtraInfo.DevAddress
+	stats.POWDiff = exp.ExtraInfo.Difficulty
+	stats.BlockReward = exp.blockData.BlockSubsidy(int64(exp.blockData.GetHeight()), 5).Total
+	stats.NextBlockReward = exp.ExtraInfo.NBlockSubsidy.Total
```
chappjc: Instead of all these `stats.<fieldname> = ` assignments, please simultaneously declare and define, as in:

```go
stats := StatsInfo{
	TotalSupply: exp.ExtraInfo.CoinSupply,
	....
}
```

explorer/explorertypes.go
```diff
@@ -533,3 +533,34 @@ func AddressPrefixes(params *chaincfg.Params) []AddrPrefix {
 func GetAgendaInfo(agendaId string) (*agendadb.AgendaTagged, error) {
 	return agendadb.GetAgendaInfo(agendaId)
 }
+
+// StatsInfo represents the all data for the stats page
+type StatsInfo struct {
+	UltimateSupply             int64
+	TotalSupply                int64
```
chappjc: `TotalSupply` is confusing to me.  This is the *current* supply?  As of what block?  Where is the best block height and hash in this structure to indicate that this is valid for?

explorer/explorertypes.go
```diff
@@ -533,3 +533,34 @@ func AddressPrefixes(params *chaincfg.Params) []AddrPrefix {
...
+	TotalSupplyPercentage      float64
+	DevFunds                   int64
```
chappjc: `Dev` -> `Project`

explorer/explorertypes.go
```diff
@@ -533,3 +533,34 @@ func AddressPrefixes(params *chaincfg.Params) []AddrPrefix {
...
+	DevAddress                 string
+	POWDiff                    float64
```
chappjc: `POW` -> `PoW`, similarly for PoS

main.go
```diff
@@ -565,6 +565,10 @@ func mainCore() error {
 	webMux.Get("/search", explore.Search)
 	webMux.Get("/charts", explore.Charts)
 	webMux.Get("/ticketpool", explore.Ticketpool)
+	webMux.Get("/statistics", explore.StatsPage)
+	webMux.Get("/stats", func(w http.ResponseWriter, r *http.Request) {
```
chappjc: I would prefer to do this the other way.  (`/statistics` -> `/stats`).  Keeping the URL short is preferable.

views/extras.tmpl
```diff
@@ -76,6 +76,8 @@
                         <a data-keynav-skip href="/mempool" title="Decred mempool">Mempool</a>
                         <a data-keynav-skip href="/ticketpool" title="Decred ticket pool">Ticket Pool</a>
                         <a data-keynav-skip href="/charts" title="Decred charts">Charts</a>
+                        <a data-keynav-skip href="/statistics" title="Decred Statistics">Statistics</a>
+                        <a data-keynav-skip href="/parameters" title="Chain Parameters">Parameters</a>
```
chappjc: You've added another /parameters entry.

explorer/explorertypes.go
McEdward: Totalsupply is CoinSuppy, I could probably put in the block height and the hash but i was following the layout. Do have an idea of where i can put this information. I felt like it might have needed a new card (But that would break the layout) or should i just add it to the totalsupply card.

McEdward:
screenshot 111

explorer/explorerroutes.go
chappjc: @papacarp This was the thing I mentioned when talking about embedding the mutex in a struct.  It's error prone when it's not clear what the mutex goes with.

explorer/explorertypes.go
chappjc: Yes the best block height and hash should definitely be on the stats page.  Maybe not a card.  Perhaps try some text above the "column" headings that gives the chain status.

chappjc:
On current iteration and comments by @xaur:

- Agreed no hyphen on proof-of-stuff.
- Agreed that small text is a bit too light to read comfortably.
- I don't agree that 0 decimals is OK for supply and project fund.  Maybe 2 or 3, but I do not want to suggest to the viewer that it's somehow exactly a multiple of 1 DCR.
- For the links "Show Chart" and "Show Address", I also agree these can be "show chart" and "show address".  It looks a bit odd in title case.
- 100% agreed with all comments on Block Reward Adjustment.
- Votes in mempool.... it is confusing to say 29.  I think only the votes that are on the current best block should be counted.
 - Ticket ROI: the yearly should go on the next line, but also I suggest a hover text (`title=''`) tag property that briefly says that it refers to continuous re-staking and is based on a variable ticket price.
- Mostly agreed with Ticket Price Adjustment comments, although "Next Estimate" -> "Next (est.)" would be better IMO.

Thanks for the feedback @xaur!  You have some very good advice on this.

chappjc:
Also it seems that the values in the main part of the box are too high.  They need to go toward the bottom a bit to look centered.

papacarp:
When I was working on the Ticket rewards a few months back the conclusion was we should avoid the term APR (annual percentage rate) and instead use the term ASR (annual stake rewards) as APR implies this is an "investment".   Along the same line of thought we probably want to change TICKET ROI back to TICKET REWARDS.

Jujhar:
Views have been updated we ran through @Xaur's todo list.
Not sure how many decimals to use as varying opinion on some cards so please be sure check those.

One thing we are still working on though are the additional details on Proof-of-Work Difficulty and Hash Rate cards if you can help or give hint how to complete those that would be appreciated:

- Proof-of-Work Difficulty
can also show 24h change
can also show next estimated change e.g. +X.X%

- Hash Rate
can also show 24h and 7d change, maybe also 30d change

xaur:
Re decimals, I still believe they are not needed _but_ if you really want them, 1-2 in _some_ places are still much better than 8. They can even be grayed out to distract less. Then the decimals will serve to convey this message: "hey user look, we are not round integers and we are not sure why we are even bothering you with this, but that is all we have to say".

chappjc:
In the future please make informative commit messages describing the change.

The challenge you are facing with rebasing is because of the merge you did with some other branch:

image

Doing a `git rebase master` on that branch is going to be quite painful unless you squash first.

First note the common ancestor:

git merge-base master stats_page
8d924bf8e669ad5bd55568f8f68a241f5b08c469

In the screenshot, that's the commit "views: style invalidated transactions on block page. (#663)" where master and stats_page branches diverged.  There are many ways to squash down your branch to a single commit (thus removing the merge commit), but the easiest may be the following (assuming you've pulled latest into master!):

```
git checkout stats_page
git reset --hard `git merge-base master stats_page`
git merge --squash HEAD@{1}
git commit # EDIT THE MESSAGE!
```

Then you can rebase stats_page on master:

    git rebase master

In this case there will be one "conflict", but you really take both changes since they are just adjacent, not conflicting.  When `git rebase master` says there are conflicts, `git status` will tell you what to do.  Edit the templates.go files, making it valid, then do `git rebase --continue` and it should succeed.  At this point you will need should check that it builds right and all the above went well.  *Then* force push: `git push -f`.

Jujhar:
Updated. Added new function.
Regarding commits did I do it right? One concern I seem to have is the McEdward/dcrdata seems to be behind the decred/dcrdata master. I tried upstream and stuff like that maybe it is a permissions thing. Will look more into it.

chappjc:
No there's still a million commits.

chappjc:
Start by trying to squash it down.  Do not do any merges.  Only squash and rebase on master.

How do you have your git workspace setup?  What are the remotes?

chappjc:
Please start a new PR with a new branch that does not have a mess of commits.  Branch off of updated master please.

Jujhar:
> Please start a new PR with a new branch that does not have a mess of commits. Branch off of updated master please.

OK

PR #647, opened 2018-08-28
Updates README for golang 1.11

logicminds:
main: updates the build instructions and dependency resolution tooling for golang 1.11.

Adds a new section for building with a docker container and cross compiling.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 [![Go Report Card](https://goreportcard.com/badge/github.com/decred/dcrdata)](https://goreportcard.com/report/github.com/decred/dcrdata)
 [![ISC License](https://img.shields.io/badge/license-ISC-blue.svg)](http://copyfree.org)
 
-The dcrdata repository is a collection of go packages and apps for [Decred](https://www.decred.org/) data collection, storage, and presentation.
+The dcrdata repository is a collection of Go packages and apps for [Decred](https://www.decred.org/) data collection, storage, and presentation.
 
 - [dcrdata](#dcrdata)
   - [Repository overview](#repository-overview)
@@ -62,7 +62,7 @@ The dcrdata repository is a collection of go packages and apps for [Decred](http
 │   ├── rebuilddb       rebuilddb utility, for SQLite backend. Not required.
 │   ├── rebuilddb2      rebuilddb2 utility, for PostgreSQL backend. Not required.
 │   └── scanblocks      scanblocks utility. Not required.
-├── dcrdataapi          Package dcrdataapi for go API clients.
+├── dcrdataapi          Package dcrdataapi for Go API clients.
 ├── db
 │   ├── agendadb        Package agendadb is a basic PoS voting agenda database.
 │   ├── dbtypes         Package dbtypes with common data types.
@@ -89,7 +89,7 @@ The dcrdata repository is a collection of go packages and apps for [Decred](http
 
 ## Requirements
 
-- [Go](http://go.org) 1.10.x or 1.11.x.
+- [Go](http://golang.org) 1.10.x or 1.11.x.
 - Running `dcrd` (>=1.3.0) synchronized to the current best block on the
   network. This is a strict requirement as testnet2 support is removed from
   dcrdata v3.0.0.
@@ -101,7 +101,7 @@ The dcrdata repository is a collection of go packages and apps for [Decred](http
 The inclusion of a dockerfile in this repo means you can use docker for either development or production usage. Although
 at this time we are not publishing images to docker hub.
 
-When developing you can utilize containers for easily swapping out go runtimes and overall complete project setup.
+When developing you can utilize containers for easily swapping out Go runtimes and overall complete project setup.
 You don't even need go installed on your system if using containers during development.
 
 The first thing you need is [docker](https://docs.docker.com/install/). So once installed you can then download this repo
@@ -171,7 +171,7 @@ need to get used to.
 3. Attached volumes on a mac are a bit slower than linux/windows (when testing)
 4. Install everything in the container, don't muck up your docker host
 5. Resist the urge to run git commands from the container
-6. You can swap out the go runtimes just by using a different docker image
+6. You can swap out the Go runtimes just by using a different docker image
 
 When working in a container you want your source code from your docker host to be in the container.
 To do this you attach a volume to the container when running and the synchronization happens automatically.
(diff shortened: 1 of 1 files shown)


logicminds:
You can review the README here: https://github.com/nwops/dcrdata/blob/golang1.11/README.md

README.md
```diff
@@ -193,24 +221,52 @@ The following instructions assume a Unix-like shell (e.g. bash).
...
+
+### Build from source for golang >= 1.11
+
+Golang 1.11 introduced a new [dependency management feature](https://github.com/golang/go/wiki/Modules) that
+negates the use of third party tooling such as `dep`. Using `Go mod` will be the dependency managment tool of choice
+for dcrdata and other Decred projects going forward.
+
+Usage is simple and nothing is required except Golang 1.11.
+
+1. `go build`
```
chappjc: must still set `GO111MODULE=on`, at least until go 1.12

README.md
```diff
@@ -193,24 +221,52 @@ The following instructions assume a Unix-like shell (e.g. bash).
...
+
+If you want to pre staging dependencies without building you can run
```
chappjc: Reads weird.  Please adjust grammar.  Also, why and when would you do this (esp. in the context of dcrdata)?   `go build` does this automatically, so it's not clear why it's mentioned here.  IMO, `go mod vendor` might be worth mentioning to prepare a vendor folder for go 1.10 or a setup that is not using modules.

README.md
```diff
@@ -193,24 +221,52 @@ The following instructions assume a Unix-like shell (e.g. bash).
...
 
 - Ensure `$GOPATH/bin` is on your `$PATH`.
+
+- Clone the dcrdata repository. It **must** be cloned into the following directory.
+
+      git clone https://github.com/decred/dcrdata $GOPATH/src/github.com/decred/dcrdata
+
+Once Golang is installed and the dcrdata project is cloned proceed to the build steps.
+
+### Build from source for golang >= 1.11
```
chappjc: For consistency and other reasons, let's refer to it as just Go (the language) or `go` (the tool).

README.md
```diff
@@ -193,24 +221,52 @@ The following instructions assume a Unix-like shell (e.g. bash).
...
+If you want to pre staging dependencies without building you can run
+
+1. `go mod download`
+
+Golang's builtin dependency management resolution will start processing
+the source code and downloading dependencies on the fly and then update the `go.mod` and `go.sum` files automatically.
+
+### Build from source for golang < 1.11
+
+Building with the `dep` dependency management tool.
```
chappjc: There are really two options with go 1.10.  The first is `dep`, as described here, and the other is `vgo`.  The go 1.11 way and and the vgo way are supported and described here: https://github.com/golang/go/wiki/Modules#installing-and-activating-module-support

It's desirable to have builds from the same dependency tree regardless of go version, so I'd advise the vgo way if the user is unable to use go 1.11.  We've not kept the dep files updated to match the module defs, and it might not be possible.

README.md
```diff
@@ -193,24 +221,52 @@ The following instructions assume a Unix-like shell (e.g. bash).
...
 
       cd $GOPATH/src/github.com/decred/dcrdata
       dep ensure -vendor-only
       # build dcrdata executable in workspace:
       go build
 
-To build with the git commit hash appended to the version, set it as follows:
+_Note_ we recommend upgrading to [Golang 1.11](#build-from-source-for-golang--111) or using the docker [container build instructions](#building-dcrdata-with-docker)
+
+### Build from commit hash
```
chappjc: The word *from* is a bit misleading as it suggests the suggestion will tell you how to choose a certain commit hash to build from.  Might as well be more specific and say it's building with the git commit hash in the version string.

README.md
```diff
@@ -249,6 +305,9 @@ necessary.
 
 ## Updating dependencies
 
+_NOTE_ This is only needed if using golang 1.10 or earlier since golang 1.11 has the `go mod` depedency management
```
chappjc: ...if using go 1.10 without vgo since...

README.md
```diff
@@ -86,7 +89,7 @@ The dcrdata repository is a collection of golang packages and apps for [Decred](
 
 ## Requirements
 
-- [Go](http://golang.org) 1.10.x or 1.11.x.
+- [Go](http://go.org) 1.10.x or 1.11.x.
```
chappjc: well, the website is golang.org

logicminds: Oops.  I did a replace all.  Let me fix.

README.md
```diff
@@ -5,17 +5,20 @@
 [![Go Report Card](https://goreportcard.com/badge/github.com/decred/dcrdata)](https://goreportcard.com/report/github.com/decred/dcrdata)
 [![ISC License](https://img.shields.io/badge/license-ISC-blue.svg)](http://copyfree.org)
 
-The dcrdata repository is a collection of golang packages and apps for [Decred](https://www.decred.org/) data collection, storage, and presentation.
+The dcrdata repository is a collection of go packages and apps for [Decred](https://www.decred.org/) data collection, storage, and presentation.
```
chappjc: Since you're going to be making another edit, note that the language is called "Go" while the tools is "go".

README.md
```diff
@@ -182,35 +212,68 @@ the DCRDATA_DCRD_CERT env variable to set the location. Or build a new container
...
 
 - Clone the dcrdata repository. It **must** be cloned into the following directory.
 
       git clone https://github.com/decred/dcrdata $GOPATH/src/github.com/decred/dcrdata
 
+Once go is installed and the dcrdata project is cloned proceed to the build steps.
+
+### Building with go 1.11
+
+Go 1.11 introduced a new [dependency management feature](https://github.com/go/go/wiki/Modules) that
```
chappjc: this link broke too.  Should be https://github.com/golang/go/wiki/Modules

README.md
```diff
@@ -249,14 +312,17 @@ necessary.
...
+_NOTE_ This is only needed if using go 1.10 or earlier since go 1.11 has the `go mod` depedency management
+built in.
+
 `dep ensure -vendor-only` fetches project dependencies, without making changes
 in manifest files `Gopkg.toml` and `Gopkg.lock`.
 
 Call `dep ensure` to update project dependencies when introduce new imports.
 
-For guides and reference materials about `dep`, see [the documentation](https://golang.github.io/dep).
+For guides and reference materials about `dep`, see [the documentation](https://go.github.io/dep).
```
chappjc: another link

PR #653, opened 2018-08-29
made changes in file explorermiddleware.go for showing future block.

SeaLightHouse:
made changes in file explorermiddleware.go for showing future block.

### DIFF
--- a/explorer/explorermiddleware.go
+++ b/explorer/explorermiddleware.go
@@ -5,9 +5,10 @@ package explorer
 
 import (
 	"context"
+	"fmt"
 	"net/http"
 	"strconv"
-        "fmt"
+	"time"
 
 	"github.com/go-chi/chi"
 )
@@ -36,18 +37,15 @@ func (exp *explorerUI) BlockHashPathOrIndexCtx(next http.Handler) http.Handler {
 				return
 			}
 		} else {
-                        //Handle blocks which are in future time
-                        maxHeight := exp.blockData.GetHeight()
-                        if height > int64(maxHeight) {
-                        maturityTime := int(float64(height - int64(maxHeight) ) * exp.ChainParams.TargetTimePerBlock.Minutes())
-                        days := maturityTime / 1440
-                        hours := (maturityTime % 1440  ) / 60
-                        minutes := maturityTime % 60
-                        message := fmt.Sprintf("This block will arrive in approximately %d days, %d hour, %d min.",days,hours,minutes)
-                        log.Errorf("Block with height %d will arrive in future.", height)
-                        exp.StatusPage(w, defaultErrorCode, message, ErrorStatusType)
-                        return
-                        }
+			//Handle future blocks
+			maxHeight := int64(exp.blockData.GetHeight())
+			if height > maxHeight {
+				expectedTime := time.Duration(height-maxHeight) * exp.ChainParams.TargetTimePerBlock
+				message := fmt.Sprintf("This block is expected to arrive in approximately in %v. ", expectedTime)
+				log.Errorf("Block with height %d will arrive in future.", height)
+				exp.StatusPage(w, defaultErrorCode, message, NotFoundStatusType)
+				return
+			}
 
 			hash, err = exp.blockData.GetBlockHash(height)
 			if err != nil {
--- a/explorer/explorermiddleware.go
+++ b/explorer/explorermiddleware.go
@@ -42,7 +42,6 @@ func (exp *explorerUI) BlockHashPathOrIndexCtx(next http.Handler) http.Handler {
 			if height > maxHeight {
 				expectedTime := time.Duration(height-maxHeight) * exp.ChainParams.TargetTimePerBlock
 				message := fmt.Sprintf("This block is expected to arrive in approximately in %v. ", expectedTime)
-				log.Errorf("Block with height %d will arrive in future.", height)
 				exp.StatusPage(w, defaultErrorCode, message, NotFoundStatusType)
 				return
 			}
--- a/explorer/explorermiddleware.go
+++ b/explorer/explorermiddleware.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"net/http"
 	"strconv"
+        "fmt"
 
 	"github.com/go-chi/chi"
 )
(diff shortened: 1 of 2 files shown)


chappjc:
Please let me know when you're ready for review.  I have several suggestions now, but I will wait until you are done.

SeaLightHouse:
ok

SeaLightHouse:
2018-08-29-181453_1366x768_scrot
Before

SeaLightHouse:
2018-08-29-181552_1366x768_scrot
after

SeaLightHouse:
OK @chappjc , Please provide your review .
Thanks

explorer/explorermiddleware.go
```diff
@@ -7,6 +7,7 @@ import (
 	"context"
 	"net/http"
 	"strconv"
+        "fmt"
```
chappjc: All .go files need to be processed with `go fmt`.  I suggest configuring your editor to do this automatically on save.

I use Visual Studio Code with the Go extension.

explorer/explorermiddleware.go
```diff
@@ -35,6 +36,19 @@ func (exp *explorerUI) BlockHashPathOrIndexCtx(next http.Handler) http.Handler {
...
+                        if height > int64(maxHeight) {
+                        maturityTime := int(float64(height - int64(maxHeight) ) * exp.ChainParams.TargetTimePerBlock.Minutes())
+                        days := maturityTime / 1440
+                        hours := (maturityTime % 1440  ) / 60
+                        minutes := maturityTime % 60
+                        message := fmt.Sprintf("This block will arrive in approximately %d days, %d hour, %d min.",days,hours,minutes)
+                        log.Errorf("Block with height %d will arrive in future.", height)
+                        exp.StatusPage(w, defaultErrorCode, message, ErrorStatusType)
+                        return
+                        }
```
chappjc: The lack of go fmt in this file to.

explorer/explorermiddleware.go
chappjc: Also, a good go extension will sort the imports, but leave them in separate blocks like we have done for std lib imports.

explorer/explorermiddleware.go
```diff
@@ -35,6 +36,19 @@ func (exp *explorerUI) BlockHashPathOrIndexCtx(next http.Handler) http.Handler {
 				return
 			}
 		} else {
+                        //Handle blocks which are in future time
```
chappjc: `// Handle future blocks.`

explorer/explorermiddleware.go
```diff
@@ -35,6 +36,19 @@ func (exp *explorerUI) BlockHashPathOrIndexCtx(next http.Handler) http.Handler {
...
+                        maxHeight := exp.blockData.GetHeight()
+                        if height > int64(maxHeight) {
+                        maturityTime := int(float64(height - int64(maxHeight) ) * exp.ChainParams.TargetTimePerBlock.Minutes())
```
chappjc: This is doing: `int(float64(int64-int64) * time.Duration)`, where `int64` is the underlying type of `time.Duration`.  There is no need for the `float64` or `int` conversions.

chappjc: `maturityTime` is confusing.  Please call this `expectedTime`.

explorer/explorermiddleware.go
```diff
@@ -35,6 +36,19 @@ func (exp *explorerUI) BlockHashPathOrIndexCtx(next http.Handler) http.Handler {
...
+                        days := maturityTime / 1440
+                        hours := (maturityTime % 1440  ) / 60
+                        minutes := maturityTime % 60
```
chappjc: In all the duration processing above, please make `maturityTime` (or `expectedTime`) be of type `time.Duration`.  Then you can do `minutes := expectedTime.Minutes()` to avoid most of the "magic numbers" above.  I usually prefer to use standard library functions when they are available.  Note `func (d Duration) Truncate(m Duration) Duration`

explorer/explorermiddleware.go
```diff
@@ -35,6 +36,19 @@ func (exp *explorerUI) BlockHashPathOrIndexCtx(next http.Handler) http.Handler {
...
+                        message := fmt.Sprintf("This block will arrive in approximately %d days, %d hour, %d min.",days,hours,minutes)
```
chappjc: The wording should be less definitive (no use of "will").  Perhaps "This block is expected to arrive in approximately..."

explorer/explorermiddleware.go
```diff
@@ -35,6 +36,19 @@ func (exp *explorerUI) BlockHashPathOrIndexCtx(next http.Handler) http.Handler {
...
+                        log.Errorf("Block with height %d will arrive in future.", height)
```
chappjc: We don't want this logged as `ERR` in the server logs.  It's a user input that did it.  In this case I would not have anything logged on the server.

explorer/explorermiddleware.go
```diff
@@ -35,6 +36,19 @@ func (exp *explorerUI) BlockHashPathOrIndexCtx(next http.Handler) http.Handler {
...
+                        exp.StatusPage(w, defaultErrorCode, message, ErrorStatusType)
```
chappjc: Since we are now returning a sensible message, I'm not sure we want this to be an error.  I do think it's appropriate to have a not found code though.

SeaLightHouse:
ok let me go through and come back by a few hours .

SeaLightHouse:
ok , @chappjc , I have implemented all , except the ErrorType . Please view this file : https://github.com/decred/dcrdata/blob/master/views/status.tmpl
The message is hardcoded when error type is "Not Found" .

SeaLightHouse:
Please advice . One option is not to use "Not Found" status type . May be use "Error status type, or create a new status type . Another option is to modyfy the status.tmpl file .

SeaLightHouse:
@chappjc I have made all the changes .

explorer/explorermiddleware.go
```diff
@@ -36,18 +37,15 @@ func (exp *explorerUI) BlockHashPathOrIndexCtx(next http.Handler) http.Handler {
...
-                        log.Errorf("Block with height %d will arrive in future.", height)
-                        exp.StatusPage(w, defaultErrorCode, message, ErrorStatusType)
-                        return
-                        }
+			//Handle future blocks
+			maxHeight := int64(exp.blockData.GetHeight())
+			if height > maxHeight {
+				expectedTime := time.Duration(height-maxHeight) * exp.ChainParams.TargetTimePerBlock
+				message := fmt.Sprintf("This block is expected to arrive in approximately in %v. ", expectedTime)
+				log.Errorf("Block with height %d will arrive in future.", height)
```
chappjc: It's still log.Error, but we don't want an ERR in the logs for this.  It's really Debugf, not an error in any way.

SeaLightHouse:
ok , sure

SeaLightHouse:
2018-08-30-222321_1366x768_scrot

SeaLightHouse:
2018-08-30-222528_1366x768_scrot

raedah:
Rather then 49801h, it should show days. Also, since this does not need to be squeezed into a column, it can use readable words like hours minutes  seconds  instead of abbreviating.

SeaLightHouse:
At @raedah, I am doing the updation with websocket , will include this there

SeaLightHouse:
In a new pr

Issue #324, opened 2018-01-27
Genesis contains invalid transaction


raedah:
Genesis contains invalid transaction. Say a msg about why instead of an error.
https://explorer.dcrdata.org/explorer/block/298e5cc3d985bfe7f81dc135f360abe089edd4396b86d2de66b0cef42b21d980

chappjc:
resolved by https://github.com/decred/dcrdata/pull/657

PR #657, opened 2018-08-31
special handling for genesis block coinbase transaction

chappjc:
This changes the block page to recognize the genesis block and handle the (invalid) coinbase transaction differently.

- Show a UTF8 attention symbol in red next to the tx hash.
- Add a title (hover text) explaining why it is special.
- Change the link to point to code (chaincfg/genesis.go).

image

This also adds a function `GenesisTxHash` in package txhelpers to retrieve the expected hash of the genesis coinbase transaction for the given network.

### DIFF
--- a/public/css/main.css
+++ b/public/css/main.css
@@ -590,7 +590,7 @@ body.darkBG .progress {
 #address-qrcode img {
   width: 280px;
 }
-.sidechain-tag {
+.attention {
   color: red;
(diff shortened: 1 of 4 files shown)


chappjc:
Resolves https://github.com/decred/dcrdata/issues/324.  Tagged @raedah for review to see if this is what he had in mind.

Issue #270, opened 2017-11-07
Favicon needs work, transparency


chappjc:
The white background on the favicon, as it appears on the browser's bookmarks has a white background that makes it look not so professional:

![image](https://user-images.githubusercontent.com/9373513/32509881-52edfa4a-c3b4-11e7-95b9-17891761f3d3.png)

The browser tab however seems to have it right:

![image](https://user-images.githubusercontent.com/9373513/32509944-8d52c86e-c3b4-11e7-94e5-698492944866.png)

s-ben:
I just tried to recreate on Brave, Chrome, and Firefox (Quantun v62, non developer edition). Favicon seems to render fine (with expected (not white) background) for me. Is it possible this is limited to Firefox Developer Edition? Already been fixed?

PR #660, opened 2018-09-04
Block  visualization display at the newhome page

dmigwi:
Creates new home page with a block visualization implementation.

screen shot 2018-09-04 at 05 16 14


dmigwi:
This PR introduces a newhome page with a block visualization created from mockup described here https://github.com/decred/dcrdata/issues/451..

Here is the final version of the newhome page.

screen shot 2018-09-10 at 21 29 02

Since the default title was also modified most of the other pages were also affect and modified accordingly. Here is **Before** and **After** modifying the default title on the address history page.


**Before**

screen shot 2018-09-10 at 21 51 06


**After**

screen shot 2018-09-10 at 21 50 44

dmigwi:
https://github.com/decred/dcrdata/issues/453

gozart1:
Have not looked at the code yet, but seems to be coming along nicely. Here are a few first issues from my first impression:

1. Decred logo in header is pixelated on high resolution display ( ideally it should be svg so it's always crisp, less ideal but better than current would be to use an image that's at least 2x the width of the image in the html. Also, it no longer links to the home page, I think we should continue giving it some utility by making it link to the home page. 

2. The sent icon should be vertically aligned with the number sent. Also, to me the current icon isn't very expressive, maybe instead of 2 gray boxes it could be 2 arrows pointing left and right to hint at exchange?
screen shot 2018-09-12 at 2 45 48 pm

3. Would move "Available Supply (of 21 mil)" to the bottom, similar to "(46.17 % of total supply)" in the Staking card, because it doesn't look good with the text wrapping like that.screen shot 2018-09-12 at 2 48 08 pm
 
4. The word "Show" in "Show BlockChain History" implies it will appear in place when clicked. Would change this to "View" since it links to a new page. Also, I don't think the word blockchain should not be in camel case, and the whole sentence doesn't really need to be in titlecase since its not a title or a heading. I'd just put "View blockchain history".
screen shot 2018-09-12 at 2 53 39 pm

5. It's a bit tedious that it takes 2 clicks to view the active state of another block when another is active. Would be nice to be able to go from one to the other in 1 click. Eg. I can just click on block 273804 once to see it's active state. Currently the first click just closes 273805, and I need to click again after that to see 273804.
screen shot 2018-09-12 at 2 42 09 pm

explorer/templates.go
```diff
@@ -107,6 +109,50 @@ var toInt64 = func(v interface{}) int64 {
...
 
+// standard golang package math in go1.9 doesn't support math.Round
+func Round(val float64, places int) (newVal float64) {
+	pow := math.Pow(10, float64(places))
+	digit := pow * val
+	newVal = math.Floor(digit) / pow
+	return
+}
+
+// boldNumPlaces defines the number of decimal places to be write with same font as the whole
```
chappjc: Please lead the comment with

`// float64Formatting formats a float64 value in multiple strings: the digits on the left of the decimal place, the....`

explorer/templates.go
```diff
@@ -120,6 +166,9 @@ func makeTemplateFuncMap(params *chaincfg.Params) template.FuncMap {
 		"divide": func(n, d int64) int64 {
 			return n / d
 		},
+		"divideFloat": func(n float64, d float64) float64 {
```
chappjc: Please elide the first `float64` as we have with the other `int64` functions.

main.go
```diff
@@ -449,12 +449,12 @@ func mainCore() error {
 
 	// Initial data summary for web ui. stakedb must be at the same height, so
 	// we get do this before starting the monitors.
-	blockData, _, err := collector.Collect()
+	blockData, val, err := collector.Collect()
```
chappjc: `val` -> `msgBlock` (or whatever doesn't clash with any other variables in `main`)

public/js/humanize.js
```diff
@@ -8,8 +8,18 @@ var humanize = (function() {
     return Math.round(value * multiplier) / multiplier
   }
   return {
-    decimalParts: function(v, useCommas, precision, asString) {
-      if (!precision || precision > 8) {
+    fmtPercentage: function(val) {
+      var sign = "+"
+      var color = "green"
+      if (val < 0 ){
+        sign = ""
```
chappjc: Should this be `"-"`?

chappjc: And 0 is technically not + or -, but that's no biggie.

dmigwi:
@gozart1 I appreciate your review, Its quite detailed.

I will into point 1 and point 2 as thats a justified anomalies.

On point 3, The mockup design that I was using  had `Available supply (of 21 mil)` located there. Its strange that there is some text wrapping on your end, that should never happen. Please assist me with your display resolution so that I can make the text to never wrap.

Point 4, is justified and I will work on it too.

Point 5. Will be a bit tricky to implement because you as you can see, the card in an active state occupies the space for other cards. Imagine a situation when every card can be activated, some cards will be blocked thereby giving a worse user experience.

db/dbtypes/conversion.go
```diff
@@ -83,3 +84,10 @@ func ChartGroupingToInterval(grouping ChartGrouping) (float64, error) {
 		return -1, fmt.Errorf(`unknown chart grouping "%d"`, grouping)
 	}
 }
+
+// CalculateHashRate calculates the block chain from the difficulty value and
+// the targetTimePerBlock in seconds. The hashrate return is in form PentaHash
```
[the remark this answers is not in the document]
dmigwi: ok. Noted

db/dcrsqlite/apisource.go
```diff
@@ -774,6 +774,15 @@ func (db *wiredDB) GetSDiff(idx int) float64 {
 	return sdiff
 }
 
+func (db *wiredDB) GetDiff(idx int64) float64 {
```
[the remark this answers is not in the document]
dmigwi: ok. Noted

db/dcrsqlite/sqlite.go
```diff
@@ -651,10 +651,17 @@ func (db *DB) RetrieveBlockSummaryByTimeRange(minTime, maxTime int64, limit int)
 	return blocks, nil
 }
 
+// RetrieveDiff returns the difficulty for block ind
```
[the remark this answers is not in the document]
dmigwi: ok. Noted

db/dcrsqlite/sqlite.go
```diff
@@ -651,10 +651,17 @@ func (db *DB) RetrieveBlockSummaryByTimeRange(minTime, maxTime int64, limit int)
...
+func (db *DB) RetrieveDiff(ind int64) (float64, error) {
+	var sdiff, diff float64
+	err := db.QueryRow(db.getSDiffSQL, ind).Scan(&sdiff, &diff)
+	return diff, err
+}
+
 // RetrieveSDiff returns the stake difficulty for block ind
```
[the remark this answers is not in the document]
dmigwi: ok. Noted

explorer/explorer.go
```diff
@@ -310,32 +311,30 @@ func (exp *explorerUI) Store(blockData *blockdata.BlockData, _ *wire.MsgBlock) e
...
-		Transactions:   bData.TxLen,
-		BlockTime:      bData.Time,
-		FormattedTime:  bData.FormattedTime,
-		FormattedBytes: humanize.Bytes(uint64(bData.Size)),
-		Revocations:    uint32(bData.Revocations),
-	}
-	exp.NewBlockData = newBlockData
-	bdHeight := newBlockData.Height
+	// Returns the Block with some more data needed for the full block visualization.
+	exp.NewBlockData = exp.blockData.GetExplorerBlock(val.BlockHash().String())
```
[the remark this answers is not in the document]

explorer/templates.go
```diff
@@ -107,6 +109,50 @@ var toInt64 = func(v interface{}) int64 {
 	}
 }
 
+// standard golang package math in go1.9 doesn't support math.Round
+func Round(val float64, places int) (newVal float64) {
```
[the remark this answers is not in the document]
dmigwi: ok. Noted. I actually added it because the Travis CI still had support for go1.9.

explorer/templates.go
dmigwi: ok. Noted

explorer/templates.go
dmigwi: ok. Noted

main.go
dmigwi: ok. Noted

public/js/humanize.js
dmigwi: ok. Noted

dmigwi: If a value is less than 0 it already has a negative sign.

dmigwi:
@gozart1 On point 2 concerning the Icon change, I can't do much on that end since the [mockup](https://xd.adobe.com/spec/e9f3f22a-7eb2-4bef-8921-9fc4b04a8b6d/screen/78a5a58f-fbfb-4dee-a4b6-0149caf1fdb7/1366x800-home-block-selected/) that I used uses that Icon.

explorer/templates.go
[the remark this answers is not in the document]
chappjc: We're now down to just go1.11. https://github.com/decred/dcrdata/pull/673
Module builds are the build method now, and I didn't think it was worth setting up vgo for go1.10, although we could do that.

chappjc:
@saender gave us the new symbol: https://github.com/decred/dcrdata/issues/670
@denyszayets That's different from the mock up.  Use the new one from @saender ?

db/dcrsqlite/sqlite.go
```diff
@@ -651,14 +651,14 @@ func (db *DB) RetrieveBlockSummaryByTimeRange(minTime, maxTime int64, limit int)
 	return blocks, nil
 }
 
-// RetrieveDiff returns the difficulty for block ind
+// RetrieveDiff returns the difficulty for block ind at the specified chain height.
```
chappjc: I meant to replace the block ind stuff. `// RetrieveDiff returns the difficulty at the specified chain height.` If you feel it needs to directly refer to the input parameter name, it would be clearest to rename it to `height` or say `... at the specified chain height, ind.`  Personally I think it's clear enough.

explorer/explorer.go
```diff
@@ -308,23 +308,23 @@ func (exp *explorerUI) Store(blockData *blockdata.BlockData, val *wire.MsgBlock)
...
 	// Returns the Block with some more data needed for the full block visualization.
-	exp.NewBlockData = exp.blockData.GetExplorerBlock(val.BlockHash().String())
-	bdHeight := exp.NewBlockData.Height
+	newBlockData := exp.blockData.GetExplorerBlock(msgBlock.BlockHash().String())
+	targetTimePerBlock := float64(exp.ChainParams.TargetTimePerBlock)
 	difficulty := blockData.Header.Difficulty
+	bdHeight := newBlockData.Height
 
+	// 248 blocks take almost a day to mine
+	last24hrDifficulty := exp.blockData.DifficultyAtHeight(bdHeight - 248)
```
chappjc: Typo here, I think.  It's 288 blocks/day.  But we want to compute this from `exp.ChainParams.TargetTimePerBlock`

db/dcrsqlite/sqlite.go
dmigwi: ok. Noted

explorer/explorer.go
dmigwi: or I could use the actual timestamp from 24hrs ago to fetch the difficulty value.

chappjc:
@dmigwi Note the new SVG package by @denyszayets: https://github.com/decred/dcrdata/issues/670#issuecomment-420997889

It looks like the symbol is indeed colored differently.

explorer/explorer.go
chappjc: Honestly I don't know what's right.  I would check in #dev and/or with @davecgh

chappjc:
While updating graphics, perhaps consider the favicon issue: https://github.com/decred/dcrdata/issues/270

chappjc:
Sorry, I meant https://github.com/decred/dcrdata/pull/689

dmigwi:
I have set the last 24hr timestamp to be used to fetch the last 24hr difficulty instead of the approximate block height at the last 24hr.

db/dcrsqlite/sqlite.go
```diff
@@ -79,29 +80,31 @@ func NewDB(db *sql.DB) (*DB, error) {
...
 		TableNameSummaries)
-	d.getWinnersByHashSQL = fmt.Sprintf(`select height, winners from %s where hash = ?`,
+	d.getWinnersByHashSQL = fmt.Sprintf(`SELECT height, winners FROM %s WHERE hash = ?`,
 		TableNameSummaries)
 
-	d.getSDiffSQL = fmt.Sprintf(`select sdiff, diff from %s where height = ?`,
+	d.getSDiffSQL = fmt.Sprintf(`SELECT sdiff FROM %s WHERE height = ?`,
 		TableNameSummaries)
-	d.getSDiffRangeSQL = fmt.Sprintf(`select sdiff from %s where height between ? and ?`,
+	d.getLast24hrsDifficulty = fmt.Sprintf(`SELECT diff FROM %s WHERE time >= ? ORDER BY time limit 1`,
```
chappjc: The name `getLast24hrsDifficulty` is in accurate and confusing since it's not 24 hours but the difficulty at or after the given time.

explorer/explorer.go
```diff
@@ -317,8 +317,9 @@ func (exp *explorerUI) Store(blockData *blockdata.BlockData, msgBlock *wire.MsgB
 	difficulty := blockData.Header.Difficulty
 	bdHeight := newBlockData.Height
 
-	// 248 blocks take almost a day to mine
-	last24hrDifficulty := exp.blockData.DifficultyAtHeight(bdHeight - 248)
+	// last 24hr timestamp
+	timestamp := time.Now().Add(-24 * time.Hour).Unix()
+	last24hrDifficulty := exp.blockData.DifficultyInLast24Hrs(timestamp)
```
chappjc: (The function itself doesn't consider 24 hours, just the time stamp.)

db/dcrsqlite/sqlite.go
dmigwi: Will replace it it `getDifficulty` Then add a comment

explorer/templates.go
```diff
@@ -117,7 +117,7 @@ var toInt64 = func(v interface{}) int64 {
 // []string{"342", "12", "132", "000"}. If its not set the returned slice should be
 // []string{"342", "12132", "000"}.
 func float64Formatting(v float64, numPlaces int, useCommas bool, boldNumPlaces ...int) []string {
-	pow := math.Pow(1.0, float64(numPlaces))
+	pow := math.Pow(10, float64(numPlaces))
```
chappjc: Ah, good catch!

explorer/explorer.go
dmigwi: will remove the 24hrs part and add a comment.

chappjc:
Milestone set to 3.1, with the understanding that this will be accessible on /nexthome (for testing).  When the home is polished and other pages like /stats are done, we can expose them and/or make them default for 4.0.

dmigwi:
Kindly take note of the change is the colors displayed in the websocket statuses.

I have replaced the actual use of colors with the svg Icons that were provided by @denyszayets.
Most notable change (not sure if it should be this way) is that the connecting Icon changes from yellow to blue... The other two have no major change.

screen shot 2018-09-21 at 01 16 15

chappjc:
Do these svg status icons still pulse slowly?

dmigwi:
Yes. But only the connecting one does not pulse. It did not pulse even initially.

chappjc:
Just wondering.  No big deal.  I'll fire this PR up tonight

chappjc:
Here's what I see after a clear browser cache and hard reload:

image

Something is amiss with the decimal formatting.

dmigwi:
Crap... I missed cross checking the home. Great catch, I am fixing that ASAP.

dmigwi:
screen shot 2018-09-21 at 21 19 06

@chappjc Fixed the issue about decimals formatting on the original home page.

chappjc:
Looks like the decimal place issue still exists on the transaction and address pages:

image


image

chappjc:
Some values on the blocks page too.  What's doing that?

dmigwi:
The prototype that I was using had css with functionality such that if `span` tags are nested inside a `span` tag then it adds space between each of the nested tags.

The solution is to just set `display:flex;` for all the places have that problem.

dmigwi:
I have modified every statement that uses decimalParts template to format the float value.
It was easier to do it this way rather that to make changes on the `nexthome` page where elements are tightly coupled.

I cross-checked again and again the decimal places formatting on all pages but I wouldn't if you gave it another look just to confirm that no new bug was introduced.

chappjc:
I'm good with merging this, as we're in the 3.1 dev phase and this is a good starting point for the next steps.  However, a cursory JS and CSS review by @gozart1 would be great.

The need for 2 clicks to go from block to block is a pain.  I agree with @gozart1 about that.  If a fix does not seem straightforward, or requires more discussion, we can save that an move on as long as an issue is created for anything not addressed in this PR.

Regarding point 3 brought up by gozart, I do not think the entire "Available supply (of 21 mil)" needs to go below the number.  "Available supply" logically should go above it as each value has a header like this, but "(of ~21 million)" could go below.  I also think the "~" and spelling out "million" are improvements here.

gozart1:
This changes the search UX, making it less convenient.

Previously the search bar would autofocus when the page loaded and remain focused after a search result loaded. But now the user must first click with the mouse to reveal the search input before being able to enter a query.

There's a good reason all the leading bitcoin block explorers load the page with an auto-focused search input. Why are we bucking the convention? I think it looks great, but maybe there's a way to get the best of both worlds?  @denyszayets can you chime in here since this is going off your spec? 

At the very least, we could throw our power users a bone by leaving the input revealed and focused upon completing a search.

Also, mobile layout needs some work:
mobile search

chappjc:
That was my thought too.  I think we should revise the spec.

dmigwi:
Concerning the mobile layout, I informed @raedah that we need a responsive design layout for the whole dcrdata application and he is doing something concerning that. On smaller screens most of the page are pretty unresponsive. Maybe we could wait till we have at least WIP responsive design for all the pages.

On the issue about search input requiring extra clicks to search stuff, @denyszayets is better placed to explain why he thought the new approach is better. IMO @gozart1 concern is valid.

denyszayets:
@dmigwi @chappjc Search field is always closed by default. By close I mean, you can only see Search icon. On click it expands. Maybe an expanded version was left out on that screen by mistake.

chappjc:
@denyszayets What you described is what we are seeing too.  It's not having it already expanded and autofocused with the cursor ready that is somewhat of a regression IMO.

denyszayets:
> @denyszayets What you described is what we are seeing too. It's not having it already expanded and autofocused with the cursor ready that is somewhat of a regression IMO.

I agree. Let's make it then already expanded and auto-focused just like you suggested.

dmigwi:
I have made the search input always expanded and auto focused. No more button clicks needed to display the search input.

screen shot 2018-09-26 at 01 55 22

raedah:
The above works for me. We should have a follow up PR ready in the next few days to add in the visual blocks component for the right side.

gozart1:
I think the alignment of the magnifying glass icon next to logo needs some work.
alignment

It would likely look less jumbled if:
- the top of the magnifying glass icon lined up with the top of the Decred logo. 
- the "Search for blocks..." text lined up with the 'ECRED" part of the logo a bit more.
- the icon was the same value as the placeholder instead of jet black ( since that competes with the logo) 

Here's just a quick photoshop sketch of the suggested tweaks:
better

Also, do we really need to make the header that much taller? This has a significant impact on the mobile experience, since screen real estate is at a premium there.  I At the very least, lets shave a bit off the top so that the whitespace between the top of logo and bottom of logo is balanced.

dmigwi:
Shaved off some space from the title bar and realigned the search Icon.

screen shot 2018-09-28 at 20 56 45

chappjc:
@dmigwi If you could write a commit message for the squashed commit and post it as a comment, we'll merge.

dmigwi:
@chappjc How is this? 👇 

This PR introduces a graphical display of the blockchain's latest blocks and some network statistics about the current environment that dcrdata is run in i.e mainnet, testnet or simnet. 

The network statistics comprises of Coin Supply, Dev reward, Project Fund, Locked Ticket Pool value in, Ticket(PoS) reward,  Ticket Price, Difficulty, PoW reward and the Hashrate. 

Mempool data is listed as the first block display on the decred Blockchain's graphical blocks view.

Issue #295, opened 2017-12-07
Show valid/invalid for blocks and transactions


chappjc:
The majority of votes mined in block A can invalidate block A-1.  That is, the regular transactions in block A-1 are invalidated.

We should show:
1. whether a block's regular transactions are valid (votebits = 0 on the next block)
2. whether a transaction is valid or invalid

raedah:
Visually indicate when a block and its transactions are invalidated by stake votes for no on the previous block.
The could be done by displaying a notification on the blocks page, and changing the colors to red, to indicate that they are not valid.

chappjc:
This is partially resolved by https://github.com/decred/dcrdata/pull/654, which added the APPROVED  true/false text on a block's summary table.

![image](https://user-images.githubusercontent.com/9373513/45057805-3788eb80-b05c-11e8-925a-d0fb8444a9ea.png).

On the page of a disapproved block, the regular transactions (including the coinbase) should be grayed out or something to indicate that they are not valid, at least in that block.

The pages for the individual transactions should have an indication of if they were mined in an approved block (with a valid regular transaction tree).  Although this may be confusing since a transaction may be in more than one block (and more than one invalidated block).  As such, for transactions that were mined in multiple blocks, there should be a list of blocks in which the transaction was mined.

chappjc:
PR https://github.com/decred/dcrdata/pull/663 addresses styling of regular transactions in invalidated blocks.

Next is adding to the transaction page a list of blocks in which the transactions was mined (at most 1 approved/valid block).

PR #663, opened 2018-09-04
views: style invalidated transactions on block page.

chappjc:
Create invalidated-tx CSS style to gray out and change background color
of regular transaction table data, including the coinbase.

Resolves https://github.com/decred/dcrdata/issues/295.

image


chappjc:
The "Blocks" page also indicated invalidated blocks:

image

PR #664, opened 2018-09-05
Cache the ticket pool data

dmigwi:
This is a fix to the data racing issue/handling multiple requests fetching similar data, we had initially with https://github.com/decred/dcrdata/pull/542 `tpEvent` implementation

Data is cached till updated data (new block) is available.

screen shot 2018-09-05 at 04 03 03

From the above screenshot the first request is a bit slow then consecutive requests become super fast since no db connection is necessary then.

### DIFF
--- a/db/dcrpg/pgblockchain.go
+++ b/db/dcrpg/pgblockchain.go
@@ -66,6 +66,12 @@ func (d *DevFundBalance) Balance() *explorer.AddressBalance {
 	return d.AddressBalance
 }
 
+// ticketPoolSync helps to manage the calls made to fetch ticket pool data from
+// the db via ticketPoolVisualization method.
+type ticketPoolSync struct {
+	updating trylock.Mutex
+}
+
 // ChainDB provides an interface for storing and manipulating extracted
 // blockchain data in a PostgreSQL database.
 type ChainDB struct {
@@ -83,6 +89,7 @@ type ChainDB struct {
 	devPrefetch        bool
 	InBatchSync        bool
 	InReorg            bool
+	tpSync             *ticketPoolSync
 }
 
 // ChainDBRPC provides an interface for storing and manipulating extracted and
@@ -257,6 +264,7 @@ func NewChainDB(dbi *DBInfo, params *chaincfg.Params, stakeDB *stakedb.StakeData
 		unspentTicketCache: unspentTicketCache,
 		DevFundBalance:     new(DevFundBalance),
 		devPrefetch:        devPrefetch,
+		tpSync:             new(ticketPoolSync),
 	}, nil
 }
 
@@ -547,12 +555,32 @@ func (pgb *ChainDB) GetTicketPoolByDateAndInterval(maturityBlock int64,
 	return retrieveTicketsByDate(pgb.db, maturityBlock, int64(val))
 }
 
-// TicketPoolVisualization fetches the following ticketpool data: tickets
+// TicketPoolVisualization helps block consecutive access to the ticketPoolVisualization
+// method till the first request is resolved.
+func (pgb *ChainDB) TicketPoolVisualization(interval dbtypes.ChartGrouping) (
+	[]*dbtypes.PoolTicketsData, *dbtypes.PoolTicketsData, error) {
+
+	isLocking := pgb.tpSync.updating.TryLock()
+
+	if isLocking {
+		defer pgb.tpSync.updating.Unlock()
+		// check if after the first request handled, the consecutive request response
+		// data was also processed and stored in the cache.
+		barcharts, donutCharts, ok := explorer.GetTicketPoolData(interval)
+		if ok {
+			return barcharts, donutCharts, nil
+		}
+		return pgb.ticketPoolVisualization(interval)
+	}
+	return nil, nil, nil
+}
+
+// ticketPoolVisualization fetches the following ticketpool data: tickets
 // grouped on the specified interval, tickets grouped by price, and ticket
 // counts by ticket type (solo, pool, other split). The interval may be one of:
 // "mo", "wk", "day", or "all". The data is needed to populate the ticketpool
 // graphs. The data grouped by time and price are returned in a slice.
-func (pgb *ChainDB) TicketPoolVisualization(interval dbtypes.ChartGrouping) (
+func (pgb *ChainDB) ticketPoolVisualization(interval dbtypes.ChartGrouping) (
 	[]*dbtypes.PoolTicketsData, *dbtypes.PoolTicketsData, error) {
 	var maturityBlock = pgb.GetTicketPoolBlockMaturity()
 
--- a/db/dcrpg/pgblockchain.go
+++ b/db/dcrpg/pgblockchain.go
@@ -560,19 +560,18 @@ func (pgb *ChainDB) GetTicketPoolByDateAndInterval(maturityBlock int64,
 func (pgb *ChainDB) TicketPoolVisualization(interval dbtypes.ChartGrouping) (
 	[]*dbtypes.PoolTicketsData, *dbtypes.PoolTicketsData, error) {
 
-	isLocking := pgb.tpSync.updating.TryLock()
-
-	if isLocking {
-		defer pgb.tpSync.updating.Unlock()
-		// check if after the first request handled, the consecutive request response
-		// data was also processed and stored in the cache.
-		barcharts, donutCharts, ok := explorer.GetTicketPoolData(interval)
-		if ok {
-			return barcharts, donutCharts, nil
+	for {
+		if pgb.tpSync.updating.TryLock() {
+			defer pgb.tpSync.updating.Unlock()
+			// check if after the first request was handled, the consecutive request
+			// response data was also processed and stored in the cache.
+			barcharts, donutCharts, ok := explorer.GetTicketPoolData(interval)
+			if ok {
+				return barcharts, donutCharts, nil
+			}
+			return pgb.ticketPoolVisualization(interval)
 		}
-		return pgb.ticketPoolVisualization(interval)
 	}
-	return nil, nil, nil
 }
 
 // ticketPoolVisualization fetches the following ticketpool data: tickets
--- a/explorer/explorer.go
+++ b/explorer/explorer.go
@@ -31,7 +31,6 @@ import (
 )
 
 const (
-	cleanUpTickerInterval        = 5
 	maxExplorerRows              = 400
 	minExplorerRows              = 20
 	defaultAddressRows     int64 = 20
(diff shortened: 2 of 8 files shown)


explorer/explorerroutes.go
```diff
@@ -239,14 +239,20 @@ func (exp *explorerUI) Mempool(w http.ResponseWriter, r *http.Request) {
...
 func (exp *explorerUI) Ticketpool(w http.ResponseWriter, r *http.Request) {
-	chartData, groupedTickets, err := exp.explorerSource.TicketPoolVisualization(
-		dbtypes.ChartGroupingFromStr("all"),
-	)
-	if err != nil {
-		log.Errorf("Template execute failure: %v", err)
-		exp.StatusPage(w, defaultErrorCode, defaultErrorMessage, ErrorStatusType)
-		return
+	var err error
+	interval := dbtypes.ChartGroupingFromStr("all")
```
chappjc: `interval := dbtypes.AllChartGrouping` is fine in this case since the input is a constant.

explorer/websockethandlers.go
```diff
@@ -106,20 +106,29 @@ func (exp *explorerUI) RootWebsocket(w http.ResponseWriter, r *http.Request) {
...
-					// TODO: Implement a cache to control websocket connections
-					// meant to serve the same data.
-					cData, gData, err := exp.explorerSource.TicketPoolVisualization(
-						dbtypes.ChartGroupingFromStr(msg.Message),
-					)
-					if err != nil {
-						if strings.HasPrefix(err.Error(), "unknown interval") {
-							log.Debugf("Invalid ticket pool interval provided via getticketpooldata: %s", msg.Message)
-							webData.Message = "Error: " + err.Error()
+					var err error
```
chappjc: Might as well limit the scope of this.  Putting it after `if !ok {` would seem to be fine.

explorer/websockethandlers.go
```diff
@@ -106,20 +106,29 @@ func (exp *explorerUI) RootWebsocket(w http.ResponseWriter, r *http.Request) {
...
+					interval := dbtypes.ChartGroupingFromStr(msg.Message)
+
+					cData, gData, ok := GetTicketPoolData(interval)
```
chappjc: There's still a gotcha with this approach.  What happens when there's another handler already running `TicketPoolVisualization`, but the cache is still out-of-date when a second client requests ticket pool data?  Any subsequent handlers will also query the DB.  This is what all the trylock and ok to update stuff is for in the DevFundBalance type.

explorer/websocket.go
```diff
@@ -188,6 +188,9 @@ func (wsh *WebsocketHub) run() {
 			switch hubSignal {
 			case sigNewBlock:
 				log.Infof("Signaling new block to %d websocket clients.", len(wsh.clients))
+
+				// drops the outdated data in the ticketpool cache.
+				cleanUpTicketPoolData()
```
chappjc: In general I prefer lazy updates or implicit expiration to this since this adds a delay, however small, before it can actually signal to the clients (`for client := range wsh.clients {` below).  Making a couple empty maps is probably no big deal, but it's also waiting for a write lock, which will be a major issue if there is a cache update in progress.

To be clear, by implicit expiration I mean the cache would have the block hash for which it is current, and it would be responsible for checking if it is current.  A more quick and dirty solution might be to simply `go cleanUpTicketPoolData()` so that if it has to wait for the lock, it wont block here.

explorer/explorerroutes.go
dmigwi: ok. Noted

explorer/websockethandlers.go
dmigwi: ok. Noted

explorer/websocket.go
dmigwi: ok. Noted

explorer/websockethandlers.go
dmigwi: So if I get your point trylock in this case scenario is supposed to block consecutive calls made to `TicketPoolVisualization` method till its first instance is resolved?

explorer/websocket.go
dmigwi: This makes lots of sense now. It allows the possibility of scaling horizontally instead of vertically. I.e. we don't overburden a single thread with too much work but we create several threads to share the workload.

main.go
```diff
@@ -527,6 +528,17 @@ func mainCore() error {
...
+	quitCleanUpTicker := make(chan int)
+	defer func() {
+		// safely exit the clean up ticker
+		quitCleanUpTicker <- 1
+	}()
+
+	// Instantiates the ticketpool data cache and set it with the current
+	// block height. Also Lazily cleans up the invalid data in the tickepool
+	// cache when a new block is added.
+	go explore.CleanUpTicketPoolData(quitCleanUpTicker)
```
chappjc: You could use the already existing `quit` channel.

explorer/explorer.go
```diff
@@ -109,27 +107,50 @@ func GetTicketPoolData(interval dbtypes.ChartGrouping) (barGraphs []*dbtypes.Poo
...
-		BarGraphsCache:  make(map[dbtypes.ChartGrouping][]*dbtypes.PoolTicketsData),
-		DonutGraphCache: make(map[dbtypes.ChartGrouping]*dbtypes.PoolTicketsData),
+			ticketPoolGraphsCache = &ticketPoolDataCache{
+				Height:          height,
+				BarGraphsCache:  make(map[dbtypes.ChartGrouping][]*dbtypes.PoolTicketsData),
+				DonutGraphCache: make(map[dbtypes.ChartGrouping]*dbtypes.PoolTicketsData),
+			}
+		case <-quit:
+			log.Info("Stopping the ticketpool cleanUp")
+			ticker.Stop()
```
chappjc: And `return`.

explorer/explorer.go
```diff
@@ -109,27 +107,51 @@ func GetTicketPoolData(interval dbtypes.ChartGrouping) (barGraphs []*dbtypes.Poo
...
-	defer ticketPoolGraphsCache.Unlock()
+// CleanUpTicketPoolData provides a thread-safe way to clean up/invalidate the
+// ticketpool cache data. It resets the ticketpool cache data and sets the
+// current block height. Check for invalid data every cleanUpTickerInterval.
+func (exp *explorerUI) CleanUpTicketPoolData(quit chan int) {
+	ticker := time.NewTicker(time.Duration(cleanUpTickerInterval) * time.Second)
+	for {
+		select {
+		case <-ticker.C:
+			height := exp.Height()
```
chappjc: Two other options to consider instead of polling as you are doing with this solution:

1. `explorer.GetTicketPoolData` could accept a height parameter, doing the clean up and responding with false when it does not have the data.
2. You can remove the channel from `CleanUpTicketPoolData` and call `CleanUpTicketPoolData` directly from `(*appContext).StatusNtfnHandler` in response to signals on either `UpdateStatusNodeHeight` or `UpdateStatusDBHeight`.

explorer/explorer.go
```diff
@@ -109,27 +107,51 @@ func GetTicketPoolData(interval dbtypes.ChartGrouping) (barGraphs []*dbtypes.Poo
...
 	}
 }
 
 // UpdateTicketPoolData updates the ticket pool cache with the latest data fetched.
-// This is a thread-safe way to update ticket pool cache data.
+// This is a thread-safe way to update ticket pool cache data. TryLock helps avoid
+// stacking calls to update the cache.
 func UpdateTicketPoolData(interval dbtypes.ChartGrouping, barGraphs []*dbtypes.PoolTicketsData,
 	donutcharts *dbtypes.PoolTicketsData) {
+	isLocking := ticketPoolGraphsCache.updating.TryLock()
```
chappjc: The "is already updating?" check needs to be where the DB query is launched since that's what we want to prevent stacking up.

main.go
dmigwi: ok. Noted

explorer/explorer.go

db/dcrpg/pgblockchain.go
```diff
@@ -547,12 +555,32 @@ func (pgb *ChainDB) GetTicketPoolByDateAndInterval(maturityBlock int64,
...
+		defer pgb.tpSync.updating.Unlock()
+		// check if after the first request handled, the consecutive request response
+		// data was also processed and stored in the cache.
+		barcharts, donutCharts, ok := explorer.GetTicketPoolData(interval)
+		if ok {
+			return barcharts, donutCharts, nil
+		}
+		return pgb.ticketPoolVisualization(interval)
+	}
+	return nil, nil, nil
```
chappjc: We're still missing the mark with these edits.  There should be no path that returns no data.

There is a data request function that ensures fresh data, and which calls an update function if the data is not fresh.

A **data request** call:

- Get cached data (with a RLock) if it is not stale.
- If the cache was fresh, RUnlock and be done.
- If the cached data was stale/absent, the caller should attempt to **update** the cache.  This is where things are tricky.

An **update request** call (remember, multiple callers may be following this path, and we do not want multiple duplicate queries *or* failed requests / empty responses):
- Callers should attempt to get the "OK to update" lock (the trylock) in case the cached data is stale.  This lock indicates if a query is already running.
- Depending on if the caller got the updating lock, there are different paths, but returning with no data is not one.
- Regardless of whether the caller got the updating lock or not, wait on the data lock.  The data lock is needed:
  - If the updating lock was not obtained, the data lock is needed because there is an update in progress and they must wait for the data in cache to be refreshed!
  - If the updating lock was obtained, the lock is needed because the caller is about to do the update themselves.
  - Either way, callers must wait for the update to complete (when the data lock is free).
- A caller that **did not get the updating lock** will eventually be unblocked when the updating caller finishes and releases the data lock.  At this point they will return from the update function (when the cache is finally updated).
- A caller that **did get the updating lock** will finish the actual query and cache update, release the updating lock *and* the data lock, and return indicating that the cache was updated.
- On returning to the data request function, which called the update function, fresh data may be obtained from the cache.

dmigwi: At that point where no data is expected, I should actually return an error that something went wrong with the trylock. Because the all the ticketPoolVisualization requests are suppose to be pilled and resolved one at a time by checking the actual cache or making the db call if the data doesn't exist.

dmigwi: `return nil, nil, nil` we don't expect it to ever execute since something will have gone wrong and therefore I should return an error that the trylock did not work as expected?

dmigwi: Maybe the question to ask is, Is trylock a blocking operation when the resource is not available for locking?

dmigwi: I am thinking then of a for loop and ticker will help here.

db/dcrpg/pgblockchain.go
```diff
@@ -547,12 +555,39 @@ func (pgb *ChainDB) GetTicketPoolByDateAndInterval(maturityBlock int64,
 	return retrieveTicketsByDate(pgb.db, maturityBlock, int64(val))
 }
 
-// TicketPoolVisualization fetches the following ticketpool data: tickets
+// TicketPoolVisualization helps block consecutive access to the ticketPoolVisualization
+// method till the first request is resolved.
+func (pgb *ChainDB) TicketPoolVisualization(interval dbtypes.ChartGrouping) (
+	[]*dbtypes.PoolTicketsData, *dbtypes.PoolTicketsData, error) {
+	ticker := time.NewTicker(1 * time.Second)
```
dmigwi: @chappjc How does a loop and timer here work?

dmigwi: Maybe one second is too long to wait?

dmigwi:
@chappjc I have used a for loop only.

chappjc:
Superseded by the joint PR by @dmigwi and @chappjc: https://github.com/decred/dcrdata/pull/664

Issue #642, opened 2018-08-28
address page, Highlight corresponding txid on hover over


raedah:
Currently one can read the link of a 'source' or 'spent' link on the address page, and then look to find the matching txid in the table list. This could be automated for the user with JS by automatically highlighting the corresponding row with the matching txid when hovering over a source or spent link.

chappjc:
Absolutely, even a "pulse" would be good.

The relevant information or output could be indicated on the url with an #anchor, similar to how insight does this.

chappjc:
Sorry I responded on the PR by accident: https://github.com/decred/dcrdata/pull/665

chappjc:
Resolved by https://github.com/decred/dcrdata/pull/696

PR #665, opened 2018-09-05
Address page tx highlighting

papacarp:
This PR is for issue #642.  There are three potential aspects to this issue - I have finished only the first one:

1.  This version will highlight the transactions on the page.  The matched transaction Vin/out is not easily available in the address page raw data and thus right now all we are doing is highlighting the transaction, but if there are multiple identical transactions they all get highlighted.

See example gif below. (address: DsiJmwqFbsCEc1H3HF8UeRjoqJtKt9ZFRKw)
recording 67

2.  I can update the address page to also pull the matched tx vin/vout index but that will likely require changes to the query and/or a secondary query.  I don't know that its that critical to be able to isolate each vin/vout index.

3.  @chappjc mentions also adding a url parameter to the source/spent links to allow the relevant in/out to be highlighted when visiting the tx page as well from that url.  To do that we would also need the vin/vout of the matched tx.

Please let me know if you want me to proceed with steps 2 and 3 here to get the matched tx vin/out index.

### DIFF
--- a/db/dcrpg/internal/addrstmts.go
+++ b/db/dcrpg/internal/addrstmts.go
@@ -1,9 +1,9 @@
 package internal
 
 const (
-	insertAddressRow0 = `INSERT INTO addresses (address, matching_tx_hash, tx_hash,
+	insertAddressRow0 = `INSERT INTO addresses (address, matching_tx_hash, matching_tx_index, tx_hash,
 		tx_vin_vout_index, tx_vin_vout_row_id, value, block_time, is_funding, valid_mainchain, tx_type)
-		VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10) `
+		VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11) `
 
 	InsertAddressRow = insertAddressRow0 + `RETURNING id;`
 
@@ -37,11 +37,12 @@ const (
 		is_funding BOOLEAN,
 		tx_vin_vout_index INT4,
 		tx_vin_vout_row_id INT8,
-		tx_type INT4
+		tx_type INT4,
+		matching_tx_index INT4
 		);`
 
 	addrsColumnNames = `id, address, matching_tx_hash, tx_hash, valid_mainchain,
-		tx_vin_vout_index, block_time, tx_vin_vout_row_id, value, is_funding`
+		tx_vin_vout_index, block_time, tx_vin_vout_row_id, value, is_funding, matching_tx_index`
 
 	SelectAddressAllByAddress = `SELECT ` + addrsColumnNames + ` FROM addresses WHERE address=$1 ORDER BY block_time DESC;`
 	SelectAddressRecvCount    = `SELECT COUNT(*) FROM addresses WHERE address=$1 AND valid_mainchain = TRUE;`
@@ -143,8 +144,8 @@ const (
 	SelectAddressIDByVoutIDAddress = `SELECT id FROM addresses WHERE address=$1 AND
 	    tx_vin_vout_row_id=$2 AND is_funding = TRUE;`
 
-	SetAddressFundingForMatchingTxHash = `UPDATE addresses SET matching_tx_hash=$1
-		WHERE tx_hash=$2 AND is_funding = TRUE AND tx_vin_vout_index=$3;`
+	SetAddressFundingForMatchingTxHash = `UPDATE addresses SET matching_tx_hash=$1, matching_tx_index=$2
+		WHERE tx_hash=$3 AND is_funding = TRUE AND tx_vin_vout_index=$4;`
 
 	SetAddressMainchainForVoutIDs = `UPDATE addresses SET valid_mainchain=$1
 		WHERE is_funding = TRUE AND tx_vin_vout_row_id=$2;`
@@ -224,6 +225,15 @@ const (
 
 	SetTxTypeOnAddressesByVinAndVoutIDs = `UPDATE addresses SET tx_type=$1 WHERE
 		tx_vin_vout_row_id=$2 AND is_funding=$3;`
+
+	UpdateMatchingFundingTxIdFromVins = `UPDATE addresses SET matching_tx_index=tx_index
+		FROM vins WHERE matching_tx_hash=vins.tx_hash AND addresses.tx_hash=prev_tx_hash
+		AND addresses.tx_vin_vout_index=prev_tx_index AND matching_tx_hash !=''
+		AND is_funding=true AND addresses.id >= $1 AND addresses.id < $2;`
+
+	UpdateMatchingSpendingTxIdFromVins = `UPDATE addresses SET matching_tx_index=vins.prev_tx_index
+		FROM vins WHERE vins.id=tx_vin_vout_row_id AND is_funding=false AND matching_tx_hash != ''
+		AND addresses.id >= $1 AND addresses.id < $2;`
 
 	IndexBlockTimeOnTableAddress   = `CREATE INDEX block_time_index ON addresses (block_time);`
 	DeindexBlockTimeOnTableAddress = `DROP INDEX block_time_index;`
--- a/db/dcrpg/internal/addrstmts.go
+++ b/db/dcrpg/internal/addrstmts.go
@@ -227,12 +227,13 @@ const (
 		tx_vin_vout_row_id=$2 AND is_funding=$3;`
 
 	UpdateMatchingFundingTxIdFromVins = `UPDATE addresses SET matching_tx_index=tx_index
-		FROM vins WHERE matching_tx_hash=vins.tx_hash AND addresses.tx_hash=prev_tx_hash
-		AND addresses.tx_vin_vout_index=prev_tx_index AND matching_tx_hash !=''
+		FROM vins WHERE addresses.tx_hash=prev_tx_hash
+		AND addresses.tx_vin_vout_index=prev_tx_index
+		AND vins.is_valid=addresses.valid_mainchain
 		AND is_funding=true AND addresses.id >= $1 AND addresses.id < $2;`
 
 	UpdateMatchingSpendingTxIdFromVins = `UPDATE addresses SET matching_tx_index=vins.prev_tx_index
-		FROM vins WHERE vins.id=tx_vin_vout_row_id AND is_funding=false AND matching_tx_hash != ''
+		FROM vins WHERE vins.id=tx_vin_vout_row_id AND is_funding=false
 		AND addresses.id >= $1 AND addresses.id < $2;`
 
 	IndexBlockTimeOnTableAddress   = `CREATE INDEX block_time_index ON addresses (block_time);`
--- a/db/dbtypes/types.go
+++ b/db/dbtypes/types.go
@@ -1,6 +1,7 @@
 package dbtypes
 
 import (
+	"database/sql"
 	"database/sql/driver"
 	"encoding/json"
 	"fmt"
@@ -357,6 +358,7 @@ type AddressRow struct {
 	// MatchingTxHash provides the relationship between spending tx inputs and
 	// funding tx outputs.
 	MatchingTxHash   string
+	MatchingTxIndex  sql.NullInt64
 	IsFunding        bool
 	TxBlockTime      uint64
 	TxHash           string
--- a/db/dcrpg/queries.go
+++ b/db/dcrpg/queries.go
@@ -515,7 +515,7 @@ func insertSpendingTxByPrptStmt(tx *sql.Tx, fundingTxHash string, fundingTxVoutI
 	// Insert the new spending tx
 	var isFunding bool
 	sqlStmt := internal.MakeAddressRowInsertStatement(checked)
-	err = tx.QueryRow(sqlStmt, newAddr, fundingTxHash, spendingTxHash,
+	err = tx.QueryRow(sqlStmt, newAddr, fundingTxHash, fundingTxVoutIndex, spendingTxHash,
 		spendingTxVinIndex, vinDbID, value, blockTime, isFunding,
 		validMainchain, txType).Scan(&rowID)
 	if err != nil {
@@ -525,7 +525,7 @@ func insertSpendingTxByPrptStmt(tx *sql.Tx, fundingTxHash string, fundingTxVoutI
 	// Update the matchingTxHash for the funding tx output. matchingTxHash here
 	// is the hash of the funding tx.
 	res, err := tx.Exec(internal.SetAddressFundingForMatchingTxHash,
-		spendingTxHash, fundingTxHash, fundingTxVoutIndex)
+		spendingTxHash, spendingTxVinIndex, fundingTxHash, fundingTxVoutIndex)
 	if err != nil || res == nil {
 		return 0, fmt.Errorf("SetAddressFundingForMatchingTxHash: %v", err)
 	}
@@ -867,7 +867,7 @@ func scanAddressQueryRows(rows *sql.Rows) (ids []uint64, addressRows []*dbtypes.
 		// Scan values in order of columns listed in internal.addrsColumnNames
 		err = rows.Scan(&id, &addr.Address, &addr.MatchingTxHash, &txHash,
 			&addr.ValidMainChain, &txVinIndex, &blockTime, &vinDbID,
-			&addr.Value, &addr.IsFunding)
+			&addr.Value, &addr.IsFunding, &addr.MatchingTxIndex)
 		if err != nil {
 			return
 		}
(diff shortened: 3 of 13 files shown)


chappjc:
I do think we'll want to implement highlighting the relevant in/out when visiting the transaction page, but this seems like a good start.  Pulling the matching tx index seems like a minor change to the query.  Are there other complications?

papacarp:
No further complications that I'm aware of.  Once we have vin/vout index for the matched TX, the other two  items should go fast.  I'll proceed.

chappjc:
The ">" that shows up when you expand a transaction details on the address page:

image

chappjc:
Oh, that's not and `#anchor`.  But whether it's an anchor or a url param (e.g `?in=1`) or a path element (e.g.  `/tx/{txid}/in/{idx}` or `/tx/{txid}/out/{idx}`) is not a big deal.  However, the path element approach fits with our current router and middleware design.

papacarp:
Thank you for pointing it out.  Yes, I see what they are doing.  I was thinking to have a URL parameter as well rather than an anchor.

papacarp:
recording 68
1.  Address page queries now pull the vout/vin of the matching transactions.  @chappjc I could use your excellent ideas on these queries.  What I have seems to work, but I'm not sure there is no situation where the result is ambiguous.
```
SELECT 
     a1.id, a1.address, a1.matching_tx_hash, a1.tx_hash, a1.valid_mainchain,
     a1.tx_vin_vout_index, a1.block_time, a1.tx_vin_vout_row_id, a1.value, a1.is_funding, 
     a2.tx_vin_vout_index as matching_tx_index 
FROM addresses AS a1
LEFT JOIN addresses AS a2
     ON a2.tx_hash=a1.matching_tx_hash AND a2.matching_tx_hash=a1.tx_hash AND a2.valid_mainchain=TRUE
WHERE a1.address=$1 AND a1.valid_mainchain = TRUE 
ORDER BY a1.block_time DESC LIMIT $2 OFFSET $3;
```
2.  The TX page now accepts optional extended url in the format of `{tx_hash}/{in or out}/{index}` as an example:  ` /tx/6727343a13d2991c256a07cf18c95c3f04d3264a07e5bdbdb2c35a123d4fbc26/in/0` which will subsequently highlight the input or output row

3.  previous outpoints, spent links and transaction entries in the address table all have been updated to use this method.

3.  Links without extended URL still work like they used to.

chappjc:
Thanks for the update.  I'll look at it first thing in the a.m.

papacarp:
No problem.  If there are no ah-ha moments on alternative solutions for the query then I'll separate out this query and processing for the address page to isolate the possible implications.  I agree its extensive, but until we had a strategy on that query it was easiest to tweak what we already have slightly globally. 

Another option, is upgrading the addresses table to include matched-tx-index as well.  I think we struggled a bit with a different query because this was lacking as well.

As I'm testing this I'm finding some errors were introduced in the unconfirmed transaction rendering - probably related to the address table updates awhile back.  I'll try to fix those issues as well.

papacarp:
I just rebased to get the latest address page fixes

chappjc:
Comparing warmed-up queries for the previous and joined queries indicates a bit more than twice the exec time:

```
dcrdata=# explain analyze select id, address, matching_tx_hash, tx_hash, valid_mainchain, tx_vin_vout_index, block_time, tx_vin_vout_row_id, value, is_funding FROM addresses WHERE address='Dcur2mcGjmENx4DhNqDctW5wJCVyT3Qeqkx' ORDER BY block_time DESC;
                                                                       QUERY PLAN                                                               
         
------------------------------------------------------------------------------------------------------------------------------------------------
---------
 Gather Merge  (cost=286752.43..325784.60 rows=320058 width=197) (actual time=204.773..338.786 rows=318097 loops=1)
   Workers Planned: 6
   Workers Launched: 6
   ->  Sort  (cost=285752.33..285885.69 rows=53343 width=197) (actual time=201.797..204.725 rows=45442 loops=7)
         Sort Key: block_time DESC
         Sort Method: quicksort  Memory: 17659kB
         ->  Parallel Bitmap Heap Scan on addresses  (cost=5618.20..281564.10 rows=53343 width=197) (actual time=58.672..187.903 rows=45442 loop
s=7)
               Recheck Cond: (address = 'Dcur2mcGjmENx4DhNqDctW5wJCVyT3Qeqkx'::text)
               Heap Blocks: exact=35621
               ->  Bitmap Index Scan on uix_addresses_address  (cost=0.00..5538.19 rows=320057 width=0) (actual time=30.720..30.720 rows=319233 
loops=1)
                     Index Cond: (address = 'Dcur2mcGjmENx4DhNqDctW5wJCVyT3Qeqkx'::text)
 Planning time: 0.078 ms
 Execution time: 345.460 ms
(13 rows)


dcrdata=# EXPLAIN ANALYZE SELECT a1.id, a1.address, a1.matching_tx_hash, a1.tx_hash, a1.valid_mainchain, a1.tx_vin_vout_index, a1.block_time, a1.tx_vin_vout_row_id, a1.value, a1.is_funding, a2.tx_vin_vout_index as matching_tx_index FROM addresses AS a1 LEFT JOIN addresses AS a2 ON a2.tx_hash=a1.matching_tx_hash AND a2.matching_tx_hash=a1.tx_hash WHERE a1.address='Dcur2mcGjmENx4DhNqDctW5wJCVyT3Qeqkx' ORDER BY a1.block_time DESC;
                                                                           QUERY PLAN                                                           
                
------------------------------------------------------------------------------------------------------------------------------------------------
----------------
 Gather Merge  (cost=681784.96..720817.13 rows=320058 width=201) (actual time=576.715..723.201 rows=318097 loops=1)
   Workers Planned: 6
   Workers Launched: 6
   ->  Sort  (cost=680784.86..680918.22 rows=53343 width=201) (actual time=574.192..577.159 rows=45442 loops=7)
         Sort Key: a1.block_time DESC
         Sort Method: quicksort  Memory: 14945kB
         ->  Nested Loop Left Join  (cost=5618.89..676596.63 rows=53343 width=201) (actual time=60.206..560.741 rows=45442 loops=7)
               ->  Parallel Bitmap Heap Scan on addresses a1  (cost=5618.20..281564.10 rows=53343 width=197) (actual time=60.085..140.659 rows=4
5442 loops=7)
                     Recheck Cond: (address = 'Dcur2mcGjmENx4DhNqDctW5wJCVyT3Qeqkx'::text)
                     Heap Blocks: exact=28167
                     ->  Bitmap Index Scan on uix_addresses_address  (cost=0.00..5538.19 rows=320057 width=0) (actual time=32.576..32.576 rows=3
19233 loops=1)
                           Index Cond: (address = 'Dcur2mcGjmENx4DhNqDctW5wJCVyT3Qeqkx'::text)
               ->  Index Scan using uix_addresses_funding_tx on addresses a2  (cost=0.69..7.40 rows=1 width=127) (actual time=0.006..0.009 rows=
0 loops=318097)
                     Index Cond: (tx_hash = a1.matching_tx_hash)
                     Filter: (matching_tx_hash = a1.tx_hash)
                     Rows Removed by Filter: 26
 Planning time: 0.765 ms
 Execution time: 729.912 ms
(18 rows)
```

Before warm-up, the second query was ~1900ms.  Keeping the second query warm is likely requiring a much larger cache, the index uix_addresses_funding_tx at least.

So there could be two entire sets of queries: where the other index is required and where it is not required.  Or we could add the matching_tx_index to the table.  I don't love data replication, but this is starting to look like the better option.

chappjc:
Now that I consider the JOIN closely, I believe this part is not sufficient: `a2.tx_hash=a1.matching_tx_hash AND a2.matching_tx_hash=a1.tx_hash`.  This could apply two multiple matching in/out inds.

papacarp:
@chappjc yes, I was worried about ambiguity on those.  Do you want me to put together the code to upgrade the db?  While we are at it, what would you think about indexing into matching txid as a row_id rather than a hash?  Intuitively it seems like integer indexes and db operations would be faster than string ones.  @dmigwi mentioned a few weeks back that tx hash's are only guaranteed unique on a per block basis rattling around in the back of my head as part of this proposal as well.

chappjc:
@papacarp Correct, there can be multiple identical tx_hash values in the transactions table because a side chain transaction and/or a pos-invalidated transaction may get re-mined in another block.  And we are recording all instances.

Recording the matching transaction for a row of the addresses table as row_id instead of tx hash is not necessary though.  In fact, we want to be able to get them all (if intentionally omitting WHERE is_mainchain = true and is_valid = true, or limit 1, etc.).  The vins table (to which `tx_vin_vout_row_id` points), and the addresses table itself, were not intended to be concerned with occurrences in blocks.  Except for the `is_valid`, `is_mainchain`, and `block_time` columns that were added recently to these tables, and which are updated any time the transaction containing the vin/address is seen, all the other data pertains only to a component of a transaction.  Notice in the "upsert" statements that the valid/mainchain and block_time columns are updated, but not the other stuff that would stay the same for a given tx with that hash.

In the case of the addresses table, there is a _unique_ index on `tx_vin_vout_row_id, address, is_funding`, which obviously corresponds to a single vin/vout row  (but there may be multiple rows in addresses with that same `tx_vin_vout_row_id`).  Rows of both the vins and vouts tables are unique on `tx_hash, tx_index, tx_tree`, which can correspond to multiple rows of the transactions table.

I don't know if this is a great design, but that's how it's set up.

dmigwi:
>While we are at it, what would you think about indexing into matching txid as a row_id rather than a hash?

This was an implementation that we considered before we decided to stick the `tx_hash`. On the address history page we have source and spent links added to the respective table rows. We never wanted to build a  new functionality i.e. use the `row_id` to show funds flow while we still had the `/tx/{tx_hash}` endpoint that was in existence.
That is why we decided against matching tx column holding the `row_id` in favour of the `tx_hash`.

db/dcrpg/internal/addrstmts.go
```diff
@@ -225,6 +226,15 @@ const (
...
 
+	UpdateMatchingFundingTxIdFromVins = `UPDATE addresses SET matching_tx_index=tx_index
+		FROM vins WHERE matching_tx_hash=vins.tx_hash AND addresses.tx_hash=prev_tx_hash
+		AND addresses.tx_vin_vout_index=prev_tx_index AND matching_tx_hash !=''
+		AND is_funding=true AND addresses.id >= $1 AND addresses.id < $2;`
+
+	UpdateMatchingSpendingTxIdFromVins = `UPDATE addresses SET matching_tx_index=vins.prev_tx_index
+		FROM vins WHERE vins.id=tx_vin_vout_row_id AND is_funding=false AND matching_tx_hash != ''
+		AND addresses.id >= $1 AND addresses.id < $2;`
+
```
papacarp: @chappjc and @dmigwi  these are the two upgrade queries.  I would appreciate your feedback on them.  On my laptop the upgrade process takes 5 hours - I'm guessing on one of your machines it would be faster.  I saw @dmigwi added temporary indexes for one of his upgrades - perhaps I could experiment with that.  

The linkages seem to be accurate and the queries appropriately fill in all lines in address table with non blank matching_tx lines with a matching_tx_index.   However, as with the other queries on this table I'm not sure if we still have ambiguity due to the sidechains or re-mined items. 

Your feedback would be appreciated.

papacarp:
This seems to be in pretty good shape functionally.  The upgrade time is a bit long and I added a comment around those queries for feedback.

I'll leave the WIP designation on this for now, but stopping work until I get feedback.

chappjc:
Thanks, @papacarp!  I'll test it too.

Gopkg.lock
```diff
@@ -3,65 +3,52 @@
 
 [[projects]]
   branch = "master"
-  digest = "1:4c5803754aedbf42ec7da8cdfea47d36337397267851828610a4f9a76c1b7fd9"
```
chappjc: We've actually stopped supporting `dep`, so let's undo this change.  In not too long, we'll remove the Gopkg files and update the instructions to explain using vgo with go1.10

db/dcrpg/internal/addrstmts.go
chappjc: Rerunning the update for `is_funding=TRUE` with `EXPLAIN ANALYZE`:

```
                                                                        QUERY PLAN                                                              
          
------------------------------------------------------------------------------------------------------------------------------------------------
----------
 Update on addresses  (cost=5816983.24..9533917.15 rows=165117827 width=217) (actual time=1842016.473..1842016.473 rows=0 loops=1)
   ->  Merge Join  (cost=5816983.24..9533917.15 rows=165117827 width=217) (actual time=26565.186..146692.090 rows=10657405 loops=1)
         Merge Cond: ((vins.prev_tx_hash = addresses.tx_hash) AND (vins.prev_tx_index = addresses.tx_vin_vout_index))
         ->  Index Scan using uix_vin_prevout on vins  (cost=0.56..839737.22 rows=12261598 width=83) (actual time=0.009..91095.270 rows=12261535
 loops=1)
         ->  Materialize  (cost=5697774.36..5788845.58 rows=18214243 width=207) (actual time=24377.136..45082.620 rows=13161750 loops=1)
               ->  Sort  (cost=5697774.36..5743309.97 rows=18214243 width=207) (actual time=24377.134..43378.516 rows=13161719 loops=1)
                     Sort Key: addresses.tx_hash, addresses.tx_vin_vout_index
                     Sort Method: external merge  Disk: 2780144kB
                     ->  Seq Scan on addresses  (cost=0.00..1386348.06 rows=18214243 width=207) (actual time=0.028..15872.825 rows=13161720 loop
s=1)
                           Filter: is_funding
                           Rows Removed by Filter: 10657385
 Planning time: 2.493 ms
 Execution time: 1842124.734 ms
(13 rows)
```

On my machine that's a little more than 30 minutes.

The large disk usage on the sort on (addresses.tx_hash, addresses.tx_vin_vout_index) is damn large (2,780,144 kB).  Honestly I don't know what PG is doing this.  I'm wondering if it might help to do a subquery to just get (id, tx_hash, tx_vin_vout_index) in a temp table seeing as it's doing a materialize anyway.

chappjc: It doesn't do the sort and materialize using all that disk space when actually doing a batch of id ranges as would be done for real, so never mind the last point.

chappjc: Anyway, I only removed what seemed like unnecessary conditionals from your upgrade queries and it came out like this:

```
2018-09-17 16:57:53.220 [INF] PSQL: Set the matching tx index in address table...
2018-09-17 16:57:53.221 [INF] PSQL: Updating address table entries 0 to 100000 of 23819112
2018-09-17 16:58:07.900 [INF] PSQL: Updating address table entries 100000 to 200000 of 23819112
2018-09-17 16:58:27.183 [INF] PSQL: Updating address table entries 200000 to 300000 of 23819112
2018-09-17 16:58:41.982 [INF] PSQL: Updating address table entries 300000 to 400000 of 23819112
2018-09-17 16:58:52.832 [INF] PSQL: Updating address table entries 400000 to 500000 of 23819112
2018-09-17 16:59:07.257 [INF] PSQL: Updating address table entries 500000 to 600000 of 23819112
2018-09-17 16:59:25.228 [INF] PSQL: Updating address table entries 600000 to 700000 of 23819112
2018-09-17 17:31:39.985 [INF] PSQL: Updating address table entries 23800000 to 23900000 of 23819112
2018-09-17 17:31:41.570 [INF] PSQL:  - 21314759 rows in addresses table (matching tx index column upgrade) were successfully upgraded in 2028.35 sec
```

So about 34 minutes.  Not so bad.

chappjc: Just this diff:

```diff
@@ -227,12 +227,12 @@ const (
                tx_vin_vout_row_id=$2 AND is_funding=$3;`
 
        UpdateMatchingFundingTxIdFromVins = `UPDATE addresses SET matching_tx_index=tx_index
-               FROM vins WHERE matching_tx_hash=vins.tx_hash AND addresses.tx_hash=prev_tx_hash
-               AND addresses.tx_vin_vout_index=prev_tx_index AND matching_tx_hash !=''
+               FROM vins WHERE addresses.tx_hash=prev_tx_hash
+               AND addresses.tx_vin_vout_index=prev_tx_index
                AND is_funding=true AND addresses.id >= $1 AND addresses.id < $2;`
 
        UpdateMatchingSpendingTxIdFromVins = `UPDATE addresses SET matching_tx_index=vins.prev_tx_index
-               FROM vins WHERE vins.id=tx_vin_vout_row_id AND is_funding=false AND matching_tx_hash != ''
+               FROM vins WHERE vins.id=tx_vin_vout_row_id AND is_funding=false
                AND addresses.id >= $1 AND addresses.id < $2;`
 
        IndexBlockTimeOnTableAddress   = `CREATE INDEX block_time_index ON addresses (block_time);`
```

chappjc: Actually, for `UpdateMatchingFundingTxIdFromVins`, we'll also want to add:

`AND vins.is_valid=addresses.valid_mainchain`

The most important rows to update are vins.is_valid=true.  For funding addresses table rows where we have to look into vins table via the prev_tx_hash:prev_tx_index, there may be multiple matching vins rows.  Matching the vins row with is_valid=true is most important, but if there is no vins row with is_valid=true (just 1+ with is_valid=false) and addresses.valid_mainchain=false, we at also want to patch the matching_tx_index to something valid even if there are multiple matching vins rows.

Gopkg.lock
papacarp: ok

PR #666, opened 2018-09-05
Transaction block list

chappjc:
This adds a new query to get details for all blocks containing a specified transaction. The details are block index, height, hash, is_valid, and is_mainchain.  `ChainDB.TransactionBlocks` gets these data.

On the transaction page, use the new `TransactionBlocks` function in full mode to retrieve and display a list of blocks (on the transaction view) that contain the transaction.  When a transaction is found in multiple blocks, list them all with unambiguous links by hash.  Also pass the transaction index via `?ind=`
URL parameter so that we can use it on the block page to highlight the transaction (not done in this work). Create "grayed out" colors with a `"grayed"` class for `<body>` and `<a>` elements.  Also define these for dark/night mode. Also simplify the signature for several template functions.

In day mode:

image

Night mode:

image

(previously there was just one block listed).

This resolves issue https://github.com/decred/dcrdata/issues/295.

### DIFF
--- a/db/dcrpg/internal/txstmts.go
+++ b/db/dcrpg/internal/txstmts.go
@@ -96,6 +96,11 @@ const (
 		FROM transactions
 		WHERE block_height BETWEEN $1 AND $2;`
 
+	SelectTxsBlocks = `SELECT block_height, block_hash, block_index, is_valid, is_mainchain
+		FROM transactions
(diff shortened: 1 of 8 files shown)


chappjc:
Tested on mainnet and testnet, for transactions in blocks that were invalidated and orphaned.

Issue #662, opened 2018-09-04
api: unknown addresses should not be an error


chappjc:
When info for an unknown but valid address is requested on /api/address/{addr}
dcrdata response with a 422 error code, Unprocessable Entity.  What should
happen is:

* If the address is valid (decodes for the active network), but it has no entries in the addresses table, a valid JSON result should be returned, just with an empty transactions array:

```json
{
"address": "Ds...................",
"address_transactions": []
}
```

* The backend should not log an error.  Right now it says:

    [ERR] PSQL: Unable to get address Ds......... history: ReduceAddressHistory failed. len(addressRows) = 0

chappjc:
resolved by #668

PR #668, opened 2018-09-06
dcrpg/dcrsqlite: no transactions found for address is not an error

chappjc:
In both dcrpg (for full mode) and dcrsqlite (for lite mode), this
responds to address history requests without error when there are no
known transactions for a given address but no other error.

For the API, this means returning a non-nil *apitypes.Address with the
Transactions field set to a non-nil zero-length slice.  The slice must
be non-nil so that when marshalled to JSON, "address_transactions"
shows as [] instead of null.

So instead of code 422 and "Unprocessible Entity", the API  now responds
with code 200 and a valid JSON struct like:

```json
{
"address": "TsmSGofXA4hqx9XRNgWoRZtNMXXshhMLByA",
"address_transactions": []
}
```

This resolves issue https://github.com/decred/dcrdata/issues/662.

### DIFF
--- a/db/dcrpg/pgblockchain.go
+++ b/db/dcrpg/pgblockchain.go
@@ -776,7 +776,7 @@ func (pgb *ChainDB) AddressHistory(address string, N, offset int64,
 	if err != nil {
 		return nil, nil, err
 	}
-	if fresh {
+	if fresh || len(addressRows) == 0 {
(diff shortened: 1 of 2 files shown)


chappjc:
We have had no objections, and I'm certain 422 is the wrong response.  Merging.

PR #669, opened 2018-09-07
cache the ticket pool data to avoid unnecessary DB queries

chappjc:
This implements a cache for the ticket pool data at the dcrpg package level.  The cache automatically updates on the first request when the data is stale.  Requests made when an update is already running are served with stale data.

So that the APIs and the explorer can easily leverage the cache but not even know about it, the cache should live entirely within `ChainDB` in the `dcrpg` package.  Also, consumers only need use `TicketPoolVisualization`, not the cache query and update functions. `ticketPoolDataCache` lives in pgblockchain.go (dcrpg) now, and it is managed entirely by `ChainDB`.
Modify tickets pool by date API response to include height of the data. Also pass `ChartsHeight` to the /ticketpool page template, for future use.
The `ticketPoolVisualization` function now ensures that the DB height hasn't changed from the beginning of the queries to the end.
Move `ChartDataCounter` out of explorertypes.go, which is intended to help importers of the explore package find useful exported types.

This also disables the ticketpool page when running in lite mode.

### DIFF
--- a/db/dcrpg/pgblockchain.go
+++ b/db/dcrpg/pgblockchain.go
@@ -66,6 +66,12 @@ func (d *DevFundBalance) Balance() *explorer.AddressBalance {
 	return d.AddressBalance
 }
 
+// ticketPoolSync helps to manage the calls made to fetch ticket pool data from
+// the db via ticketPoolVisualization method.
+type ticketPoolSync struct {
+	updating trylock.Mutex
+}
+
 // ChainDB provides an interface for storing and manipulating extracted
 // blockchain data in a PostgreSQL database.
 type ChainDB struct {
@@ -83,6 +89,7 @@ type ChainDB struct {
 	devPrefetch        bool
 	InBatchSync        bool
 	InReorg            bool
+	tpSync             *ticketPoolSync
 }
 
 // ChainDBRPC provides an interface for storing and manipulating extracted and
@@ -257,6 +264,7 @@ func NewChainDB(dbi *DBInfo, params *chaincfg.Params, stakeDB *stakedb.StakeData
 		unspentTicketCache: unspentTicketCache,
 		DevFundBalance:     new(DevFundBalance),
 		devPrefetch:        devPrefetch,
+		tpSync:             new(ticketPoolSync),
 	}, nil
 }
 
@@ -571,12 +579,31 @@ func (pgb *ChainDB) GetTicketPoolByDateAndInterval(maturityBlock int64,
 	return retrieveTicketsByDate(pgb.db, maturityBlock, int64(val))
 }
 
-// TicketPoolVisualization fetches the following ticketpool data: tickets
+// TicketPoolVisualization helps block consecutive access to the ticketPoolVisualization
+// method till the first request is resolved.
+func (pgb *ChainDB) TicketPoolVisualization(interval dbtypes.ChartGrouping) (
+	[]*dbtypes.PoolTicketsData, *dbtypes.PoolTicketsData, error) {
+
+	for {
+		if pgb.tpSync.updating.TryLock() {
+			defer pgb.tpSync.updating.Unlock()
+			// check if after the first request was handled, the consecutive request
+			// response data was also processed and stored in the cache.
+			barcharts, donutCharts, ok := explorer.GetTicketPoolData(interval)
+			if ok {
+				return barcharts, donutCharts, nil
+			}
+			return pgb.ticketPoolVisualization(interval)
+		}
+	}
+}
+
+// ticketPoolVisualization fetches the following ticketpool data: tickets
 // grouped on the specified interval, tickets grouped by price, and ticket
 // counts by ticket type (solo, pool, other split). The interval may be one of:
 // "mo", "wk", "day", or "all". The data is needed to populate the ticketpool
 // graphs. The data grouped by time and price are returned in a slice.
-func (pgb *ChainDB) TicketPoolVisualization(interval dbtypes.ChartGrouping) (
+func (pgb *ChainDB) ticketPoolVisualization(interval dbtypes.ChartGrouping) (
 	[]*dbtypes.PoolTicketsData, *dbtypes.PoolTicketsData, error) {
 	var maturityBlock = pgb.GetTicketPoolBlockMaturity()
 
--- a/db/dcrpg/pgblockchain.go
+++ b/db/dcrpg/pgblockchain.go
@@ -1,3 +1,4 @@
+// Copyright (c) 2018, The Decred developers
 // Copyright (c) 2017, The dcrdata developers
 // See LICENSE for details.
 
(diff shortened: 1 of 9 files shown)


chappjc:
@dmigwi I've made this into exactly two commits with the intent of rebasing them both onto master.

db/dcrpg/pgblockchain.go
```diff
@@ -571,37 +626,91 @@ func (pgb *ChainDB) GetTicketPoolByDateAndInterval(maturityBlock int64,
...
+	// Attempt to retrieve data for the current block from cache.
+	heightSeen := pgb.Height()
+	barcharts, donutCharts, height, ok := TicketPoolData(interval, heightSeen)
+	if ok {
+		// The cache was fresh.
+		return barcharts, donutCharts, height, nil
+	}
+
+	// Cache is stale. Attempt to gain updater status.
+	if !pgb.tpUpdatePermission.TryLock() {
```
dmigwi: Is this a blocking operation? Like will the process wait till the trylock is true?

db/dcrpg/pgblockchain.go
```diff
@@ -571,37 +626,91 @@ func (pgb *ChainDB) GetTicketPoolByDateAndInterval(maturityBlock int64,
...
+		// Another goroutine is running db query to get the updated data. Return
+		// stale data instead of waiting.
+		return barcharts, donutCharts, height, nil
```
dmigwi: I still don't get how this will work because if `ok` is false it means that there is some data missing and its mostly going to barsCharts going by this `barGraphs, isFound = ticketPoolGraphsCache.BarGraphsCache[interval]`....

The so called stale data that you will be serving, most likely you will be serving empty data set except the height.

db/dcrpg/pgblockchain.go
```diff
@@ -571,37 +626,91 @@ func (pgb *ChainDB) GetTicketPoolByDateAndInterval(maturityBlock int64,
...
-func (pgb *ChainDB) TicketPoolVisualization(interval dbtypes.ChartGrouping) (
-	[]*dbtypes.PoolTicketsData, *dbtypes.PoolTicketsData, error) {
-	var maturityBlock = pgb.GetTicketPoolBlockMaturity()
-
-	// Tickets grouped by time interval
-	ticketsByTime, err := pgb.GetTicketPoolByDateAndInterval(maturityBlock, interval)
+func (pgb *ChainDB) ticketPoolVisualization(interval dbtypes.ChartGrouping) (byTimeAndPrice []*dbtypes.PoolTicketsData, byInputs *dbtypes.PoolTicketsData, height uint64, err error) {
+	// Ensure DB height is the same before and after queries since they are not
+	// atomic. Initial height:
+	height, err = pgb.HeightDB()
```
dmigwi: you've got `height` variable checked in the above function, you could pass it as a variable instead to this method.

db/dcrpg/pgblockchain.go
```diff
@@ -571,37 +626,91 @@ func (pgb *ChainDB) GetTicketPoolByDateAndInterval(maturityBlock int64,
...
+		err = fmt.Errorf("failed to query DB height: %v", err)
+		return
 	}
 
-	// Tickets grouped by price
-	ticketsByPrice, err := retrieveTicketByPrice(pgb.db, maturityBlock)
-	if err != nil {
-		return nil, nil, err
-	}
+	for {
```
dmigwi: I don't get why we need a for loop here, the queries will obviously return the latest data as in the db. This may also go against our earlier priority of serving first and fast just incase block heights keeps changing in split seconds.

explorer/explorerroutes.go
```diff
@@ -272,16 +279,18 @@ func (exp *explorerUI) Ticketpool(w http.ResponseWriter, r *http.Request) {
...
 
 	str, err := exp.templates.execTemplateToString("ticketpool", struct {
-		Version     string
-		NetName     string
-		ChartData   []*dbtypes.PoolTicketsData
-		GroupedData *dbtypes.PoolTicketsData
-		Mempool     *dbtypes.PoolTicketsData
+		Version      string
+		NetName      string
+		ChartsHeight uint64
```
dmigwi: This will be used to indicate the latest block fetched on the frontend?

db/dcrpg/pgblockchain.go
dmigwi: And also this too..
```go
if ticketPoolGraphsCache.Height != height {
		return nil, nil, 0, false
	}
```

db/dcrpg/pgblockchain.go
chappjc: It does not block.  It just says whether or not it got the lock.

db/dcrpg/pgblockchain.go
chappjc: I've removed the clear function call

db/dcrpg/pgblockchain.go
chappjc: Good point

db/dcrpg/pgblockchain.go
chappjc: There are multiple queries.  They all need to be done when the DB is at the same height

chappjc: It's not likely to loop, but this guarantees consistency

explorer/explorerroutes.go
chappjc: The block height corresponding to the charts data.  Unused, but can be

db/dcrpg/pgblockchain.go
dmigwi: But if `ok` is false then either `barcharts` or `donutCharts` or both are empty. And after `ok` is false you are not checking if the data you send is empty.

db/dcrpg/pgblockchain.go
chappjc: We actually should be checking block hash

db/dcrpg/pgblockchain.go
dmigwi: This is when the trylock is false.

chappjc: Oh the return nil, nil, ... was a mistake.

This was my error. Should just return the stale data and false. Not nils

db/dcrpg/pgblockchain.go
dmigwi: Even though block-height and block-hash can be used interchangeably, block-height might be easier to fetch the block-hash.

db/dcrpg/pgblockchain.go
dmigwi: ok. Noted

db/dcrpg/pgblockchain.go
dmigwi: Just to be on the safe side you also confirm that you will not be sending empty data set when the trylock returns false.

db/dcrpg/pgblockchain.go
```diff
@@ -66,6 +67,59 @@ func (d *DevFundBalance) Balance() *explorer.AddressBalance {
...
+	donutChart *dbtypes.PoolTicketsData, actualHeight uint64, isFound bool) {
+	ticketPoolGraphsCache.RLock()
+	defer ticketPoolGraphsCache.RUnlock()
+
+	if ticketPoolGraphsCache.Height != height {
+		return nil, nil, 0, false
+	}
+
+	barGraphs, isFound = ticketPoolGraphsCache.BarGraphsCache[interval]
+	donutChart = ticketPoolGraphsCache.DonutGraphCache[interval]
```
dmigwi: since you are using `isFound` to check if the data is stale then you could have:
`isFound = ticketPoolGraphsCache.Height == height`
rather than what is there.

dmigwi: or maybe change the variable name to `isStale`..

db/dcrpg/pgblockchain.go
chappjc: The thing is that height cannot be reliably used to indicate if the data is current.  The block height could be the same as the hash changes in a reorg.

chappjc:
The added commit treats the caching and updating of data from the intervals separately.  Still using height though.

db/dcrpg/pgblockchain.go
```diff
@@ -244,6 +297,12 @@ func NewChainDB(dbi *DBInfo, params *chaincfg.Params, stakeDB *stakedb.StakeData
 		unspentTicketCache.SetN(unspentTicketHashes, unspentTicketDbIDs)
 	}
 
+	// For each chart grouping typle create a non-blocking updater mutex.
```
dmigwi: `// For each chart grouping type, create a non-blocking updater mutex.`

db/dcrpg/pgblockchain.go
```diff
@@ -571,37 +631,117 @@ func (pgb *ChainDB) GetTicketPoolByDateAndInterval(maturityBlock int64,
...
-func (pgb *ChainDB) TicketPoolVisualization(interval dbtypes.ChartGrouping) (
-	[]*dbtypes.PoolTicketsData, *dbtypes.PoolTicketsData, error) {
-	var maturityBlock = pgb.GetTicketPoolBlockMaturity()
-
-	// Tickets grouped by time interval
-	ticketsByTime, err := pgb.GetTicketPoolByDateAndInterval(maturityBlock, interval)
+func (pgb *ChainDB) ticketPoolVisualization(interval dbtypes.ChartGrouping) (byTimeAndPrice []*dbtypes.PoolTicketsData, byInputs *dbtypes.PoolTicketsData, height uint64, err error) {
+	// Ensure DB height is the same before and after queries since they are not
+	// atomic. Initial height:
+	height, err = pgb.HeightDB()
```
dmigwi: You may need do away with:
```go
height, err = pgb.HeightDB()
if err != nil {
	err = fmt.Errorf("failed to query DB height: %v", err)
	return
}
```
Before the for loop, you are already sending the `height` as a variable from the function above.

db/dcrpg/pgblockchain.go
```diff
@@ -66,6 +67,57 @@ func (d *DevFundBalance) Balance() *explorer.AddressBalance {
...
+
+// UpdateTicketPoolData updates the ticket pool cache with the latest data fetched.
+// This is a thread-safe way to update ticket pool cache data. TryLock helps avoid
+// stacking calls to update the cache.
+func UpdateTicketPoolData(interval dbtypes.ChartGrouping, barGraphs []*dbtypes.PoolTicketsData,
+	donutcharts *dbtypes.PoolTicketsData, height uint64) {
+	ticketPoolGraphsCache.Lock()
+	defer ticketPoolGraphsCache.Unlock()
+
+	ticketPoolGraphsCache.Height[interval] = height
```
dmigwi: I don't know if you've got a mechanism to ensure that the `height` map never grows to uncontrollable sizes because it will always be a unique entry that you make here every time.

db/dcrpg/pgblockchain.go
chappjc: `HeightDB()` here is getting the best block processed and entered entirely into all of the DB tables.  This is the block height that you can expect to get complete charts for.

In `TicketPoolVisualization`, `heightSeen := pgb.Height()` gives the best block ever seen, but not necessarily fully processed by ChainDB.  `heightSeen` is set that way there because you should be able to get that data, although it may involve waiting (or getting stale data).

db/dcrpg/pgblockchain.go
chappjc: Since the key is `interval` and there are only a handful of intervals, it is not going to grow unbounded.

db/dcrpg/pgblockchain.go
dmigwi: ok. Noted

chappjc: Now that I look closer, HeightDB and Height should both indicate the block is processed in the tables of interest.  I will change to Height()

chappjc:
Thanks for the review.  I'm going to squash a commit in and force push now.

PR #680, opened 2018-09-12
Show address page for address with ONLY unconfirmed Tx

papacarp:
This PR will allow the address page to be visible for an address with only unconfirmed transactions.  It will also insure the charts do not stall on loading and display a message that there are no CONFIRMED transactions to display (and thus no charts).

I could not find any open issues related to this.

image

image


chappjc:
tACK.  Good catch.  Looks like the bug was introduced with the addition of `GetOldestTxBlockTime`.

PR #681, opened 2018-09-12
response for tx send:  rawtx -> txid

papacarp:
Typo fix for rawtx broadcast response from Insight API


chappjc:
Resolves issue https://github.com/decred/dcrdata/issues/684

PR #683, opened 2018-09-12
dcrpg: combine spent and unspent totals queries into one

chappjc:
Add `SelectAddressSpentUnspentCountAndValue` query that gets the count and total value for spent and unspent address table rows for a given address.  It uses `GROUP BY (is_funding, matching_tx_hash='')` with aggregate functions to do it efficiently in a single scan on the addresses table.
Scanning the result, `isFunding && noMatchingTx` means unspent, `!isFunding && !noMatchingTx` means spent, while `!isFunding && noMatchingTx` is not expected so an error is logged.

Timings on one machine (project fund):

| query  | timing   |
|---|---|
| new unified query  | 230 ms  |
| old unspent query  | 380 ms  |
| old spent query | 227 ms |
| old spent+unspent queries | 607 ms |

That's a 62% reduction in time (down to 38% of the original query cost) to get both spent and unspent totals for the project fund.

Also clean-up some query formatting.
Fix the conditional on the (unused) `InsertAddressRowReturnID` query.

### DIFF
--- a/db/dcrpg/queries.go
+++ b/db/dcrpg/queries.go
@@ -724,30 +724,47 @@ func RetrieveAddressSpentUnspent(db *sql.DB, address string) (numSpent, numUnspe
 		return
 	}
 
-	var nu, tu sql.NullInt64
-	err = dbtx.QueryRow(internal.SelectAddressUnspentCountANDValue, address).
-		Scan(&nu, &tu)
+	// Query for spent and unspent totals
+	var rows *sql.Rows
+	rows, err = db.Query(internal.SelectAddressSpentUnspentCountAndValue, address)
 	if err != nil && err != sql.ErrNoRows {
 		if errRoll := dbtx.Rollback(); errRoll != nil {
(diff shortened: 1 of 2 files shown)


db/dcrpg/queries.go
```diff
@@ -724,30 +724,47 @@ func RetrieveAddressSpentUnspent(db *sql.DB, address string) (numSpent, numUnspe
...
-		Scan(&ns, &ts)
-	if err != nil && err != sql.ErrNoRows {
-		if errRoll := dbtx.Rollback(); errRoll != nil {
-			log.Errorf("Rollback failed: %v", errRoll)
+	for rows.Next() {
+		var count, totalValue int64
+		var noMatchingTx, isFunding bool
+		err = rows.Scan(&count, &totalValue, &isFunding, &noMatchingTx)
+		if err != nil {
+			break
```
dmigwi: This will exit the for loop without giving any notification to as why the loop exited prematurely.

Issue #682, opened 2018-09-12
Insight API: Websocket testing- tx room data format


papacarp:
On the websocket API, when subscribed to the `inv` room, I get `tx` messages with a value like:
```
{
  "txid": "3315a6ef44cd24183b5d98cf70ed081d533e28fcb2edcfbfbe0e2c0501c29880",
  "size": 251,
  "valueOut": 14741789,
  "vout": {
    "DsZQaCQES5vh3JmcyyFokJYz3aSw8Sm1dsQ": 13741789,
    "DsidNx5RnC7jMHdvUSEV54WANwMyKbyGtjh": 1000000
  }
}
```

The format for the `vout` is incorrect.
Here's an example from an insight server for DASH:
```
{
  "txid": "1dfc29776c4a94138b5b1aa00eb8351e05e26f9649b33fa6b20b20b603939e5c",
  "valueOut": 0.003,
  "vout": [
    {
      "XvYMppkbKgoFsud7kDYA6DkawzAjvDBzPQ": 300000
    }
  ],
  "isRBF": false,
  "txlock": false
}
```
Note that in this correct example, `vout` is an array of objects, not an object itself.
Here's another correct DASH example for a tx with two outputs, for clarity on the correct response:
```
{
  "txid": "c1d7627bfa9b1320b30f8b19ea99cf81c44a41b00b6dc3818169dbc5a078a658",
  "valueOut": 0.12439791,
  "vout": [
    {
      "XfHLznCyUixzXf5nRu5yxV6VtTh8h1gZKj": 1109710
    },
    {
      "Xe54SKBdrjz9SMGH9cNXDX5PWgRBfRcxkZ": 11330081
    }
  ],
  "isRBF": false,
  "txlock": false
}
```

chappjc:
@papacarp Unless you're already on this, I'm gonna fix it up.

papacarp:
Not working it yet. go for it.

PR #686, opened 2018-09-12
insight: fix marshal of vout values in the tx message

chappjc:
Define a new type for a vout array element, and implement `json.Marshaler` using a temporay map with one key-value pair.
Clean up socket.io.go, embedding the address mutex into a new type, `roomSubscriptionCounter`.

Resolves issue https://github.com/decred/dcrdata/issues/682

### DIFF
--- a/api/insight/socket.io.go
+++ b/api/insight/socket.io.go
@@ -4,6 +4,7 @@
 package insight
 
 import (
+	"encoding/json"
 	"regexp"
 	"sync"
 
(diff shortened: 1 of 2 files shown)


chappjc:
@papacarp Feel free to question anything!  To me, it is easy to end up with high-level structs like `SocketServer`, `ChainDB`, etc. having named mutexes, often more than one, where it's not self-explanatory to which fields of the struct the mutex is intended to guard.

The embedding of a Mutex/RWMutex helps sweeten the syntax by ridding the need to access a named field of a struct.  It also drives home that it applies to the struct as a whole, whereas it's not at all certain that `addressesMtx` guarded `watchedAddresses`.  With the `roomSubscriptionCounter`-typed field, it's clear that locking it guards its members (`c map[string]int`) and nothing else.

chappjc:
A random `tx` event that came by looks good:

```json
{
    "txid":"66a37fb196a72f5b5e93f05075fad69c862339a4cc25db2536c7a098ea98d717",
    "size":538,
    "valueOut":9788375675,
    "vout":[
        {
            "Dcb5tcwGzU4U1Pj9bAtcnoAeuq1EHxayuv3":9788375675
        },
        {
            "DsifZiZqpBDQ1GyVoPqomY72PhK69jtS3YW":0
        },
        {
            "DsQxuVRvS4eaJ42dhQEsCXauMWjvopWgrVg":0
        },
        {
            "DsbZUsmVngjGzkvCGsMyposduQh8ZReMFqm":0
        },
        {
            "DsQxuVRvS4eaJ42dhQEsCXauMWjvopWgrVg":0
        }
    ]
}
```

Issue #691, opened 2018-09-17
Remove dep files and update README.md


chappjc:
Remove the Gopkg files and update the instructions to explain using vgo with go1.10

PR #693, opened 2018-09-19
Remove unsupported dep files.

chappjc:
This removes the dep files since dep is no longer supported.

This also updates README.md accordingly, and cleans up the docker sections.

This resolves https://github.com/decred/dcrdata/issues/691.


dmigwi:
Maybe we could also add that setting `export GO111MODULE=on` on the bash shell script might come in handy.

dmigwi:
https://github.com/golang/go/wiki/Modules#how-to-define-a-module

PR #695, opened 2018-09-19
Adds Insight API endpoints to README

### LINKED ISSUE #694: Insight API Readme Documentation
Now that insight is in master we should update the [README.md ](https://github.com/decred/dcrdata/blob/master/README.md#insight-api-experimental) with at least a list of the endpoints implemented.    The implemented endpoints are listed here:  https://github.com/decred/dcrdata/issues/400

s-ben:
Have listed endpoints for Insight API, as documented [in Issue #400](https://github.com/decred/dcrdata/issues/400) ( Closes #694 ).

Changes:
- Adds table with Insight API endpoints
- Puts 'Insight API' and 'dcrd API' as subheadings under new heading 'APIs'
- Tweaks text introducing APIs

I could have worked on this further, but wanted to put what I have out for review first. Also, here are  a few Qs I had as an imaginary user:

- Do we want to provide guidance on which API to use? Is one being depricated eventually (this is often implied when a new API is added)? Is one API better than another for a particular use case?
- Is the Insight API still experimental? Remove (EXPERIMENTAL) qualifier?
- Do we want to document the datatypes, as is done with dcrdata in the 'Type' column? 
- We say “....the following endpoints are not implemented ‘status, sync, peer, email, and rates.’. Is this still true?

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,21 +53,21 @@ The dcrdata repository is a collection of Go packages and apps for [Decred](http
 ```none
 ../dcrdata              The dcrdata daemon.
 ├── api                 Package blockdata implements dcrdata's own HTTP API.
-│   ├── insight         Package insight implements the Insight API.
-│   └── types           Package types includes the exported structures used by
+│   ├── insight         Package insight implements the Insight API.
+│   └── types           Package types includes the exported structures used by
 |                         the dcrdata and Insight APIs.
 ├── blockdata           Package blockdata is the primary data collection and
 |                         storage hub, and chain monitor.
 ├── cmd
-│   ├── rebuilddb       rebuilddb utility, for SQLite backend. Not required.
-│   ├── rebuilddb2      rebuilddb2 utility, for PostgreSQL backend. Not required.
-│   └── scanblocks      scanblocks utility. Not required.
+│   ├── rebuilddb       rebuilddb utility, for SQLite backend. Not required.
+│   ├── rebuilddb2      rebuilddb2 utility, for PostgreSQL backend. Not required.
+│   └── scanblocks      scanblocks utility. Not required.
 ├── dcrdataapi          Package dcrdataapi for Go API clients.
 ├── db
-│   ├── agendadb        Package agendadb is a basic PoS voting agenda database.
-│   ├── dbtypes         Package dbtypes with common data types.
-│   ├── dcrpg           Package dcrpg providing PostgreSQL backend.
-│   └── dcrsqlite       Package dcrsqlite providing SQLite backend.
+│   ├── agendadb        Package agendadb is a basic PoS voting agenda database.
+│   ├── dbtypes         Package dbtypes with common data types.
+│   ├── dcrpg           Package dcrpg providing PostgreSQL backend.
+│   └── dcrsqlite       Package dcrsqlite providing SQLite backend.
 ├── dev                 Shell scripts for maintenance and deployment.
 ├── explorer            Package explorer, powering the block explorer.
 ├── mempool             Package mempool for monitoring mempool for transactions,
@@ -563,21 +563,48 @@ default in a reduced functionality ("lite") mode that does not require
 PostgreSQL. To enable the PostgreSQL backend (and the expanded functionality),
 dcrdata may be started with the `--pg` switch.
 
+## APIs
+
+The dcrdata block explorer is exposed by two APIs: a Decred implementation of the [Insight API](https://github.com/bitpay/insight-api) (EXPERIMENTAL), and its own JSON HTTP API. The Insight API uses the path prefix `/insight/api`. The dcrdata API uses the path prefix `/api`.
+
 ### Insight API (EXPERIMENTAL)
 
-The [Insight API](https://github.com/bitpay/insight-api) is accessible via HTTP
-at the `/insight/api` URL path prefix, and via WebSocket at the
-`/insight/socket.io` URL path prefix.
+The [Insight API](https://github.com/bitpay/insight-api) is accessible via HTTP via REST or WebSocket. 
+
+To call the REST API, use the `/insight/api` path prefix. To call the Websocket API, use the `/insight/socket.io` path prefix.
 
 The following endpoints are not implemented: `status`, `sync`, `peer`, `email`,
 and `rates`.
 
+#### Endpoint List
+
+| Method           | Path                   | 
+| -------------------- | ---------------------- | 
+| Block              | `/block/{hash}`          |  
+| Block index          | `/block-index/{height}`      |  
+| Raw block (hash)               | `/rawblock/{hash}`   | 
+| Raw block (height)                | `/rawblock/{height}`     |  
+| Raw summaries               | `/blocks (URL query params: limit, blockDate, ...) `   |              
+| Transaction                 | `/tx/{hash}`     |  
+| Raw transaction              | `/rawtx/{hash}`  | 
+| Address         | `/addr/{address}`       | 
+| Address Props   | `/balance, totalReceived, totalSent, /totalSent, unconfirmedBalance` |  
+| UTXOs | `/addr/{address}/utxo`  |  
+| Multi-address UTXOs | `/addrs/{address0, address1, etc}/utxo`  |  
+| POST-variant of multi-address UTXOs | ``  |                     
+| Transactions by block | `/txs/?block={hash}`  |   
+| Transactions by address | `/txs/?address={address}`  |   
+| Transactions for multiple addresses | `/addrs/{addr0, addr1, ...}/txs (params: from, to)`  |   
+| POST-variant for ^^ | `(includes params: noAsm, noScriptSig, noSpent)`  |  
+| Transaction broadcast | `/tx/send (raw tx in POST params) `  | 
+| Sync status (dcrdata) | `/sync`  |  
+| Sync status (dcrd) | `/peer`  |  
+| Decred network status | `/status(with each possible value for the q param)`  |  
+| Estimate fee | `/estimatefee (params: nbBlocks)`  |  
 ### dcrdata API
 
-dcrdata has its own JSON HTTP API in addition to the experimental Insight API
-implementation. **dcrdata's API endpoints are prefixed with `/api`** (e.g.
-`http://localhost:7777/api/stake`). The Insight API endpoints (not described in
-this section) are prefixed with `/insight/api`.
+The dcrdata API is a REST API accessible via HTTP. To call the dcrdata API, use the `/api` path prefix.
+
 
 #### Endpoint List
 
--- a/README.md
+++ b/README.md
@@ -573,8 +573,6 @@ The [Insight API](https://github.com/bitpay/insight-api) is accessible via HTTP 
 
 To call the REST API, use the `/insight/api` path prefix. To call the Websocket API, use the `/insight/socket.io` path prefix.
 
-The following endpoints are not implemented: `status`, `sync`, `peer`, `email`,
-and `rates`.
 
 #### Endpoint List
 


README.md
```diff
@@ -563,21 +563,48 @@ default in a reduced functionality ("lite") mode that does not require
...
 
+#### Endpoint List
+
+| Method           | Path                   | 
+| -------------------- | ---------------------- | 
+| Block              | `/block/{hash}`          |  
+| Block index          | `/block-index/{height}`      |  
+| Raw block (hash)               | `/rawblock/{hash}`   | 
+| Raw block (height)                | `/rawblock/{height}`     |  
+| Raw summaries               | `/blocks (URL query params: limit, blockDate, ...) `   |
```
papacarp: I think we should call this one `Blocks` as far as the params go, perhaps we have a third column so the params are clear?

| Method | Path | Optional Parameters |
| --- | --- | --- |
| Blocks | /blocks | ?limit=N&blockDate=YYYY-MM-DD |

README.md
```diff
@@ -563,21 +563,48 @@ default in a reduced functionality ("lite") mode that does not require
...
+| Method           | Path                   | 
+| -------------------- | ---------------------- | 
+| Block              | `/block/{hash}`          |  
+| Block index          | `/block-index/{height}`      |  
+| Raw block (hash)               | `/rawblock/{hash}`   | 
+| Raw block (height)                | `/rawblock/{height}`     |  
+| Raw summaries               | `/blocks (URL query params: limit, blockDate, ...) `   |              
+| Transaction                 | `/tx/{hash}`     |  
+| Raw transaction              | `/rawtx/{hash}`  | 
+| Address         | `/addr/{address}`       |
```
papacarp: same here.  the url params would be:
?noTxlist=1&from=N&to=M

I recommend addr/{address}/balance AND /totalReceived AND /totalSent  AND /unconfirmedBalance are all separate lines.  I know its a bit verbose, but useful clarity - especially if we ever want to add in the return details or samples.

README.md
```diff
@@ -563,21 +563,48 @@ default in a reduced functionality ("lite") mode that does not require
...
+| Raw block (hash)               | `/rawblock/{hash}`   | 
+| Raw block (height)                | `/rawblock/{height}`     |  
+| Raw summaries               | `/blocks (URL query params: limit, blockDate, ...) `   |              
+| Transaction                 | `/tx/{hash}`     |  
+| Raw transaction              | `/rawtx/{hash}`  | 
+| Address         | `/addr/{address}`       | 
+| Address Props   | `/balance, totalReceived, totalSent, /totalSent, unconfirmedBalance` |  
+| UTXOs | `/addr/{address}/utxo`  |  
+| Multi-address UTXOs | `/addrs/{address0, address1, etc}/utxo`  |  
+| POST-variant of multi-address UTXOs | ``  |
```
papacarp: For the optional parameters in the case of POST endpoints we should display them slightly differently since they need to be posted in the JSON body.  in this case:

**Required**:  `{"addrs":"[ADDR1],[ADDR2]"}`

README.md
```diff
@@ -563,21 +563,48 @@ default in a reduced functionality ("lite") mode that does not require
...
+| Raw transaction              | `/rawtx/{hash}`  | 
+| Address         | `/addr/{address}`       | 
+| Address Props   | `/balance, totalReceived, totalSent, /totalSent, unconfirmedBalance` |  
+| UTXOs | `/addr/{address}/utxo`  |  
+| Multi-address UTXOs | `/addrs/{address0, address1, etc}/utxo`  |  
+| POST-variant of multi-address UTXOs | ``  |                     
+| Transactions by block | `/txs/?block={hash}`  |   
+| Transactions by address | `/txs/?address={address}`  |   
+| Transactions for multiple addresses | `/addrs/{addr0, addr1, ...}/txs (params: from, to)`  |   
+| POST-variant for ^^ | `(includes params: noAsm, noScriptSig, noSpent)`  |
```
papacarp: here is an example of the parameters:
`{"addrs":"Dca7Vsv42RAJC6cEdw7dyhLER8QJCHiUYNL,DsTHdW8erZsD1Fqt8uCfapVR1Qsf9KaMwBN","from":3, "to": "5"}`

`{"addrs":"Dca7Vsv42RAJC6cEdw7dyhLER8QJCHiUYNL,DsTHdW8erZsD1Fqt8uCfapVR1Qsf9KaMwBN", "from":3,"to":"5", "noScriptSig":1}`

Only the addrs part is required.  `from` , `to` , `noScriptSig` and `noSpent` are all optional.  Not sure how best to cram that in.

README.md
```diff
@@ -563,21 +563,48 @@ default in a reduced functionality ("lite") mode that does not require
...
+| Address         | `/addr/{address}`       | 
+| Address Props   | `/balance, totalReceived, totalSent, /totalSent, unconfirmedBalance` |  
+| UTXOs | `/addr/{address}/utxo`  |  
+| Multi-address UTXOs | `/addrs/{address0, address1, etc}/utxo`  |  
+| POST-variant of multi-address UTXOs | ``  |                     
+| Transactions by block | `/txs/?block={hash}`  |   
+| Transactions by address | `/txs/?address={address}`  |   
+| Transactions for multiple addresses | `/addrs/{addr0, addr1, ...}/txs (params: from, to)`  |   
+| POST-variant for ^^ | `(includes params: noAsm, noScriptSig, noSpent)`  |  
+| Transaction broadcast | `/tx/send (raw tx in POST params) `  |
```
papacarp: example:

`{"rawtx":"010000000184de71690b97b4cbac6e723570a25a56295b30aa26345bfc40c0609c87e23f1d0100000000ffffffff02809698000000000000001976a9143eb656115197956125365348c542e37b6d3d259988ac00811b2c0000000000001976a914e4c9ada86ba67b2c082fac1aa09fb17fccc6833688ac000000000000000001ffffffffffffffff00000000ffffffff6b483045022100b5020004e60a4d26c99a00316ed4f51131b99838fd8fab936d55f9475719bb3b022063b86596adba004367a138f62c73d59f8f3b2cded4b10b5fc0ebe92a4cf41c840121029ec6a82a9646c090decf20806029f332cc11a5c6fa17c943ee72c31b9707a433"}`

README.md
```diff
@@ -563,21 +563,48 @@ default in a reduced functionality ("lite") mode that does not require
...
+| Multi-address UTXOs | `/addrs/{address0, address1, etc}/utxo`  |  
+| POST-variant of multi-address UTXOs | ``  |                     
+| Transactions by block | `/txs/?block={hash}`  |   
+| Transactions by address | `/txs/?address={address}`  |   
+| Transactions for multiple addresses | `/addrs/{addr0, addr1, ...}/txs (params: from, to)`  |   
+| POST-variant for ^^ | `(includes params: noAsm, noScriptSig, noSpent)`  |  
+| Transaction broadcast | `/tx/send (raw tx in POST params) `  | 
+| Sync status (dcrdata) | `/sync`  |  
+| Sync status (dcrd) | `/peer`  |  
+| Decred network status | `/status(with each possible value for the q param)`  |
```
papacarp: ?q=getInfo
?q=getDifficulty
?q=getBestBlockHash
?q=getLastBlockHash

README.md
```diff
@@ -563,21 +563,48 @@ default in a reduced functionality ("lite") mode that does not require
...
+| POST-variant of multi-address UTXOs | ``  |                     
+| Transactions by block | `/txs/?block={hash}`  |   
+| Transactions by address | `/txs/?address={address}`  |   
+| Transactions for multiple addresses | `/addrs/{addr0, addr1, ...}/txs (params: from, to)`  |   
+| POST-variant for ^^ | `(includes params: noAsm, noScriptSig, noSpent)`  |  
+| Transaction broadcast | `/tx/send (raw tx in POST params) `  | 
+| Sync status (dcrdata) | `/sync`  |  
+| Sync status (dcrd) | `/peer`  |  
+| Decred network status | `/status(with each possible value for the q param)`  |  
+| Estimate fee | `/estimatefee (params: nbBlocks)`  |
```
papacarp: ?nbBlocks=N

README.md
```diff
@@ -563,21 +563,48 @@ default in a reduced functionality ("lite") mode that does not require
...
 
-The [Insight API](https://github.com/bitpay/insight-api) is accessible via HTTP
-at the `/insight/api` URL path prefix, and via WebSocket at the
-`/insight/socket.io` URL path prefix.
+The [Insight API](https://github.com/bitpay/insight-api) is accessible via HTTP via REST or WebSocket. 
+
+To call the REST API, use the `/insight/api` path prefix. To call the Websocket API, use the `/insight/socket.io` path prefix.
 
 The following endpoints are not implemented: `status`, `sync`, `peer`, `email`,
 and `rates`.
```
papacarp: rates and email are not supported in the current insight API.  seems like we could just drop this from the comments.  Status, Sync and Peer are currently supported.

papacarp:
Thank you for the PR.  Great start!  

1. Both API's have specific use cases, so both will be supported going forward.
2. I think EXPERIMENTAL is good to leave in there for now.
3. See my recommendations inline with the readme regarding the last two questions.

s-ben:
@papacarp , thx for the feedback. It seems that we could do this in a big table, similar to how you proposed:

Method | Name | Path | Parameters
-- | -- | -- | --
GET | Blocks | /blocks | Optional: ?limit=N&blockDate=YYYY-MM-DD

I think this is going to get a little unwieldy however, once we start adding in necessary info like parameter descriptions and examples, GET and POST variants, etc. I would suggest going with a more standard approach to documenting REST APIs, as illustrated below with the Eventbrite API documentation. 
image

This is similar in structure to the existing [Insight API docs](https://github.com/bitpay/insight-api). Though they present a very "bare bones" approach, where not all methods have all the same info (some just have the endpoint, while some have parameters, examples, descriptions, etc.). This structure would provide space for all the method-specific info you give in other comments. 

I would agree to your suggestion of organizing around "resources". In your comment, you suggest `Blocks` for the `/block-index/{height}`, `/rawblock/{hash}`,`/rawblock/{height}`, and `/blocks` endpoints. This could work. Though the more standard (and therefore perhaps easier to use) approach is to organize around "base endpoints". E.g. the `/rawblock/{hash}` and `/rawblock/{height}` would go under `Raw Block`. The other block-related endpoints would have their own resource heading, as show in Eventbrite API example. If we did this, the endpoints would be organized into the following entries.

```Block (/block/)
Block index (/block-index/)
Raw block (/rawblock/)
Raw summaries (/blocks/)
Transaction (/tx/)
Raw transaction (/rawtx/)
Address (/addr/)
Address Props (/balance/)
Multi-address (/addrs/)
Transactions (/txs/)
Transaction broadcast (/tx/)
Sync (/sync/)
Decred network status (/status/)
Estimate fee (/estimatefee/)```

A couple concerns I have re: this suggested approach:

1) It may be labor intensive to pull all the info for each endpoint out of the codebase/developers' minds. A solution to that could just be doing what the Insight API does, and simply omit documentation elements if we don't have them. E.g. Maybe one endpoint just has the URL and short description, but not a parameter table showing each parameter, a description of said parameter, optionality of parameter, and data type of parameter. I think we should decide what the minimum info a developer needs to use this. For the dcrd API, that appears to be just path and datatype, with endpoints grouped by "sub domains" (wrong term, but I think you get it). 
2) Even a table with more information is going to "bloat" this README further than it already is (and I think the README could already use a major overhaul as it is, but will save that for another Issue). So I think we should put this in a separate file somewhere. Perhaps under a /docs folder in the repo, similar to what dcrd has done,
https://github.com/decred/dcrd/tree/master/docs

Thoughts?

papacarp:
I understand what you are saying.  I recommend whatever fits without making it too unwieldy.  Even just a list of endpoints is a great start.  I like the idea of a separate document eventually but seems like a bit of overkill at this point.  @chappjc, thoughts?

chappjc:
I agree making whatever improvements that do not bloat the document with every detail and example. I like your proposed API documentation approach, so if you wanted to work on that as a separate md that would be great.

By the way, I regret letting the docker stuff into the readme, and that is a candidate for a separate .md file as well.  I just merged a PR to condense it and reduce a lot of slop, but it's still a large part of the document that really is not the kind of basic knowledge that is fit for a README.  Further edits to the README that improve its readability and tighten it up are welcome.

s-ben:
Sounds good. Unless anyone has further suggestions, will start building out the API docs in a separate .md file, with the goal of avoiding bloat. I plan to just start fleshing my approach out with the existing info I have, and adjust the design as necessary. Even if we just end up with something like the original Insight API, but with a bit more info (not hard to do), I think that would be an improvement. 

I think tightening up the README and keeping it to "basic knowledge" you typically find in a README is a good idea. I think breaking out more detailed instructions into separate docs (API, Docker, etc.) and linking to them from the README is a good, best practice approach. 

Putting an editing for flow pass on the README on my TODO list. I made notes while installing the dev environment, and have a couple ideas that could make installation easier for new devs.

chappjc:
@s-ben To clarify, what's your feeling on the current PR?  I think it is sufficiently terse so that it's not cluttering the main README much.  So even if you're going to do a separate md file for Insight, I think this is good to merge, assuming you're all done with these edits.

s-ben:
@chappjc , I've made one minor change per @papacarp comment -- removing line about unsupported methods (which needed updating anyway). I think this is good to merge. If this basic info is needed now, might as well publish what we have. People can always refer back to the Insight API docs if they need to. 

I have started a more fleshed out version of the API docs in a separate file, which will likely need to go in a separate file to avoid bloating the README. Below is the public gist I'm working in if you're curious. Note: this is a fairly rough draft.
https://gist.github.com/s-ben/310aa3085f7bd4d6ad0b78e0346ff733

I've documented about half of the API methods so far, and was shooting to have a rough 1st draft done by EOD. It's looking good to me, but open to feedback if anyone has opinions.

PR #696, opened 2018-09-19
Address page tx highlighting - No db upgrade

papacarp:
This PR will implement the same functionality as #665  but without a db upgrade.  It uses a slightly different strategy to leverage the address fill function.

### DIFF
--- a/db/dcrpg/pgblockchain.go
+++ b/db/dcrpg/pgblockchain.go
@@ -981,7 +981,7 @@ func (pgb *ChainDB) FillAddressTransactions(addrInfo *explorer.AddressInfo) erro
 
 	var numUnconfirmed int64
 
-	for _, txn := range addrInfo.Transactions {
+	for i, txn := range addrInfo.Transactions {
 		_, dbTx, err := RetrieveDbTxByHash(pgb.db, txn.TxID)
 		if err != nil {
 			return err
@@ -997,6 +997,36 @@ func (pgb *ChainDB) FillAddressTransactions(addrInfo *explorer.AddressInfo) erro
 			txn.Confirmations = 0
 		}
 		txn.FormattedTime = time.Unix(dbTx.BlockTime, 0).Format("2006-01-02 15:04:05")
+
+		// Get the funding or spending transaction matching index if there is a
+		// matching tx hash already present.  During the next database
+		// restructuring we may want to consider including matching tx index
+		// along with matching tx hash in the addresses table.
+		if txn.MatchedTx != `` {
+			if !txn.IsFunding {
+				// Lookup by the database row id
+				idx, err := RetrieveFundingOutpointIndxByVinID(pgb.db, dbTx.VinDbIds[txn.InOutID])
+
+				if err != nil {
+					log.Warnf("Matched Transaction Lookup failed for %s:%d: id: %d:  %v",
+						txn.TxID, txn.InOutID, txn.InOutID, err)
+				} else {
+					addrInfo.Transactions[i].MatchedTxIndex = idx
+				}
+
+			} else {
+				// Lookup by the matching tx hash and matching tx index
+				_, _, idx, _, err := RetrieveSpendingTxByTxOut(pgb.db, txn.TxID, txn.InOutID)
+
+				if err != nil {
+					log.Warnf("Matched Transaction Lookup failed for %s:%d: %v",
+						txn.TxID, txn.InOutID, err)
+				} else {
+					addrInfo.Transactions[i].MatchedTxIndex = idx
+				}
+
+			}
+		}
 	}
 
 	addrInfo.NumUnconfirmed = numUnconfirmed
--- a/db/dcrpg/pgblockchain.go
+++ b/db/dcrpg/pgblockchain.go
@@ -1016,7 +1016,7 @@ func (pgb *ChainDB) FillAddressTransactions(addrInfo *explorer.AddressInfo) erro
 
 			} else {
 				// Lookup by the matching tx hash and matching tx index
-				_, _, idx, _, err := RetrieveSpendingTxByTxOut(pgb.db, txn.TxID, txn.InOutID)
+				_, idx, _, err := pgb.SpendingTransaction(txn.TxID, txn.InOutID)
 
 				if err != nil {
 					log.Warnf("Matched Transaction Lookup failed for %s:%d: %v",
@@ -1024,7 +1024,6 @@ func (pgb *ChainDB) FillAddressTransactions(addrInfo *explorer.AddressInfo) erro
 				} else {
 					addrInfo.Transactions[i].MatchedTxIndex = idx
 				}
-
 			}
 		}
 	}
--- a/explorer/explorerroutes.go
+++ b/explorer/explorerroutes.go
@@ -319,17 +319,13 @@ func (exp *explorerUI) TxPage(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	inout, ok := r.Context().Value(ctxTxInOut).(string)
-	if !ok {
-		inout = ""
+	inout, _ := r.Context().Value(ctxTxInOut).(string)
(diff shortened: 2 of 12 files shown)


explorer/explorerroutes.go
```diff
@@ -319,8 +319,18 @@ func (exp *explorerUI) TxPage(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	// Initially try to get the transaction information from the "lite" mode
-	// data source, which relies on RPCs to dcrd.
+	inout, ok := r.Context().Value(ctxTxInOut).(string)
+	if !ok {
+		inout = ""
```
chappjc: If this happened, that would indicate a problem with our `TransactionIoIndexCtx` middleware, not the user input, so I think we should log this.  Also, on a failed type assertion with `!ok` being the case, the value of `inout` will be the string zero value (empty string).

explorer/explorerroutes.go
```diff
@@ -319,8 +319,18 @@ func (exp *explorerUI) TxPage(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	ioid, ok := r.Context().Value(ctxTxInOutId).(string)
+	var inoutid int64
+	inoutid, err := strconv.ParseInt(ioid, 10, 0)
```
chappjc: Why first declare `var inoutid int64`?  Also, why the if statement, and why the assignment of 0 (the default value already) to `inoutid`?

explorer/explorerroutes.go
```diff
@@ -655,6 +667,8 @@ func (exp *explorerUI) TxPage(w http.ResponseWriter, r *http.Request) {
 		exp.Height() - tx.Confirmations,
 		exp.Version,
 		exp.NetName,
+		inout,
```
chappjc: Should error if `inout` is not either "in" or "out".

main.go
```diff
@@ -558,6 +558,7 @@ func mainCore() error {
 	webMux.Get("/parameters", explore.ParametersPage)
 	webMux.With(explore.BlockHashPathOrIndexCtx).Get("/block/{blockhash}", explore.Block)
 	webMux.With(explorer.TransactionHashCtx).Get("/tx/{txid}", explore.TxPage)
+	webMux.With(explorer.TransactionHashCtx,explorer.TransactionIoIndexCtx).Get("/tx/{txid}/{inout}/{inoutid}", explore.TxPage)
```
chappjc: To avoid the parsing and checking of {inout}, you could have two routes,

```go
webMux.With(explorer.TransactionHashCtx, explorer.TransactionIndexCtx, explorer.TxnIsIn).Get("/tx/{txid}/in/{inoutid}", explore.TxPage)
webMux.With(explorer.TransactionHashCtx, explorer.TransactionIndexCtx, explorer.TxnIsOut).Get("/tx/{txid}/out/{inoutid}", explore.TxPage)
```

where `explorer.TransactionIndexCtx` only checks `{inoutid}`, `explorer.TxnIsOut` sets `ctxTxInOut` to `"in"`, and  `explorer.TxnIsIn` sets `ctxTxInOut` to `"out"`.

Any path along these lines with anything other than "in" or "out" in that position would not match a handler.

db/dcrpg/pgblockchain.go
```diff
@@ -997,6 +997,36 @@ func (pgb *ChainDB) FillAddressTransactions(addrInfo *explorer.AddressInfo) erro
...
+				if err != nil {
+					log.Warnf("Matched Transaction Lookup failed for %s:%d: id: %d:  %v",
+						txn.TxID, txn.InOutID, txn.InOutID, err)
+				} else {
+					addrInfo.Transactions[i].MatchedTxIndex = idx
+				}
+
+			} else {
+				// Lookup by the matching tx hash and matching tx index
+				_, _, idx, _, err := RetrieveSpendingTxByTxOut(pgb.db, txn.TxID, txn.InOutID)
```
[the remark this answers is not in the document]
papacarp: ok

main.go
chappjc: But I'm fine with the way you have it as long as we be sure to handle incorrect paths.

db/dcrpg/pgblockchain.go
```diff
@@ -997,6 +997,36 @@ func (pgb *ChainDB) FillAddressTransactions(addrInfo *explorer.AddressInfo) erro
...
+
+				if err != nil {
+					log.Warnf("Matched Transaction Lookup failed for %s:%d: %v",
+						txn.TxID, txn.InOutID, err)
```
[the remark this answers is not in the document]
papacarp: The template triggers off of a value in matchedTx (the hash) and assuming we can get a matched tx index for every case that we have a matchedTx != '' then I think we are ok.  

Having said that as part of my testing this afternoon I confirmed that we do have some ambiguous results for this query and as such should add is_mainchain and is_valid to the selection criteria for this query.

```
SelectSpendingTxByPrevOut = `SELECT id, tx_hash, tx_index, tx_tree FROM vins
		WHERE prev_tx_hash=$1 AND prev_tx_index=$2 AND is_valid=TRUE AND is_mainchain=TRUE;`
```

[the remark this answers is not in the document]
chappjc: Ah, right I meant to bring that up.  Alternatively, `ORDER BY is_valid DESC, is_mainchain DESC, block_time DESC`.  This way, if there is no valid mainchain instance, you can at least link to the orphaned/invalidated transaction, which we are now tracking.  If there is a valid one, it will be first, and a valid mainchain one before valid side chain ones.  The block_time ordering is to go with the most recent one if there are no valid mainchain ones.

explorer/explorerroutes.go
papacarp: !ok is true if the user links to a transaction without any in/out designation.  as an example: `tx/1c5c53d3a604f4b875173e4681e3e789faeed1dd563e04970d1e5655c8cc885e`

So its not something we should log.  Its normal operation given we are sharing the route

Good to know on the empty string.  I'll try removing the entire !ok if statement.

explorer/explorerroutes.go
papacarp: My take was to let it default to no highlighting and not force an error.  Right now the only way it will highlight is if inout = "in" or "out" and the index specified is match-able on the page.  Given the way this is done sharing the same tx page as a non-highlighting link, we would need to error on anything but "in", "out" and "".  

Happy to do expicit url validity if you want however. What type of error do you want to throw?  something like:  ```No matching page, block, address or transaction that could be found
that transaction has not been recorded``` ?

explorer/explorerroutes.go
papacarp: oops, leftover from some type issue I had during development I'm sure.  I'll remove it.

Top of Tx page is now greatly simplified because the default error values are all ok.
```
inout, _ := r.Context().Value(ctxTxInOut).(string)
ioid, _ := r.Context().Value(ctxTxInOutId).(string)
inoutid, _ := strconv.ParseInt(ioid, 10, 0)
```

explorer/explorerroutes.go
chappjc: I'm sure the front end does the right thing, but the user can type in anything for the url.

I think a redirect back to the vanilla /tx/{txid} would be most user friendly, but really it should be a 404 of some kind

main.go
papacarp: Ok, i put in the check to validate the in/out/''.  I didn't do anything with the index value at the end.  right now it can be anything still.  you ok with that?  
```
...6eb590adfaf6/in/apples
...6eb590adfaf6/out/bananas
...6eb590adfaf6/in/55334343234

```

These are all valid at present

chappjc: Lol, yeah that's cool.

papacarp: Given the explicit url matching you want it might be best to just use separate routes as you recommend.  Seems like a cleaner solution.  I'm happy to do that later today or tomorrow.  Up to you.

PR #697, opened 2018-09-20
Properly test object validity on null chart data

papacarp:
Uses LoDash rather than an array length to test for presence of chart data.  Fixes PR #680


gozart1:
LGTM.

PR #706, opened 2018-09-26
dcrpg: addresses table query optimization with block_time_index sorting

chappjc:
This creates the addresses' table block_time_index *with sorting* on the
`block_time` column in order to avoid a sort when querying for rows by a
particular address.
Add `ReindexAddressesBlockTime`, to simplify dropping and re-creating the
block_time_index with this sorting.  Implement the upgrade from dcrpg
3.5.4 to 3.5.5 for this block_time reindexing.

This is motivated in part by the slowness of such queries with large numbers
of rows in the addresses table, such as
DshMNsvETDWpVoCe1re9NTAChiJagzsFV7J, which has over 116,746
transactions some of which have hundreds of inputs.
@papacarp Proposed the index sorting solution implemented here.

The reindexing is fairly quick:
```
15:01:38.196 [INF] PSQL: Reindex the addresses table on block_time
15:01:38.196 [INF] PSQL: Reindexing addresses table on block time.
15:01:51.213 [INF] PSQL: Modified the vouts table version to 3.5.5
```

### DIFF
--- a/db/dcrpg/internal/addrstmts.go
+++ b/db/dcrpg/internal/addrstmts.go
@@ -118,11 +118,13 @@ const (
 		ORDER BY addresses.block_time DESC;`
 
 	SelectAddressLimitNByAddress = `SELECT ` + addrsColumnNames + ` FROM addresses
-	    WHERE address=$1 AND valid_mainchain = TRUE ORDER BY block_time DESC LIMIT $2 OFFSET $3;`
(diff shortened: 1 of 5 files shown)


papacarp:
I can't evaluate until I get my db re-sync'd but I reviewed the code and got through the upgrade process and it all looks good to me.  I recommend you push it in.

Issue #352, opened 2018-02-15
Load web server before building the db


raedah:
Lets load the web server when launching the application, instead of waiting till after the database is built which can take a long time. We can display any message, status of the db building, or allow the interface to be used with available data.

chappjc:
It would be nice to have the profiler up too

influx6:
Am currently taking a go at this issue on my fork. See commit [a1d740ad6d153f3a6bb3f7505ea79f9a60ea3519](https://github.com/influx6/dcrdata/commit/a1d740ad6d153f3a6bb3f7505ea79f9a60ea3519). Although am able to use the http explorer api, i do wish to ask if there is away to deal with the estimator. See log below:

```2018-03-19 15:12:10.989 [WRN] BLKD: estimatestakediff is broken:  -32603: stake difficulty estimation does not currently work when the retarget interval is larger than the ticket maturity (interval 144, ticket maturity 16)
2018-03-19 15:12:11.016 [INF] BLKD: Unable to find block (0000000000bdeddd751ec946d6b2792b1e13d346dc02978ed1a2613524b8f303) in pool info cache, trying best block.
2018-03-19 15:13:01.602 [WRN] BLKD: Collected block height 78996 != stake db height 33872. Pool info will not match the rest of this block's data.
```

A you can see my changes, bring about a few issues with the estimator, my question are the ff:

1. Does this affect the way the httpapi shows and calculates the transactions being processed and loaded by the blockchain?

2. Will the issue be resolved once the full transaction has being resolved, so there would be no reason to worry on the estimator error above?

Any clearity would be great on this.

dmigwi:
Awesome stuff

dmigwi:
https://github.com/decred/dcrdata/issues/356, https://github.com/decred/dcrdata/issues/399 Relate to what this task will do.

On task completion a sync status page will be introduced which is the only page that can be loaded when the background sync is in progress... Of course, a progress bar will be added on the new sync status page.

A configurable option will be added to the`dcrdata.conf` file that shows how many blocks back will it take to display the status page ONLY on the running web server. By default it will be set to 2 and we could allow user to increase the value to maybe 5,000 blocks behind.

chappjc:
Resolved by https://github.com/decred/dcrdata/pull/710

PR #710, opened 2018-09-27
Load webserver soonest possible

dmigwi:
This PR Fixes a bug that makes dcrdata to crash immediately after stakedb is rewound to height 0.

**Screenshot showing the crash after stake db rewinding**
screen shot 2018-09-24 at 20 27 24

It also loads the web server before blockchain sync starts running on system startup. Two case scenario happen here:

**Scenario One**
 A sync status page runs by default if;  the blockchain sync is run for the first time,`sync-status-limit` value is not set, or set to a value greater than 5000 or less than 2. 
`sync-status-limit` - is the number of blocks behind the best block height past which only the sync status page that can be viewed on the running web server.
            
 to set the `sync-status-limit` add the following to the `dcrdata.conf` file.
```
; Sets the threshold past which only the syncing status page can be served on
; the running web server when blockchain sync is running on dcrdata startup.
; The maximum value that can be set is 5000. If set to more than 5000
; or less than 2 syncing status page is loaded by default till syncing is done.
sync-status-limit=1000
```

**Screenshot showing a sync status page running.**
screen shot 2018-09-26 at 05 12 44

**Scenario Two**
If the `sync-status-limit` is set to 1000 and on system startup only like 700 blocks (should be less than 1000) are behind the best block height, the user will access the whole dcrdata api and pages without the restriction of only viewing the status sync page. The 700 blocks will be synced in the background as the user does other things with the app.

**Screenshot showing a complete blockchain sync status page.**
If the status sync page was running and the blockchain sync was successfully done, the user is automatically redirected to a working home page in 30 sec. They can choose not to wait for the 30 seconds and just visit any page of their choice. 

screen shot 2018-09-27 at 17 31 11


raedah:
would close https://github.com/decred/dcrdata/issues/352

db/dbtypes/types.go
```diff
@@ -549,3 +567,87 @@ type BlockStatus struct {
...
+	blockchainSyncStatus.Lock()
+	defer blockchainSyncStatus.Unlock()
+
+	percentage := 0.0
+	if to > 0 {
+		percentage = math.Floor((float64(from)/float64(to))*10000) / 100
+	}
+
+	val := &SyncStatusInfo{
+		Change:     percentage,
```
chappjc: I'm trying to understand the semantics of the input arguments to `SyncStatusUpdate`.

When this function is called [here](https://github.com/decred/dcrdata/pull/710/files#diff-60206eba7205ea27cd7a8137d424f61cR2018), `from` is set to `int64(i)`, while `to` is set to `int64(totalVinIbIDs)`, where `i` is the current block and `totalVinIbIDs` is the number of blocks to process.  So this is a measure of total progress, not an incremental change.

The definition of `percentage` above confirms that it's "percentage complete" (as opposed to "percentage remaining" or "percentage change since last update" etc.).

So unless I'm misunderstanding, please rename `Change` to `PercentComplete` for clarity.

db/dcrpg/pgblockchain.go
```diff
@@ -2364,15 +1983,17 @@ func (pgb *ChainDB) UpdateSpendingInfoInAllAddresses() (int64, error) {
 	}
 
 	updatesPerDBTx := 500
+	totalVinIbIDs := len(allVinDbIDs)
 
-	log.Infof("Updating spending tx info for %d addresses...", len(allVinDbIDs))
+	var timeStart time.Time
```
chappjc: `timeStart` needs a non-zero initial value so the first call to `time.Since` isn't crazy.  Either that or move the `timeStart = time.Now()` in the loop before `time.Since`.

config.go
```diff
@@ -104,7 +104,8 @@ type config struct {
 	PGPass        string `long:"pgpass" description:"PostgreSQL DB password."`
 	PGHost        string `long:"pghost" description:"PostgreSQL server host:port or UNIX socket (e.g. /run/postgresql)."`
 	NoDevPrefetch bool   `long:"no-dev-prefetch" description:"Disable automatic dev fund balance query on new blocks. When true, the query will still be run on demand, but not automatically after new blocks are connected."`
-	SyncAndQuit   bool   `long:"sync-and-quit" description:"Sync to the best block and exit. Do not start the explorer or API."`
```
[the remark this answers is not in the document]
dmigwi: I did away with `SyncAndQuit` since I made the sync part of the code to be among the last thing that is executed in the mainCore function in main.go. I found that part of the code to be redundant as most of the functions are already up and running by the time blockchain sync ends. This is [where](https://github.com/decred/dcrdata/pull/710/files#diff-7ddfb3e035b42cd70649cc33393fe32cR686) that piece of code would go but it if I put it back it would be pretty much redundant especially when no sync status page is running which will be common.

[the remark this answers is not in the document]
dmigwi: ok. I could still get it back and we can try to reason with in place if we should keep it or not.

dmigwi:
To fully test this PR one might have to run a fresh data migration. No upgrade that is implemented here.

config.go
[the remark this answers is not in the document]
chappjc: Well we can discuss it now.  I may be missing something, as I don't get "redundant especially when no sync status page is running which will be common".  The `SyncAndQuit` flag was to *quit* dcrdata when sync is complete.  The new option doesn't seem relevant to this.

main.go
```diff
@@ -245,17 +247,32 @@ func mainCore() error {
...
+			expectedHeight = height
+		}
+
 		// How far auxDB is behind the node
-		blocksBehind := height - lastBlockPG
+		blocksBehind = expectedHeight - lastBlockPG
 		if blocksBehind < 0 {
 			return fmt.Errorf("Node is still syncing. Node height = %d, "+
-				"DB height = %d", height, heightDB)
+				"DB height = %d", expectedHeight, heightDB)
```
chappjc: But we know the  height of the best block... Why guess it from the best block's timestamp?

main.go
```diff
@@ -421,124 +400,139 @@ func mainCore() error {
...
+			wg.Add(1)
+			go mpm.TxHandler(dcrdClient)
 		}
+		return nil
+	}
 
-		newTicketLimit := int32(cfg.MPTriggerTickets)
-		mini := time.Duration(cfg.MempoolMinInterval) * time.Second
-		maxi := time.Duration(cfg.MempoolMaxInterval) * time.Second
+	// do not collect the data summary if the sync status page has been activated.
```
chappjc: `collectBlocksAfterSyncing` does two main things:
1. configures and starts handlers such as new blocks, reorg, mempool.
2. Initial data collection for the explorer.

The naming of `collectBlocksAfterSyncing` should change to reflect both the start of handlers, and the initial data collection.  Also, the `AfterSyncing` is not appropriate for the function name as it does not know about sync, it just performs these two functions.

// do not collect the data summary if the sync status page has been activated.

main.go
```diff
@@ -421,124 +400,139 @@ func mainCore() error {
...
+	if !explore.SyncStatus {
+		if err := collectBlocksAfterSyncing(); err != nil {
```
chappjc: How can calling this function work before `getSyncd` is run?

chappjc: First, if all the event handlers are started, and then sync is run, there's trouble when an event comes and is processed for a block beyond the DB best block.  That's why we don't start handlers until later.
Second, the Collect()/Store() parts of `collectBlocksAfterSyncing` that get data for the explorer are for an old block.  How is it going to get the fresh data when sync is done (or as it goes on)?

db/dcrsqlite/apisource.go
```diff
@@ -40,15 +40,17 @@ type wiredDB struct {
 	params   *chaincfg.Params
 	sDB      *stakedb.StakeDatabase
 	waitChan chan chainhash.Hash
+	liteMode bool
```
chappjc: wiredDB should NOT be aware of the modes of explorer.  The notion of lite/full mode stays in explorer/api.  Instead of this struct field being called `liteMode`, name it according to its purpose, e.g. `updateSyncStatus`.

main.go
dmigwi: This part was introduced help to handle the case scenarios where dcrdata data migration is initiated when dcrd is not sync'd with the latest block in the chain. That difference in sync affects how we may handle data in bulk.
We don't have trustworthy way to fetch the latest block in the chain even in case scenarios when dcrd and dcrdata are almost simultaneously run together. Adding the `expectedHeight` value that is calculated from the best block timestamp means that we can roughly be able to estimate the latest block in the chain even when dcrd is not sync.

chappjc:
```go
elapsedTime := time.Since(block.Timestamp).Minutes()
expectedHeight := int64(elapsedTime / 5)
```
Say this indicates the node's best block is 1 day old, `expectedHeight` as computed is the estimate of the *number of blocks dcrd has left to process to get up with the network*, not the networks estimated best block height.

The expected height would be `int64(elapsedTime / p.ChainParams.TargetTimePerBlock) + int64(block.Height)`.

main.go
dmigwi: I expect it to work especially in the case scenarios when its called before `getSyncd`, I mean the times when `explore.SyncStatus` is false and sync status page is not running as the only accessible page on the web server. The number of blocks behind the best block then should be quite few (can only be set to a max of 5000) which means that by the time the system expects to handle this unsynced load, much of the data may have already been loaded.

"quite few" in the above statement is not always correct because some machines will take longer to process a given load. This brings us to the option of configuring the `sync-status-limit` value that should help us control what load a given machine can process simultaneously while serving the web server an still be effective.

db/dcrsqlite/apisource.go
dmigwi: This `liteMode` is needed to determine which of the `sync.go` file either from `dcrpg` or `dcrsqlite` package should be updating the progress that is running on the sync status page.

[Here](https://github.com/decred/dcrdata/pull/710/files#diff-da97fde754870fa5c4b02913ac225e8aR249) is one instance where its important.  On Lite Mode `func (db *wiredDB) resyncDB(` method in `sync.go` from `dcrsqlite` package sends the updates  while  on full mode `func (db *ChainDB) SyncChainDB(` in `sync.go` from `dcrpg`package sends the updates.

It would be so unfortunate and confusing to have been methods sending progress updates especially when Full Mode is running since both functions must run.

dmigwi: I will change it to that name.

main.go
dmigwi: here the bigger value is used as what should be used to make decisions on how to handle the bulk data.

```go
if height > expectedHeight {
        expectedHeight = height
}
```

main.go
chappjc: > the times when explore.SyncStatus is false and sync status page is not running as the only accessible page on the web server

I get that in the above case the idea is to make then entire explorer available.  Correct?

The problems I mentioned are still a concern.  The new block and reorg handlers are up and running because of `collectBlocksAfterSyncing`, which means there will be new block notifications from dcrd for new blocks that we cannot possibly connect in dcrdata's DBs.  The handlers should not be running for this reason.

Also, when `collectBlocksAfterSyncing` runs Collect and Store, what block is it collecting and sending to the explorer UI? It will not be in the DB. Also, when sync does finish, it won't run Collect and Store after sync because it's moved up here now...

main.go
chappjc: But `expectedHeight` is not a height, it is an estimate of the remaining blocks that dcrd has to sync.  In plain words, what is it that `expectedHeight` should represent?

db/dcrsqlite/apisource.go
chappjc: The types implement different interfaces. In package api, `wiredDB` implements the `DataSourceLite` interface, while `ChainDB` implements `DataSourceAux`.  So each type has a clear role w.r.t. modes.  I think it's fair to put a comment in the docs for `InitWiredDB` explaining that a case when `wiredDB` should be responsible for updating sync status is if the auxiliary / full-mode DB is not handling the task.

main.go
dmigwi: `expectHeight` is simply an estimate of what block height value should be used to make  close to accurate decisions on how to handle bulk data.

The previous implementation was only accurate if dcrd was in sync with the latest block from the blockchain.

main.go
dmigwi: >I get that in the above case the idea is to make then entire explorer available. Correct?

Yes

> The new block and reorg handlers are up and running because of collectBlocksAfterSyncing, which means there will be new block notifications from dcrd for new blocks that we cannot possibly connect in dcrdata's DBs. 

I get your concern and you correct on this `The handlers should not be running for this reason`. I will fix that so that the handlers will be started still in the `collectBlocksAfterSyncing` function but will only trigger any further processing if the data that is needed already exist in all dbs.

The reason why i think the handlers should remain in the same position as they are, is that  some machines will be fast enough to process the "quite few" blocks that are pending.

I also need to think through this section or the again as it too error prone.

config.go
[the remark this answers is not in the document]
dmigwi: The reason why I thought adding back that code would be redundant was that, the blockchain sync will now be run as the last thing. Also I expect that the sync status page to be displayed on special occasions, once everyone understands the purpose of the PR which is to make the user to never wait for the background syncing completion whenever possible, for them to use the web app. But I have had a changed of mind.

```go
	log.Infof("All ready, at height %d.", baseDBHeight)

	if cfg.SyncAndQuit {
		return nil
	}

	// collect the data now it was not collected earlier. Set up the monitors too.
	if explore.SyncStatus {
		if err := collectBlocksAfterSyncing(); err != nil {
			return err
		}

		explore.RetrieveUpdates()
	}

	// Deactivate displaying the sync status page after the db sync was
	// completed successfully.
	explore.SyncStatus = false

	// Wait for notification handlers to quit
	wg.Wait()

	return nil
```

I have looked at this code again and I can see that there is `wg.Wait()` to used wait for any of the running goroutines to send quit status.

It now makes sense to keep the `SyncAndQuit` implementation since we may not want to wait for the quit message to be sent just as you described.

main.go
chappjc: Right, but it is not "an estimate of which height value", at least not as you are computing it now.

If `block` is dcrd's current best block that is somewhere behind the network's best block, and then you have:

```go
elapsedTime := time.Since(block.Timestamp).Minutes()
expectedHeight := int64(elapsedTime / 5)
```

Then `elapsedTime` is the time since dcrd's best block was mined.  So `elapsedTime / 5` is an estimate of the number of blocks dcrd is behind by, not the height of the network that dcrd is trying to reach.  Note in https://github.com/decred/dcrdata/pull/710#discussion_r221281566 I said I would expect you'd need to compute:

`int64(elapsedTime / p.ChainParams.TargetTimePerBlock) + int64(block.Height)`

main.go
chappjc: Sure, some machine may get done processing the pending blocks before a new one is mined, but we can't rely on that.

dmigwi: Generally for all machines whether fast or slow, the handlers will have to confirm if the needed information already exists. This means that monitors will not trigger further processing if the need block is not yet processed.

chappjc: Just to be clear, we're not talking about when dcrd is still syncing, this is regarding when dcrdata is syncing to dcrd.  Take `BlockConnectedHandler` in the `blockdata` package.  Say sync is at block 1000, and dcrd sends a notification that block 1200 was just mined, then `blockdata`'s `BlockConnectedHandler` gets the block hash in a message, collects all the data that dcrd has about the block without problem since dcrd has it, but the problems come with when `BlockConnectedHandler` calls `CollectHash`, which calls `blockdata/(*Collector).CollectBlockInfo`, for because stakedb is behind and there's no ticket pool info for that new block from dcrd.  If there were a workaround to that there would still be a problem when `BlockConnectedHandler` calls `Store` for all the `dataSaver`s.

dmigwi: This is another valid concern and I think any data update triggered by dcrd should be blocked by all means while the sync is in progress. What we could allow is that data that has been pushed to the dbs during sync could be processed for display as the latest update fetched. This is data processing that I was taking about with this.
>the handlers will have to confirm if the needed information already exists.

Also the sync code is designed to stop after initial startup  if the best block from dcrd is in sync with the best block from dcrdata. So the blocked dcrd updates will be handled eventually.

BTW, this PR only affects how dcrdata initial startup will work after that the current norm takes precedence.

chappjc: > What we could allow is that data that has been pushed to the dbs during sync could be processed for display as the latest update fetched.

I referred to this earlier, and I would suggest updating the explorer UI every ~20 blocks while syncing to dcrd.

>  So the blocked dcrd updates will be handled eventually.

While dcrd is syncing to the network, each block processed by dcrd will be notified to dcrdata.  They will stack up, yes, but the regular sync loop would have handled them.  They'll be out of order w.r.t. the DB height and/or already processed.

dmigwi: As long as the first sync to all tables except to the addresses table isn't done yet this 

>While dcrd is syncing to the network, each block processed by dcrd will be notified to dcrdata. They will stack up, yes, but the regular sync loop would have handled them. They'll be out of order w.r.t. the DB height and/or already processed.

will be handled here.

```go
// Update height, the end condition for the loop
if nodeHeight, err = client.NodeHeight(); err != nil {
	return ib, fmt.Errorf("GetBestBlock failed: %v", err)
}
```
if its done it will definately be picked up this immediately after the addresses table sync.

```go
for baseDBHeight < height {
	fetchToHeight = auxDBHeight + 1
	baseDBHeight, _, err = getSyncd(updateAllAddresses, updateAllVotes,
			newPGIndexes, fetchToHeight)
	if err != nil {
		return err
	}
	_, height, err = dcrdClient.GetBestBlock()
	if err != nil {
		return fmt.Errorf("unable to get block from node: %v", err)
	}
}
```

dmigwi: >How can calling this function work before getSyncd is run?

It now makes lots of sense why initiating the monitors before sync is complete isn't such a good idea. If I went ahead with my implementation of initiating the monitors & handlers before, I would be forced to block them all to avoid db corruption thereby piling unnecessarily pressure on the underlying hardware's processing power.

main.go
```diff
@@ -397,143 +395,88 @@ func mainCore() error {
...
+		// can view on the running webserver but the syncing of blocks behind the
+		// best block is happening in the background. No new blocks monitor from
+		// dcrd will be initiated till the best block from dcrd is in sync with
+		// the best block from dcrdata.
+		go func() {
+			// After the sync on initial startup is complete this goroutine is no
+			// longer needed, so kill it.
+			defer runtime.Goexit()
+
+			isStatusSyncingDisabled := make(chan bool)
```
chappjc: Is this just to get out of the `for` loop below?  If so, I do not think it is working as intended.

In any case, to get out of the outer loop, not just the select, you can tag the for and specify that in the break.  e.g.:

```go
out:
	for {
		select {
		case ...:
		if ... {
			break out
...
```

However, the select is not needed since with just one useful case, you can use for range over a channel:

```go
for hash := range latestBlockHash {
...
```

See https://golang.org/ref/spec#For_range

main.go
```diff
@@ -397,143 +395,88 @@ func mainCore() error {
...
+					}
+
+					// store the blockdata fetch for the explorer pages
+					if err = explore.Store(d, nil); err != nil {
+						log.Warnf("failed to store (%s) hash's blockdata for the explorer pages error: %v",
+							hash.String(), err)
+					}
+
+				case <-isStatusSyncingDisabled:
+					break
```
chappjc: This break breaks out of the `select`, not the `for`.

db/dcrpg/sync.go
```diff
@@ -218,9 +229,15 @@ func (db *ChainDB) SyncChainDB(client rpcutils.MasterBlockGetter, quit chan stru
 		if nodeHeight, err = client.NodeHeight(); err != nil {
 			return ib, fmt.Errorf("GetBestBlock failed: %v", err)
 		}
+
+		// if updating explorer is activated, update it at intervals of 20
+		if ib%20 == 0 && explorer.SyncExplorerUpdateStatus() && updateExplorer != nil && !updateAllAddresses {
```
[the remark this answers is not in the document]
dmigwi: `explorer.SyncExplorerUpdateStatus` might still be important since the channel `updateExplorer` might not be nil but the update should not go through.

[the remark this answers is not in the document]
chappjc: Sure, you keep all conditions that are needed, but by putting `updateExplorer != nil` first it won't bother to call the function if `updateExplorer==nil`

explorer/explorertypes.go
```diff
@@ -574,17 +573,19 @@ func SyncStatusUpdate(from, to, timeToComplete int64, updateType, msg string) {
...
-		blockchainSyncStatus.Update = map[string]*SyncStatusInfo{
-			updateType: val,
-		}
+		blockchainSyncStatus.Update = map[string]*SyncStatusInfo{updateType: val}
 	} else {
-		blockchainSyncStatus.Update[updateType] = val
+		if _, ok := blockchainSyncStatus.Update[updateType]; ok {
+			blockchainSyncStatus.Update[updateType] = val
+		} else {
+			blockchainSyncStatus.Update = map[string]*SyncStatusInfo{updateType: val}
```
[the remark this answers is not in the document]
dmigwi: I added the `updateType` as a field in `SyncStatusInfo` since javascript does not support maps data structure. To send this data to the javascript side via the websocket, I have to convert the map into an array and the update type is one of the things that are needed to display data accurately and in the right position.

It so happens that looping through maps in golang doesn't maintain the items insertion order. Using the `updateType` as a value in the struct, we can ensure what progress data that should appear first will always appear as first on the status page.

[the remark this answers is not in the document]
chappjc: Let's chat in slack about this.

main.go
```diff
@@ -397,143 +395,88 @@ func mainCore() error {
...
+		// This goroutines updates the blocks needed on the explorer pages. It
+		// only runs when the status sync page is not the defualt page that a user
+		// can view on the running webserver but the syncing of blocks behind the
+		// best block is happening in the background. No new blocks monitor from
+		// dcrd will be initiated till the best block from dcrd is in sync with
+		// the best block from dcrdata.
+		go func() {
+			// After the sync on initial startup is complete this goroutine is no
+			// longer needed, so kill it.
+			defer runtime.Goexit()
```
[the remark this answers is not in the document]
dmigwi: Should be at the end of the goroutine without the defer.

dmigwi:
screen shot 2018-10-01 at 19 40 53

dmigwi:
@chappjc Slack just did hang on me. Well above is the screenshot

main.go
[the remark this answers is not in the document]
chappjc: The function returning on its own after sync is good.  What does Goexit achieve here? (Honest question, I have never used Goexit and it's hard to think of legit use cases).

chappjc:
Regarding the heuristics to guess if dcrd is still syncing to the network, that may not be needed for link.  Please see the following issue, https://github.com/decred/dcrd/issues/1475#issuecomment-426041564.

dmigwi:
This awesome. Could we keep our current implementation till the new rpc endpoint is ready?

main.go
dmigwi:
```golang
for hash := range latestBlockHash {
...
```
This version isn't working.  I will revert to using the labels

chappjc:
Yeah it's good for now until that RPC is ready.

dmigwi:
I have run a another round of testing and fixed a couple of bugs.

public/js/controllers/status.js
```diff
@@ -0,0 +1,71 @@
...
+                var d = JSON.parse(evt);
+                var i;
+
+                for (i = 0; i < d.length; i ++) {
+                    var v = d[i]
+
+                    $("#"+v.ProgressBarID).html(buildProgressBar(v));
+
+                    if (v.BarSubtitle === "sync complete"){
+                        $(".alert.alert-info h5").html("Blockchain sync is complete. Redirecting to home in 30secs.");
```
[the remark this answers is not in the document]
dmigwi: Yes, 30 sec is supposed to give the system some to make all the needed information ready. 30 sec is slow but someone who has waited for a couple hours to get the system sync'd should be able to give the 30 sec.

explorer/explorertypes.go
```diff
@@ -535,3 +541,96 @@ func AddressPrefixes(params *chaincfg.Params) []AddrPrefix {
...
+	} else {
+		for i, v := range blockchainSyncStatus.ProgressBars {
+			if v.ProgressBarID == barID {
+				blockchainSyncStatus.ProgressBars[i] = val
+				return
+			}
+		}
+		// new entry
+		blockchainSyncStatus.ProgressBars = append(blockchainSyncStatus.ProgressBars, val)
+	}
```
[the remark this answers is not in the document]
dmigwi: We may have a future whereby status sync pages might be used to display progress updates in a given auxiliary db upgrade(s) in future. Setting the `ProgressBars` len to 2 IMO I think we may be limiting the usage of the sync update feature.

dmigwi:
Wherever we have the bypass implementation is what fixes the crash mentioned in this PR write up when the stake db is rewound to a height of -1. Since its this one instance only that dcrdata crashes I choose to build a custom solution about how this unique case is handled.

So when stake db is set to be rewound to a height of -1 it only goes to zero, this means at one instance we try to update a block that already exist in the stake db causing to crash the system. There was the option of blocking the update of a block that is less than or equal to the stake db height but this became a blocking operation because `waitChan` could not receive the notification of the consecutive block to be processed. `bypassWaitChan` is introduced to handle the situations whereby `waitChan` will not send notifications of the next block to be processed because the previous block wasn't sent to the stake db thus not even processed.

explorer/explorertypes.go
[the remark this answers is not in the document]
chappjc: Fair point.

db/dcrpg/pgblockchain.go
```diff
@@ -2365,69 +1984,49 @@ func (pgb *ChainDB) UpdateSpendingInfoInAllAddresses() (int64, error) {
...
 		endChunk := i + updatesPerDBTx
-		if endChunk > len(allVinDbIDs) {
-			endChunk = len(allVinDbIDs)
+		if endChunk > totalVinIbIDs {
+			endChunk = totalVinIbIDs
 		}
+
+		// fullMode is definately running so no need to check.
+		timeTakenPerBlock := (time.Since(timeStart).Seconds() / float64(endChunk-i))
+		timeToComplete := int64(timeTakenPerBlock * float64(totalVinIbIDs-endChunk))
```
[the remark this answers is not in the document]
dmigwi: Because it not possible to accurately determine how the future will be (how long a given update will actually take) I have using the word `approx.` is telling the using there may be a variation in the time this update might take.

db/dcrsqlite/sync.go
```diff
@@ -149,9 +152,19 @@ func (db *wiredDB) resyncDB(quit chan struct{}, blockGetter rpcutils.BlockGetter
...
+		explorer.SyncStatusUpdate(0, 0, 0, dbtypes.InitialDBLoad, InitialLoadSyncStatusMsg)
+	}
+
 	// Start at next block we don't have in every DB
 	startHeight++
 
+	// bypassWaitChan defines if the wait channel should be bypassed since we did
+	// not connect to the stakedb block thereby triggering the notification for
+	// the next block in line.
+	var bypassWaitChan bool
```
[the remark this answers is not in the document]
dmigwi: Let me reword this comments to something better but this actually the bug fix.

chappjc:
But it was erroring in main:

```go
// stakedbHeight is always rewinded to a height of zero even when lastBlockPG is -1.
			if stakedbHeight != lastBlockPG && stakedbHeight > 0 {
				return fmt.Errorf("failed to rewind stakedb: got %d, expecting %d",
					stakedbHeight, lastBlockPG)
			}
```

And the `stakedbHeight > 0` fixed it.

The stakedb only gets rewound to zero, not -1, because of this check:

```go
func (db *wiredDB) RewindStakeDB(toHeight int64, quit chan struct{}) (stakeDBHeight int64, err error) {
	// rewind best node in ticket db
	stakeDBHeight = int64(db.sDB.Height())
	if toHeight < 0 {
		toHeight = 0
	}
```

I'm still not understanding how in `(db *wiredDB) resyncDB` it would happen that `(blockHeight != i)`.  That indicates that the `waitChan` set in `(db *wiredDB).SyncDBAsync` was created for the wrong block height.  If you turn on debug logging, it will say something like `"Setting block gate height to %d"`.  This is where the problem needs to be addressed.

db/dcrsqlite/sync.go
```diff
@@ -189,30 +202,58 @@ func (db *wiredDB) resyncDB(quit chan struct{}, blockGetter rpcutils.BlockGetter
...
 			if err = db.sDB.ConnectBlock(block); err != nil {
 				return i - 1, err
 			}
 		}
 
+		// If the current block's height fetched earlier, using the notification
+		// waitChan is not greater than the current stakedb height, bypass the waitChan
+		// to fetch the blocks directly till we find a block whose height is greater
+		// than the stakedb height. For blocks whose height is less than the stakedb
+		// height, they exists in the other dbs so skip to the next iteration.
```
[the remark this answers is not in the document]
dmigwi: Its mostly the other way round, like stakedb is behind the wireDB because wireDb is never rewound but stakedb is always rewound to a height matching the auxiliary db. The sync now allows the stakedb to catchup with the wireDB.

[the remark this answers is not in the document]
chappjc: Ah right the wiredDB will be ahead of both.

db/dcrsqlite/sync.go
```diff
@@ -189,30 +202,58 @@ func (db *wiredDB) resyncDB(quit chan struct{}, blockGetter rpcutils.BlockGetter
...
+		bypassWaitChan = (blockHeight != i)
+		if bypassWaitChan {
+			// subtraction kicks in here because of blockGetter.WaitForHeight(i + 1).
+			// Here we want consecutive blocks to be fetched without skipping any
+			// because of (i+1) in the function above when waitChan kicks in.
+			i--
+			continue
```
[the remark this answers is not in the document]
dmigwi: `db.waitChan = blockGetter.WaitForHeight(i + 1)` only returns the notification of the current block if the previous one was processed and added to the stakedb. If by any means the previous block wasn't processed (added to stakedb) then the notification for the current block will never be sent via `db.waitChan`, the sync essentially just blocks forever.

`bypassWaitChan` is only set to true if when `i` is not equal to the new block height of the block to be sync'd. In reality this will only happen when`blockHeight > stakeDBHeight` is false thereby only running when `waitChan` expected not to send any block notifications.

If this still doesn't make sense lets have a chats about this via slack

db/dcrpg/pgblockchain.go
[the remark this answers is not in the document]
chappjc: Certainly it had to be an approximation, but the approximation can vary wildly.  When it hits a bunch of anomalous blocks, the estimate will jump up, then back down in the next more regular chunk of blocks.  It's the best way can do without smoothing out the estimate history, which would require a lot more code for little gain.

main.go
```diff
@@ -332,9 +347,196 @@ func mainCore() error {
...
+	case (usePG && auxDBheight < 40000) || wireDBheight < 40000:
+		explore.DisplaySyncStatusPage = true
+
+	// incorrect value set
+	case cfg.SyncStatusLimit < 2 || cfg.SyncStatusLimit > 5000:
+		explore.DisplaySyncStatusPage = true
+
+	// more blocks are behind the set sync status user value.
+	case blocksBehind > cfg.SyncStatusLimit:
+		explore.DisplaySyncStatusPage = true
```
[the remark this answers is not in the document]
dmigwi: used the 40,000 because we use it too to determine if a given update is a bulk on or not.

dmigwi:
lets say `db.waitChan = blockGetter.WaitForHeight(i + 1)` was initially called when the value of `i` was 1 thus `i+1 = 2` that means the current block `i` is at the height of 1. Therefore if by any case the block at height 1 is not added to the stake db, then notification for the block at height 2 will never be sent by the `waitChan`, the sync essentially blocks waiting for block at height 2 to be sent.

Using `bypassWaitChan` we detect those situations that `waitChan` can't return notifications because the previous block wasn't added to the stakedb and fetch the blocks directly. We then go back one step just to ensure that when the `waitChan` kicks in the next block requested will only have a difference of 1 with the previous block. Without `i--` the difference between the next and previous blocks being processed when `waitChan` kicks in is 2 because of that first time we did not anything after `blockHeight > stakeDBHeight` was false.

db/dcrpg/pgblockchain.go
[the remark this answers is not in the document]
dmigwi: Is there a way of doing this smoothing without resulting to a slower sync progress? Technically using the transactions estimation method was more accurate but would end up slowing down the sync.
This would be worse as it would give some more accuracy (not even close to 100%) and result to a slower sync. Consider this; syncing 500 blocks takes like 3 sec on average but the additional process of fetching the transactions count for a couple of last blocks sync'd and using that to estimate the average transaction counts per next bunch of 500 blocks will definitely take more than 5 sec to do a sync of 500 blocks on a slow machine. Take into consideration that this fetching of the transactions count for a couple of last blocks is going to happen at sync.go that doesn't have direct access to transactions count details thus slowing down the sync.
Because predicting the future close to accurate or even accurately here is almost impossible I would say the extra hustle of getting some more accurate values isn't worth it

public/js/controllers/status.js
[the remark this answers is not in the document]
dmigwi: Reduced it to 20 secs

explorer/explorermiddleware.go
```diff
@@ -66,6 +66,22 @@ func (exp *explorerUI) BlockHashPathOrIndexCtx(next http.Handler) http.Handler {
 	})
 }
 
+// SyncStatusPageActivation serves only the syncing status page until its
+// deactivated when DisplaySyncStatusPage is set to false. This page is served
+// for all the possible routes supported until the background syncing is done.
+func (exp *explorerUI) SyncStatusPageActivation(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if exp.DisplaySyncStatusPage {
```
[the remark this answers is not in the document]
dmigwi: We already have a mutex in the explorerUI struct that is what I will reuse.

db/dcrpg/pgblockchain.go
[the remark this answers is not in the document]
chappjc: Not worth it right now anyway.

chappjc:
In the interest of getting to the root of the coordinated sync / rewind issue, I've thoroughly documented the relevant pieces in the following PR, which also has two fixes for the issues:

https://github.com/decred/dcrdata/pull/720

The approach is fundamentally different from what you did, but I believe this is the "right way" since I believe the function (resyncDB) has all the information it needed to to the sync correctly when coordinating with ChainDB.

explorer/explorer.go
```diff
@@ -318,11 +322,14 @@ func (exp *explorerUI) StartSyncingStatusMonitor() {
...
 			case <-timer.C:
+				exp.NewBlockDataMtx.RLock()
+
 				if !exp.DisplaySyncStatusPage {
 					stop <- exp.DisplaySyncStatusPage
 				}
 				exp.wsHub.HubRelay <- sigSyncStatus
 
+				exp.NewBlockDataMtx.RUnlock()
 			case <-stop:
```
chappjc: Let's simplify this function.

```go
// StartSyncingStatusMonitor fires up the sync status monitor. It signals the
// websocket to check for updates after every syncStatusInterval.
func (exp *explorerUI) StartSyncingStatusMonitor() {
	go func() {
		timer := time.NewTicker(syncStatusInterval)
		for range timer.C {
			exp.NewBlockDataMtx.RLock()
			if !exp.DisplaySyncStatusPage {
				exp.NewBlockDataMtx.RUnlock()
				timer.Stop()
				return
			}
			exp.NewBlockDataMtx.RUnlock()
			exp.wsHub.HubRelay <- sigSyncStatus
		}
	}()
}
```

explorer/explorer.go
```diff
@@ -148,12 +149,70 @@ type explorerUI struct {
...
 	return exp.templates.reloadTemplates()
 }
 
+// SyncStatusUpdate receives the raw progress updates calculates the percentage
+// and updates the blockchainSyncStatus.ProgressBars.
+func (exp *explorerUI) SyncStatusUpdate(barLoad chan *dbtypes.ProgressBarLoad) {
+	// when this method is run check if it should be running in the first place.
+	exp.NewBlockDataMtx.RLock()
+	if exp.DisplaySyncStatusPage {
+		return
```
chappjc: `exp.NewBlockDataMtx` is still locked when you return.  Need to unlock it before return.

explorer/explorer.go
```diff
@@ -148,12 +149,70 @@ type explorerUI struct {
...
+	}
+	exp.NewBlockDataMtx.RUnlock()
+
+	for bar := range barLoad {
+		blockchainSyncStatus.Lock()
+
+		// For consecutive iteraions check when the sync status page is deactivated.
+		exp.NewBlockDataMtx.RLock()
+		if exp.DisplaySyncStatusPage {
+			return
```
chappjc: need unlock here too, but of both `blockchainSyncStatus` and `NewBlockDataMtx`

explorer/explorer.go
```diff
@@ -148,12 +149,70 @@ type explorerUI struct {
...
+		} else {
+			for i, v := range blockchainSyncStatus.ProgressBars {
+				if v.ProgressBarID == bar.BarID {
+					if len(bar.Subtitle) > 0 && bar.Timestamp == 0 {
+						// Handle case scenario when only subtitle should be updated.
+						blockchainSyncStatus.ProgressBars[i].BarSubtitle = bar.Subtitle
+					} else {
+						blockchainSyncStatus.ProgressBars[i] = val
+					}
+					return
```
chappjc: blockchainSyncStatus still locked

explorer/explorer.go
```diff
@@ -249,10 +308,17 @@ func New(dataSource explorerDataSourceLite, primaryDataSource explorerDataSource
 		}
 	}
 
-	if !exp.liteMode {
+	// Do not fetch charts updates when on liteMode or when blockchain syncing
+	// is running in the background.
+	exp.NewBlockDataMtx.RLock()
+
+	isSyncRunning := exp.DisplaySyncStatusPage || SyncExplorerUpdateStatus()
```
chappjc: If an `atomic.Value` is not an attractive solution (it has drawbacks), how about:

```go
func (exp *explorerUI) DisplaySyncStatusPage() bool {
	exp.SomeMtx.RLock()
	defer exp.SomeMtx.RUnock()
	return exp.displaySyncStatusPage
}
```

This is much less error prone.

db/dcrpg/sync.go
```diff
@@ -113,11 +114,25 @@ func (db *ChainDB) SyncChainDB(client rpcutils.MasterBlockGetter, quit chan stru
...
-	explorer.SyncStatusUpdate(0, 0, 0, dbtypes.InitialDBLoad, InitialLoadSyncStatusMsg)
-	// Addresses table sync should only run if bulk update is enabled.
-	if updateAllAddresses {
-		explorer.SyncStatusUpdate(0, 0, 0, dbtypes.AddressesTableSync, AddressesSyncStatusMsg)
+	if barLoad != nil {
+		// Add the various updates that should run on successful sync.
+		barLoad <- &dbtypes.ProgressBarLoad{
+			From:      0,
+			To:        0,
+			Timestamp: 0,
```
chappjc: I would elide all the zero value items (From, To, Timestamp) as this is done implicitly.

dmigwi:
@chappjc I can now say that this PR is ready for a further review and probably testing. I have tested it quite widely, found a number of bugs and fixed them. Kindly give it another look.

explorer/explorermiddleware.go
```diff
@@ -66,6 +66,22 @@ func (exp *explorerUI) BlockHashPathOrIndexCtx(next http.Handler) http.Handler {
...
 }
 
+// SyncStatusPageActivation serves only the syncing status page until its
+// deactivated when DisplaySyncStatusPage is set to false. This page is served
+// for all the possible routes supported until the background syncing is done.
+func (exp *explorerUI) SyncStatusPageActivation(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if exp.DisplaySyncStatusPage() {
+			exp.StatusPage(w, "Database Update Running. Please Wait...",
+				"Blockchain sync is running. Please wait ...", BlockchainSyncingType)
```
chappjc: In testing, I discovered that `BlockchainSyncingType` as the `statusType` results in `http.StatusServiceUnavailable` (503), however I'm not certain we should return an error code.  I think that 503 Service Unavailable does make sense though:

> The server is currently unable to handle the request due to a temporary overloading or maintenance of the 
server. The implication is that this is a temporary condition which will be alleviated after some delay. If known, the length of the delay MAY be indicated in a Retry-After header. If no Retry-After is given, the client SHOULD handle the response as it would for a 500 response.

My only concern here is how a client would handle the response.  We have seen websocket connections broken when pages reload with turbolinks if the initial response code was not a 200.  @SeaLightHouse ran into this.

We might consider 202: https://httpstatuses.com/202

> The request has been accepted for processing, but the processing has not been completed. The request might or might not eventually be acted upon, as it might be disallowed when processing actually takes place.

> There is no facility in HTTP for re-sending a status code from an asynchronous operation.

> The 202 response is intentionally noncommittal. Its purpose is to allow a server to accept a request for some other process (perhaps a batch-oriented process that is only run once per day) without requiring that the user agent's connection to the server persist until the process is completed. **The representation sent with this response ought to describe the request's current status and point to (or embed) a status monitor that can provide the user with an estimate of when the request will be fulfilled.**

public/css/main.css
```diff
@@ -1005,4 +1005,13 @@ body.darkBG .keynav-target {
 }
 .fs15 span.rm-spacing-tb .int {
   line-height: 1.6rem;
+}
+.sync-progress-bar {
+  animation: glowGreen 3000ms infinite;
+}
```
chappjc: We may wish to limit the width of the progress bar.  It looks a bit silly when the full width of a maximized window.
Perhaps adding to `.progress` way up on line 823 with `width: 480px` or something?

public/css/main.css
```diff
@@ -1005,4 +1005,13 @@ body.darkBG .keynav-target {
...
+.sync-progress-bar span {
+  vertical-align: -webkit-baseline-middle;
+}
+.container div.sync-progress {
```
chappjc: If we reduce/limit the width of the progress bar itself, we might want to center the sync-progress div with `margin: auto`.

explorer/explorermiddleware.go
dmigwi: Honestly I haven't given much thought to the http status code 503. When the user tries to access a page that is restricted while the sync in progress and the server doesn't serve it, we should not classify that as an error with 503 usage but 202 makes more sense in this case scenario.

chappjc:
The sync status (progress bars) seems to be working very well so far.  For a follow-up change, we might want to add a desktop notification for when it completes sync.

When the sync gets within 5000 blocks, I'll see how it works having the explorer functioning.  Actually, this brings up two things:
1. I don't think the explorer should be available, regardless of the sync status blocks config setting or block progress, unless the indexing is complete.  So this would be a catch up mode, not an initial sync.  Is it setup to not allow explorer access if indexing and/or address table spending data updates are not done?
2. There is no indication on the progress bars or their subtitles of what block it is on and how many blocks are left.  Certainly percentage and a guess of remaining time are good, but it should probably say what block it's on.  This item can wait for another day so we can get this merged.

main.go
```diff
@@ -332,9 +347,197 @@ func mainCore() error {
...
+	})
+	cacheControlMaxAge := int64(cfg.CacheControlMaxAge)
+	FileServer(webMux, "/js", http.Dir("./public/js"), cacheControlMaxAge)
+	FileServer(webMux, "/css", http.Dir("./public/css"), cacheControlMaxAge)
+	FileServer(webMux, "/fonts", http.Dir("./public/fonts"), cacheControlMaxAge)
+	FileServer(webMux, "/images", http.Dir("./public/images"), cacheControlMaxAge)
+
+	webMux.With(explore.SyncStatusPageActivation).Group(func(r chi.Router) {
+		r.NotFound(explore.NotFound)
+		r.Mount("/api", apiMux.Mux)
```
chappjc: I missed this until testing, but `r.Mount("/api", apiMux.Mux)` shouldn't be used with the `explore.SyncStatusPageActivation` middleware.  There should be another middleware that either redirects to the api status endpoint (returning a JSON structure indicating sync progress) or returns an appropriate code like 202, along with  the sync status as JSON.

db/dcrsqlite/sync.go
```diff
@@ -278,10 +330,25 @@ func (db *wiredDB) resyncDB(quit chan struct{}, blockGetter rpcutils.BlockGetter
...
 
-		// update height, the end condition for the loop
+		// Update height, the end condition for the loop
 		if _, height, err = db.client.GetBestBlock(); err != nil {
 			return i, fmt.Errorf("GetBestBlock failed: %v", err)
 		}
+
+		// If updating explore is activated, update it at intervals of 200 blocks.
+		if updateExplorer != nil && i%200 == 0 && explorer.SyncExplorerUpdateStatus() && db.updateStatusSync {
+			updateExplorer <- &blockhash
```
chappjc: Here's a spot where you could try to send the status update to the api too.

```go
select {
	case db.updateStatusChan <- summary.Height:
	default:
	}
```

dmigwi: On this part I don't quite follow. On the code I can see where channel `db.updateStatusChan` is written with data but no where its data is read from the channel. What is the purpose of `db.updateStatusChan` if no data it ever read from it?

chappjc: https://github.com/decred/dcrdata/blob/acd8da6aecb0dd4821b4e70117fe8803ed8bbd73/main.go#L154

main.go
```diff
@@ -332,9 +347,197 @@ func mainCore() error {
...
+	// Starts a goroutine that fetches the raw updates, process them and
+	// set them as the latest sync status progress updates. The goroutine exits
+	// when no sync is running.
+	explore.SyncStatusUpdate(barLoad)
+
 	blockDataSavers = append(blockDataSavers, explore)
 
-	// Sync up with the blockchain
+	// Register for notifications from dcrd
+	cerr := notify.RegisterNodeNtfnHandlers(dcrdClient)
```
chappjc: This should still be move way down near `initiateHandlersAndCollectBlocks`.  With this here dcrdata unnecessarily gets notifications for new blocks and transactions, although it has not yet launched any of the monitors (that's what the goroutines in `initiateHandlersAndCollectBlocks` do).

The handlers defined in `MakeNodeNtfnHandlers` and subsequently used when creating the RPC client just shuffle the data to the monitors, which are started in `initiateHandlersAndCollectBlocks`.  Note the design of the `rpcclient.NotificationHandlers` defined in `notification.MakeNodeNtfnHandlers`, and how the generally do not hang if the monitor is not receiving on the channel (the select-case-default).

The one handler that does send successfully is `OnBlockConnected` this is because the `(*collectionQueue).ProcessBlocks()` is always started, just with no synchronous handler functions defined for it to use (they get set via `collectionQueue.SetSynchronousHandlers` in `initiateHandlersAndCollectBlocks`).  This is evident with debug logging turned on as we see the following kind of message through the entire process:

    [DBG] NTFN: Synchronous handlers of collectionQueue.ProcessBlocks() completed in 242ns

While all of the above seems mostly harmless, it is not because a large number of the channel sends will "succeed" in filling up the channel buffer, only to be pulled out of the buffer loooong after the main sync is done.  When they do get pulled out of the buffer, these blocks and transactions are all already processed.  They should not be processed again.

For these reasons, it is necessary to not register for notifications from dcrd until we are ready (sync and indexing are completed).

main.go
dmigwi: Do we do the same thing for the insight api?

chappjc: Yeah they both need to return sensible json, not a web page

dmigwi: How about this as the return json?
```json
{ 
 "message":  "blockchain sync is in progress",
 "sync_complete": "1 of 2",
 "syncs": [ { 
       "progress_bar_id": "initial-load",
       "bar_note": "(Full Mode) Syncing stake, base and auxiliary DBs...",
       "best_block_height": 289673,
       "progress": 100.0,
       "time_to_complete": "0 sec",
       "subtitle": "",
      }, { 
       "progress_bar_id": "address-table-sync",
       "bar_note": "Syncing addresses table with spending info...",
       "best_block_height": 1967,
       "progress": 10.0,
       "time_to_complete": "1hr 20 sec",
       "subtitle": "",
      } ],
}
```

chappjc: Parsing it is tricky.  A program might want to know what stage, and stage progress.

So instead of `"sync_complete": "1 of 2"`, perhaps `"stage": 2`. <-- numeric

And instead of `"data"`, `"stages"`.

Then it's obvious and easy to get data for the active stage by indexing the stages array.

dmigwi: I have updated the return json above, you can also update it too just in case something else is missing.

chappjc: Lgtm

main.go
```diff
@@ -394,216 +597,153 @@ func mainCore() error {
...
+	explore.DisplaySyncStatusPage = false
+
+	// set that newly sync'd blocks should no longer be stored in the explorer.
+	// Monitors that fetch the latest updates from dcrd will be launched next.
+	explorer.SetSyncExplorerUpdateStatus(false)
+
+	// initiateHandlersAndCollectBlocks configures and starts handlers that monitor
+	// for new blocks, change in the mempool and handle chain reorg. It also
+	// initiates data collection for the explorer.
+	initiateHandlersAndCollectBlocks := func() error {
```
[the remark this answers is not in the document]
dmigwi: Just seen this now. I am adding it.

dmigwi:
>The sync status (progress bars) seems to be working very well so far. For a follow-up change, we might want to add a desktop notification for when it completes sync.

Let me see how fast I can be able to a desktop notification.

>I don't think the explorer should be available, regardless of the sync status blocks config setting or block progress, unless the indexing is complete. So this would be a catch up mode, not an initial sync. Is it setup to not allow explorer access if indexing and/or address table spending data updates are not done?

When the indexing is activated the explorer isn't available till the sync is done. I mean once the sync status page is activated it will only be deactivated once the sync is complete.

chappjc:
> When the indexing is activated the explorer isn't available till the sync is done. I mean once the sync status page is activated it will only be deactivated once the sync is complete.

So say you interrupt an initial sync with 500 blocks to go, so there are no indexes yet and the addresses table isn't updated yet.  Then you start dcrdata configured go straight to the explorer for that height.  I'm wondering if there's an exception when there are no indexes so that it must do the progress bars.

Actually, I'm not sure off hand how dcrdata behaves if the sync is almost done but there are no indexes.  Hopefully absence of indexes is recognized and the sync runs as appropriate.

dmigwi:
I get your point fixing it now.

main.go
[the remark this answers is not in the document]
chappjc: I won't push hard for this change, but do notice how I removed `initiateHandlersAndCollectBlocks := func() error` and left the `{`/`{` brackets.  This is what I meant by "syntactic block".  There's really no need for a function, except perhaps for the error handling you did in `fmt.Errorf("RPC client error: %v (%v)", cerr.Error(), cerr.Cause())`.  This does make sense, do it's OK to leave it how you have it... just wanted to make my point clear.

dmigwi:
@chappjc This PR has quite a huge conversation that is super important is there a specific that I may have overlooked?

Here is how desktop notifications look like...

screen shot 2018-10-06 at 00 49 59

main.go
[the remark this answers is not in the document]
dmigwi: ok. I actually did not notice the "syntactic block". Let me remove this `nitiateHandlersAndCollectBlocks := func() error`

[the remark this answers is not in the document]
dmigwi: Done

db/dcrsqlite/sync.go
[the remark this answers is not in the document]
dmigwi: I guess I should remove this part because its already fixed by #720

chappjc:
No I think this is good to merge after removing the rewind exception stuff.  There may be some rough edges here and there, but we'll get this in and move forward.  Good work.

dmigwi:
I have dropped the rewind stuff.

dmigwi:
This is the biggest and the most complicated PR that I think I have raised to date. With your help @chappjc I have understood a lot about the drcdata blockchain syncing, Thank you for your time and help.

Issue #652, opened 2018-08-29
agendas page: redundant time spec


xaur:
+0000 and UTC are redundant in Start Time and Expire time columns. I suggest keeping UTC.

dcrdata v3.1.0-pre+ee54aec

chappjc:
UTC.  I agree.

chappjc:
Resolved by https://github.com/decred/dcrdata/pull/712

PR #712, opened 2018-09-29
agendas page: redundant time spec #652

ademuanthony:
removed redundant time spec in agenda page by updating the `TimeConversion` template function to return a string with the format, `2006-01-02 15:04:05 MST`
https://github.com/decred/dcrdata/issues/652

### DIFF
--- a/explorer/templates.go
+++ b/explorer/templates.go
@@ -251,8 +251,9 @@ func makeTemplateFuncMap(params *chaincfg.Params) template.FuncMap {
 			result = int(a) * b
 			return
 		},
-		"TimeConversion": func(a uint64) (result time.Time) {
-			result = time.Unix(int64(a), 0)
+		"TimeConversion": func(a uint64) (result string) {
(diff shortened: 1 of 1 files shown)


explorer/templates.go
```diff
@@ -251,8 +251,9 @@ func makeTemplateFuncMap(params *chaincfg.Params) template.FuncMap {
 			result = int(a) * b
 			return
 		},
-		"TimeConversion": func(a uint64) (result time.Time) {
-			result = time.Unix(int64(a), 0)
+		"TimeConversion": func(a uint64) (result string) {
+			dateTime := time.Unix(int64(a), 0)
+			result = dateTime.Format("2006-01-02 15:04:05 MST")
```
chappjc: Note that this will only say "UTC" if the host machines clock is on UTC.  To force UTC display (desirable, especially for public machines):

    dateTime := time.Unix(int64(155341541), 0).UTC()

Issue #678, opened 2018-09-11
correct MaximumBlockSize is missing from /parameters


raedah:
MaximumBlockSizes
https://github.com/decred/dcrd/blob/566bb6625b83738cb97ab826d5d77e862d44e92c/chaincfg/mainnetparams.go#L33

chappjc:
This issue should remain open until https://github.com/decred/dcrd/pull/1479 is merged, and dcrdata is updated to use dcrd's new `getblockchaininfo` RPC, which will provide the correct block size.

chappjc:
Resolved by https://github.com/decred/dcrdata/pull/923

PR #715, opened 2018-09-29
show MaximumBlockSizes parameter on the /parameters page

ademuanthony:
Displayed MaximumBlockSizes in the view to fix https://github.com/decred/dcrdata/issues/678

### DIFF
--- a/explorer/explorerroutes.go
+++ b/explorer/explorerroutes.go
@@ -1155,6 +1155,7 @@ func (exp *explorerUI) ParametersPage(w http.ResponseWriter, r *http.Request) {
 	actualTicketPoolSize := int64(cp.TicketPoolSize * cp.TicketsPerBlock)
 	ecp := ExtendedChainParams{
 		Params:               cp,
+		MaximumBlockSize:     cp.MaximumBlockSizes[0],
 		AddressPrefix:        addrPrefix,
 		ActualTicketPoolSize: actualTicketPoolSize,
 	}
--- a/views/parameters.tmpl
+++ b/views/parameters.tmpl
@@ -57,7 +57,7 @@
                             <tr>
                                 <td class="mono text-right pr-2 nowrap p03rem0">MaximumBlockSize</td>
                                 <td class="mono">{{.MaximumBlockSize}} bytes</td>
-                                <td>Maximum sizes of a block that can be generated on the network</td>
+                                <td>Maximum size of a block that can be generated on the network</td>
                             </tr>
                             <tr>
                                 <td class="mono text-right pr-2 nowrap p03rem0">MaxTxSize</td>
--- a/views/parameters.tmpl
+++ b/views/parameters.tmpl
@@ -55,7 +55,7 @@
                                 <td>Whether or not CPU mining is allowed</td>
                             </tr>
                             <tr>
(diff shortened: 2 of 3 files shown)


chappjc:
I do not think we should show the *slice* `MaximumBlockSizes` but rather which max block size is currently applicable to the network.  There are only >1 values when there is a vote to increase the block size.  Only one size is currently applicable.

We can either take `MaximumBlockSizes[0]`, or we can find a way to determine which is presently active.

In terms of wording, I suggest just "Maximum block size."

chappjc:
@davecgh Confirmed that a new RPC would need to be added to make this information easily accessible.  Presently you would need an instance of `github.com/decred/dcrd/blockchain.BlockChain` to get this information.  There is no RPC for this presently, but the suggestion from davecgh is to add one:

> davecgh [2:10 PM]
> Yeah, there probably needs to be a `getchaininfo` added that includes info such as that

That may be too much effort for now (what other fields would go into `getchaininfo`?), although a `getmaxblocksize` would be very simple.  So, either take `MaximumBlockSizes[0]` as the max block size and let it be out of date _very, very_ occasionally, or look into adding a new RPC to dcrd.

chappjc:
See https://github.com/decred/dcrd/issues/1475

ademuanthony:
> Maximum block size

The other labels on this page are in camel case. Should I update them are well?

explorer/explorerroutes.go
```diff
@@ -1155,6 +1155,7 @@ func (exp *explorerUI) ParametersPage(w http.ResponseWriter, r *http.Request) {
 	actualTicketPoolSize := int64(cp.TicketPoolSize * cp.TicketsPerBlock)
 	ecp := ExtendedChainParams{
 		Params:               cp,
+		MaximumBlockSize:     cp.MaximumBlockSizes[0],
```
chappjc: This is not necessary.  Just index it when executing the template.

views/parameters.tmpl
```diff
@@ -56,7 +56,7 @@
                             </tr>
                             <tr>
                                 <td class="mono text-right pr-2 nowrap p03rem0">Maximum block size</td>
-                                <td class="mono">{{.Params.MaximumBlockSize}}</td>
```
chappjc: `{{index .Params.MaximumBlockSizes 0}}`

views/parameters.tmpl
```diff
@@ -54,6 +54,11 @@
                                 <td class="mono">{{.Params.GenerateSupported}}</td>
                                 <td>Whether or not CPU mining is allowed</td>
                             </tr>
+                            <tr>
+                                <td class="mono text-right pr-2 nowrap p03rem0">MaximumBlockSize</td>
+                                <td class="mono">{{.MaximumBlockSize}} bytes</td>
+                                <td>Maximum sizes of a block that can be generated on the network</td>
```
chappjc: please edit the comment appropriately.

explorer/explorerroutes.go
ademuanthony: The reason I added this is because there a possibility for this value to something other than cp.MaximumBlockSizes[0], according to an earlier comment, 

> We can either take MaximumBlockSizes[0], or we can find a way to determine which is presently active.

Although I am unable to determine the block that is currently active, I am thinking the same field will pass that value when it is sorted out

Issue #601, opened 2018-08-10
startup should not assume DBs on main chain


chappjc:
Since there is no reorg notification if not connected to dcrd during the reorg, starting up dcrdata cannot rely on an explicit signal from dcrd.

On startup, dcrdata should first check that the hash of it's best block matches what dcrd says is at that height.  If there is disagreement, dcrdata should self-trigger a reorg.

PR #716, opened 2018-09-30
startup should not assume DBs on main chain #601

ademuanthony:
- Moved out the anonymous function for handling the OnReorganization
-Triggered a reorg notification in the main core if the hash of the best block is not the same as the hash at the same height on dcrd

### DIFF
--- a/main.go
+++ b/main.go
@@ -323,6 +323,28 @@ func mainCore() error {
 	default:
 	}
 
+	// Trigger a reorg notification if the hash of the best block is not the same as the hash at the same height on dcrd
+	func() {
+		bestBlock := baseDB.GetBestBlockSummary()
+		dcrdHash, err := dcrdClient.GetBlockHash(int64(bestBlock.Height))
+		if err != nil{
+			return
+		}
+		if bestBlock.Hash == dcrdHash.String() {
+			return
+		}
+		dcrdBlockHeight, err := baseDB.GetBlockHeight(dcrdHash.String())
+		if err != nil {
+			return
+		}
+		bestBlockHash, err := chainhash.NewHashFromStr(bestBlock.Hash)
+		if err != nil {
+			return
+		}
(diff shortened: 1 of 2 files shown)


chappjc:
Did you test this?  Please always attempt to test your work.  It is complicated with reorganization, but it is all the more important.

Addressing this issue is not quite this simple.  It requires a more complicated solution than what is proposed.   The comment I made in https://github.com/decred/dcrdata/issues/601:
> On startup, dcrdata should first check that the hash of it's best block matches what dcrd says is at that height. If there is disagreement, dcrdata should self-trigger a reorg.
The "self-trigger a reorg" is what you're attempting here, but manually running the notification handler is not what I meant.

I would not suggest taking this issue if you are not intimately familiar with:
- how reorganization works on the Decred blockchain,
- how dcrd's notifications work, especially with regard to reorganization, and
- how dcrdata handles the reorganization in response to the notifications.

A reorg notification does call the code you factored out into a separate function, but that is just the beginning of the reorganization process.  It informs the various dcrdata chain monitors that a reorganization is going to happen that will change the current best block from old height/hash to new height/hash.  That's all it does.

The rest of the process happens as dcrd sends new block notifications for the new mainchain (old side chain) blocks one-by-one.  Since it already got the reorg notification, it knows to treat these blocks differently.  Only when the block corresponding to the new chain tip is received in a new block notification can the dcrdata part of the reorganization happen.

The dcrdata part of the reorganization generally involves "disconnecting" the previous main chain blocks back to the common ancestor block, connecting the new main chain blocks one-by-one, and exiting reorganization mode when they are all connected.

ademuanthony:
I will need to get more into the way things work under the hood

> how reorganization works on the Decred blockchain,
how dcrd's notifications work, especially with regard to reorganization, and
how dcrdata handles the reorganization in response to the notifications.

Issue #338, opened 2018-02-07
Display information about orphan blocks and reorgs


raedah:
Research how we can display information about orphan blocks and sidechains.

chappjc:
This may be doable with SPV, when that's added to dcrd.  Right now all an RPC client knows about is the main chain and what it once though was the main chain prior to reorg.

raedah:
related to https://github.com/dcrdata/dcrdata/issues/56

chappjc:
Now doable with the recently-added `getchaintips` RPC:

````
getchaintips

Returns information about all known chain tips the in the block tree.

The statuses in the result have the following meanings:
active: The current best chain tip.
invalid: The block or one of its ancestors is invalid.
headers-only: The block or one of its ancestors does not have the full block data available which also means the block can't be validated or connected.
valid-fork: The block is fully validated which implies it was probably part of the main chain at one point and was reorganized.
valid-headers: The full block data is available and the header is valid, but the block was never validated which implies it was probably never part of the main chain.

Arguments:
None

Result:
[{
 "height": n,       (numeric) The height of the chain tip
 "hash": "value",   (string)  The block hash of the chain tip
 "branchlen": n,    (numeric) The length of the branch that connects the tip to the main chain (0 for the main chain tip)
 "status": "value", (string)  The status of the chain (active, invalid, headers-only, valid-fork, valid-headers)
},...]
````

chappjc:
This can now be done properly given:

1. Side chain recording in https://github.com/decred/dcrdata/pull/580
2. The `getchaintips`.

I'm assigning myself.

chappjc:
PR https://github.com/decred/dcrdata/pull/654 uses new queries to show side chain blocks.  These blocks are presently only known because they were previously main chain prior to a reorg.

The next step in this task is to use the `getchaintips` RPC to get information about side chain blocks known to dcrd.  Some of these may intersect with those recorded by dcrdata.  Others returned by dcrd that were not already recorded by dcrdata may be then be retrieved (hopefully).  My understanding is that transactions in blocks with `"status"` of `"valid-headers"` or `"valid-fork"` may be retrieved for processing.

chappjc:
The best fix possible is implemented in https://github.com/decred/dcrdata/pull/717.  With that, any side chain blocks may be loaded in dcrdata.

[...]

PR #717, opened 2018-09-30
Full side chain import from dcrd on startup

chappjc:
This adds the ability to get all known side chains from dcrd and import all side chain blocks into the aux (dcrpg) database.

- [x] dcrd/rpcclient: add `GetChainTips` to RPC client (https://github.com/decred/dcrd/pull/1469)
- [x] rpcutils: add SideChainFull to get all blocks in a side chain
- [x] dcrpg: Modify StoreBlock and storeTxns to work correctly with side chain blocks.
- [x] dcrsqlite and dcrpg: Check for preexistence of side chain blocks in the DBs.
- [x] Capability to import all side chain blocks into dcrsqlite and dcrpg on start up.
- [x] Update/verify `/side` page for imported blocks, as currently it only shows side chain blocks from reorgs.

This resolves issue https://github.com/decred/dcrdata/issues/338.

For a follow-up PR:  Do the check and import periodically *while running* so that new forks that were not reorganized may be added without having to restart dcrdata.

### DIFF
--- a/api/types/apitypes.go
+++ b/api/types/apitypes.go
@@ -285,8 +285,15 @@ type BlockDataBasic struct {
 	StakeDiff  float64 `json:"sdiff,omitemtpy"`
 	Time       int64   `json:"time,omitemtpy"`
 	NumTx      uint32  `json:"txlength,omitempty"`
-	//TicketPoolInfo
-	PoolInfo TicketPoolInfo `json:"ticket_pool,omitempty"`
+	// TicketPoolInfo may be nil for side chain blocks.
+	PoolInfo *TicketPoolInfo `json:"ticket_pool,omitempty"`
+}
+
+// NewBlockDataBasic constructs a *BlockDataBasic with pointer fields allocated.
+func NewBlockDataBasic() *BlockDataBasic {
+	return &BlockDataBasic{
+		PoolInfo: new(TicketPoolInfo),
+	}
 }
 
 // BlockExplorerBasic models primary information about block at height Height
@@ -346,7 +353,15 @@ type StakeInfoExtended struct {
 	StakeDiff        float64              `json:"stakediff"`
 	PriceWindowNum   int                  `json:"window_number"`
 	IdxBlockInWindow int                  `json:"window_block_index"`
-	PoolInfo         TicketPoolInfo       `json:"ticket_pool"`
+	PoolInfo         *TicketPoolInfo      `json:"ticket_pool"`
+}
+
+// NewStakeInfoExtended constructs a *StakeInfoExtended with pointer fields
+// allocated.
+func NewStakeInfoExtended() *StakeInfoExtended {
+	return &StakeInfoExtended{
+		PoolInfo: new(TicketPoolInfo),
+	}
 }
 
 // StakeInfoExtendedEstimates is similar to StakeInfoExtended but includes stake
@@ -356,7 +371,7 @@ type StakeInfoExtendedEstimates struct {
 	StakeDiff        StakeDiff            `json:"stakediff"`
 	PriceWindowNum   int                  `json:"window_number"`
 	IdxBlockInWindow int                  `json:"window_block_index"`
-	PoolInfo         TicketPoolInfo       `json:"ticket_pool"`
+	PoolInfo         *TicketPoolInfo      `json:"ticket_pool"`
 }
 
 // MempoolTicketFeeInfo models statistical ticket fee info at block height
--- a/blockdata/blockdata.go
+++ b/blockdata/blockdata.go
@@ -27,7 +27,7 @@ type BlockData struct {
 	FeeInfo          dcrjson.FeeInfoBlock
 	CurrentStakeDiff dcrjson.GetStakeDifficultyResult
 	EstStakeDiff     dcrjson.EstimateStakeDiffResult
-	PoolInfo         apitypes.TicketPoolInfo
+	PoolInfo         *apitypes.TicketPoolInfo
 	ExtraInfo        apitypes.BlockExplorerExtraInfo
 	PriceWindowNum   int
 	IdxBlockInWindow int
@@ -97,7 +97,7 @@ func (b *BlockData) ToBlockExplorerSummary() apitypes.BlockExplorerBasic {
 
 // Collector models a structure for the source of the blockdata
 type Collector struct {
-	mtx          sync.Mutex
+	sync.Mutex
 	dcrdChainSvr *rpcclient.Client
 	netParams    *chaincfg.Params
 	stakeDB      *stakedb.StakeDatabase
@@ -107,7 +107,6 @@ type Collector struct {
 func NewCollector(dcrdChainSvr *rpcclient.Client, params *chaincfg.Params,
 	stakeDB *stakedb.StakeDatabase) *Collector {
 	return &Collector{
-		mtx:          sync.Mutex{},
 		dcrdChainSvr: dcrdChainSvr,
 		netParams:    params,
 		stakeDB:      stakeDB,
@@ -142,6 +141,7 @@ func (t *Collector) CollectAPITypes(hash *chainhash.Hash) (*apitypes.BlockDataBa
 func (t *Collector) CollectBlockInfo(hash *chainhash.Hash) (*apitypes.BlockDataBasic,
 	*dcrjson.FeeInfoBlock, *dcrjson.GetBlockHeaderVerboseResult,
 	*apitypes.BlockExplorerExtraInfo, *wire.MsgBlock, error) {
+	// Retrieve block from dcrd.
 	msgBlock, err := t.dcrdChainSvr.GetBlock(hash)
 	if err != nil {
 		return nil, nil, nil, nil, nil, err
@@ -149,6 +149,9 @@ func (t *Collector) CollectBlockInfo(hash *chainhash.Hash) (*apitypes.BlockDataB
 	height := msgBlock.Header.Height
 	block := dcrutil.NewBlock(msgBlock)
 	txLen := len(block.Transactions())
+
+	// Coin supply and block subsidy. If either RPC fails, do not immediately
+	// return. Attempt acquisition of other data for this block.
 	coinSupply, err := t.dcrdChainSvr.GetCoinSupply()
 	if err != nil {
 		log.Error("GetCoinSupply failed: ", err)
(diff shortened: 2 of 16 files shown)


chappjc:
Tested on mainnet.  Successfully (apparently) imported 19 blocks from the 19 side chains known to the dcrd node used.  Blocks show up on the `/side` path and are all individually viewable.

Needs testing on testnet, and on other longer running nodes.

main.go
```diff
@@ -400,14 +400,78 @@ func mainCore() error {
...
+				blockHash, err := chainhash.NewHashFromStr(hash)
+				if err != nil {
+					log.Errorf("Invalid block hash %s: %v.", hash, err)
+					continue
+				}
+
+				// Collect block data.
+				blockData, msgBlock, err := collector.CollectHash(blockHash)
+				if err != nil {
+					// Do not quite if unable to collect side chain block data.
```
dmigwi: quite -> quit

rpcutils/rpcclient.go
```diff
@@ -261,6 +276,50 @@ func sideChainTips(allTips []dcrjson.GetChainTipsResult) (sideTips []dcrjson.Get
...
+
+	// Reverse side chain order so that last element is tip.
+	reverseStringSlice(sideChain)
+
+	return sideChain, nil
+}
+
+func reverseStringSlice(s []string) {
+	N := len(s)
+	for i := 0; i <= N/2-1; i++ {
```
dmigwi: I think its less confusing when you use bracket where you do multiple different operations;
 N/2-1 -> (N/2)-1

api/types/apitypes.go
```diff
@@ -285,8 +285,15 @@ type BlockDataBasic struct {
...
 	Time       int64   `json:"time,omitemtpy"`
 	NumTx      uint32  `json:"txlength,omitempty"`
-	//TicketPoolInfo
-	PoolInfo TicketPoolInfo `json:"ticket_pool,omitempty"`
+	// TicketPoolInfo may be nil for side chain blocks.
+	PoolInfo *TicketPoolInfo `json:"ticket_pool,omitempty"`
+}
+
+// NewBlockDataBasic constructs a *BlockDataBasic with all fields allocated.
+func NewBlockDataBasic() *BlockDataBasic {
```
dmigwi: "with all fields allocated" is confusing as you are just initializing `PoolInfo` field.

api/types/apitypes.go
```diff
@@ -346,7 +353,15 @@ type StakeInfoExtended struct {
 	StakeDiff        float64              `json:"stakediff"`
 	PriceWindowNum   int                  `json:"window_number"`
 	IdxBlockInWindow int                  `json:"window_block_index"`
-	PoolInfo         TicketPoolInfo       `json:"ticket_pool"`
+	PoolInfo         *TicketPoolInfo      `json:"ticket_pool"`
+}
+
+// NewStakeInfoExtended constructs a *StakeInfoExtended with all fields
+// allocated.
```
dmigwi: Same to this "all fields allocated" is confusing unless further explained.

blockdata/blockdata.go
```diff
@@ -142,31 +141,39 @@ func (t *Collector) CollectAPITypes(hash *chainhash.Hash) (*apitypes.BlockDataBa
...
 	}
 	height := msgBlock.Header.Height
 	block := dcrutil.NewBlock(msgBlock)
 	txLen := len(block.Transactions())
+
+	// Coin supply and block subsidy
 	coinSupply, err := t.dcrdChainSvr.GetCoinSupply()
 	if err != nil {
 		log.Error("GetCoinSupply failed: ", err)
+		// Keep going to attemp acquisition of other block data.
```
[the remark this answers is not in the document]
dmigwi: This comment might also not be placed at the incorrect position.

db/dcrpg/pgblockchain.go
```diff
@@ -180,6 +181,62 @@ func (db *ChainDBRPC) Store(blockData *blockdata.BlockData, msgBlock *wire.MsgBl
 	return db.ChainDB.Store(blockData, msgBlock)
 }
 
+// MissingSideChainBlocks identifies side chain blocks that are missing from the
+// DB. Side chains known to dcrd are listed via the getchaintips RPC. Each block
```
dmigwi: Each block's presence is checked in the postgres DB

db/dcrpg/pgblockchain.go
```diff
@@ -1826,10 +1885,15 @@ func (pgb *ChainDB) StoreBlock(msgBlock *wire.MsgBlock, winningTickets []string,
 	// Convert the wire.MsgBlock to a dbtypes.Block
 	dbBlock := dbtypes.MsgBlockToDBBlock(msgBlock, pgb.chainParams)
 
-	// Get the previous winners (stake DB pool info cache have this info)
+	// Get the previous winners (stake DB pool info cache have this info). If
+	// the previous block is side chain, stakedb will not have the
+	// winners/validators. Since Validators are only used to identify misses in
+	// InsertVotes, we will leave the Validators empty and assume there are no
+	// misses. TODO: ensure this wouldn't break reorg if this block becomes main
+	// chain at some point.
```
dmigwi: If a block that is in a side chain can be brought back as a mainchain block what is the likelihood of this index being violated? 
```golang
IndexBlockTableOnHash = `CREATE UNIQUE INDEX uix_block_hash ON blocks(hash);`
```

db/dcrpg/pgblockchain.go
```diff
@@ -1912,6 +1976,23 @@ func (pgb *ChainDB) StoreBlock(msgBlock *wire.MsgBlock, winningTickets []string,
 	lastBlockHash := msgBlock.Header.PrevBlock
 	// Only update if last was not genesis, which is not in the table (implied).
 	if lastBlockHash != zeroHash {
+		// Ensure previous block has the same main/sidechain status. If the
+		// curent block being added is side chain, do not invalidate the
```
dmigwi: curent -> current

main.go
```diff
@@ -334,30 +333,31 @@ func mainCore() error {
 
 	blockDataSavers = append(blockDataSavers, explore)
 
-	// Sync up with the blockchain
+	// Coordinated sync up both sqlite and postgres DBs with the network. This
```
dmigwi: Coordinated -> Coordinate. 
feels better.
"Coordinate the sync of both sqlite and postgres DBs with the network..."

If one day we decided to change the auxiliary db from Postgres to maybe Mysql or even a graph database, shall we be forced to rename the all Postgres occurrences to that specific db then. IMO I think we should use "auxiliary db" when referencing Postgres db now.

main.go
```diff
@@ -334,30 +333,31 @@ func mainCore() error {
...
-		// available blocks to the baseDB. In return, baseDB maintains a
+		// Now that stakedb is either catching up or waiting for a block, start
+		// the auxDB sync, which is the master block getter, retrieving and
+		// making available blocks to the baseDB. In return, baseDB maintains a
 		// StakeDatabase at the best block's height.
 		go auxDB.SyncChainDBAsync(pgSyncRes, smartClient, quit,
 			updateAddys, updateVotes, newPGInds)
 
-		// Wait for the results
+		// Wait for the results from both DBs.
```
dmigwi: They are actually three database especially when full mode is running.

main.go
```diff
@@ -400,14 +400,78 @@ func mainCore() error {
 		return nil
 	}
 
-	// Register for notifications from dcrd
+	// Ensure all side chains known by dcrd are also present in postgresql and
```
dmigwi: How about "auxiliary db" instead of "postgresql"?

db/dcrpg/pgblockchain.go
```diff
@@ -180,6 +181,62 @@ func (db *ChainDBRPC) Store(blockData *blockdata.BlockData, msgBlock *wire.MsgBl
...
+		// For each block in the side chain, check if it already stored.
+		for is := range sideChain {
+			// Check for the block hash in the DB.
+			sideHeightDB, err := db.BlockHeight(sideChain[is])
+			if err != nil && err != sql.ErrNoRows {
+				// Unexpected error
+				log.Debugf("Failed to retrieve block %s.", sideChain[is])
+			} else if err == nil {
+				// This block is alredy in the DB.
+				log.Tracef("Found block %s in postgres at height %d.",
```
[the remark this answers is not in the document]
dmigwi: I am talking about the comment

api/types/apitypes.go
dmigwi: "with valid pointer fields" feels better unless you've got a better option.

db/dcrpg/pgblockchain.go
chappjc: Being "brought back" does not involve the hash changing, so it would be violated if we tried to re-insert it, but we update the row with that hash so it won't be violated in a reorg.

The issue I was worried about is that there is no ticket pool info data ( for blocks imported this way (as side chain to start), and I was not certain that would be an issue if such a block somehow became mainchain via a reorg.

So, talking this through, the primary "pool info" not available to imported side chain blocks are the winning tickets (that vote in the next block) and the validators (the winning tickets that should be spent on votes in the block).  The validators are used in InsertVotes to identify missed votes as validators that did not vote.  Imported side chain blocks will just have no misses recorded in the misses table.  But if that block became main chain, the reorganization process will actually connect these blocks in stakedb, thus providing this information when `dcrpg.(*ChainMonitor).switchToSideChain` calls `(*ChainDB).StoreBlock` again, this time with `isMainchain=true`.  This time, `InsertVotes` will have the validators and will thus be able to insert any new misses.

To summarize, a reorganization will involve connecting the blocks in stakedb, allowing the validators to be known, and all the inserts will be run again, including the misses that were not known while the block was an imported side chain block.

main.go
chappjc: Referring to the agenda db?  `getSyncd` only runs the sync routines of `wiredDB` and `ChainDB`.  Note that this keeps the `StakeDatabase` at the same height as these too, but that is managed by `wiredDB` so I'm only mentioning the two DBs with sync routines.
Do we need to be doing something with agenda db on startup, while catching up?

rpcutils/rpcclient.go
chappjc: Personally I think this order of operations is pretty clear, but it's good to help readability if it doesn't affect code semantics.

main.go
dmigwi: I would suggest we ignore agenda db for now as we have a lot of loose end that will be addressed when the new version of `getvoteinfo` rpc endpoint on dcrd is finalized.

dmigwi: But updates are always checked on startup https://github.com/decred/dcrdata/blob/master/main.go#L288-L294

db/dcrpg/pgblockchain.go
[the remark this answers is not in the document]
chappjc: oh, lol :)

chappjc:
Let me know if the last few commits address the issues you've identified. Thanks for the review.

chappjc:
This resolves issue https://github.com/decred/dcrdata/pull/717.

PR #689, opened 2018-09-17
Block display bar for newhome page enhancement 

Jujhar:
Added @xaur's bar (Please update @denyszayets, or send back what you have)
screenshot_2018-09-16 decred block edxplorer by dcrdata org

screenshot_2018-09-16 decred block edxplorer by dcrdata org 1

Addresses #451

PR #718, opened 2018-10-03
Visual Blocks for new home page design

itswisdomagain:
This PR is related to the new home design initiative in issues #451 and #453.
This integrates with the rest of the work recently merged into master in #660.
The initial work for this was done in #535 but was incomplete. It was then picked up in #689, which this PR picked up and took to finish.

Each block is displayed as a collection of 4 rows of visual data. Hovering over each cell in each row reveals information about that cell/block.

new blocks display

- The first row represents block rewards (PoW/PoS/Dev/TxFees)
- Second row represents votes (usually between 3 and 5)
- Third row shows Tickets and Revocations (if present)
- Fourth row shows regular transactions

This PR only applies to the right side visual blocks. Improvements to the left panel can be fixed in another PR.

**EDIT (Design and Implementation):**
As highlighted in #451, the goal is to make the explorer home page "more average-user friendly". This PR is an attempt to make the **Decred Blockchain** section more visually appealing while displaying relevant information.

The implementation is as follows:
- load data for mempool and most recent 30 blocks
- only data required to display the blocks as defined below is loaded to prevent unused data wasting memory space in the dom and causing performance issues (the `NextHome()` handler in explorerroutes.go was modified to retrieve just the required data)
- each block is displayed as a rectangular box with 5 parallel, horizontal rows consisting of small cells
- the first row displays the block's height and the block time, with a link to the block/block-height page
- the second row shows the block's MiningFee and subsidy
- the subsidy cells (pow, pos and fund) have backgrounds that are as wide as the ratio of `the number of votes in the block` to `5`. An example is if there are 3 votes, the background for each of those cells would only fill 3/5th the available width of the cell
- hovering over the cells in the first row shows the amount for PoW, PoS, fund and block MiningFee respectively
- clicking on any of those cells opens the transaction page for the block's reward tx
- the third row shows the block votes, 5 pre-defined cells, if there are less than 5 votes in the block, there'd be some visible white space on the right end of the row
- hovering over the cells in the votes row shows the amount in the vote tx while clicking on the cells opens the transaction page for the vote tx
- the fourth row displays the tickets and revocations of the block in at most 50 cells. if there are more than 50 tickets in the block, only the first 30 are shown as cells leaving space on the right end of the row to display how many more tickets and revs are in the block
- the fifth row displays the regular transactions in at most 50 cells, following same pattern as the tickets+revs row
- each cell in the tickets/rev/regular tx row has an opacity proportional to the number of inputs and outputs in the tx
- hovering over each cell in the tickets/rev/regular tx row shows the total amount of the tx as well as the number of inputs and outputs while clicking on each cell opens the transaction page for that tx


chappjc:
That is looking very nice.  Thanks for the descriptive PR.

dmigwi:
Here is my initial review of the UI.  I use photoshop to highlight places with issues.
The work is awesome.

nexthome-page-review

**Box 1**
This is wrong.. The search Icon should never occupy the search bar space.

**Box 2 and Box 3**
- Lack of symmetry on the left and the right side doesn't look okay to me. 
- The spacing between left and right blocks seems to be affected by the screen size unlike the spacing between the top and bottom blocks.
- some text on box 2 is overlapping because of the reduced box 2 size. This is making items look misplaced. Take a look at the PoS reward change. Consider changing the font size in box 2 if width can't be increased.
- The view Blockchain History bar is extending past the blocks coverage total width.
- I like the Idea that when the screen display changes box 3 height changes too but sadly box 2 isn't changing.  Would you mind implementing something similar on box 2?
- Alternatively you could ensure that box 3 never retreats to a total height less than box 2. box 3 height should be at least be greater than or equal box 2 height.
- Tickets on the mempool seem to have a defined area size, be advised that tickets in the mempool can get to 500 and plus. Does your mempool implementation consider that?
- Concerning the new blocks display, some help link with more details about what each part of the block represents should be added somewhere on the screen.
- You seem to have set a static block width but don't mention at all what the value in DCR represents, @denyszayets this your part.

Will look at the code later.

raedah:
-Box 2 should not be shrunken, it should be kept at its normal size so if it is too small this can be corrected. Box 2 can not easily be resized with the page though because the data in it is static and it would simply add in large useless white space. The extra screen space can be utilized with more room to show visual blocks though. Perhaps this can be resolved with minimum requirements of extra space available, so that unless the larger threshold is reached, the page would simply maintain symmetry and still maintain the left and right side page borders.
-Blockchain history bar size issue does need to be resolved.
-Regarding help link with details, this should all be included already in the hover over details. Same for the DCR values, in hover over.

public/css/main.css
```diff
@@ -1,15 +1,15 @@
 ﻿/*icons*/
+
```
dmigwi: Adding new characters to the various classes in my opinion results to an increase in the file size that is loaded to the frontend. Some of this css files may run to quite a number of classes. We could argue that added size isn't that huge but taking into consideration network latency issues and speed of loading pages, I think we should take every opportunity we get to make our pages faster. In this case I think we would be better off without the new line characters in the css file.

public/css/nexthome.css
```diff
@@ -1,530 +1,742 @@
...
+header .container {
     display: flex;
-    justify-content: space-between;
     align-items: center;
-    height: 40px; }
-  header .right {
+    justify-content: space-between;
+    height: 40px;
+}
+
```
dmigwi: Same comment made in the previous css file applies here too.

public/js/controllers/mempool.js
```diff
@@ -73,13 +73,15 @@
...
                 var m = JSON.parse(evt)
+                console.log("update", evt);
                 this.renderLatestTransactions(m.latest, false)
                 $(this.numTicketTarget).text(m.num_tickets)
                 $(this.numVoteTarget).text(m.num_votes)
                 keyNav(evt, false, true)
                 ws.send("getmempooltxs", "")
             });
             ws.registerEvtHandler("getmempooltxsResp", (evt) => {
+                console.log("resp", evt);
```
dmigwi: This logging aren't necessary unless you were using them to test. @chappjc Why do we even log websocket responses on the frontend?

public/js/controllers/mempool.js
```diff
@@ -73,13 +73,15 @@
             })
             ws.registerEvtHandler("mempool", (evt) => {
                 var m = JSON.parse(evt)
+                console.log("update", evt);
```
dmigwi: Not important

public/js/controllers/nexthome-socket-update.js
```diff
@@ -0,0 +1,51 @@
+function handleMempoolUpdate(evt) {
+	const mempool = JSON.parse(evt);
+	if (mempool.Votes.length === 0) {
+		console.log('mempool update discarded, can votes be 0?');
```
dmigwi: Remove please?

views/extras.tmpl
```diff
@@ -46,9 +46,7 @@
 <header>
 <div class="top-nav">
     <div class="container">
-        <div class="logo">
-            <a class="no-underline" href="/"><img src="/images/logo-2018.svg" alt="dcr-icon"></a>
-        </div>
+        <a class="no-underline logo" href="/"><img src="/images/logo-2018.svg" alt="dcr-icon"></a>
```
dmigwi: Some of this css especially in the nexthome is heavily chained together. Changing the order without changing the css used will affect the items alignment. In this case this css will no longer work..
```css
.container .logo a img {
        width: 70px !important;
    }
```
And there are many more just like this.

views/nexthome.tmpl
```diff
@@ -2,11 +2,35 @@
 <!DOCTYPE html>
 <html lang="en">
 {{ template "html-head" "Decred Block Explorer by dcrdata.org"}}
+<script src="https://unpkg.com/tippy.js@2.5.4/dist/tippy.all.min.js"></script>
```
dmigwi: You could download this js file and added to /public/js folder in the project.

views/nexthome.tmpl
```diff
@@ -2,11 +2,35 @@
...
+<style>
+.block-rows > div > span > a {
+    display: block;
```
dmigwi: You could move this css to one of the files.

views/nexthome.tmpl
```diff
@@ -2,11 +2,35 @@
...
+    color: #26323d;
+    box-shadow: 0 0 20px 4px rgba(154,161,177,.15), 0 4px 80px -8px rgba(36,40,47,.25), 0 4px 4px -2px rgba(91,94,105,.15);
+    background-color: #fff;
+}
+.tippy-popper[x-placement^=top] .tippy-arrow {
+    border-top-color: white;
+}
+</style>
+<script type="text/javascript" src="/js/controllers/blocks.js"></script>
+<script type="text/javascript" src="/js/controllers/nexthome-socket-update.js"></script>
```
dmigwi: You will need to add `data-turbolinks-eval="false"` to your external js imports. This ensures that your js file is only loaded once and then reused by the browser when needed. i.e.
```html
...
<script type="text/javascript" src="/js/controllers/blocks.js" data-turbolinks-eval="false"></script>
...
```

chappjc:
@itswisdomagain, bump.  Just a friendly reminder that there is an open review that requires changes.

public/css/main.css
itswisdomagain: Could we minify CSS and JS files before deployment too?

public/js/controllers/mempool.js
itswisdomagain: Found those comments in the source code I inherited. Will take them all out. Didn't really like seeing them either.

itswisdomagain:
UPDATE:
The current implementation actually shows ALL the tickets on the mempool and revocations (if any exists). This can look bad for high numbers of mempool tickets. I'm thinking of capping tickets display at 30 tickets. Any other thoughts?

views/extras.tmpl
itswisdomagain: Well, the CSS does need some work. I found this and similar styles in the code I built off from. I'm suggesting doing CSS cleanup in a separate PR

itswisdomagain:
UPDATE:
After conversing with @raedah, this is the implementation for cases where mempool tickets/transactions are more than 50

screen shot 2018-10-12 at 4 11 02 am

views/extras.tmpl
```diff
@@ -194,36 +194,26 @@
         console.log('Starting websocket');
         console.log("createdWebSocket", uri, ws)
         ws.registerEvtHandler("open", function() {
```
dmigwi: Just for clarification, this console logs are used especially in debugging. That said some of the logging available now are fairly useless and some are too important to be deleted. 
We could do away with quite a number of them but logging for New Blocks and New Transactions notifications should be maintained.

views/extras.tmpl
```diff
@@ -194,36 +194,26 @@
...
-            console.log("ping. users online: ", evt)
-            //ws.send("pong", "copy")
-        });
         ws.registerEvtHandler("decodetxResp", function(evt) {
-            console.log("Got message: ", evt);
             $("#decode_header").text("Decoded tx");
             $("#decoded_tx").text(evt);
         })
         ws.registerEvtHandler("sendtxResp", function(evt) {
-            console.log("Got message: ", evt);
```
dmigwi: Keep this

views/extras.tmpl
```diff
@@ -194,36 +194,26 @@
...
             $("#decode_header").text("Sent tx");
             $("#decoded_tx").text(evt);
         })
         var updateBlockData = function (event) {
-            console.log("Received newblock message", event);
```
dmigwi: And this. For now..

public/js/controllers/mempool.js
dmigwi: Kindly delete this.

public/js/controllers/mempool.js
dmigwi: You still need to do away with this

dmigwi:
@itswisdomagain Do rebase your PR with the current master version and then test how https://github.com/decred/dcrdata/pull/710 will affect the nexthome page. 
The above PR basically tries to load the web server soonest possible.

itswisdomagain:
@dmigwi Thanks again for the review. I've effected the requested changes wrt console logging and also rebased with master.

I did create a new branch off PR #710 and merged this PR #718 into the created branch. Resolved the merge conflicts (basically retained all the changes from PR #710 except for changes to the nexthome.tmpl file). The new branch is [here](https://github.com/itswisdomagain/dcrdata/tree/pr710_718). There was an error while trying to run it though, that's when I realized PR #710 was already merged into master and there was no longer a need to see how it plays with the changes in this PR.

explorer/websockethandlers.go
```diff
@@ -95,7 +129,47 @@ func (exp *explorerUI) RootWebsocket(w http.ResponseWriter, r *http.Request) {
 
 				case "getmempooltxs":
 					exp.MempoolData.RLock()
-					msg, err := json.Marshal(exp.MempoolData)
+
```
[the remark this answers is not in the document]
chappjc: Unfortunately, this defeats the purpose of the lock.  `exp.MempoolData` is a pointer.  So when you read the data via `mempoolData`, it's the same data the lock was trying to protect.

If all the code that is using the data is too slow (e.g. `GetExplorerTx` accesses the DB), you will have to make a deep copy of the mempool data, not just of the pointer.  This includes a deep copy of the transaction slices since copying the mempool data struct does not allocate new underlying buffers for the slice data, they will be shared with the old and new slices.

public/css/nexthome.css
```diff
@@ -644,99 +570,70 @@ footer .container .right span.green {
...
     .container {
         max-width: 1280px !important;
-    }
-    .container .logo a img {
+    }    .container .logo a img {
         width: 100px !important;
-    }
-}
-
+    }}
```
chappjc: This was much more legible as:

```css
@media (min-width: 1400px) {
     .container {
         max-width: 1280px !important;
    }
     .container .logo a img {
         width:100px !important;
    }
}
```

The byte savings is minimal.

public/css/nexthome.css
```diff
@@ -644,99 +570,70 @@ footer .container .right span.green {
...
+    }    .container .logo a img {
         width: 70px !important;
-    }
-    .col-md-6 {
+    }    .col-md-6 {
         flex: 0 0 50%;
         max-width: 100%;
-    }
-    .col-sm-4 {
+    }    .col-sm-4 {
```
chappjc: same all over the CSS files.  I get removing the numerous empty lines, but starting a new style on the same line like this makes it far less legible.

public/js/controllers/blocks.js
```diff
@@ -45,7 +50,8 @@ function displayBlocks() {
     const blockHtmlElements = blocks.slice(1).map(newBlockHtmlElement).join("\n");
     blocksHolder.append(blockHtmlElements);
 
-    figureNumberOfBlocksToDisplay();
+    // allow time for left panel to load in order to properly estimate available space on the right
+    setTimeout(figureNumberOfBlocksToDisplay, 500);
```
chappjc: This feels like an anti-pattern.  @gozart1 Is there a better way to handle this? Something with `async` perhaps?

views/extras.tmpl
```diff
@@ -194,36 +194,26 @@
         console.log('Starting websocket');
         console.log("createdWebSocket", uri, ws)
         ws.registerEvtHandler("open", function() {
-            console.log('Connected')
```
chappjc: Please keep this.

views/extras.tmpl
```diff
@@ -194,36 +194,26 @@
...
             updateConnectionStatus('Connected', true);
         });
 
         ws.registerEvtHandler("close", function() {
-            console.log('Disconnected')
```
chappjc: Please keep this.

views/extras.tmpl
```diff
@@ -194,36 +194,26 @@
...
             updateConnectionStatus('Disconnected', false);
         });
 
         ws.registerEvtHandler("error", function(evt) {
-            console.log("WebSocket error:", evt);
```
chappjc: Please keep this.

views/extras.tmpl
```diff
@@ -194,36 +194,26 @@
...
             updateConnectionStatus('Disconnected', false);
         });
 
-        ws.registerEvtHandler("ping", function(evt) {
-            console.log("ping. users online: ", evt)
```
chappjc: Please keep this.  It can be made into Verbose/Debug level however.

public/js/controllers/blocks.js
```diff
@@ -181,6 +187,7 @@ function makeTicketAndRevoctionElements(tickets, revocations) {
         totalDCR += ticket.Total;
         return makeTxElement(ticket, "block-ticket", "Ticket");
     });
+    ticketElements.splice(30);
```
chappjc: What's the significance of 30?

public/css/nexthome.css
```diff
@@ -1,530 +1,649 @@
...
+  width: 100%;
+  display: flex;
+  flex-wrap: nowrap;
+  align-items: stretch;
+  border-top: 1px solid white;
+}
+.decBlockchainWrapper .decredblockWrap .decredblock .block-rows > *:first-child {
+    border-top: none;
+}
+.decBlockchainWrapper .decredblockWrap .decredblock .block-rows > * > * {
```
gozart1: In general wildcard `*` selectors should be avoided in css. They are mostly just used for doing global resets to smooth out browser base-level styling/layout differences. (Eg https://www.paulirish.com/2012/box-sizing-border-box-ftw/ )

Also, I'm not a huge fan of long nested chains like this, since they are less performant and brittle in the face of markup changes. It's better to use a descriptive class name in the markup instead of a long nested chain of classes and dom traversal operators(the > are also not efficient because the browser can't reason about them from the bottom up). While css is smoking fast these days, css cruft builds up over time and still makes a noticeable difference in user experience on mobile and other resource constrained clients to keep an eye on performance best practices.

Being liberal with classes in your markup also makes it easier to understand and change things, since complexity is localized and not co-mingled with other elements.

Also, I would drop "dec" and "decred" from the classnames, I don't see how that adds any value, it's just dead weight.

public/css/nexthome.css
```diff
@@ -1,530 +1,649 @@
...
+        margin-right: 1% !important;
+        min-height: fit-content;
+    }    .dcr-icon {
+        width: 100px;
+    }    .dcr-icon-container {
+        width: 100px;
+    }
+}
+@media only screen and (max-width: 576px) {
+    .decBlockchainWrapper {
```
gozart1: This formatting is totally non-standard. 
Please just format it like this
https://css-tricks.com/different-ways-to-format-css/#article-header-id-0
Since that's by far the most common.

public/js/controllers/nexthome-socket-update.js
```diff
@@ -0,0 +1,46 @@
+function handleMempoolUpdate(evt) {
+	const mempool = JSON.parse(evt);
```
gozart1: Would be great to keep the javascript indentation consistent ( 4 spaces )

public/js/controllers/nexthome-socket-update.js
```diff
@@ -0,0 +1,46 @@
...
+function handleBlockUpdate(block) {
+    const trimmedBlockInfo = trimBlockInfo(block);
+    blocks.push(trimmedBlockInfo);
+    $(newBlockHtmlElement(trimmedBlockInfo)).insertAfter($('.blocks-holder > *:first-child'));
+    $('.blocks-holder > *:last-child').remove();
+    setupTooltips();
+}
+
+function trimTxInfo(txs) {
+	const trimmedTxs = [];
```
gozart1: Would just use `map` here

public/js/controllers/blocks.js
itswisdomagain: @gozart1 any thoughts here?

public/js/controllers/nexthome-socket-update.js
itswisdomagain: Totally missed that

gozart1:
When I try running this, /nexthome crashes chrome. ( CPU maxes out and never stops , so sounds like some js caught in an infinite loop)
Any ideas what could be happening here? 

I just did a fresh database sync on master, then switched to this PR
I'm on the latest commit 79dcb9c8eceb24f75dbfd32704c91413c2a48c8e

crash

itswisdomagain:
I think it's the issue @chappjc raised about the possible anti-pattern in the `figureNumberOfBlocksToDisplay()` method of public/js/controllers/blocks.js. I had similar issues at a time when the amount of data returned from the database was much. That's what gave birth to the `trimTxInfo()` functions, in a bid to reduce the size of the data being stored and processed client-side

views/extras.tmpl
```diff
@@ -187,21 +185,18 @@
     var ws; // websocket global
     async function createWebSocket(loc) {
         // wait a bit to prevent websocket churn from drive by page loads
-        console.log('Waiting to create websocket');
         $('#connection').removeClass('hidden');
         var uri = getSocketURI(loc)
         ws = new MessageSocket(uri);
         await sleep(3000);
```
itswisdomagain: `await sleep(3000);`
What does this do?

itswisdomagain:
@gozart1 The latest commit should fix your issue. Displaying blocks dynamically on the frontend performs better now.
All other requested changes have been effected

views/nexthome.tmpl
```diff
@@ -2,253 +2,147 @@
...
 
-            $(".decredblockWrap").click(function() {
-                if ($( ".decredblockWrap" ).hasClass( "expanded" )) {
-                    $( ".decredblockWrap" ).siblings().addClass("fade");
-                } else {
-                    $( ".decredblockWrap" ).siblings().removeClass("fade");
-                }
-            });
-        })
+        blocks = {{.Blocks}};
```
gozart1: Please fetch JSON data via the api. 
Inlining data into the DOM is inefficient and we also don't want to create globals for data that is related to a specific view. 
By fetching the data aync the page can load quickly and show a nice loading animation instead of instead of freezing up.

gozart1: Please fetch JSON data via the api. 
Inlining data into the DOM is inefficient and we also don't want to create globals for data that is related to a specific view. 
By fetching the data async, the page can load quickly and show a nice loading animation instead of instead of freezing up.

gozart1: Please fetch JSON data via the api. 
Inlining data into the DOM is inefficient and we also don't want to create globals for data that is related to a specific view. 
By fetching the data async, the page can load quickly and show a nice loading animation instead of instead of freezing up.

gozart1: Please fetch JSON data via the api. 
Inlining data into the DOM is inefficient and we also don't want to create globals for data that is related to a specific view. 
By fetching the data async, the page can load quickly and show a nice loading animation instead of instead of freezing up.

gozart1: Please fetch JSON data via the api. 
Inlining data into the DOM is inefficient and we also don't want to create globals for data that is related to a specific view. 
By fetching the data async, the page can load quickly and show a nice loading animation instead of instead of freezing up.

gozart1:
In views/nexthome.tmpl line 142 
(Github errors out for some reason when I try to leave this comment inline)

Please fetch JSON data like blocks via the api.
Inlining data into the DOM is inefficient and we also don't want to create globals for that data since it's is related to a specific view, and not relevant to the rest of the application.
By fetching the data async, the page can load quickly and show a nice loading animation instead of instead of freezing up.

views/nexthome.tmpl
itswisdomagain: I discussed this with @raedah. I'll move all data fetching operations to API calls for the /nexthome page. I recommended the same be done for other pages but in a separate PR

chappjc: For larger data objects, that makes sense, but the site needs to work without JS.  There's not much point in doing an API request for a handful of values when there available at template execution time.  An array of blocks data, yes that makes sense.

chappjc:
There are issue in dark mode.

image

The header looks terrible, but I'm referring to the text that turns white and disappears.

public/js/controllers/blocks.js
```diff
@@ -0,0 +1,314 @@
...
+                    <span class="fund"><span class="paint" style="width:100%;"></span></span>
+                    <span class="fees" title='{"object": "Tx Fees", "total": "${fee}"}'></span>
+                </div>`;
+    }
+
+    const pow = subsidy.pow / conversionRate;
+    const pos = subsidy.pos / conversionRate;
+    const fund = (subsidy.developer || subsidy.dev) / conversionRate;
+    
+    const backgroundColorRelativeToVotes = `style="width: ${voteCount * 20}%"`; // 5 blocks = 100% painting
```
gozart1: This shrinks all the reward divs when votes are missing, that seems off to me.
What's the intent here?
bg

public/js/controllers/blocks.js
```diff
@@ -0,0 +1,314 @@
...
+        for (let i = currentlyDisplayedBlockCount; i < maxBlockElements; i++) {
+            const newBlockElement = newBlockHtmlElement(blocks[i]);
+            blocksHolder.append(newBlockElement);
+        }
+    }
+
+    setupTooltips();
+}
+
+async function displayBlocks() {
```
gozart1: Dom manipulation is typically all synchronous, so there should be no need to make this an async function or to use set time outs.

Are you seeing any issues with just doing this?
```
function displayBlocks() {
    blocksHolder = $('.blocks-holder');
    blocksHolder.append(makeMempoolBlock(blocks[0]));
    const $mempoolBlock = $('.block');
    const maxBlockElements = calculateMaximumNumberOfBlocksToDisplay($mempoolBlock);
    const blocksHTML = _.times(maxBlockElements,(n) => {
        return newBlockHtmlElement(blocks[n+1])
    }).join("\n")
    blocksHolder.append(blocksHTML)
    setupTooltips();
    window.addEventListener('resize', refreshBlocksDisplay);
}
```

gozart1:
The pure white to for the blank slots looks a bit jarring to me
white

How about a light gray #eeeeee ?
gray

public/js/controllers/blocks.js
itswisdomagain: The updated PR description explains the intent. The partial background filling is a factor of the number of votes in the block. The background currently fills 4/5th the available width for each cell because there are 4 votes in the block

itswisdomagain:
Because the cells have opacity proportionate to the number of inputs/outputs in each transaction, using #eeeeee for the blank slots could make the empty slot blend with another cell that has some opacity value

public/css/nexthome.css
```diff
@@ -1,530 +1,660 @@
...
-  }
-  .navbar-fixed-bottom{
-    font-size: 12px;
-  }
+    header .right .searchbox input {
+        width: 150px;
+    }
+}
+@media only screen and (max-width: 768px) {
+    .container {
```
gozart1: On small screens, we should not add extra margin to the side of the content.
This is where screen real estate is at a premium, we should use as much of it as possible for content, not whitespace!

itswisdomagain: That particular style applies to other pages. I had to override that property styling for the /nexthome page using `#nexthome .container` in line 116.
There's still too much space lying fallow on small devices though, caused by a padding of 10% across all screen widths. In line with your comment, I'll reduce the padding in smaller screens to allow the content take up much space.

public/js/controllers/blocks.js
gozart1: Oh interesting, I didn't realize the dev fund also gets reduced when there's less than 5 tickets! Proof is in the pudding, visual blocks are educational!

public/js/controllers/blocks.js
itswisdomagain: Yeah, when I performed the block rendering synchronously, the page goes unresponsive for a brief moment. Sometimes it gets stuck completely and won't even allow me to refresh.

I use the sleep(ms) async function to pause temporarily after adding each block to the dom to give the browser some time to actually show the user something rather than keeping the page stale for a while

views/nexthome.tmpl
gozart1: Think @chappjc chappjc has a good point here about minimizing implementation complexity. I'm ok with either, follow your gut!

public/js/controllers/blocks.js
gozart1: Do you experience those same issues with the code I pasted?

public/css/nexthome.css
```diff
@@ -1,530 +1,660 @@
...
+  align-items: stretch;
+  border-top: 1px solid white;
+}
+.block-rewards {
+    border-top: none;
+}
+.block-rewards > span,
+.block-votes > span,
+.block-tickets > span,
+.block-transactions > span:not(:first-child) {
```
gozart1: The negative psuedo class ":not" isn't widely supported, only works in safari.
https://developer.mozilla.org/en-US/docs/Web/CSS/:not#Browser_compatibility
So we'll have to over-ride to remove the border for first-child with a separate style declaration

itswisdomagain: Noted

public/js/controllers/blocks.js
itswisdomagain: Performance is okay, but there's 1 issue. In order to calculate the number of blocks that can fit into available screen space, the mempool block is first added to the dom and its size (width and height) read.

The size (especially width) of the rectangular boxes is not fixed, only a minimum width is set. The browser does some calculation to determine what width to assign to each rectangular box such that no box is smaller than the minimum width and there's no unused white space after placing the boxes in each row.

The issue is, when we try to calculate the width of the mempool block shortly after appending it to the dom, the value read is incorrect. The width is roughly the same as the width of the entire blocks section.

Waiting a couple milliseconds gives the browser some time to properly size the mempool box and then the correct width can be read.

chappjc:
I am going to echo some of my comments from Slack:

We are discussing a couple related issues:
- The page must look right when JS is disabled (OK if no websocket, but there should not be blank space where there should be values).  Doing the layout in go can be in a PR in the very near future.  @gozart1's comment that we can experiment first using client rendering before making the effort of doing the server side rendering too makes sense.
- Embedding the block data in the JS, specifically in the `<script>` sections of the html template is potentially problematic.  A minor issue is the extra payload that never changes and may not be used.  A bigger issue in my mind is that the JS can never be extracted into a .js file if it is modified at time of template execution.
**- The JSON stamped into the JS is still dead weight even with JS enabled (think when we wire the page with websockets to update on new blocks).**
- The API does not yet provide the data in the needed format.  When it does, the embedding should be removed.  However, an initial go driven layout should still be done.

Especially in the cryptocurrency community, many people browse with JS disabled.  I am not one of those, but we have to be aware of this and not have everything rely on JS.  Even though it makes things harder for us, the page should show values everywhere and be reasonably formatted without JS.

itswisdomagain:
Great summary @chappjc 
- Immediate next PR would deal with finding a way to render the blocks server-side so users who disable JS don't see a blank space
- When the API endpoint that supplies the data used in this page is prepared, another PR will be needed to request the blocks data via API

About this:
> The JSON stamped into the JS is still dead weight even with JS enabled (think when we wire the page with websockets to update on new blocks).

Not necessarily. If the browser window is resized (such as when the user's device orientation changes from portrait to landscape and vice-versa), a couple blocks may need to be added to or removed from the dom to maintain usage of available space. That's one use-case for keeping the blocks array data in memory.

chappjc:
On that last point, if 40 blocks get mined while the home page is opened, and we have the websocket update the page, would the original blocks that were stamped into the JS still be hanging around in that variable this PR is setting while executing the template?

itswisdomagain:
Currently, yes. And wrongly so, I've seen. The code that handles websocket updates on this /nexthome page is here `public/js/controllers/nexthome-socket-update.js`

What it does right now is append new blocks to the blocks array; at the end of the block, ruining the original order of the array where newer blocks are at the start of the array; which would in turn reflect in the page when a page resize event causes the page to add blocks to or remove blocks from the display.

And it does not remove older blocks from the array, adding to the dead weight.

I'm working on a fix and if my thought process is leading me properly, there just may be a way to do the initial rendering with go without missing the original target.

itswisdomagain:
Fix is ready. Here are the changes in the new implementation approach:
- instead of assigning data for mempool and the top 30 blocks to JavaScript variables, the data is processed server-side with go and rendered to the dom during template execution
- during template execution, only the first 10 blocks data is made visible in the dom, the remaining 20 are hidden by default
- when the page loads, JS code is fired to refresh the display, i.e. calculate how many block elements can fit in the blocks-section of the browser screen. If this number is more than 10, more block elements are made visible while if the reverse is the case some block elements are hidden such that only the number of block elements that can fit in the screen is visible
- for JS-disabled browsers, the first 10 blocks element remain visible ensuring that the blocks-section does not appear empty for such scenario
- when new block data is received from websocket, a new element for it is added to the dom just after mempool, and this new element is made visible.
- to ensure that the number of visible blocks is maintained when new block data is received from websocket, the last visible block is hidden. Also, to ensure there are not too many block elements in the dom, the last block element in the dom is removed.
- with the approach defined in the last bulleted point, if 40 blocks updates are received from websocket, the initial 30 block elements will have been removed from the dom to make room for the newly received data. Also, since the 40 blocks updates received from websocket is received incrementally, the first 10 block updates received would also have been removed from the dom ensuring that there only remain 30 block elements (aside mempool) in the dom at any point in time
- 30 block elements are kept in the dom at every point in time though not all may be visible to enable JS-enabled environments respond appropriately to changes in screen size by making more blocks visible as needed

chappjc:
Thank you for the detailed description of changes.  That accurately reflects how I envisioned it working.

explorer/templates.go
```diff
@@ -158,9 +158,15 @@ func makeTemplateFuncMap(params *chaincfg.Params) template.FuncMap {
 		"add": func(a, b int64) int64 {
 			return a + b
 		},
+		"intAdd": func(a, b int) int {
```
chappjc: Note that there is already a built-in `add` function.  Unfortunately the "add" above masked the built-in, so to keep with the `int64` versions being the simplest names, I acknowledge `intAdd` is necessary now.  Just please take note.

explorer/templates.go
```diff
@@ -259,5 +268,18 @@ func makeTemplateFuncMap(params *chaincfg.Params) template.FuncMap {
 		"theme": func() string {
 			return netTheme
 		},
+		"emptyArray": func(length int) []int {
+			if length < 0 {
+				length = 0
+			}
+			return make([]int, length)
+		},
+		"first30Txs": func(txs []TrimmedTxInfo) []TrimmedTxInfo {
```
chappjc: I'd prefer to generalize this to something like "clipSlice" and give it a second param, `N`.

explorer/templates.go
```diff
@@ -259,5 +268,18 @@ func makeTemplateFuncMap(params *chaincfg.Params) template.FuncMap {
...
+			first30 := make([]TrimmedTxInfo, 0, 30)
+			for i := 0; i < 30; i++ {
+				first30[i] = txs[i]
```
chappjc: Please use the built-in `copy` instead of a loop.  Also note that this panics if `txs` is shorter than 30 (or `N`).

Or even better, just create the new array with indexing: `front := txs[:N]` (when len(txs)>=N).

chappjc: Get comfy with all the concepts used in https://github.com/golang/go/wiki/SliceTricks

explorer/websockethandlers.go
```diff
@@ -21,6 +21,49 @@ import (
...
+			TxID:      tx.TxID,
+			Total:     tx.Total,
+			Fees:      tx.Fee.ToCoin(),
+			Coinbase:  tx.Coinbase,
+		}
+		trimmedTxs = append(trimmedTxs, trimmedTx)
+	}
+	return trimmedTxs
+}
+func trimBlockInfo(block *BlockInfo) *TrimmedBlockInfo {
```
chappjc: Please add a new line between these functions as per the style used in the rest of the project.

explorer/websockethandlers.go
```diff
@@ -21,6 +21,49 @@ import (
...
+	// check tx info
+	txCount := len(block.Tx)
+	transactions := make([]*TxInfo, 0, txCount)
+	for _, tx := range block.Tx {
+		if !tx.Coinbase {
```
chappjc: Why does it exclude coinbase?  I think this function needs docs.

explorer/websockethandlers.go
```diff
@@ -97,9 +140,54 @@ func (exp *explorerUI) RootWebsocket(w http.ResponseWriter, r *http.Request) {
 
 				case "getmempooltxs":
 					exp.MempoolData.RLock()
-					msg, err := json.Marshal(exp.MempoolData)
+					mempoolData := exp.MempoolData
 					exp.MempoolData.RUnlock()
```
chappjc: As I said before, this defeats the purpose of the lock. `exp.MempoolData` is a pointer. So when you read the data via the `mempoolData` variable, it's the same data the lock was trying to protect.  You either need a deep copy, or keep it locked until you are done with the underlying shared data.

explorer/explorerroutes.go
```diff
@@ -142,26 +142,80 @@ func (exp *explorerUI) DisapprovedBlocks(w http.ResponseWriter, r *http.Request)
 
 // NextHome is the page handler for the "/nexthome" path.
 func (exp *explorerUI) NextHome(w http.ResponseWriter, r *http.Request) {
+	// get top 30 blocks and trim each block to have just the fields required for this page
 	height := exp.blockData.GetHeight()
-
-	blocks := exp.blockData.GetExplorerFullBlocks(height, height-11)
-
+	blocks := exp.blockData.GetExplorerFullBlocks(height, height-30)
```
chappjc: 30 should be a variable somewhere, package-level constant perhaps

explorer/explorerroutes.go
```diff
@@ -142,26 +142,80 @@ func (exp *explorerUI) DisapprovedBlocks(w http.ResponseWriter, r *http.Request)
 
 // NextHome is the page handler for the "/nexthome" path.
 func (exp *explorerUI) NextHome(w http.ResponseWriter, r *http.Request) {
+	// get top 30 blocks and trim each block to have just the fields required for this page
```
chappjc: All comments in sentence form should begin with a capital letter and end with punctuation (.)

explorer/explorerroutes.go
```diff
@@ -142,26 +142,80 @@ func (exp *explorerUI) DisapprovedBlocks(w http.ResponseWriter, r *http.Request)
...
+	// fetch revocations for mempool
+	revCount := len(exp.MempoolData.Revocations)
+	mempoolRevs := make([]*TxInfo, 0, revCount)
+	for _, tx := range exp.MempoolData.Revocations {
+		exptx := exp.blockData.GetExplorerTx(tx.Hash)
+		mempoolRevs = append(mempoolRevs, exptx)
+	}
+
+	// fetch regular tx for mempool, and calculate total fees for mempool block
+	var mempoolFees float64 = 0
```
chappjc: Just `var mempoolFees float64`.

explorer/explorerroutes.go
```diff
@@ -142,26 +142,80 @@ func (exp *explorerUI) DisapprovedBlocks(w http.ResponseWriter, r *http.Request)
...
+	txCount := len(exp.MempoolData.Transactions)
+	mempoolTxs := make([]*TxInfo, 0, txCount)
+	for _, tx := range exp.MempoolData.Transactions {
```
chappjc: `tx` is a *copy` of the `MempoolInfo` (not just a pointer, a large struct).  Instead, use `for i := range` and index the array.

explorer/explorerroutes.go
```diff
@@ -142,26 +142,80 @@ func (exp *explorerUI) DisapprovedBlocks(w http.ResponseWriter, r *http.Request)
...
+		Subsidy:	  exp.pageData.HomeInfo.NBlockSubsidy,
+		Transactions: trimTxInfo(mempoolTxs),
+		Tickets:      trimTxInfo(mempoolTickets),
+		Votes:        trimTxInfo(mempoolVotes),
+		Revocations:  trimTxInfo(mempoolRevs),
+		Total:        exp.MempoolData.TotalOut,
+		Time:         exp.MempoolData.LastBlockTime,
+		Fees:		  mempoolFees,
+	}
+
```
chappjc: All of the new code in `NextHome` above is extremely inefficient.  I suggest to identify just what is missing in the `explorer.MempoolTx` type.  Do not use `GetExplorerTx` to get a `TxInfo` that is converted with `trimTxInfo`.  Instead only get the missing data.  If this requires a new function like `GetExplorerTxTrimmed` then you can add that, but I am **extremely** averse to running all these DB queries just for the home page.  Please have a careful look at how to be efficient about data handling and queries.  Before it was just the call to `GetExplorerFullBlocks`, and it is many times more demanding now.

explorer/explorerroutes.go
```diff
@@ -142,26 +142,80 @@ func (exp *explorerUI) DisapprovedBlocks(w http.ResponseWriter, r *http.Request)
...
 	str, err := exp.templates.execTemplateToString("nexthome", struct {
 		Info    *HomeInfo
-		Mempool *MempoolInfo
-		Blocks  []*BlockInfo
+		Mempool MempoolData
```
chappjc: Why this change from pointer to copy?

explorer/websockethandlers.go
```diff
@@ -97,9 +140,54 @@ func (exp *explorerUI) RootWebsocket(w http.ResponseWriter, r *http.Request) {
...
 
+					txCount := len(mempoolData.Transactions)
+					mempoolTxs := make([]*TxInfo, 0, txCount)
+					for _, tx := range mempoolData.Transactions {
+						exptx := exp.blockData.GetExplorerTx(tx.Hash)
```
chappjc: All this code reflects big design issues.  There should not be DB queries inside this websocket loop!  All of this code is done for each websocket client.

Note how previously it was just a `json.Marshal`.   The data was already there.  It needs to be again.

public/css/main.css
```diff
@@ -1,15 +1,13 @@
 ﻿/*icons*/
 @font-face {
   font-family: 'icomoon';
-  src:  url('/fonts/icomoon.eot?g003m6');
-  src:  url('/fonts/icomoon.eot?g003m6#iefix') format('embedded-opentype'),
-    url('/fonts/icomoon.ttf?g003m6') format('truetype'),
-    url('/fonts/icomoon.woff?g003m6') format('woff'),
-    url('/fonts/icomoon.svg?g003m6#icomoon') format('svg');
+  src: url('/fonts/icomoon.eot?g003m6');
+  src: url('/fonts/icomoon.eot?g003m6#iefix') format('embedded-opentype'), url('/fonts/icomoon.ttf?g003m6') format('truetype'), url('/fonts/icomoon.woff?g003m6') format('woff'), url('/fonts/icomoon.svg?g003m6#icomoon') format('svg');
```
chappjc: Why is this being compacted?  It's minimal byte savings but it's far less legible.

public/css/main.css
```diff
@@ -405,24 +386,21 @@ body.darkBG .progress {
   border-radius: 3px;
   z-index: 1;
   transform-origin: top right;
-  transition: transform 0.5s cubic-bezier(0.77,0.2,0.05,1.0),
-              background 0.5s cubic-bezier(0.77,0.2,0.05,1.0),
-              width .3s,
-              opacity 0.55s ease;
+  transition: transform 0.5s cubic-bezier(0.77, 0.2, 0.05, 1.0), background 0.5s cubic-bezier(0.77, 0.2, 0.05, 1.0), width .3s, opacity 0.55s ease;
```
chappjc: Same legibility issue here.  I don't think it's worth the gain.  Perhaps I'm wrong though.

db/dcrsqlite/apisource.go
```diff
@@ -1105,11 +1105,12 @@ func (db *wiredDB) GetExplorerBlock(hash string) *explorer.BlockInfo {
 		NextHash:              data.NextHash,
 		StakeValidationHeight: db.params.StakeValidationHeight,
 		AllTxs:                (uint32(b.Voters) + uint32(b.Transactions) + uint32(b.FreshStake)),
+		Subsidy:               db.BlockSubsidy(b.Height, b.Voters),
 	}
 
-	votes := make([]*explorer.TxBasic, 0, block.Voters)
-	revocations := make([]*explorer.TxBasic, 0, block.Revocations)
-	tickets := make([]*explorer.TxBasic, 0, block.FreshStake)
+	votes := make([]*explorer.TxInfo, 0, block.Voters)
```
chappjc: Going from TxBasic to TxInfo makes this way bigger.

db/dcrsqlite/apisource.go
```diff
@@ -1119,28 +1120,24 @@ func (db *wiredDB) GetExplorerBlock(hash string) *explorer.BlockInfo {
 		}
 		switch stake.DetermineTxType(msgTx) {
 		case stake.TxTypeSSGen:
-			stx := makeExplorerTxBasic(tx, msgTx, db.params)
+			stx := db.GetExplorerTx(tx.Txid)
```
chappjc: And this makes it way slower.  `makeExplorerTxBasic` involved no DB queries, now this involves tons of queries.

What all is needed?  All of TxInfo?

explorer/explorertypes.go
```diff
@@ -334,6 +359,18 @@ type BlockSubsidy struct {
...
+// MempoolData models data to update mempool info on the home page
+type MempoolData struct {
+	Transactions []*TrimmedTxInfo
+	Tickets      []*TrimmedTxInfo
+	Votes        []*TrimmedTxInfo
+	Revocations  []*TrimmedTxInfo
+	Subsidy      BlockSubsidy
+	Total        float64
+	Time         int64
+	Fees		 float64
```
chappjc: go fmt everything please

public/css/main.css
itswisdomagain: You're not IMO. Will correct

db/dcrsqlite/apisource.go
itswisdomagain: Not all. This is what I need:
```
VinCount:  len(tx.Vin),
VoutCount: len(tx.Vout),
VoteValid: voteValid, // tx.Type really
TxID:      tx.TxID,
Total:     tx.Total,
Fees:      tx.Fee.ToCoin(),
Coinbase:  tx.Coinbase,
```

I can't get all of that from TxBasic, so I get full Tx, pass into `trimTxInfo()` in `explorer/websockethandlers.go` to get just what I need before passing into template.

Looking deeper at `GetExplorerTx(txid string)` in `db/dcrsqlite/apisource.go`, I think there may be a way to get the data I need without running all those many queries. I'll refactor (create another method  in `db/dcrsqlite/apisource.go` similar to `GetExplorerTx(txid string)` that returns what I need with fewer to 0 queries)

explorer/websockethandlers.go
itswisdomagain: I'm not sure why. I adapted the pattern I saw in the block page. When looping through the transactions in a block to display in the transactions table of the block page, the code excluded coinbase transactions. (`lines 385 to 400 of /views/block.tmpl`)

I just borrowed a leaf with the impression that it was normal.

itswisdomagain:
I've attended to most of the requested changes @chappjc. The changes left are those relating to mempool - on page load and on websocket update.

I'll look out how to optimize those with fewer to no queries as I've done for blocks.

dmigwi:
@chappjc ok. Noted

dmigwi:
I think this version of the night mode looks better to me.
screenshot from 2018-10-29 03-59-01

public/css/nexthome.css
```diff
@@ -1,542 +1,697 @@
...
+}
+.block.visible {
+    display: block;
+}
+.block-info {
+    width: 100%;
+    display: flex;
+    justify-content: space-between;
+    padding: .2em .5em;
     background: white;
```
dmigwi: ` background: white;`
Delete this.

public/css/nexthome.css
```diff
@@ -1,542 +1,697 @@
...
+.block-rows .block-tickets,
+.block-rows .block-transactions {
+  flex-basis: 5px;
+  flex-grow: 1;
+  width: 100%;
+  display: flex;
+  flex-wrap: nowrap;
+  align-items: stretch;
+  border-top: 1px solid white;
+  background: white;
```
dmigwi: `background: white;`
Delete this.

public/css/nexthome.css
```diff
@@ -1,542 +1,697 @@
...
+    margin-bottom: 30px;
+}
+.blocks-holder {
+    display: grid;
+    grid-template-columns: repeat(auto-fill,minmax(250px, 1fr));
+    grid-gap: 20px;
+}
+.block {
+    display: none;
+    background: white;
```
dmigwi: ` background: white;`
Delete this

chappjc:
Issues identified while testing:

1. Block subsidies are displayed in atoms instead of DCR, but the named units are DCR.  The numbers should be converted to DCR.
2. Revocations are taking up a ticket slot, but they aren't related to the limit (20).
3. The votes say "Vote (no)".  What is the "no" indicating? These votes are all yes on the previous block, and almost all abstains on the issue items. "no" seems incorrect in any case.
4. Tx Fees for mempool is a large negative number.  If there is a bug needing to be fixed in a different PR, then these fees should be set to zero until it is fixed.
5. It is not clear what the width of the gray blocks (regular txns) indicate.  It seems to indicate the proportion of the total DCR.  Specifically the total DCR across just the regular transactions.  Is that correct?  Another interpretation might be size of the tx in bytes.
6. numbers jumbled up

image

7. extra space after decimal place
8. no units for hashrate
9. rename Dev Reward to Project Subsidy to match Project Fund name

chappjc:
Some more:

10. The votes in mempool should only include the unique votes for the current block.
11. The DCR amount for the votes does not seem important at all.
12. The "Decred Blockchain" heading would be better as "Latest Blocks" since the Decred part is obvious.
13. Mainnet Network Statistics stutters a bit.  I think Mainnet Statistics or Network Statistics would be OK.

dmigwi:
@itswisdomagain It would be great for you to post some screenshots here at least for documentation purposes once you make a relatively huge UI/graphics change.

db/dcrsqlite/apisource.go
```diff
@@ -1180,12 +1184,62 @@ func (db *wiredDB) GetExplorerBlock(hash string) *explorer.BlockInfo {
...
+		TxBasic:   txBasic,
+		Fees:      txBasic.Fee.ToCoin(),
+		VinCount:  len(txraw.Vin),
+		VoutCount: len(txraw.Vout),
+		VoteValid: voteValid,
+	}
+	return tx
+}
+
+func (db *wiredDB) GetExplorerTxTrimmed(txid string) *explorer.TrimmedTxInfo {
```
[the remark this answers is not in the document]
chappjc: In  fact, I'm not sure why `makeExplorerTxBasic`, `makeExplorerBlockBasic`, and `makeExplorerAddressTx` are even in the `dcrsqlite` package as they would be better suited to `explorer` (explorertypes.go).  But that refactoring can be saved for later.

explorer/explorerroutes.go
```diff
@@ -140,30 +143,90 @@ func (exp *explorerUI) DisapprovedBlocks(w http.ResponseWriter, r *http.Request)
...
+// show only regular tx in block.Transactions, exclude coinbase (reward) transactions
+// for use in NextHome handler and websocket response to getmempooltxs event
+func filterRegularTx(txs []*TrimmedTxInfo) []*TrimmedTxInfo {
+	transactions := make([]*TrimmedTxInfo, 0)
+	for _, tx := range txs {
+		if !tx.Coinbase {
+			transactions = append(transactions, tx)
+		}
+	}
+	return transactions
```
chappjc: A pattern I prefer for functions like this is to use a named return parameter and do no `make`.  i.e.:

```go
func filterRegularTx(txs []*TrimmedTxInfo) (transactions []*TrimmedTxInfo) {
	for _, tx := range txs {
		if !tx.Coinbase {
			transactions = append(transactions, tx)
		}
	}
	return
}
```

When appropriate, named parameters are nice as they are self-documenting, even if it's already clear in this case.  But when the slice is OK to be nil, it also eliminates the unnecessary make or other declaration of the variable.

explorer/explorerroutes.go
```diff
@@ -140,30 +143,90 @@ func (exp *explorerUI) DisapprovedBlocks(w http.ResponseWriter, r *http.Request)
...
+
 // NextHome is the page handler for the "/nexthome" path.
 func (exp *explorerUI) NextHome(w http.ResponseWriter, r *http.Request) {
+	// Get top N blocks and trim each block to have just the fields required for this page.
 	height := exp.blockData.GetHeight()
+	blocks := exp.blockData.GetExplorerFullBlocks(height, height-homePageBlocksMaxCount)
+
+	// trim unwanted data in each block
+	trimmedBlocks := make([]*TrimmedBlockInfo, 0, len(blocks))
+	for index := range blocks {
```
chappjc: Since it is a slice of pointers, best to use the second range value (a copy of the element):

```go
for _, block := range blocks {
```

and omit `block := blocks[index]` of course.

explorer/explorerroutes.go
```diff
@@ -140,30 +143,90 @@ func (exp *explorerUI) DisapprovedBlocks(w http.ResponseWriter, r *http.Request)
...
+	getTotalFee := func(txs []*TrimmedTxInfo) (total float64) {
+		for _, tx := range txs {
+			total += tx.Fees
+		}
+		return
+	}
+	mempoolFees := getTotalFee(mempoolTxs) + getTotalFee(mempoolRevs) + getTotalFee(mempoolTickets) +
+		getTotalFee(mempoolVotes)
+
+	// construct mempool object with properties required in template
```
chappjc: The `exp.pageData.RLock()` can be moved to just above this comment.

explorer/explorerroutes.go
```diff
@@ -140,30 +143,90 @@ func (exp *explorerUI) DisapprovedBlocks(w http.ResponseWriter, r *http.Request)
...
 	str, err := exp.templates.execTemplateToString("nexthome", struct {
 		Info    *HomeInfo
-		Mempool *MempoolInfo
-		Blocks  []*BlockInfo
+		Mempool *MempoolData
+		Blocks  []*TrimmedBlockInfo
 		Version string
 		NetName string
 	}{
 		exp.pageData.HomeInfo,
```
chappjc: must move `exp.pageData.RUnlock()` after `execTemplateToString`

explorer/explorerroutes.go
```diff
@@ -140,30 +143,90 @@ func (exp *explorerUI) DisapprovedBlocks(w http.ResponseWriter, r *http.Request)
...
+	}
 
 	exp.pageData.RLock()
 	exp.MempoolData.RLock()
 
+	mempoolVotes := exp.blockData.GetTrimmedMempoolTx(exp.MempoolData.Votes)
+	mempoolTickets := exp.blockData.GetTrimmedMempoolTx(exp.MempoolData.Tickets)
+	mempoolRevs := exp.blockData.GetTrimmedMempoolTx(exp.MempoolData.Revocations)
+	mempoolTxs := exp.blockData.GetTrimmedMempoolTx(exp.MempoolData.Transactions)
+
```
chappjc: If you grab the `exp.MempoolData.TotalOut` and `.MempoolData.LastBlockTime` values here, you can unlock `exp.MempoolData` much earlier.

explorer/explorertypes.go
```diff
@@ -334,6 +357,18 @@ type BlockSubsidy struct {
 	Dev   int64 `json:"dev"`
 }
 
+// MempoolData models data to update mempool info on the home page
```
chappjc: (on the new home page).  Reading the same doc as `MempoolInfo` is a bit confusing.

explorer/explorertypes.go
```diff
@@ -462,12 +497,13 @@ type TicketPoolInfo struct {
...
-	Time     int64     `json:"time"`
-	Size     int32     `json:"size"`
-	TotalOut float64   `json:"total"`
-	Type     string    `json:"Type"`
-	VoteInfo *VoteInfo `json:"vote_info"`
+	Hash      string    `json:"hash"`
+	Time      int64     `json:"time"`
+	Size      int32     `json:"size"`
+	TotalOut  float64   `json:"total"`
+	Type      string    `json:"Type"`
```
chappjc: Idiomatic JSON would have this tagged as "type".  Please change it while you are here.

explorer/explorertypes.go
```diff
@@ -462,12 +497,13 @@ type TicketPoolInfo struct {
...
+	VoteInfo  *VoteInfo `json:"vote_info"`
```
chappjc: Please also change to `json:"vote_info,omitempty"` while here.

explorer/explorertypes.go
```diff
@@ -462,12 +497,13 @@ type TicketPoolInfo struct {
...
+	VoteValid bool
```
chappjc: Needs json tag for consistency

explorer/templates.go
```diff
@@ -303,5 +312,18 @@ func makeTemplateFuncMap(params *chaincfg.Params) template.FuncMap {
 		"theme": func() string {
 			return netTheme
 		},
+		"emptyArray": func(length int) []int {
```
chappjc: Another nit pick: array and slice are quite different, and this makes a slice of a certain length with zero values, so this function is better named "zeroSlice".

explorer/websockethandlers.go
```diff
@@ -21,6 +21,26 @@ import (
 // close an already closed connection.
 var ErrWsClosed = "use of closed network connection"
 
+func trimTxInfo(txs []*TxInfo) []*TrimmedTxInfo {
+	trimmedTxs := make([]*TrimmedTxInfo, 0, len(txs))
+	for _, tx := range txs {
+		voteValid := false
```
chappjc: This is fine, but the style used in our code is usually to declare like `var v someType` when the zero value for `someType` is desired for the value of the new variable `v` (i.e. `var voteValid bool`)

explorer/websockethandlers.go
```diff
@@ -21,6 +21,26 @@ import (
 // close an already closed connection.
 var ErrWsClosed = "use of closed network connection"
 
+func trimTxInfo(txs []*TxInfo) []*TrimmedTxInfo {
```
chappjc: Please move this function to explorertypes.go.  Generally for functions that just perform simple conversions without any major dependencies, it is preferable to have them in a conversions.go or types.go kind of file.

explorer/websockethandlers.go
```diff
@@ -96,10 +116,41 @@ func (exp *explorerUI) RootWebsocket(w http.ResponseWriter, r *http.Request) {
...
 
 				case "getmempooltxs":
+					exp.pageData.RLock()
 					exp.MempoolData.RLock()
-					msg, err := json.Marshal(exp.MempoolData)
+
+					mempoolVotes := exp.blockData.GetTrimmedMempoolTx(exp.MempoolData.Votes)
+					mempoolTickets := exp.blockData.GetTrimmedMempoolTx(exp.MempoolData.Tickets)
+					mempoolRevs := exp.blockData.GetTrimmedMempoolTx(exp.MempoolData.Revocations)
+					mempoolTxs := exp.blockData.GetTrimmedMempoolTx(exp.MempoolData.Transactions)
```
chappjc: This is all way better than before thanks to the conversion functions you added to eliminate DB access.  However, the code here is identical to the code in `NextHome`.  I suggest factoring out the common code, into something like:

```go
func (exp *explorerUI) MempoolData() *MempoolData {
	// convert the *MempoolInfo in exp.MempoolData to *MempoolData
...
}
```

explorer/websockethandlers.go
```diff
@@ -96,10 +116,41 @@ func (exp *explorerUI) RootWebsocket(w http.ResponseWriter, r *http.Request) {
...
+
+					// calculate total fees for mempool block
+					getTotalFee := func(txs []*TrimmedTxInfo) (total float64) {
+						for _, tx := range txs {
+							total += tx.Fees
+						}
+						return
+					}
+					mempoolFees := getTotalFee(mempoolTxs) + getTotalFee(mempoolRevs) + getTotalFee(mempoolTickets) +
```
chappjc: I'm still quite uneasy about adding a large number of RPCs (getrawtransaction verbose) under the control of the websocket client.  It seems to me that just like the MempoolInfo struct, the MempoolData should be *updated* in `(exp *explorerUI) mempoolMonitor`, when there is a mempool event.  I say updated because there is no reason to repeatedly call getrawtransaction for the same transactions over and over.

This change must be made for this to be OK.

public/css/main.css
```diff
@@ -35,51 +35,45 @@
...
 }
-
 /*elements*/
 html {
   color: #052f4b;
   -moz-osx-font-smoothing: grayscale;
   -webkit-font-smoothing: antialiased !important;
   -moz-font-smoothing: antialiased !important;
   text-rendering: optimizelegibility !important;
-  padding-top: 35px;
```
chappjc: Please describe all the UI changes.

views/extras.tmpl
```diff
@@ -188,14 +186,11 @@
...
```
chappjc: You asked why we have this.  The reason is there are many bouncing visitors who load the page and do not stay long.  I think crawlers do this often.  There is very high websocket connection churn without a short delay.

explorer/websockethandlers.go
itswisdomagain: This function is actually no longer used/needed, I'm taking it out

itswisdomagain:
I added more fields to MempoolTx struct and update them in `(exp *explorerUI) mempoolMonitor` when there is a mempool event. This change also eliminates the need to export new functions from dcrsqlite.

I'll describe UI changes in a separate comment. Here's the new (current) look of /nexthome as @dmigwi advised I share. There's really no new change except for the background color properties I removed from some elements to maintain the default night/day mode background colors.

screenshot 2018-10-31 at 4 36 26 am

screenshot 2018-10-31 at 4 44 26 am

public/css/main.css
itswisdomagain: html had padding top, body had padding bottom. I took off both and gave body a padding top and bottom. All page visible content should be in body, that's what makes sense to pad.

itswisdomagain: Most other CSS changes were eliminating unnecessary nesting of selectors

gozart1: In nexthome.css you can go to town, since the css is majorly screwed up and you've effectively taken over ownership of that. But for main.css which several current contributors have worked on I think it's valuable to preserve the git history. 

Although I think the stylistic changes you're making to main.css are mostly fine, I don't see them as necessary  and that us all striving for good git hygiene is much more beneficial.

itswisdomagain: Noted. Will undo changes to main.css

chappjc: Thanks, @gozart1.  I strongly agree with the points you made.

explorer/mempool.go
```diff
@@ -76,25 +90,42 @@ func (exp *explorerUI) mempoolMonitor(txChan chan *NewMempoolTx) {
 			}
 		}
 
+		// get txraw
+		txraw, err := exp.blockData.DecodeRawTransaction(ntx.Hex)
```
chappjc: Really `DecodeRawTransaction` shouldn't even exist.  I think it was used in explorer/websockethandlers.go because it actually needed the struct in a dcrjson format.  But here all that is needed is the `Vin` slice, the number of vins and vouts, and the tx hash (`Txid`).  In general, we should use `txhelpers.MsgTxFromHex` so we do not have to make an RPC.

But in the function, that was already done way above.  So you just need to get rid of lines 93-107 and edit the MempoolTx assignment to:

```go
tx := MempoolTx{
	TxID:      msgTx.TxHash().String(),
	Hash:      hash,
	Time:      ntx.Time,
	Size:      int32(len(ntx.Hex) / 2),
	TotalOut:  txhelpers.TotalOutFromMsgTx(msgTx).ToCoin(),
	Type:      txhelpers.DetermineTxTypeString(msgTx),
	VoteInfo:  voteInfo,
	VinCount:  len(msgTx.TxIn),
	VoutCount: len(msgTx.TxOut),
	Coinbase:  blockchain.IsCoinBaseTx(msgTx),
}
```

itswisdomagain:
@chappjc re the issues you identified...
 
> 1. Block subsidies are displayed in atoms instead of DCR, but the named units are DCR.  The numbers should be converted to DCR.

Fixed

> 2. Revocations are taking up a ticket slot, but they aren't related to the limit (20).

Yes. They're pretty rare and few. Should this change?

> 3. The votes say "Vote (no)".  What is the "no" indicating? These votes are all yes on the previous block, and almost all abstains on the issue items. "no" seems incorrect in any case.

Fixed

> 4. Tx Fees for mempool is a large negative number.  If there is a bug needing to be fixed in a different PR, then these fees should be set to zero until it is fixed.

Can you point me to where you found this anomaly please?

> 5. It is not clear what the width of the gray blocks (regular txns) indicate.  It seems to indicate the proportion of the total DCR.  Specifically the total DCR across just the regular transactions.  Is that correct?  Another interpretation might be size of the tx in bytes.

The width is a proportion of the total DCR across just the regular transactions, yes.

> 6. numbers jumbled up

How'd you mean?

> 7. extra space after decimal place

I did some hunting and found that this issue was introduced in [PR 721](https://github.com/decred/dcrdata/pull/721), specifically [line 1079 of main.css](https://github.com/itswisdomagain/dcrdata/commit/bb157229af5b27ecf2d46b9dc835adc32cfa51cb#diff-4fab20fc9b7238f5f5307386f2b163ffR1079) which added a margin-left of -2px to `.decimal` elements. I added a style in nexthome.css to reset the margin-left of `.decimal` to 0 for all `.decimal` elements in `#mainContainer` to avoid potentially messing up some other page(s) if I remove the negative margin from main.css.

> 2. no units for hashrate

Don't remember fixing this, but the unit is showing now

> 3. rename Dev Reward to Project Subsidy to match Project Fund name

Fixed

chappjc:
Re: jumbled numbers, I'm talking about the red rectangles with rounded edges in this:

image

chappjc:
Regarding revokes taking up one of the 20 slots for tickets, what happens if there are 20 mined tickets and 1+ revokes?

public/css/nexthome.css
```diff
@@ -1,396 +1,683 @@
 @import url("https://fonts.googleapis.com/css?family=Inconsolata:400,700|Source+Sans+Pro:400,600");
```
gozart1: All assets should be locally hosted, including fonts.
We want to protect out users from 3rd party telemetry to the extend that we can.
Also, main.css already loads Inconsolata.
But if we need additional weights they should be added accordingly.

itswisdomagain: added to main.css or kept in nexthome.css but hosted locally?
as per previous conversations wrt keeping main.css untouched unless absolutely required, I'll check if the other font weights are needed/used and if so, host them locally and load them in nexthome.css

chappjc:
Regarding " It could get to as many as 50 (tickets are truncated if they're more than 50 and the first 30 are shown with an additional box providing summary of the numbers not displayed)":
There cannot be more than 20.  This is a network rule set in chaincfg.Params.

But it's fine to put any revoke (s). In the right.  Thanks for clarifying.

dmigwi:
@chappjc I think mempool is a special case in regards to the number of tickets. I once saw 400 tickets and later came to learn that this isn't a rare occurrence. Blocks will never have more than 20 tickets each but Mempool will certainly have once in while.

chappjc:
Yes, absolutely mempool has no limit.  I was referring to blocks.  Thanks for pointing this out.

views/nexthome.tmpl
```diff
@@ -2,253 +2,350 @@
...
                         </div>
                     </div>
-                    {{end}}
-  		        </div>
-                <a class="no-underline" href="/blocks"><div class="showmore"><p>View BlockChain History</p></div></a>
+                {{end}}
+                </div>
+                {{end}}
             </div>
+            <a class="no-underline" style="display: block; margin-top: 25px;" href="/blocks"><div class="showmore"><p>View BlockChain History</p></div></a>
```
gozart1: Try to use Bootstrap classes instead of inline styles, whenever possible.

It improves render performance (inline styles cause repaints as the browser gets to them)
It also improves versatility and consistency of codebase when combined together with other utility classes to produce outcomes. 

Also, it's also a very effective way to learn the framework. Whenever you notice yourself wanting to write a inline style, or declare a new classes for minor or one-off position/layout reasons, go check the bootstrap docs and see if you can't to it declaratively by composing a set of existing utility classes to achieve the end goal.  You will be amazed to find you often don't need to write any css at all.

In this case `d-block` will apply `"display: block;`

IMO that inline padding can most likely be handled via Boostrap classes as well, so that we can easily adjust it for responsive layouts without having to write any css.

gozart1:
Thanks for making the requested changes. I think this can be safely be merged now.

I think we'll likely keep iterating on this and refining the layout. But one last suggestion I have before merging this initial version is that there is way too much whitespace at the top:

screen shot 2018-11-06 at 12 52 24 am


How about something more like this?
screen shot 2018-11-06 at 12 50 45 am

PR #720, opened 2018-10-04
fix stakedb rewind error on startup

chappjc:
Fix 1: Rewinding of stakedb never goes below 0, do not use `lastBlockPG`,
which will be -1 when auxDB is empty. This addresses:
`"failed to rewind stakedb: got 0, expecting -1"`

Fix 2: Prevent the startHeight of the block scan in `(*wiredDB).resyncDB`
from beginning beyond `fetchToHeight`, as this is the height at which the
`MasterBlockGetter` in `(*ChainDB).SyncChainDB` begins notifying of new
blocks it fetches via RPC, and begins waiting for blocks to be connected
in stakedb in `(*wiredDB).resyncDB`.

Note that the stakedb rewind in main ensures stakedb is not beyond the
height of `auxDB` (a `ChainDB`).

Thoroughly document semantics of `fetchToHeight` and the relationship of
`BlockGetter` in `baseDB` and `MasterBlockGetter` in `auxDB`.

Add an early loop continue in `(*wiredDB).resyncDB` when that iteration's
only purpose was to advance the stakedb (catching up).

### DIFF
--- a/db/dcrsqlite/sync.go
+++ b/db/dcrsqlite/sync.go
@@ -176,7 +176,7 @@ func (db *wiredDB) resyncDB(quit chan struct{}, blockGetter rpcutils.BlockGetter
 		}
 		return startHeight, nil
 	}
-	
+
 	if barLoad != nil && db.updateStatusSync {
 		barLoad <- &dbtypes.ProgressBarLoad{
 			Msg:   InitialLoadSyncStatusMsg,
--- a/db/dcrsqlite/sync.go
+++ b/db/dcrsqlite/sync.go
@@ -1,3 +1,4 @@
+// Copyright (c) 2018, The Decred developers
 // Copyright (c) 2017, Jonathan Chappelow
 // See LICENSE for details.
 
@@ -95,13 +96,13 @@ func (db *wiredDB) resyncDB(quit chan struct{}, blockGetter rpcutils.BlockGetter
 	// pace rather than waiting on other consumers to get done with the stakedb.
 	master := blockGetter == nil || blockGetter.(*rpcutils.BlockGate) == nil
 
-	// Get chain servers's best block
+	// Get chain servers's best block.
 	_, height, err := db.client.GetBestBlock()
 	if err != nil {
 		return -1, fmt.Errorf("GetBestBlock failed: %v", err)
 	}
 
-	// Time this function
+	// Time this function.
 	defer func(start time.Time, perr *error) {
 		if *perr == nil {
 			log.Infof("resyncDBWithPoolValue completed in %v", time.Since(start))
@@ -110,7 +111,8 @@ func (db *wiredDB) resyncDB(quit chan struct{}, blockGetter rpcutils.BlockGetter
 		log.Errorf("resyncDBWithPoolValue failed: %v", err)
 	}(time.Now(), &err)
 
-	// Check and report heights of the DBs
+	// Check and report heights of the DBs. startHeight is the lowest of the
+	// heights, and may be -1 with an empty SQLite DB.
 	startHeight, summaryHeight, stakeInfoHeight, stakeDBHeight, err := db.DBHeights()
 	if err != nil {
 		return -1, fmt.Errorf("DBHeights failed: %v", err)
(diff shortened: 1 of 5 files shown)


db/dcrsqlite/sync.go
```diff
@@ -136,8 +139,29 @@ func (db *wiredDB) resyncDB(quit chan struct{}, blockGetter rpcutils.BlockGetter
...
+		if fetchToHeight < stakeDBHeight {
+			return startHeight, fmt.Errorf("fetchToHeight may not be less than stakedb height")
+		}
+
+		// Start at the next block we don't have in both SQLite and stakedb, but
+		// do not start beyond fetchToHeight if there is an external
+		// MasterBlockGetter, the owner of which should already be configured to
+		// send the block at fetchToHeight over the waitChan (e.g. the call to
+		// UpdateToBlock in (*ChainDB).SyncChainDB).
+		if fetchToHeight > startHeight {
```
chappjc: This is the crux of Fix 2.

chappjc:
The challenge was to keep base and aux DBs syncing at the same height.  One of the main reasons is that there is one StakeDatabase, which is shared between the two, and both DBs need access to it for each height.

`rpcutils.BlockGetter` is an interface with basic accessor methods like `Block` and `WaitForHash` that can request data and channels for future data, but do not update the state of the object.
The `rpcutils.MasterBlockGetter` is an interface that embeds the regular `BlockGetter`, adding with functions that can change state and signal to the `BlockGetter`s, such as `UpdateToHeight`, which will get the block via RPC and signal to all channels configured for that block.

In full mode, `ChainDB` has the `MasterBlockGetter` and `wiredDB` has the `BlockGetter`.  The way `ChainDB` is in charge of requesting blocks on demand from RPC without getting ahead of `wiredDB` during sync is that `StakeDatabase` has a very similar coordination mechanism (`WaitForHeight`). 

1. In main, we make a new "rpcutils.BlockGate`, a concrete type that implements `MasterBlockGetter` and thus `BlockGetter` too.  This "smartClient" is provided to `baseDB` (as a `BlockGetter`) and to `auxDB` (as a `MasterBlockGetter`).
2. `baseDB` makes a channel using `BlockGetter.WaitForHeight` and starts waiting for the current block to come across the channel.
3. `auxDB` makes a channel using `StakeDatabase.WaitForHeight`, which instructs the stakeDB to send notification when it connects the specified block. `auxDB` does not start waiting for the signal yet.
4a. `auxDB` requests via the `MasterBlockGetter` that the same current block be retrieved via RPC using `MasterBlockGetter.UpdateToBlock`.
4b. `auxDB` immediately begins waiting for the signal that `StakeDatabase` has connected the block.
5. The call to `UpdateToBlock` causes the underlying (shared) smartClient to send a signal on all channels registered for that block.
6. `baseDB`gets notification on the channel and retrieves the block (which the channel signaled is now available) via `BlockGetter.Block`.
7. Before connecting the block in the `StakeDatabase`, `baseDB` gets a new channel for the following (i+1) block, so that when `auxDB` requests it later the channel will be registered already.
8. `baseDB` connects the block in `StakeDatabase`.
9. `StakeDatabase` signals to all waiters (`auxDB`, see 4b) that the stake db is ready at the needed height.
10. `baseDB` finishes block data processing/storage and goes back to 2. for the next block.
11. Concurrent with `baseDB` processing in 10., `auxDB` receives the notification from `StakeDatabase` sent in 9. and continues block data processing/storage.  When done processing, `auxDB` goes back to step 3. for the next block.  As with the previous iteration, it sets the pace with `UpdateToBlock`.

With the above approach, the DBs share a single StakeDatabase, the DBs are in sync (tightly coupled), there is ample opportunity for concurrent computations, and the shared `blockGetter` (as a `MasterBlockGetter` in `auxDB`, and a `BlockGetter` in `baseDB`) makes it so a given block will only be fetched via RPC once and stored for the `BlockGetters` that are waiting for the block.

chappjc:
@dmigwi Please see https://github.com/decred/dcrdata/pull/720/commits/211d141e5576880ddc73439e1ed5875863c694ad

dmigwi:
This new information definitely helps a lot more than I envisioned. Just realized that stakebase db pretty much sets the syncing pace. I also did not know for sure that fetching the next block is done once for dbs when running in full mode.
Thanks for adding this extra information.

PR #721, opened 2018-10-04
New Stats Page

Jujhar:
stats page

Includes and built off Macaulay's commits from #641, approximately half of work was his.
Addresses #451


explorer/templates.go
```diff
@@ -161,6 +161,9 @@ func makeTemplateFuncMap(params *chaincfg.Params) template.FuncMap {
 		"subtract": func(a, b int64) int64 {
 			return a - b
 		},
+		"floatsubtract": func(a float64, b float64) float64 {
```
chappjc: For consistency, please elide the first `float64`, like the other functions.

explorer/templates.go
```diff
@@ -228,6 +231,35 @@ func makeTemplateFuncMap(params *chaincfg.Params) template.FuncMap {
...
+					log.Errorf("amountAsDecimalParts comma formatting failed. Input: %v Error: %v", v, err.Error())
+					integer = "ERROR"
+					dec := "VALUE"
+					zeros := ""
+					return []string{integer, dec, zeros}
+				}
+				integer = humanize.Comma(integerAsInt64)
+			}
+			dec := strings.TrimRight(amt[len(amt)-8:], "0")
+			dec = string(dec[0:numPlaces])
```
chappjc: There are numerous ways this function can panic.  Please look at all the indexing statements.  And write tests for this function in templates_test.go

chappjc: Some other ideas:

```go
package main

import (
	"fmt"
	"strconv"
	"strings"
)

func main() {
	f := func(v int64, numPlaces int64) []string {
		intpart := v / 1e8
		decpart := v % 1e8

		left := strconv.FormatInt(intpart, 10)
		rightTail := fmt.Sprintf("%08d", decpart)
		right := strings.TrimRight(rightTail, "0")
		tail := strings.Repeat("0", len(rightTail)-len(right))
		return []string{left, right, tail}
	}

	fmt.Println(f(4321, 8))
	fmt.Println(f(87654321, 8))
	fmt.Println(f(76543210, 8))
	fmt.Println(f(6543210, 8))
	fmt.Println(f(987654321, 8))
	fmt.Println(f(987654320, 8))
	fmt.Println(f(907654321, 8))
	fmt.Println(f(907654320, 8))
	fmt.Println(f(907654320, 8))
	fmt.Println(f(12345907654320, 8))
}
```

views/statistics.tmpl
```diff
@@ -0,0 +1,234 @@
+{{define "statistics"}}
+<!DOCTYPE html>
+<html lang="en">
+    {{ template "html-head" printf "Decred Statistics Page"}}
```
chappjc: Remove the word "Page".  That's self evident.  I see this word was used on the agendas page, but it should not have been.

explorer/templates.go
Jujhar: Know how to test a sub function within and under the makeTemplateFuncMap function? How do I call that from the unit test?

chappjc: You can either write the function in the package's scope and set it in `makeTemplateFuncMap`, or in the test function you can just run `funcMap := makeTemplateFuncMap(params)` with the params from any network defined in dcrd's chaincfg package.

Jujhar: Done

explorer/templates_test.go
```diff
@@ -0,0 +1,59 @@
+package explorer
+
+import (
+	"testing"
+)
+
+func TestamountAsDecimalPartsTrimmed(t *testing.T) {
```
chappjc: This test will never run.  The first letter after "Test" must not be lowercase.  Rename to `TestAmountAsDecimalPartsTrimmed`

explorer/templates_test.go
```diff
@@ -0,0 +1,59 @@
...
+	test1 := amountAsDecimalPartsTrimmed(314159000, 2, false)
+	test2 := amountAsDecimalPartsTrimmed(76543210000, 2, false)
```
chappjc: Run the function only after testing the output of the previous function.

explorer/templates_test.go
```diff
@@ -0,0 +1,59 @@
...
+	test4 := amountAsDecimalPartsTrimmed(654321000, 1, false)
+	test5 := amountAsDecimalPartsTrimmed(987654321, 8, false)
+	test6 := amountAsDecimalPartsTrimmed(987654321, 2, false)
+	test7 := amountAsDecimalPartsTrimmed(90765432100, 2, false)
+	test8 := amountAsDecimalPartsTrimmed(9076543200000, 2, false)
+	test9 := amountAsDecimalPartsTrimmed(907654320, 7, false)
+	test10 := amountAsDecimalPartsTrimmed(1234590700, 2, false)
+
+	if test1[0]+"."+test1[1] != "3.14" {
+		t.Error("Wrong output, got" + test1[0] + "." + test1[1])
```
chappjc: The error message should state expected and actual.  e.g. "amountAsDecimalPartsTrimmed failed for %d (%d decimals).  Got %s.%s, expected %s.%s.".

explorer/templates_test.go
```diff
@@ -0,0 +1,59 @@
...
+	}
+
+	if test2[0]+"."+test2[1] != "765.43" {
```
chappjc: Please rewrite this test with an array of input data and expected outputs.  Along these lines:

```go
	in := []struct {
		amt    int64
		n      int64
		commas bool
	}{
		{314159000, 2, false},
		{76543210000, 2, false},
	}

	expected := []struct {
		whole, frac, tail string
	}{
		{"3", "14", ""},
		{"765", "43", ""},
	}

	for i := range in {
		out := amountAsDecimalPartsTrimmed(in[i].amt, in[i].n, in[i].commas)
		if out[0] != expected[i].whole || out[1] != expected[i].frac ||
			out[2] != expected[i].tail {
			t.Errorf("amountAsDecimalPartsTrimmed failed for "+
				"%d (%d decimals, commas=%v). Got %s.%s%s, expected %s.%s%s.",
				in[i].amt, in[i].n, in[i].commas,
				out[0], out[1], out[2],
				expected[i].whole, expected[i].frac, expected[i].tail)
		}
	}
```

explorer/templates.go
```diff
@@ -151,6 +151,39 @@ func float64Formatting(v float64, numPlaces int, useCommas bool, boldNumPlaces .
...
+		if err != nil {
+			log.Errorf("amountAsDecimalParts comma formatting failed. Input: %v Error: %v", v, err.Error())
+			integer = "ERROR"
+			dec := "VALUE"
+			zeros := ""
+			return []string{integer, dec, zeros}
+		}
+		integer = humanize.Comma(integerAsInt64)
+	}
+	dec := strings.TrimRight(amt[len(amt)-8:], "0")
```
chappjc: This function is still broken is several ways.  Fix the test and reference my previous comment on how to simplify this function and make it less error prone: https://github.com/decred/dcrdata/pull/721#discussion_r222837818

Jujhar:
Updated. Please resolve if passing all should be okay.

chappjc:
Thank you.  Please `go fmt ./...` the repo and we can merge.  Travis used to do a gofmt check, and it will again in the future when everyone has updated to Go 1.11+

Jujhar:
OK

explorer/templates_test.go
```diff
@@ -7,48 +7,48 @@ import (
...
+		{"3", "14", ""},
+		{"765", "43", ""},
+		{"766", "43", ""},
+		{"6", "5", ""},
+		{"9", "87654321", ""},
+		{"9", "87", ""},
+		{"907", "65", ""},
+		{"90765", "43", ""},
+		{"9", "0765432", ""},
+		{"12", "34", ""},
```
chappjc: Sorry that I'm just now noticing this, but we need to add at least one test where there are trailing zeros, and one with commas=true.

Jujhar: Whoops yeah right test has both now application runs as usual.

explorer/templates.go
```diff
@@ -151,6 +151,42 @@ func float64Formatting(v float64, numPlaces int, useCommas bool, boldNumPlaces .
...
+	amt := strconv.FormatInt(v, 10)
+	intpart := v / 1e8
+	decpart := v % 1e8
+
+	if len(amt) <= 8 {
+		right := strings.TrimRight(amt, "0")
+		left := ""
+		tail := strings.Repeat("0", 8-len(amt))
+		if (8 - len(amt)) > 2 {
+			tail = "00"
```
chappjc: What is this about?  If the idea is to limit to 2 zeros max, that needs to be documented.

explorer/templates.go
```diff
@@ -151,6 +151,42 @@ func float64Formatting(v float64, numPlaces int, useCommas bool, boldNumPlaces .
...
+		}
+		return []string{"0", left + right, tail}
```
chappjc: why left + right if left is ""?  Why set left="" at all?

explorer/templates.go
```diff
@@ -151,6 +151,42 @@ func float64Formatting(v float64, numPlaces int, useCommas bool, boldNumPlaces .
...
+	}
+
+	left := strconv.FormatInt(intpart, 10)
+	right := fmt.Sprintf("%08d", decpart)
+	right = string(right[0:numPlaces])
+
+	tail := ""
```
chappjc: ???
Please comment the code more.

e.g. Everything before the `if useCommas` can be simplified and made more clear and robust:


```go
func amountAsDecimalPartsTrimmed(v, numPlaces int64, useCommas bool) []string {
	// Separate into int.dec parts.
	intpart := v / 1e8
	decpart := v % 1e8

	// Format left side.
	left := strconv.FormatInt(intpart, 10)

	// Format right side with 8 digits, including leading and trailing zeros.
	rightWithTail := fmt.Sprintf("%08d", decpart)
	// Reduce precision according to numPlaces.
	if len(rightWithTail) > int(numPlaces) {
		rightWithTail = rightWithTail[0:numPlaces]
	}

	// Separate trailing zeros.
	right := strings.TrimRight(rightWithTail, "0")
	tail := strings.TrimPrefix(rightWithTail, right) // Repeat("0", len(rightWithTail) - len(rightWithTail))
```

explorer/templates.go
```diff
@@ -151,6 +151,42 @@ func float64Formatting(v float64, numPlaces int, useCommas bool, boldNumPlaces .
...
+		tail := strings.Repeat("0", 8-len(amt))
+		if (8 - len(amt)) > 2 {
+			tail = "00"
+		}
+		return []string{"0", left + right, tail}
+	}
+
+	left := strconv.FormatInt(intpart, 10)
+	right := fmt.Sprintf("%08d", decpart)
+	right = string(right[0:numPlaces])
```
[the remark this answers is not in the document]
Jujhar: No need for trimming over 8 decimals will restrict to 8 places max
```
func amountAsDecimalPartsTrimmed(v, numPlaces int64, useCommas bool) []string {

	// Filter numPlaces to only allow up to 8 decimal places trimming eg. 1.12345678
	if numPlaces > 8 {
		numPlaces = 8
	}
```

explorer/templates.go
Jujhar: Documented, it was so it would not leave value as just _1._ with nothing trailing

explorer/templates.go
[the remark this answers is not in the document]
Jujhar: Unless you have any better ideas?

explorer/templates_test.go
```diff
@@ -0,0 +1,56 @@
...
+		{"765", "43", ""},
+		{"7,664", "32", ""},
+		{"6", "5", ""},
+		{"9", "87654321", ""},
+		{"9", "87", ""},
+		{"907", "65", ""},
+		{"90765", "43", ""},
+		{"9", "0765432", ""},
+		{"12", "34", ""},
+		{"1", "00", ""},
```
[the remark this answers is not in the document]
Jujhar: Added and passing

explorer/templates.go
Jujhar: Removed left

explorer/templates.go
Jujhar: 173 no longer used

explorer/templates.go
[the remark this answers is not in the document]
chappjc: nvm, I just saw the following commit.

explorer/templates.go
```diff
@@ -151,6 +151,65 @@ func float64Formatting(v float64, numPlaces int, useCommas bool, boldNumPlaces .
...
+
+		right := strings.TrimRight(rightWithTail, "0")
+		tail := strings.TrimPrefix(rightWithTail, right)
+
+		// Do not leave value as 1. if it is case with nothing trailing
+		if (8 - len(amt)) > 2 {
+			tail = "00"
+		}
+
+		return []string{"0", right, tail}
```
chappjc: The code still needs to be simplified a log.  Almost all of the code inside this `if` statement are duplicated after it.

Note that `strconv.FormatInt(0, 10)` is the same thing as `"0"`.

explorer/templates.go
```diff
@@ -151,6 +151,65 @@ func float64Formatting(v float64, numPlaces int, useCommas bool, boldNumPlaces .
...
+	if len(rightWithTail) > int(numPlaces) {
+		rightWithTail = rightWithTail[0:numPlaces]
+	}
+
+	// Separate trailing zeros.
+	right := strings.TrimRight(rightWithTail, "0")
+	tail := strings.TrimPrefix(rightWithTail, right)
+
+	// Add commas (eg. 3,444.33)
+	if useCommas {
```
chappjc: If you wanted to avoid this then there are too few left digits (although it doesn't hurt to run it anyway), you could have it be `if useCommas && len(left)>3`

explorer/templates.go
```diff
@@ -151,6 +151,65 @@ func float64Formatting(v float64, numPlaces int, useCommas bool, boldNumPlaces .
...
+		rightWithTail := fmt.Sprintf("%08d", decpart)
+		if len(rightWithTail) > int(numPlaces) {
+			rightWithTail = rightWithTail[0:numPlaces]
+		}
+
+		right := strings.TrimRight(rightWithTail, "0")
+		tail := strings.TrimPrefix(rightWithTail, right)
+
+		// Do not leave value as 1. if it is case with nothing trailing
+		if (8 - len(amt)) > 2 {
```
chappjc: I find it confusing to be checking the length of `amt` when we have just separated the left side as `intpart` and the right as `decpart`.  With this condition, it's hard to reason about the case where there's nothing on decimal side.  Could a check on `decpart` or `rightWithTail` be more clear?

explorer/templates.go
Jujhar: Done yes I wanted to refractor this but seemed to crash live site I think for some reason before (I think because I removed it rather than moving it I don't remember).

explorer/templates.go
Jujhar: Updated.

explorer/templates.go
Jujhar: Updated

Jujhar:
OK it should be production ready now

explorer/templates.go
```diff
@@ -151,6 +151,58 @@ func float64Formatting(v float64, numPlaces int, useCommas bool, boldNumPlaces .
...
+	if len(rightWithTail) > int(numPlaces) {
+		rightWithTail = rightWithTail[0:numPlaces]
+	}
+
+	// Separate trailing zeros.
+	right := strings.TrimRight(rightWithTail, "0")
+	tail := strings.TrimPrefix(rightWithTail, right)
+
+	// For shorter values sent from system
+	amt := strconv.FormatInt(v, 10)
```
chappjc: I mentioned previously that there's no need to format `amt` and check it's length.  You can simply `if intpart == 0`

explorer/templates.go
```diff
@@ -151,6 +151,58 @@ func float64Formatting(v float64, numPlaces int, useCommas bool, boldNumPlaces .
...
+	if len(amt) <= 8 {
+		left = "0"
```
chappjc: It will already be `"0"` because of `left := strconv.FormatInt(intpart, 10)`

explorer/templates.go
```diff
@@ -151,6 +151,58 @@ func float64Formatting(v float64, numPlaces int, useCommas bool, boldNumPlaces .
...
+
+		// Do not leave value as 1. if it is case with nothing trailing
```
chappjc: But this block is only reached if left is 0.  Did you mean to put the `if (len(right)) == 0 {` checkt outside of the left=0 check?

explorer/templates.go
Jujhar: I do not think we are on the same page I am formatting it because I want to get the original input which is a string turning it into an int in the case that the number of inputs passed in is less than 8 digits in which case I assume the intpart decpart wont work.

chappjc: https://play.golang.org/p/T1_9G7qGfro

explorer/templates.go
```diff
@@ -151,6 +151,56 @@ func float64Formatting(v float64, numPlaces int, useCommas bool, boldNumPlaces .
...
+	// Separate trailing zeros.
+	right := strings.TrimRight(rightWithTail, "0")
+	tail := strings.TrimPrefix(rightWithTail, right)
+
+	// For shorter values sent from system
+	if intpart == 0 {
+
+		// Do not leave value as 1. if it is case with nothing trailing
+		if (len(right)) == 0 {
+			tail = "00"
```
chappjc: I'm still wondering if you meant to put `if (len(right)) == 0 {
			tail = "00"` by itself (not in the `if intpart == 0`).  Consider `left="1"` and `right="0"`.  It will never get to this `tail = "00"` statement because of `intpart` will be 1.

Jujhar: You are correct I don’t think this code gets reached anymore mainly kept it for the 0.0 case to be 0.00 but I think that is fine. Shall I remove?

chappjc: I do think it is a good idea to change "1." into "1.00", and "0." into "0.00".  It'll get reached in both cases by moving the len(right) check or eliminating the outer check.

PR #722, opened 2018-10-06
Display indexing subtitle notification before it starts

dmigwi:
Indexing subtitle notifications should be sent before the actual indexing starts. Currently, they are sent after indexing ends, which makes one not know what is happening in the background when the first indexing is running.


chappjc:
utACK.  Although I wouldn't say it's a "critical" fix.

PR #726, opened 2018-10-08
Register for dcrd notifications before waiting for monitors to exit

dmigwi:
Register for new blocks, new transactions and new winning tickets notifications before monitors start immediately after blockchain sync is complete.
Currently the registration is happening when the system is exiting.

### DIFF
--- a/main.go
+++ b/main.go
@@ -711,6 +711,12 @@ func mainCore() error {
 	// blocks, change in the mempool and handle chain reorg. It also initiates
 	// data collection for the explorer.
 	{
+		// Register notifications from dcrd.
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -834,13 +840,7 @@ func mainCore() error {
 		wg.Wait()
```
chappjc: Double oops.  The call to `RegisterNodeNtfnHandlers` should be fine after starting the monitors as long as it is before `wg.Wait()`.  Anyway, we've always previously called `RegisterNodeNtfnHandlers` before starting the monitors, so I don't think it's an issue as long as these things are done back-to-back.

chappjc: I've revised the PR title to reflect this.

PR #728, opened 2018-10-09
Change the API return status code from 202 to 503

dmigwi:
For the API endpoints the status 503 (Service Temporarily Unavailable)  should be used tell the user to resend the request later when the sync is complete.


dmigwi:
This works for API routes too.

chappjc:
Ah, yeah they share the same middleware.  Perfect

papacarp:
Looks good.  Thank you.

PR #729, opened 2018-10-09
dcrsqlite: properly compute blocks to check in sync

chappjc:
In dcrsqlite/sync.go, `minBlocksToCheck` was computed using the
"start height" of the sync loop, which is normally one past the best block
in the DB.  But it did not include a necessary +1.

This change instead compares node height to DB height in deciding if the
DB is already sync'd or if the chain server is behind.  `startHeight` may
still be `dbHeight` or `dbHeight+1` depending on the state of aux DB
coordination, but the computation of `minBlocksToCheck` is now correct
regardless.

### DIFF
--- a/db/dcrsqlite/sync.go
+++ b/db/dcrsqlite/sync.go
@@ -111,13 +111,13 @@ func (db *wiredDB) resyncDB(quit chan struct{}, blockGetter rpcutils.BlockGetter
 		log.Errorf("resyncDBWithPoolValue failed: %v", err)
 	}(time.Now(), &err)
 
-	// Check and report heights of the DBs. startHeight is the lowest of the
+	// Check and report heights of the DBs. dbHeight is the lowest of the
 	// heights, and may be -1 with an empty SQLite DB.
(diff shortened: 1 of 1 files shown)


db/dcrsqlite/sync.go
```diff
@@ -131,18 +131,24 @@ func (db *wiredDB) resyncDB(quit chan struct{}, blockGetter rpcutils.BlockGetter
...
-			return startHeight, fmt.Errorf("RewindStakeDB failed: %v", err)
+			return dbHeight, fmt.Errorf("RewindStakeDB failed: %v", err)
 		}
 	}
 
+	// Start syncing at or after DB height depending on whether an external
+	// MasterBlockGetter is already configured to relay the current best block,
+	// in which case we receive and discard it to maintain synchronization with
+	// the auxiliary DB.
+	startHeight := dbHeight
```
dmigwi:
```golang
dbHeight, summaryHeight, stakeInfoHeight, stakeDBHeight, err := db.DBHeights()
```
Apart from change in name, is possible for `startHeight` to be assigned a different value than was originally assigned to `dbHeight` in the code snippet above?

Issue #730, opened 2018-10-09
changing appdata does not change datadir


chappjc:
When changing appdata (`-A`/`--appdata`), this should be used as a base path for the `datadir`.  However, when `datadir` is *not* explicitly specified (`-b`/`--datadir`) and it should be putting it in the `appdata`/data folder, where `appdata` uses the user setting, `datadir` actually ends up in the default `appdata` folder

PR #736, opened 2018-10-10
config: adjust config/log/data paths if appdata is changed

chappjc:
This resolves https://github.com/decred/dcrdata/issues/730


chappjc:
This should not be added to 3.0.3 without further testing.

Issue #618, opened 2018-08-16
Detect alternate address formats in search


raedah:
If a testnet address is detected in the search box for mainnet, we could display a message "You pasted a testnet address, you can find it here." with a link redirecting the user to the testnet dcrdata instance page with their address information, and vice versa.

If a bitcoin address is pasted, we could display the message with a link to a bitcoin block explorer.

chappjc:
Resolved by https://github.com/decred/dcrdata/pull/846

PR #743, opened 2018-10-14
Detecting alternate addresses formats in search

McEdward:
Resolving issue #618

### DIFF
--- a/db/dcrsqlite/apisource.go
+++ b/db/dcrsqlite/apisource.go
@@ -1336,7 +1336,7 @@ func (db *wiredDB) GetExplorerAddress(address string, count, offset int64) (*exp
 
 	// This is here to detect a bitcoin type address
 	if (strings.HasPrefix(address, "bc") || strings.HasPrefix(address, "1") || strings.HasPrefix(address, "3")) && len(address) >= 26 {
-		return nil, fmt.Errorf("You may have pasted a Bitcoin address type on a Decred Explorer")
+		return nil, fmt.Errorf("Looks like you are searching for a bitcoin address")
 	}
 
 	addr, err := dcrutil.DecodeAddress(address)
--- a/db/dcrsqlite/apisource.go
+++ b/db/dcrsqlite/apisource.go
@@ -1365,14 +1365,14 @@ func (db *wiredDB) GetExplorerAddress(address string, count, offset int64) (*exp
 		}
 	}
 
-	maxcount := explorer.MaxAddressRows
-	txs, err := db.client.SearchRawTransactionsVerbose(addr,
-		int(offset), int(maxcount), true, true, nil)
-
 	// Dectection of when an address belonging to a different network is inputed
 	if addr.Net().Name != db.params.Name {
 		return nil, fmt.Errorf("Wrong Network. You pasted a %s address on %v", addr.Net().Name, db.params.Name)
 	}
+
+	maxcount := explorer.MaxAddressRows
+	txs, err := db.client.SearchRawTransactionsVerbose(addr,
+		int(offset), int(maxcount), true, true, nil)
 
 	if err != nil && err.Error() == "-32603: No Txns available" {
 		log.Tracef("GetExplorerAddress: No transactions found for address %s: %v", addr, err)
--- a/db/dcrsqlite/apisource.go
+++ b/db/dcrsqlite/apisource.go
@@ -1333,15 +1333,17 @@ func (db *wiredDB) GetExplorerTx(txid string) *explorer.TxInfo {
 }
 
 func (db *wiredDB) GetExplorerAddress(address string, count, offset int64) (*explorer.AddressInfo, error) {
-
-	// This is here to detect a bitcoin type address
-	if (strings.HasPrefix(address, "bc") || strings.HasPrefix(address, "1") || strings.HasPrefix(address, "3")) && len(address) >= 26 {
-		return nil, fmt.Errorf("Looks like you are searching for a bitcoin address")
-	}
-
 	addr, err := dcrutil.DecodeAddress(address)
 	if err != nil {
 		log.Infof("Invalid address %s: %v", address, err)
+		if strings.HasPrefix(address, "Dk") || strings.HasPrefix(address, "Tk") || strings.HasPrefix(address, "Sk") {
+			return nil, fmt.Errorf("Looks like you are searching for an address of type P2PK")
+		}
+
+		// This is here to detect a bitcoin type address
+		if (strings.HasPrefix(address, "bc") || strings.HasPrefix(address, "1") || strings.HasPrefix(address, "3")) && len(address) >= 25 && len(address) <= 34 {
+			return nil, fmt.Errorf("Looks like you are searching for a bitcoin address")
+		}
 		return nil, nil
 	}
 
--- a/db/dcrsqlite/apisource.go
+++ b/db/dcrsqlite/apisource.go
@@ -1344,7 +1344,7 @@ func (db *wiredDB) GetExplorerAddress(address string, count, offset int64) (*exp
 
 		// This is here to detect a bitcoin type address
 		if (strings.HasPrefix(address, "bc") || strings.HasPrefix(address, "1") || strings.HasPrefix(address, "3")) && len(address) >= 25 && len(address) <= 34 {
-			return nil, fmt.Errorf("Bitcoin address detected")
+			return nil, fmt.Errorf("Possible Bitcoin address detected.")
 		}
 		return nil, nil
 	}
--- a/db/dcrsqlite/apisource.go
+++ b/db/dcrsqlite/apisource.go
@@ -1335,41 +1335,46 @@ func (db *wiredDB) GetExplorerTx(txid string) *explorer.TxInfo {
 func (db *wiredDB) GetExplorerAddress(address string, count, offset int64) (*explorer.AddressInfo, error) {
 	addr, err := dcrutil.DecodeAddress(address)
 	if err != nil {
-		return nil, err
+		log.Infof("Invalid address %s: %v", address, err)
+		return nil, nil
 	}
 
-	// Short circuit the transaction and balance queries if the provided address
-	// is the zero pubkey hash address commonly used for zero value
-	// sstxchange-tagged outputs.
-	isDummyAddress := IsZeroHashP2PHKAddress(address, db.params)
-	if isDummyAddress {
-		return &explorer.AddressInfo{
-			Address:         address,
-			Balance:         new(explorer.AddressBalance),
-			UnconfirmedTxns: new(explorer.AddressTransactions),
-			IsDummyAddress:  true,
-			Fullmode:        true,
-		}, nil
-	}
-
-	// Handle invalid address.
-	if !ValidateNetworkAddress(addr, db.params) {
-		return nil, fmt.Errorf("wrong network: address %s is not valid for %v",
-			address, db.params.Name)
+	{
+		// Short circuit the transaction and balance queries if the provided address
+		// is the zero pubkey hash address commonly used for zero value
+		// sstxchange-tagged outputs.
+		isDummyAddress := IsZeroHashP2PHKAddress(address, db.params)
+		if isDummyAddress {
+			return &explorer.AddressInfo{
+				Address:         address,
+				Balance:         new(explorer.AddressBalance),
+				UnconfirmedTxns: new(explorer.AddressTransactions),
+				IsDummyAddress:  true,
+				Fullmode:        true,
+			}, nil
+		}
 	}
 
 	maxcount := explorer.MaxAddressRows
 	txs, err := db.client.SearchRawTransactionsVerbose(addr,
 		int(offset), int(maxcount), true, true, nil)
+	if addr.Net().Name != db.params.Name {
+		return nil, fmt.Errorf("You pasted a %s address on %v", addr.Net().Name, db.params.Name)
+	}
 	if err != nil && err.Error() == "-32603: No Txns available" {
 		log.Tracef("GetExplorerAddress: No transactions found for address %s: %v", addr, err)
+
+		if !ValidateNetworkAddress(addr, db.params) {
+			log.Warnf("Address %s is not valid for this network", address)
+			return nil, nil
+		}
 		return &explorer.AddressInfo{
 			Address:    address,
 			MaxTxLimit: maxcount,
 		}, nil
 	} else if err != nil {
-		return nil, fmt.Errorf("SearchRawTransactionsVerbose failed for address %s: %v",
-			addr, err)
+		log.Warnf("GetExplorerAddress: SearchRawTransactionsVerbose failed for address %s: %v", addr, err)
+		return nil, nil
 	}
 
 	addressTxs := make([]*explorer.AddressTx, 0, len(txs))
--- a/db/dcrsqlite/apisource.go
+++ b/db/dcrsqlite/apisource.go
@@ -1346,7 +1346,7 @@ func (db *wiredDB) GetExplorerAddress(address string, count, offset int64) (*exp
 		if (strings.HasPrefix(address, "bc") || strings.HasPrefix(address, "1") || strings.HasPrefix(address, "3")) && len(address) >= 25 && len(address) <= 34 {
 			return nil, fmt.Errorf("Possible Bitcoin address detected.")
 		}
-		return nil, nil
+		return nil, err
 	}
 
 	{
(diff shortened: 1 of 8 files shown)


McEdward:
screenshot 118
screenshot 119

views/status.tmpl
```diff
@@ -11,6 +11,16 @@
             {{if eq .StatusType "Not Found"}}
                 <h5>No matching page, block, address or transaction that could be found</h5>
                 <h5>{{.Message}}</h5>
+            {{else if eq .StatusType "Wrong Network"}}
+                {{if eq .Message "Looks like you are searching for a bitcoin address, you can "}}
+                    <h5>{{.Message}}<a href="https://insight.bitpay.com">find that here.</a></h5>
+                {{else}}
+                    {{if eq .NetName "Mainnet"}}
```
chappjc: There's also Simnet and now Regnet, so I don't think this is general enough.

db/dcrsqlite/apisource.go
```diff
@@ -1307,11 +1307,17 @@ func (db *wiredDB) GetExplorerTx(txid string) *explorer.TxInfo {
 	return tx
 }
 
-func (db *wiredDB) GetExplorerAddress(address string, count, offset int64) *explorer.AddressInfo {
+func (db *wiredDB) GetExplorerAddress(address string, count, offset int64) (*explorer.AddressInfo, error) {
+
+	// This is here to detect a bitcoin type address
+	if (strings.HasPrefix(address, "bc") || strings.HasPrefix(address, "1") || strings.HasPrefix(address, "3")) && len(address) >= 26 {
+		return nil, fmt.Errorf("Looks like you are searching for a bitcoin address")
```
chappjc: I'm not really sure we care about this TBH.  Plus the address recognition is too basic to be correct all the time.  Just "3" would be recognized as a bitcoin address...
And how is this new error going to affect the search?  Part of the search is requesting address info.  The error returned should not stop the search from checking if it is a hash or a block height (e.g. "1").

chappjc: OK, I see @raedah's comments in the original issue, so I'm not going to fight the bitcoin address thing, but the other parts of my comment apply.  The check should be smarter about the content of the `"address"` string before declaring it a bitcoin address.  And throwing an error should not interfere with search or any other functionality.

McEdward:
screenshot 121
Current error message for P2PK addresses

McEdward:
- Change the checks for bitcoin addresses and P2PK address to only when there is an error from dcrutil.DecodeAddress() for invalid address.
- Updated the way you get redirected to mainnet and testnet when a different network address is inserted, leaving just a message to notify for any simnet or regnet address

explorer/explorerroutes.go
```diff
@@ -1121,7 +1121,11 @@ func (exp *explorerUI) Search(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	if err != nil {
-		exp.StatusPage(w, "search failed", err.Error()+", you can ", WrongNetworkStatusType)
+		if err.Error() == "Looks like you are searching for an address of type P2PK" {
```
chappjc: When possible, always avoid comparing long string literals against an expected value.  This is very error prone as changing the other message will break this one.  This is why we have codes and enumerations defined at the package level.

explorer/explorerroutes.go
```diff
@@ -1121,7 +1121,11 @@ func (exp *explorerUI) Search(w http.ResponseWriter, r *http.Request) {
...
+			exp.StatusPage(w, "search failed", err.Error(), NotFoundStatusType)
```
chappjc: Instead of doing exp.StatusPage followed by return here and outside the if, just set the status type in a variable first, then change the variable for this condition, then do exp.Status page and return once.

views/status.tmpl
```diff
@@ -12,13 +12,15 @@
...
             {{else if eq .StatusType "Wrong Network"}}
-                {{if eq .Message "Looks like you are searching for a bitcoin address, you can "}}
-                    <h5>{{.Message}}<a href="https://insight.bitpay.com">find that here.</a></h5>
+                {{if eq .Message "Looks like you are searching for a bitcoin address"}}
+                    <h5>{{.Message}}, you can <a href="https://insight.bitpay.com">find that here.</a></h5>
                 {{else}}
-                    {{if eq .NetName "Mainnet"}}
-                        <h5>{{.Message}}<a href="http://testnet.dcrdata.org/">switch to testnet here.</a></h5>
+                    {{if redirectToTestnet .NetName .Message}}
+                        <h5>{{.Message}}, you can <a href="http://testnet.dcrdata.org/">switch to testnet here.</a></h5>
```
chappjc: {{.Message}}. You can ...

It is a new sentence ^

views/status.tmpl
```diff
@@ -12,13 +12,15 @@
                 <h5>No matching page, block, address or transaction that could be found</h5>
                 <h5>{{.Message}}</h5>
             {{else if eq .StatusType "Wrong Network"}}
-                {{if eq .Message "Looks like you are searching for a bitcoin address, you can "}}
-                    <h5>{{.Message}}<a href="https://insight.bitpay.com">find that here.</a></h5>
+                {{if eq .Message "Looks like you are searching for a bitcoin address"}}
```
chappjc: The long string comparison is an issue here too.

views/status.tmpl
chappjc: We should build the link with the address in it.  If it is not available on the status page template, it can be passed in the data structure.

db/dcrsqlite/apisource.go
```diff
@@ -1308,15 +1308,17 @@ func (db *wiredDB) GetExplorerTx(txid string) *explorer.TxInfo {
 }
 
 func (db *wiredDB) GetExplorerAddress(address string, count, offset int64) (*explorer.AddressInfo, error) {
-
-	// This is here to detect a bitcoin type address
-	if (strings.HasPrefix(address, "bc") || strings.HasPrefix(address, "1") || strings.HasPrefix(address, "3")) && len(address) >= 26 {
-		return nil, fmt.Errorf("Looks like you are searching for a bitcoin address")
-	}
-
 	addr, err := dcrutil.DecodeAddress(address)
```
chappjc: A valid p2pk address will not cause an error.

chappjc:
For ref: https://github.com/dcrlabs/pktopkhaddr
No need to import this, just copy/paste.

McEdward:
New Additions
- Changes to how the link works for wrong network (Now makes a search for the search string on the appropriate network)
- Adding a link when a P2PK address is given. This does the conversion to P2PKH and makes a new search for the new address
- Removing the comparing of long strings and other related changes

views/status.tmpl
```diff
@@ -11,6 +11,31 @@
...
+                        {{end}}
+                    </h5>
+                {{else}}
+                    <h5>{{.Message}}</h5>
+                {{end}}
+            {{else if eq .StatusType "Bitcoin Wrong Network"}}
+                <h5>{{.Message}}. You can <a href="https://insight.bitpay.com">find that here.</a></h5>
+            {{else if eq .StatusType "P2PK Address Type"}}
+                <h5>{{.Message}}. 
+                    {{if and (ne .AdditionalInfo "") (ne (PKAddr2PKHAddr .AdditionalInfo) "")}}
```
chappjc: I'd prefer not to run the function twice.  Perhaps a variable?

db/dcrsqlite/apisource.go
```diff
@@ -1307,11 +1307,21 @@ func (db *wiredDB) GetExplorerTx(txid string) *explorer.TxInfo {
...
+	if strings.HasPrefix(address, "Dk") || strings.HasPrefix(address, "Tk") || strings.HasPrefix(address, "Sk") {
+		return nil, fmt.Errorf("P2PK address detected")
+	}
 	if err != nil {
 		log.Infof("Invalid address %s: %v", address, err)
-		return nil
+
+		// This is here to detect a bitcoin type address
+		if (strings.HasPrefix(address, "bc") || strings.HasPrefix(address, "1") || strings.HasPrefix(address, "3")) && len(address) >= 25 && len(address) <= 34 {
+			return nil, fmt.Errorf("Bitcoin address detected")
```
chappjc: Possible Bitcoin address detected.

We do no validation of the suspected bitcoin address, so we should soften our language.

views/status.tmpl
```diff
@@ -8,6 +8,7 @@
     <div class="container" data-controller="status">
         <br/>
         <div class="alert alert-info">
+            {{$P2PKHAddr := PKAddr2PKHAddr .AdditionalInfo}}
```
chappjc: I think this should be done in the ` {{else if eq .StatusType "P2PK Address Type"}}` block where it applies only.

db/dcrsqlite/apisource.go
```diff
@@ -1309,42 +1309,60 @@ func (db *wiredDB) GetExplorerTx(txid string) *explorer.TxInfo {
...
-			Balance:         new(explorer.AddressBalance),
-			UnconfirmedTxns: new(explorer.AddressTransactions),
-			IsDummyAddress:  true,
-			Fullmode:        true,
-		}, nil
+		// This is here to detect a bitcoin type address
+		if (strings.HasPrefix(address, "bc") || strings.HasPrefix(address, "1") || strings.HasPrefix(address, "3")) && len(address) >= 25 && len(address) <= 34 {
+			return nil, fmt.Errorf("Possible Bitcoin address detected.")
+		}
+		return nil, nil
```
chappjc: This needs to return a non-nil `error` or the address page will try to display for whatever this is.

db/dcrsqlite/apisource.go
```diff
@@ -1309,42 +1309,60 @@ func (db *wiredDB) GetExplorerTx(txid string) *explorer.TxInfo {
...
+				Balance:         new(explorer.AddressBalance),
+				UnconfirmedTxns: new(explorer.AddressTransactions),
+				IsDummyAddress:  true,
+				Fullmode:        true,
+			}, nil
+		}
 	}
 
 	maxcount := explorer.MaxAddressRows
 	txs, err := db.client.SearchRawTransactionsVerbose(addr,
```
chappjc: Using `GetExplorerAddress` even in full mode is not an elegant solution, mainly because we really don't want to call the RPC `SearchRawTransactionsVerbose`.

Instead, the address validation and network checking code should be factored out into an appropriate common location.  Maybe txhelpers or another place any packages that need it can use it without creating circular dependencies.

db/dcrsqlite/apisource.go
```diff
@@ -1309,42 +1309,60 @@ func (db *wiredDB) GetExplorerTx(txid string) *explorer.TxInfo {
...
 		int(offset), int(maxcount), true, true, nil)
+
+	// Dectection of when an address belonging to a different network is inputed
+	if addr.Net().Name != db.params.Name {
```
chappjc: Note that this could/should be before `SearchRawTransactionsVerbose` anyway, so when pulling out the validation/checking code, this can come too.

db/dcrsqlite/apisource.go
```diff
@@ -1309,42 +1309,60 @@ func (db *wiredDB) GetExplorerTx(txid string) *explorer.TxInfo {
...
+		return nil, fmt.Errorf("Wrong Network. You pasted a %s address on %v", addr.Net().Name, db.params.Name)
+	}
+
 	if err != nil && err.Error() == "-32603: No Txns available" {
 		log.Tracef("GetExplorerAddress: No transactions found for address %s: %v", addr, err)
+
+		if !ValidateNetworkAddress(addr, db.params) {
```
chappjc: Same with this.  `addr` and `params` are general pieces of data, and not generated by `SearchRawTransactionsVerbose`

explorer/explorerroutes.go
```diff
@@ -873,19 +873,28 @@ func (exp *explorerUI) AddressPage(w http.ResponseWriter, r *http.Request) {
...
-		addrData, err = exp.blockData.GetExplorerAddress(address, limitN, offsetAddrOuts)
-		if err != nil && strings.HasPrefix(err.Error(), "wrong network") {
-			exp.StatusPage(w, wrongNetwork, "That address is not valid for "+exp.NetName, NotSupportedStatusType)
-			return
-		}
-		if err != nil {
-			log.Errorf("Unable to get address %s: %v", address, err)
-			exp.StatusPage(w, defaultErrorCode, "Unexpected issue locating data for that address.", ErrorStatusType)
-			return
+	addrData, err = exp.blockData.GetExplorerAddress(address, limitN, offsetAddrOuts)
```
chappjc: Calling this for full mode is the issue to which I referred.

db/dcrsqlite/apisource.go
```diff
@@ -1309,42 +1309,60 @@ func (db *wiredDB) GetExplorerTx(txid string) *explorer.TxInfo {
...
-	isDummyAddress := IsZeroHashP2PHKAddress(address, db.params)
-	if isDummyAddress {
-		return &explorer.AddressInfo{
-			Address:         address,
-			Balance:         new(explorer.AddressBalance),
-			UnconfirmedTxns: new(explorer.AddressTransactions),
-			IsDummyAddress:  true,
-			Fullmode:        true,
-		}, nil
+	{
```
chappjc: Why have this block?

db/dcrsqlite/apisource.go
```diff
@@ -1309,42 +1309,60 @@ func (db *wiredDB) GetExplorerTx(txid string) *explorer.TxInfo {
...
+		}
 	}
 
-	// Handle invalid address.
-	if !ValidateNetworkAddress(addr, db.params) {
-		return nil, fmt.Errorf("wrong network: address %s is not valid for %v",
-			address, db.params.Name)
+	// Dectection of when an address belonging to a different network is inputed
+	if addr.Net().Name != db.params.Name {
+		return nil, fmt.Errorf("Wrong Network. You pasted a %s address on %v", addr.Net().Name, db.params.Name)
```
chappjc: `error`s should not start with caps.  Also, I would prefer not to say things like "You pasted".  Firstly, the "You" part should be something less personal, just stating that what condition got us here (address is valid on network X, not network Y).  Also, the "pasted" action is too closely tied to UI, when a program or other package could simply be using `(*wiredDB).GetExplorerAddress)` in whatever way they choose.

db/dcrsqlite/apisource.go
chappjc: I don't understand why you are validating the address *after* calling `SearchRawTransactionsVerbose`.  This should happen much farther up in this function.

explorer/explorerroutes.go
```diff
@@ -835,7 +835,26 @@ func (exp *explorerUI) AddressPage(w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		st := WrongNetworkStatusType
+		message := err.Error()
+		if err.Error() == "P2PK address detected" {
+			st = P2PkAddressStatusTypes
+			message = "Looks like you are searching for an address of type P2PK"
+		} else if err.Error() == "Possible Bitcoin address detected." {
+			st = BitcoinWrongNetworkStatusType
+			message = "Looks like you are searching for a bitcoin address"
+		} else if strings.HasPrefix(err.Error(), "Wrong Network") {
```
chappjc: Please replace `"Wrong Network"` with `WrongNetworkStatusType` globally, if possible.

txhelpers/txhelpers.go
```diff
@@ -892,3 +893,28 @@ func TotalVout(vouts []dcrjson.Vout) dcrutil.Amount {
...
+		// This is here to detect a bitcoin type address
+		if (strings.HasPrefix(address, "bc") || strings.HasPrefix(address, "1") || strings.HasPrefix(address, "3")) && len(address) >= 25 && len(address) <= 34 {
+			return fmt.Errorf("Possible Bitcoin address detected.")
+		}
+		return err
+	}
+
+	// Dectection of when an address belonging to a different network is inputed
+	if strings.ToLower(addr.Net().Name) != strings.ToLower(netName) {
+		return fmt.Errorf("Wrong Network. You pasted a %s address on %v", addr.Net().Name, netName)
```
chappjc: I see now that the error has to start like this because it is checked against `WrongNetworkStatusType`.

db/dcrsqlite/apisource.go
```diff
@@ -1309,42 +1309,60 @@ func (db *wiredDB) GetExplorerTx(txid string) *explorer.TxInfo {
 
 func (db *wiredDB) GetExplorerAddress(address string, count, offset int64) (*explorer.AddressInfo, error) {
 	addr, err := dcrutil.DecodeAddress(address)
+
+	// checks for P2Pk type addresses
+	if strings.HasPrefix(address, "Dk") || strings.HasPrefix(address, "Tk") || strings.HasPrefix(address, "Sk") {
```
chappjc: I've been ignoring this line, but the proper way to do this is via CheckDecode:

```go
~, netID, err := base58.CheckDecode(addr)
if err != nil {
	if err == base58.ErrChecksum {
		return nil, dcrutil.ErrChecksumMismatch
	}
	return nil, fmt.Errorf("decoded address is of unknown format: %v", err.Error())
}

switch netID {
case net.PubKeyAddrID:
	...
case net.PubKeyHashAddrID:
	...
case net.PKHEdwardsAddrID:
	...
case net.PKHSchnorrAddrID:
	...
case net.ScriptHashAddrID:
	...
default:
	// dcrutil.ErrUnknownAddressType
}
```

McEdward:
Re-basing doesn't seem to be going well on my side. My recently made commit doesn't seem to even be here after the rebase. I might need to close this PR and start afresh.

chappjc:
This is a start. I'm wrapping it up on my end, and will push to a new PR including both @McEdward's and my own changes.

PR #745, opened 2018-10-17
side chain import fixes and block sync/store refactoring

chappjc:
These changes focus on dcrpg and correcting the side chain import process.  However, there is a considerable amount of added documentation, and tangential refactoring that supports these changes.

An key change is support for 3 different insert statements in several tables:
1. Insert without checking for constraint conflicts.  This can result in a constraint violation error, and is intended to be used during initial sync (before creation of the unique indexes).
2. Upsert (insert or update existing).  If there is a constraint conflict, this is considered an existing record, which will be updated with the upsert statement.  The updates columns are generally sidechain/valid flags, or other data that may change for side chains.
3. Insert or do nothing on conflict.  If there is no constraint conflict, the new record is inserted, otherwise there is no insert or update.  This statement will still return the row id of the conflicting record.

To support the 3 insert strategies where appropriate, the `MakeXInsertStatement` functions (e.g. `MakeAddressRowInsertStatement`) are modified with a flag to indicate how to handle conflicts (upsert or do nothing).

`ChainDB`'s `StoreBlock` and `storeTxns` functions will take one of the strategies depending on where and how they are called:
a. Initial start-up sync (before index creation): The insert without checks strategy (1) is used.
b. Start-up sync (with existing indexes): The upsert strategy (2) is used.
c. Regular block storage in response to new block notifications from dcrd: The upsert strategy (2) is used.
d. Side chain block insertion: The insert or do nothing strategy (3) is used.
e. Reorganization, when side chain is moved to main chain (switchToSideChain): The upsert strategy (2) is used since these are changing to main chain blocks.  Moving the old main chain to side chain (TipToSideChain) does not use StoreBlock, instead updating flags without performing insertions.

The **insert or do nothing strategy** (3) is necessary for side chain block insertion because when side chain blocks are imported and stored, dcrdata has synchronized with dcrd's best block, and the imported blocks may already have main chain entries.  These main chain entries should not be modified (e.g. vins and addresses table rows have is_mainchain flags that indicate if the record exists in a main chain block).

A major fixed issue for the import side chain operation is that after side chain import, which can take a few minutes, there was not a final check for new blocks prior to starting notification monitors. This resulted in missed blocks, and inconsistencies in the ticket pool. The introduction of the `ensureSync` closure, and calling it after indexing and again after importing side blocks (right before staring ntfn handlers) ensures dcrdata is in sync with the node before beginning normal operation.

Other related fixes:
- Fix MissingSideChainBlocks to start at lowest block in side chain, and count up to the tip. Document the expected order in rpcutils.SideChainFull.
- Fix InsertVotes to always close both prepared statements and rollback on an error.
- Return an error from GetExplorerAddress so errors can be handled according to context (e.g. when called from Search, it is not appropriate to log most errors and failures to find addresses since the search string can be garbage from the user).
- Only after storing a block received in response to a notification will the project fund update be run asynchronously.  In SyncChainDB, run the project fund update after the last block is stored, but do it
synchronously so it does not interfere with subsequent start-up operations such as importing side chain blocks.

NOTE: Blocks from imported side chains are not added to SQLite since the tables use height as primary key (should be hash) and they do not even have flags for is_mainchain or is_valid.  This should be addressed in a future release.

### DIFF
--- a/api/apiroutes.go
+++ b/api/apiroutes.go
@@ -79,7 +79,7 @@ type DataSourceLite interface {
 	GetAddressTransactionsWithSkip(addr string, count, skip int) *apitypes.Address
 	GetAddressTransactionsRawWithSkip(addr string, count, skip int) []*apitypes.AddressTxRaw
 	SendRawTransaction(txhex string) (string, error)
-	GetExplorerAddress(address string, count, offset int64) *explorer.AddressInfo
+	GetExplorerAddress(address string, count, offset int64) (*explorer.AddressInfo, error)
 }
 
 // DataSourceAux specifies an interface for advanced data collection using the
--- a/cmd/rebuilddb2/rebuilddb2.go
+++ b/cmd/rebuilddb2/rebuilddb2.go
@@ -336,9 +336,10 @@ func mainCore() error {
 		}
 
 		var numVins, numVouts int64
-		isValid, isMainchain := true, true
+		isValid, isMainchain, updateExistingRecords := true, true, true
 		numVins, numVouts, err = db.StoreBlock(block.MsgBlock(), winners,
-			isValid, isMainchain, cfg.AddrSpendInfoOnline, !cfg.TicketSpendInfoBatch)
+			isValid, isMainchain, updateExistingRecords,
+			cfg.AddrSpendInfoOnline, !cfg.TicketSpendInfoBatch)
 		if err != nil {
 			return fmt.Errorf("StoreBlock failed: %v", err)
 		}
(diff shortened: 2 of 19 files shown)


chappjc:
PR description updated.

chappjc:
The optimizations in the last commit were very effective.  Full-mode block sync (excluding indexing) completed in under an hour, with the processing rate barely dropping from the starting rate:

```
[INF] PSQL: ( 86 blk/s, 1493 tx/s, 3336 vin/sec, 5455 vout/s)
[INF] PSQL: Processing blocks 284001 to 284084...
[INF] SQLT: Scanning blocks 284001 to 284084 (41012 live)...
[INF] SQLT: Rescan finished successfully at height 284084.
[INF] SQLT: resyncDBWithPoolValue completed in 57m6.225612785s  ****
[INF] DATD: SQLite sync ended at height 284084
[INF] PSQL: Avg. speed: 1469 tx/s, 5444 vout/s  ****
```

I'm going to pull this optimization commit out into a separate PR.

PR #749, opened 2018-10-17
update sample configs

chappjc:
nginx.conf: update TLS, ciphers, dhparam, ecdh_curve, etc.
dcrdata.conf: list current logging subsystems and add import-side-chains


chappjc:
While I'm updating the sample nginx.config, @dajohi could you please have a look at the ssl stuff?

Issue #538, opened 2018-07-04
Window list view


raedah:
/windows, similar to the /blocks , but with data grouped by window number (groups of 144 blocks on mainnet). Rows will list data useful for windows, such as sdiff, total tickets purchased, total transactions. The window # will be clickable and go to a page.. either /window/x  or  /blocks?height=x&rows=144 that shows the block for that window.  On the /blocks view, we will want to add a column specifying the window number.

Similar views can also be added for block groupings of  /days   /months   /years

chappjc:
Resolved by https://github.com/decred/dcrdata/pull/750

PR #750, opened 2018-10-17
Window list view

dmigwi:
Based on issue https://github.com/decred/dcrdata/issues/538.

A window list view on route /window is introduced.

screen shot 2018-10-18 at 00 30 35

Also changed the block page navigation from:
screen shot 2018-10-17 at 07 37 40

To:
screen shot 2018-10-17 at 07 36 00

### DIFF
--- a/db/dbtypes/conversion.go
+++ b/db/dbtypes/conversion.go
@@ -91,3 +91,20 @@ func ChartGroupingToInterval(grouping ChartGrouping) (float64, error) {
 func CalculateHashRate(difficulty, targetTimePerBlock float64) float64 {
 	return ((difficulty * math.Pow(2, 32)) / targetTimePerBlock) / 1000000
 }
+
+// CalculateWindowIndex calculates the window index from the quotient of a block
+// height and the chainParams.StakeDiffWindowSize. A window being a group of
+// blocks whose count is defined by chainParams.StakeDiffWindowSize, the first
+// window starts from block 1 to block 144 instead of block 0 to block 143.
+// To obtain the accurate window index value, we should add 1 to the quotient
+// obtained by dividing the block height with the chainParams.StakeDiffWindowSize
+// value; if the float precision is greater than zero. The precision is equal to
+// zero only when the block height value is divisible by the window size.
+func CalculateWindowIndex(height, stakeDiffWindowSize int64) int64 {
+	windowVal := float64(height) / float64(stakeDiffWindowSize)
+	index := int64(windowVal)
+	if windowVal != math.Floor(windowVal) || windowVal == 0 {
+		index += 1
+	}
+	return index
+}
--- a/db/dbtypes/conversion.go
+++ b/db/dbtypes/conversion.go
@@ -93,14 +93,15 @@ func CalculateHashRate(difficulty, targetTimePerBlock float64) float64 {
 }
 
 // CalculateWindowIndex calculates the window index from the quotient of a block
-// height and the chainParams.StakeDiffWindowSize. A window being a group of
-// blocks whose count is defined by chainParams.StakeDiffWindowSize, the first
-// window starts from block 1 to block 144 instead of block 0 to block 143.
-// To obtain the accurate window index value, we should add 1 to the quotient
-// obtained by dividing the block height with the chainParams.StakeDiffWindowSize
-// value; if the float precision is greater than zero. The precision is equal to
-// zero only when the block height value is divisible by the window size.
+// height and the chainParams.StakeDiffWindowSize.
 func CalculateWindowIndex(height, stakeDiffWindowSize int64) int64 {
+	// A window being a group of blocks whose count is defined by
+	// chainParams.StakeDiffWindowSize, the first window starts from block 1 to
+	// block 144 instead of block 0 to block 143. To obtain the accurate window
+	// index value, we should add 1 to the quotient obtained by dividing the block
+	// height with the chainParams.StakeDiffWindowSize value; if the float precision
+	// is greater than zero. The precision is equal to zero only when the block
+	// height value is divisible by the window size.
 	windowVal := float64(height) / float64(stakeDiffWindowSize)
 	index := int64(windowVal)
 	if windowVal != math.Floor(windowVal) || windowVal == 0 {
--- a/db/dbtypes/types.go
+++ b/db/dbtypes/types.go
@@ -117,14 +117,17 @@ type ProgressBarLoad struct {
 	Timestamp int64
 }
 
-// BlocksGroupedInfo contains the grouped information about the fetched window.
+// BlocksGroupedInfo contains the data about a stake difficulty (ticket price) window,
+// including intrinsic properties (e.g. window index, ticket price, start block, etc.),
+// and aggregate transaction counts (e.g. number of votes, regular transactions,
+// new tickets, etc.)
 type BlocksGroupedInfo struct {
-	Window        int64
-	StartBlock    int64
-	Voters        uint16
-	Transactions  uint16
-	FreshStake    uint16
-	Revocations   uint16
+	WindowIndx    int64
+	EndBlock      int64
+	Voters        uint64
+	Transactions  uint64
+	FreshStake    uint64
+	Revocations   uint64
 	BlocksCount   int64
 	Difficulty    float64
 	TicketPrice   int64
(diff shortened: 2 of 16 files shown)


dmigwi:
/blocks page
screen shot 2018-10-18 at 01 09 33

xaur:
Please add a description of what a Window is, e.g.:

windows-desc

xaur:
The "Also changed the block page navigation from" is a change in a different view and perhaps belongs to a separate PR.

views/windows.tmpl
```diff
@@ -0,0 +1,93 @@
...
+        <div class="mb-1 fs13">
+            {{if gt $windowsCount 0}}
+            {{ $topWindow := (int64 (index .Data 0).Window) }}
+            <div class="d-flex flex-wrap-reverse align-items-center justify-content-end list-display">
+                <span class="fs12 nowrap text-right">
+                    {{intComma (add $.OffsetWindow 1)}} &mdash; {{intComma (add $.OffsetWindow $windowsCount)}} of {{ intComma (add .BestWindow 1) }} windows
+                </span>
+                <nav aria-label="blocks navigation" data-limit="{{.Limit}}" class="ml-2">
+                    <ul class="pagination mb-0 pagination-sm">
+                        <li class="page-item {{if le (index .Data (subtract $windowsCount 1)).Window 1}}disabled{{end}}">
```
chappjc: We need to be super careful with these `index` statements as it's a good way to cause a panic.  I didn't think through the logic here to know if the value `(subtract $windowsCount 1)` will ever be out of range of `.Data`, but we need to look closely.

dmigwi: `{{if gt $windowsCount 0}}` This line takes care when the `.Data` is empty.

dmigwi:
@xaur The blocks page navigation issue is not within this task's scope but I feel its small enough not deserve its own PR. I also got to make a number of changes on the blocks page that are directly related to this PR so IMO I think we can keep the blocks page navigation change here.

db/dbtypes/types.go
```diff
@@ -117,6 +117,22 @@ type ProgressBarLoad struct {
...
+type BlocksGroupedInfo struct {
+	Window        int64
+	StartBlock    int64
+	Voters        uint16
+	Transactions  uint16
+	FreshStake    uint16
+	Revocations   uint16
+	BlocksCount   int64
+	Difficulty    float64
+	TicketPrice   int64
```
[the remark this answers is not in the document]
chappjc: I didn't mean grouping them with another struct, just order in this struct.  It's a very minor issue.

chappjc:
The blocks navigation thing is small, but it should still be in a separate PR, primarily because it deserves discussion.

db/dbtypes/conversion.go
```diff
@@ -91,3 +91,20 @@ func ChartGroupingToInterval(grouping ChartGrouping) (float64, error) {
...
 }
+
+// CalculateWindowIndex calculates the window index from the quotient of a block
+// height and the chainParams.StakeDiffWindowSize. A window being a group of
+// blocks whose count is defined by chainParams.StakeDiffWindowSize, the first
+// window starts from block 1 to block 144 instead of block 0 to block 143.
+// To obtain the accurate window index value, we should add 1 to the quotient
+// obtained by dividing the block height with the chainParams.StakeDiffWindowSize
+// value; if the float precision is greater than zero. The precision is equal to
+// zero only when the block height value is divisible by the window size.
```
chappjc: It is good to have these kinds of comments on the implementation, but they should go inside the function.  The comment that begins with `// FunctionName` is special because it is used to describe the function's purpose and the meaning of the parameters to potential consumers of the function.  When reading https://godoc.org/github.com/decred/dcrdata/api/types, you need to think if the comment on the function is helpful in the context of API documentation.

db/dbtypes/types.go
```diff
@@ -117,6 +117,22 @@ type ProgressBarLoad struct {
...
+	StartTime     int64
+	FormattedTime string
+	FormattedSize string
```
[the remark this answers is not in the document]
chappjc: Please add `Size int64` as you did for `StartTime`.

db/dcrpg/pgblockchain.go
```diff
@@ -696,6 +696,11 @@ func (pgb *ChainDB) GetTicketPoolByDateAndInterval(maturityBlock int64,
 	return retrieveTicketsByDate(pgb.db, maturityBlock, int64(val))
 }
 
+// WindowBlocks retrieves the blocks at the respective stakebase windows interval.
+func(pgb *ChainDB) WindowBlocks(limit, offset uint64) ([]*dbtypes.BlocksGroupedInfo, error) {
```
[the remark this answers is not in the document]
chappjc: I still do not care for the function name.  We may end up wanting a similar function for PoW diff windows, or rule change intervals, etc. and those will necessarily have more descriptive names.  This one should be specific too.

explorer/explorertypes.go
```diff
@@ -35,6 +35,7 @@ var blockchainSyncStatus = new(syncStatus)
 
 // BlockBasic models data for the explorer's explorer page
 type BlockBasic struct {
+	WindowIndx     int64  `json:"windowIndex"`
```
chappjc: Please move this down below next to `FreshStake`, but mainly just not at the top.  Also, same comment about window ambiguity.

main.go
```diff
@@ -516,6 +516,7 @@ func mainCore() error {
 
 		r.Mount("/explorer", explore.Mux)
 		r.Get("/blocks", explore.Blocks)
+		r.Get("/windows", explore.StakeDiffWindows)
```
chappjc: Same thing with "/windows" needing to be something else like "/pos-intervals" unless we intend to have /windows describe all kinds of Decred windows/intervals.

views/explorer.tmpl
```diff
@@ -6,33 +6,46 @@
...
+            <span class="fs12 nowrap text-right list-display">no confirmed blocks found</span>
+            {{end}}
         </div>
 
         <div class="row">
             <div class="col-md-12">
                 <table class="table striped table-responsive" id="explorertable">
                     <thead>
                         <tr>
+                            <th>Window Index</th>
```
chappjc: Same here.  This should be called "Ticket Price Window Number"

views/windows.tmpl
```diff
@@ -0,0 +1,145 @@
+{{define "windows"}}
+<!DOCTYPE html>
+<html lang="en">
+
+{{template "html-head" "Decred Blocks Windows"}}
+<body class="{{ theme }}">
+    {{template "navbar" . }}
+    <div class="container main" data-controller="main">
+        <h4>Windows</h4><h6>A window is a range of {{$.WindowSize}} blocks or approximately 12hrs. After each window mining; difficulty and ticket price are adjusted.</h6>
```
chappjc: This page is for PoS price windows, not PoW, correct?  The work mining diff window is set separately (`WorkDiffWindowSize`)

main.go
raedah: The window is for both PoS and PoW. What other kinds of windows are we considering? Voting activation? I dont think those would get confused here, but in any case will still need a simple url if /windows is not to be used.  /diffwindows ? The url should also match the page title which is still set to Windows.

views/windows.tmpl
raedah: There are two variables, but for all 4 net types they are set to identical values. Both sdiff and diff value can be shown on the windows list.

main.go
chappjc: As I pointed out in (my other comment)[https://github.com/decred/dcrdata/pull/750#discussion_r227439237], they need not be the same even though they currently are.   But I can assume they will always be for this PR.

However, another windows I am talking about is the block reward.  The subsidy windows/intervals.

views/windows.tmpl
chappjc: Hopefully they are never changed.

views/explorer.tmpl
chappjc: Disregard if this is going to be for both difficulty windows.  Although personally I would still make this difficulty window index.

dmigwi:
Here is the latest update.
screenshot from 2018-10-25 23-43-20

if the rows count is 20 or less than only one navigation widget that appears on top otherwise a second navigation widget appears at the bottom

dmigwi:
It Introduces a new page "/ticketpricewindows" that lists all the ticket price windows in descending order.
Every window row has intrinsic properties (end block, window number, ticket price, difficulty, start time and age) and aggregate properties(transactions, tickets, votes, revocations and total size).

Uses a new page navigation on both "/blocks" and "/ticketpricewindows"

dmigwi:
:point_up: @chappjc

Issue #758, opened 2018-10-24
Charts on address pages fails to load


jholdstock:
1. Visit the page for an address, eg: https://explorer.dcrdata.org/address/DsjZTtatQUatzhUGdwHtdCB24R9yMC7JpY3
2. Click the "Charts" button
3. Click the "Zoom" > "Day" button

Infinite spinning wheel is displayed, and a Javascript error is printed to the console:

```
stimulus.umd.js:5 
Error invoking action "click->address#onZoom"

TypeError: Cannot read property 'resetZoom' of undefined
    at e.onZoom (address.js:226)
    at t.invokeWithEvent (stimulus.umd.js:5)
    at t.handleEvent (stimulus.umd.js:5)

{identifier: "address", controller: e, element: div.container.main, event: MouseEvent}
```

buck54321:
It looks like it's just a missing check for an initialized chart.

Other than hiding the controls until the chart is initialized, should be find to just add

```
if (this.graph == undefined) {
    return
}
```
at line 225 in public/js/controllers/address.js

I'm sort of new to github, so I'm not certain whether I should create a pull request on an issue that has been assigned. I don't want to step on anybody's toes.

dmigwi:
@buck54321 Sure please do. I couldn't exactly replicate the bug above but if you have found a fix please create a PR for it.

chappjc:
Resolved by PR #760

PR #760, opened 2018-10-25
fix chart zoom for uninitialized charts

buck54321:
bug in address.js cause the loading animation to hang when chart isn't initialized

### DIFF
--- a/public/js/controllers/address.js
+++ b/public/js/controllers/address.js
@@ -222,6 +222,9 @@
         }
 
         onZoom(){
+            if (this.graph == undefined) {
+                return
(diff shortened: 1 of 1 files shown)


chappjc:
Thanks for the PR.  Could you please squash and rebase the branch?

buck54321:
I think I corrected it, but I'm not certain.

chappjc:
Yup, that did it, thanks!

buck54321:
Oh. I think I messed this up. This was for https://github.com/decred/dcrdata/issues/758.

PR #762, opened 2018-10-25
Disabled buttons out of range

dmigwi:
Disabling of buttons out of range wasn't working

**Before**
before

**After**
after fix


dmigwi:
A bug fix that disables respective "zoom" and "group by" buttons that are out of range(too huge of a grouping) in relation to the oldest transaction timestamp for a given address .

PR #714, opened 2018-09-29
'confirms' formatting on websocket update in views/transaction

ademuanthony:
This is a fix for https://github.com/decred/dcrdata/issues/637
Modified the `updateBlockData` function in the extras.tmpl file to append the parenthesis and the text 'confirmations' during WebSocket update.
the span with `data-confirmation-block-height` attribute should no longer the wrapped in parenthesis as the function will provide the full content during an update

Issue #637, opened 2018-08-27
views/transaction: confirms formatting incorrect on websocket update


chappjc:
On the initial load of the transaction page, the top will say either:

> Transaction ( unconfirmed )
> Transaction (x confirmations )

where x is the number of confirmations, x>0.

When a block is mined, these will change to:

> Transaction x

The parenthesis and the "confirmations" text disappear.

When addressing this bug, the formatting should be cleaned up (there are extra/missing spaces abound).

chappjc:
resolved by https://github.com/decred/dcrdata/pull/714

Issue #603, opened 2018-08-10
partial websocket updates to addresses with unconfirmed


chappjc:
When an address page loads with an unconfirmed transaction, it rightly shows the unconfirmed one at the top of the list with `UNCONFIRMED   1 transactions` by the sums.  But it gets confirmed, the table and the top of the page are not updated appropriately:

![image](https://user-images.githubusercontent.com/9373513/43969071-998cd72e-9c8e-11e8-94ba-1004a9327f19.png)

1. The unconfirmed count at the top.
2. The "Credit DCR" and "Time UTC" columns remain as "N/A" and "Unconfirmed" although the "Confirms" column is correctly updated.

dmigwi:
Fixed here https://github.com/decred/dcrdata/pull/597/commits/3f70e4fefb9558c9af1ffc0a48617ff9fd638ea1

dmigwi:
What is fixed in that commit was the aligning of the transactions in the row Credit DCR.

chappjc:
The updates of the actual credit and amount and Time UTC still need fixing though.

ademuanthony:
I am working on this

buck54321:
It's actually a little more subtle, it seems. This is incrementing on every block no matter what. So if an unconfirmed transaction is displayed, and a block is created that **does not** include the transaction, the confirmation counter increments anyway.

chappjc:
@ademuanthony Please coordinate with @buck54321 on the confirmations issue (both the formatting issue and the counting issue).  @buck54321 are you on Slack or matrix?  There is a #dcrdata channel where we often discuss.

ademuanthony:
From my understanding of the confirmation counter.
The confirmation count for a transaction is gotten the difference between its block height and the explorer block height, which means whenever a new block is mined (which will result in this web socket signal), the confirmation count for all transactions on this page will increase by at least 1 and therefore the number of the unconfirmed transaction will become zero. This implies that we will hide the unconfirmed transactions count after the web socket update.
This is what I have done in this [PR](https://github.com/decred/dcrdata/pull/766) and that should resolve this issue.
I have also worked on the Time UTC update in the same pull request since they all follow the same set of operation.

buck54321:
[...]

PR #766, opened 2018-10-26
Update /tx page on websocket update  

ademuanthony:
This is a follow-up issue of the issue [https://github.com/decred/dcrdata/issues/637](637) and PR [https://github.com/decred/dcrdata/pull/714#pullrequestreview-160453911](714)

The /tx page is refreshed silently with Turbolinks on every web socket 'block received event' of the page while maintaining the states (e.g., scroll-position) of the page before the update


public/js/turbolinks.js
```diff
@@ -2,5 +2,835 @@
...
...bled=function(t){var e,r,n;return this.enabled&&this.clickEventIsSignificant(t)&&(r=this.getVisitableLinkForNode(t.target))&&(n=this.getVisitableLocationForLink(r))&&this.applicationAllowsFollowingLinkToLocation(r,n)?(t.preventDefault(),e=this.getActionForLink(r),this.visit(n,{action:e})):void 0},r.prototype.applicationAllowsFollowingLinkToLocation=function(t,e){var r;return r=this.notifyApplicationAfterClickingLinkToLocation(t,e),!r.defaultPrevented},r.prototype.applicationAllowsVisitingLocation=function(t){var e;return e=this.notifyApplicationBeforeVisitingLocation(t),!e.defaultPrevented},r.prototype.notifyApplicationAfterClickingLinkToLocation=function(e,r){return t.dispatch("turbolinks:click",{target:e,data:{url:r.absoluteURL},cancelable:!0})},r.prototype.notifyApplicationBeforeVisitingLocation=function(e){return t.dispatch("turbolinks:before-visit",{data:{url:e.absoluteURL},cancelable:!0})},r.prototype.notifyApplicationAfterVisitingLocation=function(e){return t.dispatch("turbolinks:visit",{data:{url:e.absoluteURL}})},r.prototype.notifyApplicationBeforeCachingSnapshot=function(){return t.dispatch("turbolinks:before-cache")},r.prototype.notifyApplicationBeforeRender=function(e){
```
chappjc: This file was "minified" intentionally.  Please undo this change.

views/extras.tmpl
```diff
@@ -225,22 +227,36 @@
...
+            var currentTxHash = $('#spanTxID').data('tx-page-txid');
+            if(currentTxHash) { // we are on the tx page
+                Turbolinks.reload()
+                /*if((b.Tx.filter(function (tx) {
+                    return tx.TxID === currentTxHash
+                })).length > 0){
+                    // the current tx is in the received block
+                    Turbolinks.visit(window.location)
+                }*/
+            }else {
```
chappjc: I think our JS style is a little wonky.  @gozart1 What style guide would you recommend we follow?

views/extras.tmpl
```diff
@@ -225,22 +227,36 @@
...
 
+            var currentTxHash = $('#spanTxID').data('tx-page-txid');
+            if(currentTxHash) { // we are on the tx page
+                Turbolinks.reload()
+                /*if((b.Tx.filter(function (tx) {
+                    return tx.TxID === currentTxHash
+                })).length > 0){
+                    // the current tx is in the received block
+                    Turbolinks.visit(window.location)
+                }*/
```
chappjc: Does this comment have value? (honest question)

views/extras.tmpl
```diff
@@ -547,6 +565,17 @@
...
 
     $("#keynav-toggle .text").text(keyNavEnabled() ? "Disable Hot Keys":"Enable Hot Keys" )
+
+    var scrollPosition
+
+    Turbolinks.reload = function () {
+        scrollPosition = [window.scrollX, window.scrollY]
+        Turbolinks.visit(window.location, { action: 'replace' })
+        if(!ws){
+            createWebSocket(window.location)
```
chappjc: please add a console log inside this if statement so we know if this is happening.

views/extras.tmpl
ademuanthony: > Does this comment have value? (honest question)

The essence was to check that the current tx id is contained in the received block before updating the view.
What I noticed is that there were some block received events (from the websocket) that were updating the confirmation counts wrongly.

The number becomes high but when you do a full page reload you see a different number (the actual count).
I attributed this to the fact that the current transaction was not present in the received block but commented it out since I am now silently reloading the page and not doing a partial update.

public/js/turbolinks.js
ademuanthony: This is an oversight, I just wanted to check the source for debugging purpose

views/extras.tmpl
chappjc: That makes sense.  TBH, it would be nice to have like 2 sentences to that extent in the vicinity of the comment.  Such comments are helpful to prevent similar gotchas in the future, and justify the fill silent reload rather than partial update.

chappjc:
Changes look good to me.  Needs testing, and preferably also review from a JS ninja.  @gozart1 Would you mind a glance at this PR?

chappjc:
BTW, I noticed that the "(confirmations X)" fix on the transactions page also affected the table of transactions on the addresses page. The same confirms function was apparently used to set the Confirms column values.

ademuanthony:
> BTW, I noticed that the "(confirmations X)" fix on the transactions page also broke the table of transactions on the addresses page. The same confirms function was apparently used to set the Confirms column values.

Yes, it is using the same function. My plan is to work on the [https://github.com/decred/dcrdata/issues/603](issue-603) and reuse the solution on the transaction page for the address page if the tx page solution is suitable

views/extras.tmpl
gozart1: https://crockford.com/javascript/code.html
Is the oldest, simplest, most common I know of

views/extras.tmpl
```diff
@@ -225,22 +227,33 @@
...
+                Turbolinks.reload()
+            }else {
+                var confirms = $('[data-confirmation-block-height]');
+                $.each(confirms, function(i, v) {
+                    if (isNaN($(v).text())) {
+                        $(v).text("0");
+                    }
+                    var confirmations = newBlock.block.height - $(v).data('confirmation-block-height');
+                    // todo number of confirmations on the tx page should be handled seperately
+                    $(v).text("(" + confirmations + (confirmations > 1? " confirmations": " confirmation") + ")")
```
gozart1: In general as our javascript should be **idempotent**. Meaning the function can be called many times and the result is the same. 

So here for example, instead of incrementing, the function could just do current height - block height to get the # confirmations. 

We could use stimulus and make it a generic data target, similar to how age is handled.
So anywhere we need a dynamic # confirmations we can use set data-block-height and data-confirmations

Something like this:
https://github.com/decred/dcrdata/blob/master/views/block.tmpl#L126

With a stimulus controller that hooks into new block notifications and then does the delta calc and dom update
https://github.com/decred/dcrdata/blob/master/public/js/controllers/main.js#L35

views/extras.tmpl
gozart1: Since there is a lot of edge js starting to be used in our code, we might consider using a compiler like https://babeljs.io/ 
Which would let us go a futuristic as we want while supporting old browsers, and have all sorts of other benefits from code-minification, to enforcing a particular formatting style.

The main reason I haven't suggested it is because it adds a bit of overhead to the dev env. But it's not too bad.
Can open an issue of it, it there's interest.

views/extras.tmpl
gozart1: I'm all for keeping things simple and just reloading the page when determining accuracy is complex on the client side. Especially since Turbolinks makes that feel seamless to the end user.

How about keeping this decoupled?
We could use stimulus instead of jquery for a generic confirmations updater. 

And in situations when a user is viewing a block or transaction that is close to the tip ( eg. within 12 blocks ) the code governing those views can simply reload the page via Turbolinks new blocks come in as a straightforward way to ensure accuracy. 

I think it will really benefit us to avoid using imperative jQuery coding style when possible. I know it's super convenient and terse, but in the long run it inevitably leads to a spaghetti codebase. That is why I am suggesting to use stimulus for implementation. The logic of your approach seems good to me!

chappjc:
This is still approved from my perspective, but please do address [gozart1's comments](https://github.com/decred/dcrdata/pull/766#discussion_r228726405).  Thanks.

public/js/controllers/extras.js
```diff
@@ -0,0 +1,68 @@
...
+        }
+
+        connect() {
+            this.confirmationsTargets.forEach((el,i) => {
+                this.setConfirmationText(el, el.dataset.confirmations)
+            })
+        }
+
+        setConfirmationText(el, confirmations) {
+            if(confirmations > 0) {
```
gozart1: Please use whitespace consistently, eg
`if (confirmations > 0) {`
and 
`} else {`
ala https://crockford.com/javascript/code.html

gozart1:
Thank you for taking on a stimulus based approach. 
Looks like it's a WIP, but it's looking pretty good.
Please @ mention me when you'd like a final review on this PR.

My main nit at the moment is just the name "extras" for the controller. There must be something more semantic. 

Since the primary function of the controller is to hook into "newBlock" events, I think calling the controller "newBlock" seems fitting. This sets things up nicely for adding other targets into it down the line that are also related to updating some common html when new blocks arrive.

buck54321:
I might be confused, but I've convinced myself that there is a bug that affects this PR. In the following line

https://github.com/decred/dcrdata/blob/4e6ac133842d45b57f7aae65fe913cb2224f8504/explorer/explorerroutes.go#L898

I think it should be `ConfirmHeight:     exp.Height() - tx.Confirmations + 1`. Taking the difference of heights doesn't include the mined block in the count, but it is included in the height of `tx.Confirmations` and calculations of ticket maturity.

ademuanthony:
> 
> 
> I might be confused, but I've convinced myself that there is a bug that affects this PR. In the following line
> 
> https://github.com/decred/dcrdata/blob/4e6ac133842d45b57f7aae65fe913cb2224f8504/explorer/explorerroutes.go#L898
> 
> I think it should be `ConfirmHeight: exp.Height() - tx.Confirmations + 1`. Taking the difference of heights doesn't include the mined block in the count, but it is included in the height of `tx.Confirmations` and calculations of ticket maturity.

There is a possibility of a bog, I think.
But I can't tell for sure as I have only migrated what we had as jquery to stimulusjs
https://github.com/decred/dcrdata/blob/4e6ac133842d45b57f7aae65fe913cb2224f8504/views/extras.tmpl#L239
I will need more clarification on this, as in how is the confirmations count calculated. @chappjc please help

chappjc:
`ConfirmHeight` should indicate in which block the transaction was mined.  So take a tx with 1 confirmation, the `ConfirmHeight` should equal the best block height.  Note that there is a `tx.BlockHeight` field already, so I'm not sure why we this was computed for `ConfirmHeight`.  I suggest checking against `BlockHeight` to verify we have the correct understanding here, and then  switching to `BlockHeight` instead of computing it.

buck54321:
Sounds good. I can get the fix in #777, but there's going to be conflicts with this work. I'm happy to wait and work them out.

chappjc:
@buck54321 I think PR #777 can go in.  Wanna fix it there, rebase/resolve?

buck54321:
@chappjc I will, though there are other pages besides /tx that have `.ConfirmHeight`. Should I try to follow those through on the same P.R., or create a new one? 

I'm starting to think that it is actually by design, but not well documented or named. Perhaps it should be called `.ConfirmSubtrahend`, because that's how it's used, i.e. `.BestBlockHeight - .ConfirmSubtrahend = .Confirmations`.  instead of `.BestBlockHeight - .TxBlockHeight + 1 = .Confirmations`. Either way, the confusion is probably not worth the savings of the `+1`.

chappjc:
Given the uncertainty about this parameter's semantics, let's do a new PR. Can you clarify the naming of `ConfirmSubtrahend`?  I agree that we should do whatever gives the least confusing result.  Better comments going forward should help with that too.

ademuanthony:
Since the ConfirmHeight was originally gotten from exp.Height() - tx.Confirmations,
https://github.com/decred/dcrdata/blob/4e6ac133842d45b57f7aae65fe913cb2224f8504/explorer/explorerroutes.go#L898
 confirmations is simply newBlock.block.height - confirmHeight

ademuanthony:
@gozart1 can you please take a look once again

chappjc:
@ademuanthony Please rebase/resolve so others can review appropriately.  Thanks.

PR #768, opened 2018-10-26
Address page quick fix for the panic experienced in lite mode

ademuanthony:
when running the decrdata in lite mode, a panic is experienced in the address page causing the page not to load.
This panic is as a result of an err check in the AddressPage function that tries to check the if err.Error() has a prefix with checking that err is not nill and so causing a nil pointer panic.

Here i checked that there is an error before checking the prefix of the error


.gitignore
```diff
@@ -37,4 +37,5 @@ test-data-repo
 dcrdata-testdata-gitlfs
 dcrdata-testdata
 .vendor-new
-v3
\ No newline at end of file
+v3
+tmp/runner-build
```
chappjc: What's this for, out of curiosity?

chappjc: cool.  nice tool

Issue #746, opened 2018-10-17
Search should check aux DB for blocks, etc.


chappjc:
The Search page presently checks for blocks by querying SQLite (dcrsqlite), but side chain blocks will not be found this way since only main chain is presently stored in SQLite (related issue).

chappjc:
This should go into 3.1, along with the added side chain features.

Issue #747, opened 2018-10-17
SQLite should not use height as primary key, should allow side chain blocks


chappjc:
The SQLite tables do not support side chains at all.  The primary key in the block summary table is height, which precludes the possibility of inserting blocks at the same height but with a different hash.

Changes needed:
- The primary key should not be height.  It should either be hash (TEXT) or a new separate row id.
- is_valid and is_mainchain should be added to the block summary table.
- Consider related changes to the stake info table.

buck54321:
Wouldn't the new column in the summary table need to be something more flexible to accommodate an arbitrary number of side chains. Perhaps a VARCHAR chain identifier `chain_name`, i.e. "main" or "politeia"

chappjc:
Block hash should be unique.

buck54321:
Please excuse me if I'm not understanding 
correctly. Block hash is appropriate for the primary key, but to select a block by height, you would need to identify which chain you want. Or is that beyond the scope of this issue?

chappjc:
That's a good question.  We recently addressed this in the postgres backed by having a is_mainchain flag.  So you can query for all chains, or main.  The side chains are generally dead and is little interest, unless there is a reorganization that turns one into main chain (and main into a side chain).  It's essentially impossible for an ancient side chain to ever grow and overtake main.

chappjc:
To clarify, Decred does not have persistent side chains.

buck54321:
Ah. I see now. 

For SQLite, you can't change the primary key, so a routine would be needed to check the table structure and recreate the table as necessary. 

There are currently 7 queries which select by height

`getPoolSQL`
`getWinnersSQL`
`getSDiffSQL`
`getBlockSQL`
`getBlockHashSQL`
`getStakeInfoExtendedSQL`
`getStakeInfoWinnersSQL`

It would be trivial to add a `WHERE is_mainchain = TRUE` to all of those queries, but new queries and routines would be needed if you ever wanted to return an array of all blocks for a given height. 

Then I think the only other thing is that the `StoreBlockSummary` function would need to properly update the `is_mainchain` column

The PSQL table actually looks fine. Everything already querys on hash. It uses an auto-incrementing row id as its primary key and no unique constraint on height for newly built tables. It looks like there used to be one though, so I'm not sure if older tables would need to be rebuilt or not. 

I'd love to work on this.

chappjc:
That's all correct.  You seem to know what needs to be done.  Assigned to you.

[...]

Issue #640, opened 2018-08-27
Contributing readme needs more details


chappjc:
More about best practices and suggested workflow.  e.g. do not rebase/squash until review complete or unless requested.

Current CONTRIBUTING.md: https://github.com/decred/dcrdata/blob/master/CONTRIBUTING.md
The main README.md refers to CONTRIBUTING.md, but it is not a link: https://github.com/decred/dcrdata/blob/master/README.md#contributing

chappjc:
Resolved by https://github.com/decred/dcrdata/pull/774

PR #772, opened 2018-10-27
SQLite structure and routines updated for orphaned blocks

buck54321:
For #747
The solution got a little bigger than I expected. Let me know if you want anything changed. The database went from 250 to 430 MB, which is more than I would have expected for essentially adding 3 columns in total. Maybe there's some optimization I'm missing. 

I did not duplicate the `is_mainchain` and `is_valid` columns on the stake table, but that means that a few queries required some `JOIN`s with the block summary table.

### DIFF
--- a/api/apiroutes.go
+++ b/api/apiroutes.go
@@ -56,7 +56,7 @@ type DataSourceLite interface {
 	GetFeeInfo(idx int) *dcrjson.FeeInfoBlock
 	//GetStakeDiffEstimate(idx int) *dcrjson.EstimateStakeDiffResult
 	GetStakeInfoExtended(idx int) *apitypes.StakeInfoExtended
-	//needs db update: GetStakeInfoExtendedByHash(hash string) *apitypes.StakeInfoExtended
+	GetStakeInfoExtendedByHash(hash string) *apitypes.StakeInfoExtended
 	GetStakeDiffEstimates() *apitypes.StakeDiff
 	//GetBestBlock() *blockdata.BlockData
 	GetSummary(idx int) *apitypes.BlockDataBasic
@@ -731,15 +731,21 @@ func (c *appContext) getBlockFeeInfo(w http.ResponseWriter, r *http.Request) {
 }
 
 func (c *appContext) getBlockStakeInfoExtended(w http.ResponseWriter, r *http.Request) {
-	idx := c.getBlockHeightCtx(r)
-	if idx < 0 {
+	hash := c.getBlockHashCtx(r)
+	if hash == "" {
 		http.Error(w, http.StatusText(422), 422)
 		return
 	}
 
-	stakeinfo := c.BlockData.GetStakeInfoExtended(int(idx))
+	// idx := c.getBlockHeightCtx(r)
+	// if idx < 0 {
+	// 	http.Error(w, http.StatusText(422), 422)
+	// 	return
+	// }
+
+	stakeinfo := c.BlockData.GetStakeInfoExtendedByHash(hash)
 	if stakeinfo == nil {
-		apiLog.Errorf("Unable to get block %d fee info", idx)
+		apiLog.Errorf("Unable to get block fee info for %s", hash)
 		http.Error(w, http.StatusText(422), 422)
 		return
 	}
--- a/api/apiroutes.go
+++ b/api/apiroutes.go
@@ -55,7 +55,7 @@ type DataSourceLite interface {
 	GetTransactionsForBlockByHash(hash string) *apitypes.BlockTransactions
 	GetFeeInfo(idx int) *dcrjson.FeeInfoBlock
 	//GetStakeDiffEstimate(idx int) *dcrjson.EstimateStakeDiffResult
-	GetStakeInfoExtended(idx int) *apitypes.StakeInfoExtended
+	GetStakeInfoExtendedByHeight(idx int) *apitypes.StakeInfoExtended
 	GetStakeInfoExtendedByHash(hash string) *apitypes.StakeInfoExtended
 	GetStakeDiffEstimates() *apitypes.StakeDiff
 	//GetBestBlock() *blockdata.BlockData
@@ -730,22 +730,36 @@ func (c *appContext) getBlockFeeInfo(w http.ResponseWriter, r *http.Request) {
 	writeJSON(w, blockFeeInfo, c.getIndentQuery(r))
 }
 
-func (c *appContext) getBlockStakeInfoExtended(w http.ResponseWriter, r *http.Request) {
+// getBlockStakeInfoExtendedByHash retrieves the apitype.StakeInfoExtended
+// for the given blockhash
+func (c *appContext) getBlockStakeInfoExtendedByHash(w http.ResponseWriter, r *http.Request) {
 	hash := c.getBlockHashCtx(r)
 	if hash == "" {
 		http.Error(w, http.StatusText(422), 422)
 		return
 	}
 
-	// idx := c.getBlockHeightCtx(r)
-	// if idx < 0 {
-	// 	http.Error(w, http.StatusText(422), 422)
-	// 	return
-	// }
-
 	stakeinfo := c.BlockData.GetStakeInfoExtendedByHash(hash)
 	if stakeinfo == nil {
 		apiLog.Errorf("Unable to get block fee info for %s", hash)
+		http.Error(w, http.StatusText(422), 422)
+		return
+	}
+
+	writeJSON(w, stakeinfo, c.getIndentQuery(r))
+}
+
+// getBlockStakeInfoExtendedByHeight retrieves the apitype.StakeInfoExtended
+// for the given blockheight on mainchain
+func (c *appContext) getBlockStakeInfoExtendedByHeight(w http.ResponseWriter, r *http.Request) {
+	idx := c.getBlockHeightCtx(r)
+	if idx < 0 {
+		http.Error(w, http.StatusText(422), 422)
+		return
+	}
+	stakeinfo := c.BlockData.GetStakeInfoExtendedByHeight(int(idx))
+	if stakeinfo == nil {
+		apiLog.Errorf("Unable to get block fee info for height %d", idx)
 		http.Error(w, http.StatusText(422), 422)
 		return
 	}
--- a/db/dcrsqlite/sqlite.go
+++ b/db/dcrsqlite/sqlite.go
@@ -1180,7 +1180,7 @@ func (db *DB) JustifyTableStructures(dbInfo *DBInfo) error {
 	}
 
 	// Clean up the file a little bit
-	_, err = db.Query("VACUUM;")
+	_, err = db.Exec("VACUUM;")
 	if err != nil {
 		log.Error("Failed to VACUUM SQLite database.")
 	}
--- a/db/dcrsqlite/sqlite.go
+++ b/db/dcrsqlite/sqlite.go
@@ -12,6 +12,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/decred/dcrd/chaincfg/chainhash"
 	"github.com/decred/dcrd/wire"
 	apitypes "github.com/decred/dcrdata/v3/api/types"
 	"github.com/decred/dcrdata/v3/blockdata"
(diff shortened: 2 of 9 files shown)


chappjc:
@buck54321 Instead of doing a merge with master (or resolving conflicts with github's webpage, which causes such a merge) please next time rebase and resolve the conflicts on your local machine.  Thanks for the PR!

buck54321:
@chappjc No problem. I'll get it right next time. Still figuring out this workflow.

chappjc:
The CONTRIBUTING doc needs lots of work.  It neglects a lot of important details.

buck54321:
I see issue #640. I'm mostly confused about the exact sequence for squashing. Intuition tells me to 

checkout branch x -> do stuff -> commit  -> pull from up-to-date master -> push to branch x on remote -> pull request

and I'm not certain where exactly certain where the squash/rebase is performed.  If you can point me in the right direction, I'm happy to update CONTRIBUTING.md.

chappjc:
After you work on your branch and commit, it may not be building on remote master if it has changed.  And that may be OK unless there are conflicts.  Where things differ from your intuition is instead of merging in changes from master to your branch, you rebase your branch on master.  Basically you'll need a local master branch, check that out and pull lastest into that local master.  Then check out your branch again.  `git rebase master`.  Just like with merging, there may be conflicts to resolve.  You resolve one commit at a time (edit, git add [files], git rebase --continue, repeat for other commits that have conflicts).  Then force push.

buck54321:
Actually, I think this resolves #746 as well.

db/dcrsqlite/sync.go
```diff
@@ -405,3 +409,54 @@ func (db *wiredDB) getBlock(ind int64) (*dcrutil.Block, *chainhash.Hash, error)
 
 	return block, blockhash, nil
 }
+
+// Import all side chains. Similar to pgblockchain.MissingSideChainBlocks
```
chappjc: Please begin all function docs, especially for exported functions, with the function name (i.e. `// ImportSideChains ...`).  This is just convention, not my silly rule.  :)
https://blog.golang.org/godoc-documenting-go-code
https://golang.org/doc/effective_go.html#commentary

main.go
```diff
@@ -630,15 +633,26 @@ func _main(ctx context.Context) error {
 		return nil
 	}
 
+	log.Info("Mainchain sync complete.")
+
+	// Ensure all side chains known by dcrd are also present in the base DB
+	// and import them if they are not already there.
+	if cfg.ImportSideChains {
+		log.Info("Primary DB -> Now retrieving side chain blocks from dcrd...")
+		err := baseDB.ImportSideChains(collector)
```
chappjc: It's a minor issue in most cases, but the approach of doing sqlite first then possibly postgres is inefficient. That is,

```go
if cfg.ImportSideChains {
	// retrieve all side chains and import into sqlite
}
if usePG && cfg.ImportSideChains {
	// retrieve all side chains and import into sqlite
}
```

vs.

```go
if cfg.ImportSideChains {
	// retrieve all side chains

	// import into sqlite

	if usePG {
		// import into postgres
	}
}
```

Changing to the latter requires some refactoring, so it's not a big deal right now.  Just take note.

db/dcrsqlite/sqlite.go
```diff
@@ -912,6 +1076,73 @@ func (db *DB) RetrieveStakeInfoExtended(ind int64) (*apitypes.StakeInfoExtended,
...
+	// Each tuple is a row with columns [cid, name, type, notnull, dflt_value, pk]
+	rows, err := db.Query(db.getBlockSummaryTableInfo)
+	if err != nil {
+		log.Errorf("Query failed: %v", err)
+		return err
+	}
+	defer rows.Close()
+	rowCounter := 0
+	for rows.Next() {
+		rowCounter += 1
```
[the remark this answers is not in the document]
buck54321: The way I implemented this, which is probably wrong, is to use a SQL query 

`d.getBlockSummaryTableInfo = fmt.Sprintf(`PRAGMA table_info(%s);`, TableNameSummaries)`

which returns column information as rows. That way, if you wanted a stronger check, you could check to see which column has the primary key or check the column order. I'm happy to either implement the PK check or to find a different way of counting columns.

main.go
buck54321: This was part of the reason I tried to implement this as a single function call in main.go. Is there any reason both of these sidechain import routines can't be implemented in asynchronously?

chappjc: It's not an issue to do it separately for sqlite and pg, but the retrieval gets run twice.  Usually this is just a handful of blocks, so no worries.  On testnet3 there was a crazy 4k block reorg early on, and I my node still remembers it -- which is great for testing these things.  But this is certainly not something that will be an issue on mainnet.

db/dcrsqlite/sqlite.go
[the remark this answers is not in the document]
chappjc: Oh, I see!  I just read the query incorrectly.  The pragma approach is probably perfect.  I suppose that's standard sqlite, so that's great.

chappjc:
Another thing that is entirely my fault for not including in CONTRIBUTING.md: when there is a review and edits in response, it's helpful to just add new commits instead of squashing down. Github is good about showing changes since a review, but only if the git history remains.

Then when all is well and the PR can be merged, it's cool to squash the branch.  However, it's really OK to leave it since there is a "Squash and merge" option that I usually use.  Just in this case it's helpful to have the PR author comment here with a message for the squashed commit so I can put that message in.

main.go
buck54321: K. Maybe as a separate issue, but I'm happy to do this. Probably have a wrapper function that pulls the blocks once and feeds them through a checker method of each DB object. Can that be run as a goroutine?

buck54321:
I'll keep my CONTRIBUTING.md branch alive and update as I learn the system here. 

Do I have to do anything else once I've pushed my changes, or do you get a notification of some kind?

Also, I was a little confused on the relation of this issue to #601. I may not understand the reorg correctly, but doesn't the PGSQL sidechain check (and that amazing new SQLite check someone just did) already resolve that issue on startup?

api/apiroutes.go
```diff
@@ -727,15 +727,21 @@ func (c *appContext) getBlockFeeInfo(w http.ResponseWriter, r *http.Request) {
 }
 
 func (c *appContext) getBlockStakeInfoExtended(w http.ResponseWriter, r *http.Request) {
-	idx := c.getBlockHeightCtx(r)
-	if idx < 0 {
+	hash := c.getBlockHashCtx(r)
```
chappjc: In (`NewAPIRouter`) api/apirouter.go, note that `getBlockStakeInfoExtended` is bound to several URL paths:

    rd.Get("/pos", app.getBlockStakeInfoExtended)

Depending on the full path, different middleware is used to embed the height/hash.

- Under /block/best, it is using the `BlockIndexLatestCtx` middleware, which embeds only the best block index in the request context (no hash).
- Under /block/hash/{blockhash}, it is using `BlockHashPathAndIndexCtx`, which embeds the hash from the URL path, looks up the corresponding height, and embeds that too.
- Under /block/{idx}, it is using `BlockIndexPathCtx`, which embeds the index from the path only.
- Under /block/range/{idx0}/{idx}, it is using `m.BlockIndex0PathCtx` and `m.BlockIndexPathCtx`, which embed both indexes from the path only.  Although this path prefix doesn't bind to `getBlockStakeInfoExtended`.

Basically, the `(*appContext).getWhateverCtx(*http.Request)` type functions try to pull out a value.  The middleware set for different paths decides what/when/how the values get put into those context values.

chappjc:
OK, thanks for updating CONTRIBUTING.md as you go.

The startup reorg check is not resolved.  These changes won't import the side chain that was mainchain because it's already in the DB.  But doing so would not be correct because the mainchain needs to be disconnected carefully as in TipToSide chain so as to keep the other DBs (stakedb and a ticket database) correct.

main.go
chappjc: I don't see why they can't be run concurrently.

api/apiroutes.go
buck54321: OK. This is still a little confusing, but I think I've got it. There are now two distinct methods for the different contexts, and I think I've assigned them correctly. I did see one weird error

`2018-10-29 12:51:32.745 [ERR] RPCC: GetBlockVerbose(0000000000000000000000000000000000000000000000000000000000000000) failed: -5: Block not found: 0000000000000000000000000000000000000000000000000000000000000000`

but I wasn't able to reproduce it. I noticed that in `apirouter`, some index-only-context routes are calling `apiroutes.getBlockVerbose` which uses `getBlockHashCtx`, so I think that this might be another instance of the same error I made.

chappjc: The genesis (block 0) has no previous block, so it is set to that long string, `0000000000000000000000000000000000000000000000000000000000000000`
https://explorer.dcrdata.org/block/0

db/dcrsqlite/sync.go
```diff
@@ -405,3 +409,54 @@ func (db *wiredDB) getBlock(ind int64) (*dcrutil.Block, *chainhash.Hash, error)
...
 }
+
+// ImportSideChains imports all side chains. Similar to pgblockchain.MissingSideChainBlocks
+// plus the rest from main.go
+func (db *wiredDB) ImportSideChains(collector *blockdata.Collector) error {
+	tips, err := rpcutils.SideChains(db.client)
+	if err != nil {
+		return err
+	}
+	hashlist := make([]*chainhash.Hash, 0)
```
chappjc: When not allocating the slice, we generally do just `var hashlist []*chainhash.Hash`.

chappjc: However, we know the length of `tips`, so we should actually do `hashlist := make([]*chainhash.Hash, 0, len(tips))` to have the right amount of space allocated.

db/dcrsqlite/sync.go
```diff
@@ -405,3 +409,54 @@ func (db *wiredDB) getBlock(ind int64) (*dcrutil.Block, *chainhash.Hash, error)
...
+	for it := range tips {
+		sideHeight := tips[it].Height
+		log.Tracef("Primary DB -> Getting base DB side chain with tip %s at %d.", tips[it].Hash, sideHeight)
+		sideChain, err := rpcutils.SideChainFull(db.client, tips[it].Hash)
+		if err != nil {
+			log.Errorf("Primary DB -> Unable to get side chain blocks for chain tip %s: %v", tips[it].Hash, err)
+			return err
+		}
+		sideHeight -= int64(len(sideChain)) - 1
```
chappjc: The computed value does not appear to be used.

chappjc: Seems like this might be copy pasta, but I want to be sure a detail wasn't left out here.

chappjc:
Code changes approved.  Just needs testing on my end.

db/dcrsqlite/sqlite.go
```diff
@@ -912,6 +1037,99 @@ func (db *DB) RetrieveStakeInfoExtended(ind int64) (*apitypes.StakeInfoExtended,
...
+	}
+
+	// It simply checks whether there are enough columns for the current structure
+	if rowCounter >= 12 {
+		return nil
+	}
+
+	log.Info("Detected old SQLite table structure. Updating.")
```
chappjc: It occurred to me that we should really backup the sqlt.db file first.

buck54321: Should I delete the backup if the swap is successful? All table mods are wrapped in a transaction right now, so a failed table copy should be rolled back.

chappjc: It's not a corrupted db I'm concerned with (as you said, failures are rolled back), but devs like us losing our DBs while testing and such.  Although there may be a need to revert to an old version of dcrdata which won't support the new DB format. So I would suggest keeping the old DB copy around regardless of the the migration result.

buck54321:
Though there was some uncertainty regarding `os.Read` on the SQLite file, I tested on both Windows and Linux with no apparent issues.

db/dcrsqlite/sqlite.go
```diff
@@ -929,6 +1074,120 @@ func (db *DB) RetrieveStakeInfoExtended(ind int64) (*apitypes.StakeInfoExtended,
...
+	}
+
+	err = transaction.Commit()
+	if err != nil {
+		log.Error("Failed to commit SQL transaction after table reorg.")
+		return err
+	}
+
+	// Clean up the file a little bit
+	_, err = db.Query("VACUUM;")
```
chappjc: This needs to be `_, err = db.Exec("VACUUM;")`

chappjc: After the table upgrade with no VACUUM, the files size was 451538944 B.
With VACUUM, the file size is reduced to 323424256 B.

Issue #764, opened 2018-10-26
Header logo needs update.


saender:
![oldlogo](https://user-images.githubusercontent.com/29069168/47566792-7426e400-d935-11e8-87f7-134776cad10b.jpg)

New logo symbol can be found at:
https://github.com/decred/dcrdesign/releases -> Identity -> Logo package -> 
https://github.com/decred/dcrdesign/files/2160933/decred.-.logo.package.zip
-> Decred - Logo Package -> RGB -> Symbol

chappjc:
I believe this has already been updated.  See https://alpha.dcrdata.org/

However, it's not a drop-in replacement.  Making it look OK in night mode will take some additional work.  @gozart1 ^

saender:
Utility sites should have monotone logo/symbol and mentioned logo package contains both negative and positive versions for light or dark theme.

chappjc:
Resolved by https://github.com/decred/dcrdata/pull/770

PR #773, opened 2018-10-27
navbar and logo dark theming

buck54321:
for #764


buck54321:
Oops. Somebody already got it

PR #777, opened 2018-10-29
Mempool transactions on /tx update (mostly) correctly now

### LINKED ISSUE #309: Unconfirmed transaction display
TYPE and MATURE are out of alignment.
INCLUDED IN BLOCK is missing. Instead of being removed entirely, can say 'mempool'.

buck54321:
Resolves #309

There is a related issue with the confirmations counter at the top that I didn't touch because is looks like it's being worked on. see https://github.com/decred/dcrdata/pull/766#issuecomment-433546786

The only difference between a confirmed transaction's /tx page and a mempool /tx page which is confirmed after page/turbolinks load is that the "Pool Status"  is not displayed after a mempool tx is confirmed. 

@gozart1 I put  the JS in the `updateBlockData` websocket callback. I wonder if we could build a little signal system for block data callbacks, ala

```
var blockCallbacks = {}
function registerBlockCallback(key, f){
    blockCallbacks[key] = f;
}
function signalBlock(block){
    for(key in blockCallbacks){
        blockCallbacks[key](block);
    }
}

...

// In createWebSocket()
    var newBlock = JSON.parse(event);

    ...

    signalBlock(newBlock.block);
}
 ...

function doBlockStuff(block){
    ...
}
registerBlockCallback("doblockstuff", doBlockStuff)
```

Edit: I see now that you have `desktopNotifyNewBlock` which might serve a similar purpose.

### DIFF
--- a/views/tx.tmpl
+++ b/views/tx.tmpl
@@ -107,7 +107,7 @@
                                             aria-valuenow="{{.Confirmations}}"
                                             aria-valuemin="0"
                                             aria-valuemax="{{.TicketInfo.TicketMaturity}}"
-                                            data-confirm-height="{{$.ConfirmHeight}}"
+                                            data-confirm-height="{{add $.ConfirmHeight 1}}"
                                             data-tx-type="{{.Type}}"
                                         >
                                             <span class="nowrap pl-1 pr-1">
@@ -153,7 +153,7 @@
                                             aria-valuenow="{{.Confirmations}}"
                                             aria-valuemin="0"
                                             aria-valuemax="{{.Maturity}}"
-                                            data-confirm-height="{{$.ConfirmHeight}}"
+                                            data-confirm-height="{{add $.ConfirmHeight 1}}"
                                             data-tx-type="{{.Type}}"
                                         >
                                             <span class="nowrap pl-1 pr-1">
@@ -218,7 +218,7 @@
                                                     aria-valuenow="{{.Confirmations}}"
                                                     aria-valuemin="0"
                                                     aria-valuemax="{{.TicketInfo.TicketExpiry}}"
-                                                    data-confirm-height="{{add $.ConfirmHeight (uint16toInt64 $.ChainParams.TicketMaturity)}}"
+                                                    data-confirm-height="{{add (add $.ConfirmHeight 1) (uint16toInt64 $.ChainParams.TicketMaturity)}}"
                                                     data-tx-type="LiveTicket"
                                                 >
                                                     <span class="nowrap pl-1 pr-1">
--- a/views/tx.tmpl
+++ b/views/tx.tmpl
@@ -79,7 +79,7 @@
                     </tr>
                 {{end}}
                 {{if eq .Mature "True"}}
-                    <tr><td class="text-right pr-2 h1rem p03rem0">MATURITY</td><td>Fully Mature</td></tr>
+                    <tr><td class="text-right pr-2 h1rem p03rem0">MATURITY</td><td>Mature</td></tr>
                 {{end}}
                 {{if and (eq .Mature "False") (eq .Type "Ticket")}}
                     <tr>
@@ -99,7 +99,7 @@
                                         >
                                             <span class="nowrap pl-1 pr-1">
                                               {{if gt .Confirmations 0}}
-                                                Immature, {{ if eq .Type "Ticket"}}eligible to vote{{else}}spendable{{end}} in {{ if eq .Confirmations .TicketInfo.TicketMaturity }}next block{{else}}{{subtract (add .TicketInfo.TicketMaturity 1) .Confirmations}} blocks ({{printf "%.1f" .TicketInfo.TimeTillMaturity}} hours remaining){{end}}
+                                                Immature, {{ if eq .Type "Ticket"}}eligible to vote{{else}}spendable{{end}} in {{ if eq .Confirmations .TicketInfo.TicketMaturity }}next block{{else}}{{subtract .TicketInfo.TicketMaturity .Confirmations}} blocks ({{printf "%.1f" .TicketInfo.TimeTillMaturity}} hours remaining){{end}}
                                               {{else}}
                                                 Awaiting confirmation
                                               {{end}}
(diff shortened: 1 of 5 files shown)


views/tx.tmpl
```diff
@@ -77,13 +79,9 @@
                     </tr>
                 {{end}}
                 {{if eq .Mature "True"}}
-                    <tr><td class="text-right pr-2 h1rem p03rem0">MATURE</td><td>{{.Mature}}</td></tr>
-                {{else}}
-                {{if and (eq .Confirmations 0) (or (eq .Type "Ticket") (eq .Type "Vote")) }}
-                    <tr><td class="text-right pr-2 h1rem p03rem0">MATURE</td><td>N/A</td></tr>
-                {{end}}
+                    <tr><td class="text-right pr-2 h1rem p03rem0">MATURITY</td><td>Fully Mature</td></tr>
```
chappjc: The phrase "Fully Mature" does not carry much meaning beyond just "mature" or "true", and it might be confusing (what is partially mature?).

buck54321: Yeah. I wasn't sure about that. But then again, there is a progress bar that can be partially full. Perhaps just a "Mature" row with no value that only shows when it is mature, i.e. drop the "True" from the current implementation.

chappjc: Perhaps, but it might look like there is a value missing.  You could post a screen shot and solicit opinions.  @xaur and @raedah might wanna comment.

chappjc: For labeling mature tickets/votes, I personally dislike "MATURITY Fully Mature" and just "MATURE ".  It looks a bit repetitive, but "MATURITY Mature" is clear and it is consistent with the "MATURITY {progress bar}" that precedes it, so that's my vote.

xaur: Definitely not "Fully mature". Either "Maturity: \[progress bar\]", or "Mature: true|false" or "Maturity: Mature|Immature". Sorry I'm out of context, are we deciding progress bar vs text label?

Also this discussion does not belong to specific code review block, but rather the root PR or even the issue.

buck54321: Progress bar is now there for all immature tickets including unconfirmed mempool tickets. Just trying to work out what to do once it's mature. 

Sorry for the misplaced discussion. Should I re-comment in the related issue to elicit more feedback, or just go with "Maturity Mature"?

buck54321:
Now using Maturity: Mature. Also, the progress bar updates live now.

buck54321:
I think the colors used on the color bar could be better.

image

The contrast is not great in the gray section.

views/tx.tmpl
```diff
@@ -99,7 +99,7 @@
                                         >
                                             <span class="nowrap pl-1 pr-1">
                                               {{if gt .Confirmations 0}}
-                                                Immature, {{ if eq .Type "Ticket"}}eligible to vote{{else}}spendable{{end}} in {{ if eq .Confirmations .TicketInfo.TicketMaturity }}next block{{else}}{{subtract (add .TicketInfo.TicketMaturity 1) .Confirmations}} blocks ({{printf "%.1f" .TicketInfo.TimeTillMaturity}} hours remaining){{end}}
+                                                Immature, {{ if eq .Type "Ticket"}}eligible to vote{{else}}spendable{{end}} in {{ if eq .Confirmations .TicketInfo.TicketMaturity }}next block{{else}}{{subtract .TicketInfo.TicketMaturity .Confirmations}} blocks ({{printf "%.1f" .TicketInfo.TimeTillMaturity}} hours remaining){{end}}
```
chappjc: I had to check with others to be sure, but I believe a ticket is not considered mature until 257 confirmations (256 blocks after it is mined).

chappjc: @davecgh I saw you posted something to this effect in Slack, but I'd like if you would confirm the above.

buck54321: So the `TicketMaturity` comes from `chaincfg.Params` and is `256`. With this new information about 257 needed confirmation (inc. first block), I can now see why it was that way. Just to make things more confusing, the `aria-valuemax` of the progress bar is set to `256`, where I guess it should be 257 as well.

chappjc: mm hmm, please see https://decred.slack.com/archives/C3ZNV8QTX/p1540938352047200 (Dave's response).
I think we've been inaccurate in some frontend spots regarding maturity.

davecgh: Yes, the maturity is 256, but stake transactions don't affect the current block in terms of things like the ticket pool.  So, it does not actually become eligible to vote until the 257th block.  In other words, it matures in the 256th block and gets added to the pool _after_ the winning tickets and such have been selected, ergo it isn't actually eligible to be selected until the 257th.

chappjc: Alright, so we have to decide if the progress bar should maturity, or blocks until eligibility to vote.  :laughing:  Because it sounds like it would be correct to have the last step in the progress bar be "Mature, eligible to vote in next block" with 100% progress, where the previous block is "Immature, eligible to vote in two blocks" with 99.61% progress (255/256).

davecgh: I think that would be the best approach. Show maturity and clarify eligibility.

xaur:
@buck54321 agreed on contrast, either the background or font color could be tweaked.

views/tx.tmpl
chappjc: @buck54321 Does that suggestion make sense, and would it be straightforward to implement?

buck54321: Yeah. I'm kinda in a fog of `+1`'s and `-1`'s, but I think I get it.

buck54321:
@xaur, @chappjc
#1 

image

#2

image

#3 

image

#4: somethin' else

chappjc:
Each option looks fine to me, but I'd choose 1 if pressed (as long as nothing is wrong in night mode).

xaur:
@buck54321 #1 looks good.

buck54321:
Progress bars progress live as blocks are connected. Various text field update to appropriate values as tickets and votes move through their stages.

Issue #241, opened 2017-10-18
Number inputs and outputs


chappjc:
The inputs and outputs of a transaction should have their index displayed somehow, if even just by moving the mouse over the i/o.

Also, the same applies to blocks, although since we split into categories, it doesn't make sense to have the number always visible, but a hover would be nice.

PR #778, opened 2018-10-30
Elided text, vout/vin index, and table-layout: auto on tx page tables.

buck54321:
See #241

This makes some changes to the table layout on the transactions page. The abridged/elided text is created using the `template.FuncMap`, so it can be used on any other page as well. 

image

### DIFF
--- a/public/css/main.css
+++ b/public/css/main.css
@@ -152,6 +152,18 @@ body.darkBG a,
 body.darkBG a:hover {
   color: #94ffca;
 }
+body.darkBG a.elidedhash,  body.darkBG a.elidedhash:hover{
+  color:transparent;
+  text-decoration-color:#94ffca;
+}
+body.darkBG a.elidedhash::before,
+body.darkBG a.elidedhash::after {
+  color: #94ffca;
+}
+body.darkBG div.elidedhash::before,
+body.darkBG div.elidedhash::after {
+  color: #fdfdfd;
+}
 body.darkBG a.grayed,
 body.darkBG a.grayed:hover {
   color: rgba(148, 255, 202, 0.75) !important;
@@ -594,9 +606,11 @@ body.darkBG .progress {
   -moz-user-select: all;
   user-select: all;
   z-index:20;
-  cursor:pointer;
   text-overflow:clip;
   overflow:hidden;
+}
+a.elidedhash {
+  cursor:pointer;
 }
 a.elidedhash:visited {
   color:transparent;
@@ -624,7 +638,6 @@ div.elidedhash:before, div.elidedhash:after {
 }
 .elidedhash:after {
   content:attr(data-tail);
-  color:#2e75ff;
   position:absolute;
   right:0;
   top:0;
@@ -633,7 +646,7 @@ div.elidedhash:before, div.elidedhash:after {
   text-align:right;
   direction:rtl;
 }
-.elidedhash:hover {
+a.elidedhash:hover {
   color:transparent;
   text-decoration: underline;
   text-decoration-color:#2e75ff;
--- a/public/css/main.css
+++ b/public/css/main.css
@@ -553,12 +553,6 @@ body.darkBG .progress {
   padding: .1rem .5rem;
   height: 25px;
 }
-.table .tx-type-stub{
-  max-width: 80px;
-  text-overflow: ellipsis;
-  white-space: nowrap;
-  overflow:hidden;
-}
 .table.flush-left tr td:first-child {
   padding-left: 0px;
 }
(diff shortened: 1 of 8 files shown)


chappjc:
I'm all for eliding characters of the hashes, but we're only saving 2 characters of the address.  I would support forcing the address not to wrap so it looks better.

buck54321:
Oh yeah. I guess just the transaction id in the previous outpoint column could be elided, and then no wrap on the address columns.

raedah:
Related to this, some implementations being done at the following link. Not entirely complete yet as we were trying to get better support for large page widths.
https://github.com/raedahgroup/design/pull/3

chappjc:
This is looking very good now, except something broke how the page reactos to changing window width.

### Before

image

### After

image

public/css/main.css
```diff
@@ -538,11 +538,11 @@ body.darkBG .progress {
...
-.tx-col-l{
-  max-width:54%;
-}
-.tx-col-r{
-  max-width:46%;
+.tx-type-stub{
+  max-width: 80px;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+  overflow:hidden;
```
chappjc: Please see the discussion with gozart [here](https://github.com/decred/dcrdata/pull/718/files/ec2db4682fe91adac7f957aec171be7e76a7e37f#discussion_r229499754), and the issue [here](https://github.com/decred/dcrdata/issues/782).  I don't pretend to be a competent frontend developers, so I almost always defer to gozart or whoever is working on a new design.  But we've gotten to a point where we need to be very careful and communicate all changes to main.css so as to keep it stable.

gozart1:
This is an interesting approach to formatting the hashes to fit a single line. 

When looking into ways for solving this, I explored a similar approach as the one you are taking. But ended up an a purely css based approach by using flexbox instead of tables for the layout. 

https://github.com/decred/dcrdata/blob/master/views/home.tmpl#L207
flexbox-truncation

The nice thing about a pure css approach is that transaction hash is still fully intact in the html, so cmd+f still works and I think that is a fairly common operation so it's worth preserving that functionality. 

Have you considered this approach? If not would you be willing to give it a shot? If yes, are there reasons why this way preferable?

buck54321:
I like a pure-CSS approach. I'm even using it on a different column, but when somebody is visually scanning for a particular hash, they check only the first few or the last few characters, which means we should elide the string parts in the middle to leave the ends intact. I've seen pure-CSS implementations of middle eliding, but they are a bit tedious, and having it all in a table isn't going to make it easier. 

I'm happy to do it that way if you that's the consensus, though.

chappjc:
One thing to think about is how this would render to a web search crawler.  If the transaction hashes don't end up fully on the page, I think it's a sub-optimal approach.

buck54321:
That's a good point. Let me try a couple things. Maybe we can have it all somehow.

buck54321:
Trying pure-CSS middle eliding. 
I was trying to accomplish the following four goals.

1. Looks good
2. Full text in HTML
3. Selectable
4. Both ends of string are visible

I think I've hit these goals, but there are a couple of things to note.

1. This will only work on very long strings that will always be trimmed. It's great in the "Previous Outpoint" column, but wouldn't work in the address column, unfortunately. 

2. If you click drag to select the text across the ellipsis, you actually get the string twice, with a space in between. If you only select only one side, you will get the correct value. If this is a deal breaker, I fully understand. 

image

gozart1:
Sounds like progress on a thorny problem.
We could put `user-select: none` on the right side to avoid duplicate text when selecting from copy paste.

Something else to explore is pseudo elements and data attributes. That solves duplication of text. Hacked around on that concept a bit here:
https://codepen.io/gozart1/pen/EOxVzB

I don't think putting an ellipsis into the middle of string is actually something that fully been solved in web dev, so this is novel territory!

buck54321:
@gozart1 Your comment and codepen put me right on track. I think it's just about as far as CSS can take it right now. 

I want to make a quick argument here that you don't actually want the entire hash to be a link because it makes it hard to select. In this newest attempt, the right side is a fixed length unselectable link. The other side is a `user-select:all` hash with ellipsis. I've matched the color and just underlined the link part as a visual.

image

chappjc:
That's something to think about, but TBH I think most people will recoil at that.

buck54321:
Is it the underline part? I can just keep the whole thing a link then. It's just so functional though, but yeah, it looks weird.

chappjc:
Partly the underline, but more that I can imagine trying to click on the left side that isn't clickable.  Pretty much everyone would I bet.

buck54321:
No problem. Easy to switch back. I'll switch it up, resolve that conflict, and push in the morning.

gozart1:
Something got messed up with how Stakebase is displayed, it's displayed as `Stakebase: 0000` in master.

stakebase

buck54321:
It's the `direction:rtl`. No problem to fix, but I should ask, do we really need the `0000`?

chappjc:
Here's the tx page, with the window width maximized:

image

Narrowed window:

image

Narrowed until In and Out are on top/bottom instead of side-by-side:

image

Some comments:
- There is sometimes space after the ellipses (e.g. `abc...  d`), but not always.  It doesn't look right with a space.
- When the window is wide enough, I think the addresses should be fully displayed.  If that means changing `max-width` for the largest `@media` container, I think we should do that unless there's a better solution.
- The only way to get the full address is to narrow the screen a ton until In and Out tables are arranged above/below each other.  Just a reason to fix the previous point.

buck54321:
I've modified the table layout a little bit so that when possible, full addresses are shown. This comes at the cost of some of the width on the previous outpoint column, which is too long to ever show full width anyway. Also adjusted the alignment at the ellipsis to remove that annoying space. 

@dmigwi @itswisdomagain As per #782, I ran into an issue with one of the bootstrap overwrites. I made it a custom class instead.

dmigwi:
@buck54321 Nice work on this task.

chappjc:
Yah, this is really cool.  Gonna retest today.

PR #779, opened 2018-10-30
Timebased blocks listing

dmigwi:
PR introduce time-based blocks list view pages below. Grouping is done using the block time value.
"/years" - Groups blocks into years
"/months" - groups blocks into months
"/weeks' - Groups blocks into weeks
"/days' - Groups blocks into days

All the above pages follow this similar format.
screenshot from 2018-10-30 16-26-23


db/dbtypes/types.go
```diff
@@ -139,32 +142,32 @@ type BlocksGroupedInfo struct {
...
 }
 
-// ChartGroupings helps maping a given chart grouping to its standard string value.
-var ChartGroupings = map[ChartGrouping]string{
-	AllChartGrouping:   "all",
-	YearChartGrouping:  "yr",
-	MonthChartGrouping: "mo",
-	WeekChartGrouping:  "wk",
-	DayChartGrouping:   "day",
+// TimeBasedGroupings helps mapping a given time grouping to its standard string value.
```
chappjc: TimeBasedGroupings *maps* a given time grouping to its standard string value

explorer/explorer.go
```diff
@@ -684,3 +685,14 @@ func (exp *explorerUI) simulateASR(StartingDCRBalance float64, IntegerTicketQty
...
 }
+
+// genesisBlockTime returns the genesis block time.
+func (exp *explorerUI) genesisBlockTime() (int64, error) {
+	hash, err := exp.blockData.GetBlockHash(0)
+	if err != nil {
+		return -1, err
+	}
+
+	data := exp.blockData.GetExplorerBlock(hash)
```
chappjc: Also might as well skip the `GetBlockHash(0)` and use `params.GenesisHash`.  Even better, use `params.GenesisBlock.Header.Timestamp` and avoid any DB look ups or issues getting the value before genesis is in the db.

explorer/explorerroutes.go
```diff
@@ -224,7 +226,114 @@ func (exp *explorerUI) StakeDiffWindows(w http.ResponseWriter, r *http.Request)
...
+}
+
+// TimeBasedBlocksListing is the main handler for "/day", "/week", "/month" and "/year".
+func (exp *explorerUI) timeBasedBlocksListing(val string, w http.ResponseWriter, r *http.Request) {
+	if exp.liteMode {
+		exp.StatusPage(w, fullModeRequired,
+			"Time based blocks listing page cannot run in lite mode.", NotSupportedStatusType)
+	}
+
+	grouping := dbtypes.TimeGroupingFromStr(val)
```
chappjc: Please pass a `TimeBasedGrouping` instead of a string to `timeBasedBlocksListing`.  This should be simpler, and will eliminate the potential future bug where TimeGroupingFromStr and the strings in the functions like `YearBlocksListing` may no longer agree.

db/dcrpg/queries.go
```diff
@@ -599,25 +599,66 @@ func retrieveWindowBlocks(db *sql.DB, windowSize int64, limit uint64,
...
+			IndexVal:           indexVal,
+			EndBlock:           endBlock,
+			Voters:             votes,
+			Transactions:       txs,
+			FreshStake:         tickets,
+			Revocations:        revocations,
+			BlocksCount:        blocksCount,
+			StartTime:          startTime,
+			EndTime:            endTime,
+			Size:               int64(blockSizes),
```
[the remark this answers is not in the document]
dmigwi: `humanize.Bytes(blockSizes)` takes an parameter of `uint64`.

explorer/explorerroutes.go
dmigwi: The string is needed on the front end since the four routes share a template. I could pass the string and the `TimeBasedGrouping` object but that might result to duplication and also create bugs when there are inconsistencies between the two values set.

db/dcrpg/queries.go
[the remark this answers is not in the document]
chappjc: gotcha

dmigwi:
Latest Update
screenshot from 2018-10-31 01-34-20

explorer/explorerroutes.go
chappjc: The nice thing about objects with methods is that you can call the method from the template.  :D  I'm referring to `func (g TimeBasedGrouping) String() string`.  In fact, since this type implements the `Stringer` interface, I'd be surprised if you even need to call String() manually, but you can.

explorer/templates.go
```diff
@@ -303,6 +304,37 @@ func makeTemplateFuncMap(params *chaincfg.Params) template.FuncMap {
...
+			case dbtypes.WeekGrouping:
+				matchedGrouping = "days"
+			// for dbtypes.DayGrouping and any other groupings default to blocks.
+			default:
+				return fmt.Sprintf("/blocks?offset=%d", interval)
+			}
+			matchingVal := dbtypes.TimeGroupingFromStr(matchedGrouping)
+			intervalVal, err := dbtypes.TimeBasedGroupingToInterval(matchingVal)
+			if err != nil {
+				log.Errorf("Resolving the new group interval failed: error : %v", err)
```
chappjc: Whenever the user has control over the input, and can cause a non-nil error at will, I prefer to have the log as Debug[f] or Trace[f]

explorer/templates.go
```diff
@@ -303,6 +304,37 @@ func makeTemplateFuncMap(params *chaincfg.Params) template.FuncMap {
 		"toLowerCase": func(a string) string {
 			return strings.ToLower(a)
 		},
+		"fetchListNavURL": func(groupingStr string, interval int64) string {
+			// This allows all links in years page to be linked to months i.e.
+			// /year -> /months -> weeks -> /days -> /blocks
```
chappjc: I don't follow these comments and what the function should do.  e.g. the switch statement will take a MonthGrouping and set `matchedGrouping = "weeks"`, which is then converted to a time interval value.  Sorry, I'm being dense here...

explorer/templates.go
```diff
@@ -303,6 +304,37 @@ func makeTemplateFuncMap(params *chaincfg.Params) template.FuncMap {
...
+			intervalVal, err := dbtypes.TimeBasedGroupingToInterval(matchingVal)
+			if err != nil {
+				log.Errorf("Resolving the new group interval failed: error : %v", err)
+				return "/blocks?offset=0"
+			}
+
+			latestTime := time.Now().Unix()
+			offsetVal := (latestTime - interval) / int64(intervalVal)
+			if offsetVal < 0 {
+				return "/blocks?offset=0"
```
chappjc: Would you document the logic in this check and exception?  I guess the idea is to default to /blocks with zero offset if the passed `interval` is in the future?  Seems like we should check `latestTime < interval` at the top of the function instead of down here.

dmigwi:
`fetchRowLinkURL` template function simply helps to hierarchically link the list view pages i.e. `/years` -> `/months` -> `/weeks` -> `/days` -> `/blocks`
This simply implies that 
- every row in "/years" page has a link to "/months" page
- every row in "/months" page has a link to "/weeks" page
- every row in "/weeks" page has a link to "/days" page
- every row in "/days" page has a link to "/blocks" page

explorer/explorerroutes.go
chappjc: This can be revisited, but please note by above comment.

dmigwi: This function `func (g TimeBasedGrouping) String() string` is going to be very important in relation to the next PR that I am currently testing. 
Comment noted though.

dmigwi:
Adds /years, /months, /weeks and /days pages that hierarchically linked to each other i.e /years -> /months -> /weeks -> /days -> blocks.

This implies every row in "/years" page has a link to "/months" page, every row in "/months" page has a link to "/weeks" page, every row in "/weeks" page has a link to "/days" page and every row in "/days" page has a link to "/blocks" page.

Reduces the font-size of the drop down menu and rearranges the navigation buttons so that the right most buttons returns the history data.

:point_up:

dmigwi:
Bootstrap class `table-responsive` was squeezing the tables. I removed it to fix that issue. Its not even being used in most of the other tables.

**Before**
screenshot from 2018-11-01 20-51-07

**After**
screenshot from 2018-11-01 20-51-33

views/windows.tmpl
```diff
@@ -78,7 +78,7 @@
 
         <div class="row">
             <div class="col-md-12">
-                <table class="table striped table-responsive">
+                <table class="table striped table-sm">
```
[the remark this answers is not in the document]
dmigwi: his version seems better. I will revert this.

chappjc:
This is working very well in my testing now.  At first I was seeing the blocks list pages automatically reload the homepage on some websocket event, but that seems to be sorted out after clearing cache.

The only real style change is the font size of the menu going from 20 pt to 15 pt.  It's quite easy to revert that and it is not too disruptive, so I say we merge this and see how people like it during the 3.2 dev cycle.

Issue #399, opened 2018-03-18
display site wide warning if last block is > x min old


raedah:
Display a site wide warning in the header if the most recent block is older then an x min, an hour would do it.

WARNING: Last block last more then an hour ago. dcrdata.org may not be properly connected to the network.

chappjc:
We should work on this soon. Takers?

buck54321:
Resolved by #780?

PR #780, opened 2018-10-30
Block time counter in the footer

buck54321:
Possible solution for #399. This is a persistent block time counter in the footer. The text comes red when the time is longer than my somewhat arbitrarily defined 8*blocktime = 40 minutes. 

It is not ready, but I need some advice. Right now, I don't see a straightforward way of getting basic block information into the footer on every page, but I might be missing something. There is no welcome message from the websocket with basic block info. And the page data struct used by `templates` is different for every page, so if I wanted to grab it there, I would have to add it to every page's handler in `explorerroutes`, I think. Is that the right approach? Is the basic best block information available somewhere that I'm not seeing?

### DIFF
--- a/explorer/explorerroutes.go
+++ b/explorer/explorerroutes.go
@@ -55,17 +55,17 @@ func (exp *explorerUI) Home(w http.ResponseWriter, r *http.Request) {
 	exp.pageData.RLock()
 
 	str, err := exp.templates.execTemplateToString("home", struct {
+		*CommonPageData
 		Info    *HomeInfo
 		Mempool *MempoolInfo
 		Blocks  []*BlockBasic
-		Version string
 		NetName string
 	}{
-		exp.pageData.HomeInfo,
-		exp.MempoolData,
-		blocks,
-		exp.Version,
-		exp.NetName,
+		CommonPageData: exp.commonData(),
+		Info:           exp.pageData.HomeInfo,
+		Mempool:        exp.MempoolData,
+		Blocks:         blocks,
+		NetName:        exp.NetName,
 	})
 
 	exp.MempoolData.RUnlock()
@@ -91,13 +91,13 @@ func (exp *explorerUI) SideChains(w http.ResponseWriter, r *http.Request) {
 	}
 
 	str, err := exp.templates.execTemplateToString("sidechains", struct {
+		*CommonPageData
 		Data    []*dbtypes.BlockStatus
-		Version string
 		NetName string
 	}{
-		sideBlocks,
-		exp.Version,
-		exp.NetName,
+		CommonPageData: exp.commonData(),
+		Data:           sideBlocks,
+		NetName:        exp.NetName,
 	})
 
 	if err != nil {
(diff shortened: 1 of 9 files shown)


chappjc:
Your observations are correct.  In fact we had the same issue getting the version into the footer so I manually put it in each struct passed to `execTemplateToString`.

Some options:

- Refactor how template data so that it is less ad hoc.  I might suggest keeping each struct anonymous and defined ad hoc in the handler, but if each struct inherits from a common struct, this might be doable.  e.g.:

```go
// package-level shared struct
type commonPageData struct {
	Version string
	NetName string
	BestBlockInfo SomeBestBlockDataType
	...
}
```

Then in the page handler:

```go
pageData := struct {
	common commonPageData
	TxData              *TxInfo
	Blocks            []*dbtypes.BlockStatus
	BlockInds         []uint32
	HasValidMainchain bool
	ConfirmHeight     int64
	HighlightInOut    string
	HighlightInOutID  int64
}
{ ... }
```

- A variation of the above where there is no anonymous structure, but an interface{} to hold the page specific struct which would have a concrete named type somewhere for reference.  This might get hairy though.
- Pass this data in an initial websocket message, but this won't play nice with no-JS clients, which are many in this space.

explorer/explorerroutes.go
```diff
@@ -1502,3 +1502,19 @@ func (exp *explorerUI) StatsPage(w http.ResponseWriter, r *http.Request) {
...
+// are used on every page
+func (exp *explorerUI) commonData() *CommonPageData {
+	var cd CommonPageData
+	cd.Version = exp.Version
+	cd.ChainParams = exp.ChainParams
+	cd.BlockTimeUnix = int64(exp.ChainParams.TargetTimePerBlock.Seconds())
+	var err error
+	cd.Tip, err = exp.blockData.GetTip()
+	if err != nil {
+		log.Errorf("Failed to get retreive the chain tip from the database.: %v", err)
```
chappjc: typos: `get retreive`

buck54321:
@chappjc So I went ahead and implemented something like your first suggestion. I maybe should have done this in a separate PR, because I'd like to use `CommonPageData` in another PR. What do you suggest?

public/js/controllers/main.js
```diff
@@ -30,11 +30,24 @@
         }
 
         setAges() {
+            if(this.data.has("lastblocktime")){
+              // This next line is apparently wrong :
+              // var lbt = this.data.get("lastblocktime")
+              // console.log("-- stimulus lastblocktime: "+lbt)
+              // It doesn't give the correct value if the data attribute has
+              // been updated elsewhere. Using jQuery until I can figure it out
```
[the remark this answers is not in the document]
chappjc: I wish I could answer definitively, but I'm getting to know it too.  I have seen that `stuffTarget.data("blah")` will get `"thisvalue"` from the element with `data-target="thing.stuff" data-blash="thisvalue"`

chappjc:
This looks really good in testing.  Here's a screenshot for everyone (note the bottom left):

image

The timer resets as expected when a new block is mined.

explorer/explorertypes.go
```diff
@@ -578,6 +593,13 @@ type StatsInfo struct {
 	RewardWindowSize           int64
 }
 
+type CommonPageData struct {
```
chappjc: Please add docs to all exported functions, starting like `// CommonPageData`.

chappjc:
Some minor conflict resolution needed.  Please do via rebase in your local workstation.  Thanks.

buck54321:
Adds a time since last block counter to the footer. Implements CommonPageData struct for HTML templates.

PR #786, opened 2018-11-01
scope table-responsive class to narrow screens in /blocks table

gozart1:
This fixes /blocks table not expanding to full width after upgrading to latest Bootstrap.

**Before**
screen shot 2018-11-01 at 11 10 15 am

**After
screen shot 2018-11-01 at 11 08 23 am
**
image

### DIFF
--- a/views/explorer.tmpl
+++ b/views/explorer.tmpl
@@ -78,10 +78,10 @@
 
         <div class="row">
             <div class="col-md-12">
-                <table class="table striped table-responsive" id="explorertable">
+                <table class="table striped table-responsive-sm" id="explorertable">
                     <thead>
                         <tr>
-                            <th>Ticket Price Window #</th>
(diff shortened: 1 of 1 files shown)


chappjc:
Would you mind making the same switch in rejects.tmpl, sidechains.tmpl, and windows.tmpl part of this PR too?  It seems like the Latest Blocks table on home.tmpl doesn't get this, but I just wanna be sure about that too.

gozart1:
Good point, updated for rejects.tmpl, sidechains.tmpl, and windows.tmpl.

Made a minor change to sidechains.tmpl as well to keep the `true/false` & `none` values from getting chopped into multiple lines.

For home.tmpl I think it makes sense to keep the globally scoped class since that table is always in a narrow space. Looking at it makes my think we could rename "Revocations" to "Revokes" so that column isn't as wide.

chappjc:
I'm OK with Revocations -> Revokes if you wanna sneak that in too.

PR #787, opened 2018-11-01
images: shrink svg files with svg cleaner

chappjc:
Shrink all svg images with svgcleaner, except for sprites.svg.

Shrinking sprites.svg using svgcleaner with the `--remove-unused-defs=no` option seems possible, but I don't want to risk messing it up.

The total size of the 20 shrunk SVGs in this PR:
- Before: 38,623 bytes
- After: 27,794 bytes

10 kB is a relatively small savings, but it's for free.

Note: new lines are kept, with tab indentation, to keep visual diffs meaningful in the future.

This also renames abstain.svg to connecting.svg since this is not the abstain dot but the 
WebSocket connecting" dot.  main.css is updated accordingly.


gozart1:
LGTM

PR #788, opened 2018-11-01
rename revocations to revokes

gozart1:
Minor change that renames revocations to revokes in the front end.

Since our tables can have a lot of columns and space can be at a premium, we should always choose the shorter version of a word if it's just as clear as the longer version. This helps with using that horizontal space efficiently.

I think it's best to be consistent with the choice of words so I've renamed it in all places exposed to the end user.

### DIFF
--- a/views/block.tmpl
+++ b/views/block.tmpl
@@ -145,7 +145,7 @@
                         </td>
                     </tr>
                     <tr>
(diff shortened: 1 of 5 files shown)


gozart1:
Ugh! So sorry about that terrible commit message. I usually don't use the github ui to squash and merge so I missed putting the title :(

PR #789, opened 2018-11-01
upgrade to jquery v3.3.1

gozart1:
This upgrades jquery from v1 to v3, which has some bug fixes and more efficient execution due to dropping ie8 support ( which is fine since we don't support that anyways ).
https://www.dotnetcurry.com/jquery/1309/jquery-3-new-features

There where a few breaking changes, and this PR should address them all.
Unfortunately dygraphs depends on 1.x jQuery. But the good news is that we can use jquery-migrate.js to solve the backwards compatibility issue. 

At 85kb for the minified jquery.js + 9kb for the jquery-migrate.js shim we are at roughly the same file size as a minified 1.x ( we're currently serving 1.x un-minified, so we'll see some nice page weight savings from this change ).


chappjc:
Looks like there's a (trivial) conflict to resolve.  But I'm good to merge if someone else has tested too.  @ademuanthony or @dmigwi ?

Issue #776, opened 2018-10-29
Set block time data type to be timestamp instead of big int.


dmigwi:
Currently, to do a time-based grouping in a given table (i.e group by month, week, days, hours e.t.c) we use integer division and modulus to group records instead on the builtin date/time functionality. Grouping by months accurately becomes almost impossible.

**Cost of switching from big int to timestamp data type**
- Both timestamp and big int occupy occupy 8 bytes of memory space thus no space change.

- Grouping by builtin date/time functions is more expensive but more accurate.
![screenshot from 2018-10-29 06-06-30](https://user-images.githubusercontent.com/22055953/47643779-7b403300-db7d-11e8-91da-9b6ec583994b.png)

- Current db block time insertion methods will integrate seamlessly.

- Time formatting implementation done in golang could be partly done by default by the db.

PR #791, opened 2018-11-02
Set block time datatype to timestamp.

dmigwi:
As described in #776, This PR converts the block time values stored in the Auxiliary DB (postgresql) from `big int` data type to `timestamp` data type.

One of the reason behind this PR is that months and years don't have a constant number of days throughout but when grouping data we need it to be grouped as accurate as possible. The needed accuracy can only be obtained optimally when the builtin date/time functions e.g. `date_trunc` are used.

Performance wise no significant decrease that has been noted.

This PR results to a db upgrade (3.6.0), when testing one can use a separate db to avoid corrupting all the data available without a backup. To create a backup: `create database dcrdata_backup template dcrdata owner dcrdata;`

### DIFF
--- a/db/dbtypes/types.go
+++ b/db/dbtypes/types.go
@@ -14,6 +14,28 @@ import (
 	"github.com/decred/dcrdata/v3/db/dbtypes/internal"
 )
 
+// TimeDef is time.Time wrapper that formats time by default as a string without
+// a timezone. The time Stringer interface formats the time into a string
+// with a timezone.
+type TimeDef struct {
+	T time.Time
+}
+
+func (t *TimeDef) String() string {
+	return t.T.Format("2006-01-02 15:04:05")
+}
+
+// TimeAPI is a fall back dbtypes.TimeDef wrapper that allows API endpoint that
+// previously accepted timestamp instead of a formatted string time to continue
+// working.
+type TimeAPI struct {
+	S TimeDef
+}
+
+func (t *TimeAPI) String() int64 {
+	return t.S.T.Unix()
+}
+
 // Tickets have 6 states, 5 possible fates:
 // Live -...---> Voted
 //           \-> Missed (unspent) [--> Revoked]
@@ -129,9 +151,9 @@ type BlocksGroupedInfo struct {
 	EndBlock           int64
 	Difficulty         float64
 	TicketPrice        int64
-	StartTime          time.Time
+	StartTime          TimeDef
 	FormattedStartTime string
-	EndTime            time.Time
+	EndTime            TimeDef
 	FormattedEndTime   string
 	Size               int64
 	FormattedSize      string
@@ -410,7 +432,7 @@ type AddressRow struct {
 	// funding tx outputs.
 	MatchingTxHash   string
 	IsFunding        bool
-	TxBlockTime      time.Time
+	TxBlockTime      TimeDef
 	TxHash           string
 	TxVinVoutIndex   uint32
 	Value            uint64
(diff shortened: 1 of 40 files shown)


systemstesting/serverPerformanceTest.go
```diff
@@ -0,0 +1,79 @@
+package main
+
```
dmigwi: This file attempts to create an easy way we can measure the server performance by checking the API status codes and the response time from the selected API endpoints.

chappjc:
Just started looking and it's good so far.  One concern is that the Time field in many apitypes structs is now `time.Time` instead of `int64`, which is a breaking change.  Unfortunately we've not versioned the API yet, so we can't simply bump to a /v2 path.

Two possible solutions:
1. Keep the api responses returning `int64` (unix time).
2. Add a new field to the types instead of modifying `Time`, perhaps `TimeString` or `FormattedTime`.

db/dcrpg/internal/stakestmts.go
```diff
@@ -118,10 +118,10 @@ const (
 		WHERE pool_status = 0 AND tickets.is_mainchain = TRUE
 		GROUP BY price ORDER BY price;`
 
-	SelectTicketsByPurchaseDate = `SELECT (transactions.block_time/$1)*$1 as timestamp,
+	selectTicketsByPurchaseDate = `SELECT %s as timestamp,
```
chappjc: Soooo much nicer.  :)

dmigwi:
@chappjc I suspect not all dcrdata API endpoints are in use by external parties e.g `/chart` and `/ticketpool` may not be in use by others. Do we have specific endpoints that I can focus on in retaining the time value as a timestamp?

dmigwi:
Rather than adding an extra field to display time in the required format, I used wrappers to print time into the required format when the time stringer is invoked.

dmigwi:
I have fixed all the APIs but I appreciate if someone confirmed again that all is well.

dmigwi:
**Before this PR the start date wasn't consistent with the expected norm.**

before-nov-3


**After.**

after-fix-nov-3

chappjc:
I'll try to apply papacarp's Postman test suite.

db/dbtypes/types.go
```diff
@@ -9,10 +9,43 @@ import (
...
+
+// TimeAPI is a fall back dbtypes.TimeDef wrapper that allows API endpoint that
+// previously accepted timestamp instead of a formatted string time to continue
+// working.
+type TimeAPI struct {
+	S TimeDef
+}
+
+func (t TimeAPI) String() string {
+	return t.S.T.Format("2006-01-02 15:04:05")
```
chappjc: I'd suggest to return `t.S.String()` instead.

However if `TimeAPI` embedded `TimeDef` as in

```go
type TimeAPI struct {
    TimeDef
}
```

(without naming it `S`), then you would not need to implement `(t TimeAPI) String()` as it would get it from `TimeDef`.  You would still need to implement `(t *TimeAPI) MarshalJSON` of course, but it would be just `return json.Marshal(t.TimeDef.T.Unix())`.

I understand defining `TimeAPI` in this package though if it helps deal with dbtypes depending on apitypes.

db/dcrpg/queries.go
```diff
@@ -741,18 +741,19 @@ func RetrieveTicketIDsByHashes(db *sql.DB, ticketHashes []string) (ids []uint64,
 // retrieveTicketsByDate fetches the tickets in the current ticketpool order by the
 // purchase date. The maturity block is needed to identify immature tickets.
 // The grouping interval size is specified in seconds.
```
chappjc: I noticed several comments describing the old parameters and types.

db/dbtypes/types.go
dmigwi:
```
type TimeAPI struct {
    TimeDef
}
```
This version of the struct made my linter  to complain that I was doing unkeyed struct assignment. The keyed entry made it easier to access much of the time functionality that is accessible via the wrapper.

db/dbtypes/types.go
```diff
@@ -9,10 +9,43 @@ import (
...
+
+// MarshalJSON is set as the default marshalling function for TimeDef struct.
+func (t *TimeDef) MarshalJSON() ([]byte, error) {
+	return json.Marshal(t.String())
+}
+
+// TimeAPI is a fall back dbtypes.TimeDef wrapper that allows API endpoint that
+// previously accepted timestamp instead of a formatted string time to continue
+// working.
+type TimeAPI struct {
```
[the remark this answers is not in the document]
dmigwi: Having both struct share a name even though they will located in different packages is quite confusing. I propose we stick with `dbtypes.TimeDef` and `apitypes.TimeAPI` just to help differentiate the structs on inspection.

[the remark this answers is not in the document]
dmigwi: I would also not advocate we name either struct as `Time` since the default time field on several structs is `Time`.
```golang
Time  []Time `json:"time,omitempty"`
```
This looks not right to me.

db/dbtypes/types.go
chappjc: I won't push my opinion further, but note that the solution is

`TimeAPI{TimeDef: t}`

Regarding accessing time functionality, you still do stuff like `t.TimeDef.T.Unix()`.  The reason I suggested this is because the definition of String is redundant.

dmigwi: Let me try it. It looks easier than my current implementation.

db/dbtypes/types.go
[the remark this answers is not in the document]
chappjc: No problem keeping the names.  Personally I find the import prefixes enough to disambiguate, but it's a minor style issue. I'm also not personally bothered by `Time []Time`, but I understand it can be hard to swallow.
PR looks good.

db/dbtypes/types.go
dmigwi: Just discovered that this might not work as intended because I moved `TimeAPI` struct to `apitypes` package.

chappjc: Oh, right.  Well let's just revert back and save it for another day, unless you wanted to try something else.

dmigwi:
The last commit above fixes a bug that was resulting to a timestamp of value 0 to be saved in the sqlite db for all the records. This implies that all the sqlite db created before the last commit is already corrupted and should be deleted.
Otherwise everything else should be fine including the fine tuned time-based row links generated.

chappjc:
Approving the changes, but given the nature of the DB upgrade, it needs testing from more folks.  Also, we are not in a hurry to push this in.

chappjc:
Upgrade was smooth:

```
2018-11-14 09:24:39.613 [DBG] PSQL: Table agendas: v3.5.5
2018-11-14 09:24:39.613 [DBG] PSQL: Table vins: v3.5.5
2018-11-14 09:24:39.613 [DBG] PSQL: Table vouts: v3.5.5
2018-11-14 09:24:39.613 [DBG] PSQL: Table block_chain: v3.5.5
2018-11-14 09:24:39.613 [DBG] PSQL: Table tickets: v3.5.5
2018-11-14 09:24:39.613 [DBG] PSQL: Table votes: v3.5.5
2018-11-14 09:24:39.613 [DBG] PSQL: Table blocks: v3.5.5
2018-11-14 09:24:39.613 [DBG] PSQL: Table transactions: v3.5.5
2018-11-14 09:24:39.613 [DBG] PSQL: Table addresses: v3.5.5
2018-11-14 09:24:39.613 [DBG] PSQL: Table misses: v3.5.5
2018-11-14 09:24:39.614 [INF] PSQL: Dropping block time indexes
2018-11-14 09:24:39.629 [INF] PSQL: Setting the addresses table block_time column data type to TIMESTAMP. Please wait...
2018-11-14 09:29:28.666 [INF] PSQL: Setting the blocks table time column data type to TIMESTAMP. Please wait...
2018-11-14 09:29:33.797 [INF] PSQL: Setting the agendas table block_time column data type to TIMESTAMP. Please wait...
2018-11-14 09:29:38.129 [INF] PSQL: Setting the transactions table time column data type to TIMESTAMP. Please wait...
2018-11-14 09:30:24.413 [INF] PSQL: Setting the transactions table block_time column data type to TIMESTAMP. Please wait...
2018-11-14 09:31:09.586 [INF] PSQL: Setting the vins table block_time column data type to TIMESTAMP. Please wait...
2018-11-14 09:32:58.306 [INF] PSQL: Modified the tickets table version to 3.6.0
2018-11-14 09:32:58.306 [INF] PSQL: Modified the votes table version to 3.6.0
2018-11-14 09:32:58.306 [INF] PSQL: Modified the misses table version to 3.6.0
2018-11-14 09:32:58.306 [INF] PSQL: Modified the vins table version to 3.6.0
2018-11-14 09:32:58.306 [INF] PSQL: Modified the vouts table version to 3.6.0
2018-11-14 09:32:58.306 [INF] PSQL: Modified the block_chain table version to 3.6.0
2018-11-14 09:32:58.306 [INF] PSQL: Modified the agendas table version to 3.6.0
2018-11-14 09:32:58.306 [INF] PSQL: Modified the blocks table version to 3.6.0
2018-11-14 09:32:58.306 [INF] PSQL: Modified the transactions table version to 3.6.0
2018-11-14 09:32:58.306 [INF] PSQL: Modified the addresses table version to 3.6.0
2018-11-14 09:32:58.306 [INF] PSQL: Creating the dropped block time indexes
2018-11-14 09:33:11.938 [INF] PSQL: Table upgrades completed.
```

dmigwi:
It converts all the block time data columns stored in the Auxiliary db to TIMESTAMP datatype instead on BIG INT(int64) datatype. This conversion guarantees high accuracy especially in time-based grouping queries.

Two wrappers; TimeDef and TimeAPI are introduced. TimeDef, a time.Time wrapper helps in automating the time formatting into human-readable time format without adding the timezone into the formatted time. It also helps doing away with the FormattedTime field in most payloads.

TimeAPI, a TimeDef wrapper guarantees that all API endpoint that previously returned a timestamp instead of a formatted time, continue to return a timestamp.

:point_up: @chappjc

PR #794, opened 2018-11-05
Extended search to check aux DB for blocks

ademuanthony:
Resolve [Issue 746](https://github.com/decred/dcrdata/issues/746)

Extended the search page to search for block and address in the chaindb if the specified block height index or hash is not found in the sqlite db and the app is running the full mode.

Transaction uses rpc and so was not altered as that should cover all cases in both full and lite mode

### DIFF
--- a/explorer/explorerroutes.go
+++ b/explorer/explorerroutes.go
@@ -1403,14 +1403,21 @@ func (exp *explorerUI) Search(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	// Attempt to get a block hash by calling GetBlockHash to see if the value
-	// is a block index and then redirect to the block page if it is.
+	// Attempt to get a block hash by calling GetBlockHash of wiredDB and BlockHash of ChainDB (if not int lite) to see
+	// if the value is a block index and then redirect to the block page if it is.
 	idx, err := strconv.ParseInt(searchStr, 10, 0)
 	if err == nil {
 		_, err = exp.blockData.GetBlockHash(idx)
 		if err == nil {
 			http.Redirect(w, r, "/block/"+searchStr, http.StatusPermanentRedirect)
 			return
+		}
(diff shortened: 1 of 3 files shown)


chappjc:
Transactions have the same issue as the orphaned blocks they are in.

chappjc:
You seem to have pulled in a bunch of commits from master.  Please squash those out and rebase.

explorer/explorerroutes.go
```diff
@@ -1444,6 +1444,14 @@ func (exp *explorerUI) Search(w http.ResponseWriter, r *http.Request) {
 		http.Redirect(w, r, "/tx/"+searchStr, http.StatusPermanentRedirect)
 		return
 	}
+	if !exp.liteMode {
+		// Search for occurrences of the transaction in the database.
+		dbTxs, _ := exp.explorerSource.Transaction(searchStr)
```
chappjc: Still need to check error.  The only errors we should be OK with are `nil` (when we should redirect) and `sql.ErrNoRows` (which should not throw an internal server error, but also not redirect).  When `err != nil && err != sql.ErrNoRows`, that is an internal server error.  Note that `dbTxs==nil` when `err == sql.ErrNoRows`.

chappjc:
This PR requires the following fixes to be merged first:
- https://github.com/decred/dcrdata/pull/819
- https://github.com/decred/dcrdata/pull/820

This PR is intended to search for data relating to orphaned blocks.  Thus, testing can be performed by browsing the blocks listed on the /side page.

chappjc:
Rebased and ready for testing.  I recommend checking the `/side` page and browsing through all the block data for a side chain block listed there.

Since the rebase involved a force push, please `git fetch {{thisremote}} && git checkout issue-746-search && git reset --hard {{thisremote}}/issue-746-search` to avoid an unintentional merge.

Issue #649, opened 2018-08-29
Chain Params page feedback


xaur:
Feedback for https://explorer.dcrdata.org/parameters, footer says v2.1.2+release.

* Page and tab title: "Decred ChainParams" -> "Decred Chain Parameters"
* first label: "MainNet Parameters [github](https://github.com/decred/dcrd/blob/master/chaincfg/params.go)" -> "Parameters for Decred **mainnet** as declared in chaincfg/params.go ([github](https://github.com/decred/dcrd/blob/master/chaincfg/params.go))".
  * Anyone agrees "MainNet" is ugly?
* Left-aligned first column may look better, worth trying.
* "TicketPoolSize 8192 (40960 actual)" please explain in Description column
* Both PubKeyHashAddrID and PKHEdwardsAddrID have "P2PKH address prefix" in description, while PKHSchnorrAddrID adds the detail about Schnorr

chappjc:
Yeah, "MainNet" is ugly.

chappjc:
bump @McEdward

chappjc:
Resolved by https://github.com/decred/dcrdata/pull/795.

PR #795, opened 2018-11-05
Parameters page tweaks.

buck54321:
As per #649

### DIFF
--- a/explorer/explorertypes.go
+++ b/explorer/explorertypes.go
@@ -513,9 +513,9 @@ type AddrPrefix struct {
 func AddressPrefixes(params *chaincfg.Params) []AddrPrefix {
 	Descriptions := []string{"P2PK address",
 		"P2PKH address prefix. Standard wallet address. 1 public key -> 1 private key",
-		"Ed25519 high-speed signature",
+		"Ed25519 P2PKH address prefix",
 		"secp256k1 Schnorr P2PKH address prefix",
-		"P2SH Stake Pool (multi-sig 1-of-2) address",
+		"P2SH address prefix",
 		"WIF private key prefix",
 		"HD extended private key prefix",
 		"HD extended public key prefix",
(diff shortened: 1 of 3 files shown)


explorer/explorertypes.go
```diff
@@ -510,10 +510,10 @@ type AddrPrefix struct {
 // AddressPrefixes generates an array AddrPrefix by using chaincfg.Params
 func AddressPrefixes(params *chaincfg.Params) []AddrPrefix {
 	Descriptions := []string{"P2PK address",
-		"P2PKH address prefix",
-		"P2PKH address prefix",
+		"P2PKH address prefix. Standard wallet address. 1 public key -> 1 private key",
+		"Ed25519 high-speed signature",
 		"secp256k1 Schnorr P2PKH address prefix",
-		"P2SH address prefix",
+		"P2SH Stake Pool (multi-sig 1-of-2) address",
```
chappjc: Stake pool is the most common, but P2SH can be any purpose.  I'm not sure if we can same much more beyond "P2SH address prefix" except perhaps to spell out "pay to script hash".

explorer/explorertypes.go
```diff
@@ -510,10 +510,10 @@ type AddrPrefix struct {
 // AddressPrefixes generates an array AddrPrefix by using chaincfg.Params
 func AddressPrefixes(params *chaincfg.Params) []AddrPrefix {
 	Descriptions := []string{"P2PK address",
-		"P2PKH address prefix",
-		"P2PKH address prefix",
+		"P2PKH address prefix. Standard wallet address. 1 public key -> 1 private key",
+		"Ed25519 high-speed signature",
```
chappjc: Still should say it's a P2PKH address (just using the edwards sig algo), similar to the description "secp256k1 Schnorr P2PKH address prefix"

PR #805, opened 2018-11-11
Convert Javascript to ES6 modules loaded & linted via webpack

gozart1:
This converts the javascript codebase to ES6 modules and adds [webpack](https://webpack.js.org/) for front end dev tooling and production asset bundling.

- Javascript code style is enforced via [ESlint](https://eslint.org/docs/rules/) + [standardJS](https://standardjs.com/) 

In order to convert to modules, several small refactors had to be done.

- Created new folders for organizing code into services, helpers, and controllers.
- All inlined javascript has been removed from extras.tmpl and moved into index.js. This is a temporary measure and much of this logic will continue to be organized properly into controllers, services and helpers in subsequent PRs.

Stimulus controller filenames renamed to `public/js/controllers/[name]_controller.js`
- /rawtx logic has been converted to a Stimulus controller
- mempool and homepageMempool have been split into their own dedicated modules containing the respective controllers

The following concerns have been moved into `public/js/services/<name>_service.js`:
- keyboard_navigation
- theme ( dark / light switching )
- cookie (`setCookie()`)
- event_bus (`EventBus` structured as a class exporting it's instance).
- message_socket (`MessageSocket` structured as a class exporting it's instance)

The following common functions have been moved into `public/js/helpers/<name>_helper.js`
- animation `animationFrame()`
- chart `barChartPlotter()`
- humanize functions

The README has also been updated to give an overview of the front end development philosophy and workflow.


views/extras.tmpl
```diff
@@ -94,429 +87,14 @@
...
-<script src="/js/notify.min.js"></script>
 {{end}}
 
 {{define "footer"}}
 <footer class="navbar-fixed-bottom">
     <div class="container d-flex justify-content-between align-items-center">
         <ul class="nav justify-contents-left nowrap blockcounter">
           <li class="nav-item text-left">
-            <span data-controller="main" data-target="main.age" data-main-lastblocktime="{{$.Tip.Time}}"></span> <span class="block-msg">since last block</span>
+            <span data-controller="time" data-target="time.age" data-main-lastblocktime="{{$.Tip.Time}}"></span> <span class="block-msg">since last block</span>
```
buck54321: The `data-main-lastblocktime` here is actually part of the stimulus naming convention, so it too should be changed to `data-time-lastblocktime`, so that stimulus can use the `data.has` function in `time_controller.js`. Appropriate updates are needed in other places where this attribute is referenced as well.

gozart1:
Your PR is merged @buck54321 

This PR is ready for some review and PR's to fix minor issues and catch up to master.

I've been working through this and have had to refactor lots of things to eliminate globals.

@dmigwi https://github.com/decred/dcrdata/blob/e65493cfb18d8bd3aca304cfcdf539d32772d120/public/js/controllers/ticketpool_controller.js
I'd like you to note how the state for onZoom and onBarsChange is now modeled by binding data to the this.zoom and this.bars on the controller instead of using css classes. CSS should never be used to model application state, it's only domain is to represent it visually. It's important to keep a separation of concerns so that the code is less brittle and easier to reason about.

views/extras.tmpl
```diff
@@ -83,7 +76,7 @@
                         {{else}}
                             <a data-keynav-skip href="https://explorer.dcrdata.org/" title="Home">Switch To Mainnet</a>
                         {{end}}
-                        <a data-keynav-skip data-turbolinks="false" href="javascript:toggleSun()" class="jsonly"><span id="sun-icon" class="dcricon-sun-fill no-underline pr-2"></span> Night Mode</a>
+                        <span data-keynav-skip data-action="click->menu#onSunClick" class="jsonly"><span id="sun-icon" class="dcricon-sun-fill no-underline pr-2"></span> Night Mode</span>
```
buck54321: This was behaving weird in Brave. The span needed `display:inline-block;`

public/js/services/event_bus_service.js
```diff
@@ -0,0 +1,33 @@
...
+      eventCallbacksPair.callbacks.push(callback)
+    } else {
+      eventCallbacksPairs.push(new EventCallbacksPair(eventType, callback))
+    }
+  }
+
+  publish (eventType, args) {
+    const eventCallbacksPair = findEventCallbacksPair(eventType)
+    if (!eventCallbacksPair) {
+      console.error('no subscribers for event ' + eventType)
```
buck54321: I don't think this is an error.

public/js/controllers/homepageMempool_controller.js
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  disconnect () {
+    ws.deregisterEvtHandlers('newtx')
+    ws.deregisterEvtHandlers('mempool')
+    ws.deregisterEvtHandlers('getmempooltxsResp')
+  }
+
+  renderLatestTransactions (txs, incremental) {
+    console.log('render', txs)
```
buck54321: Can we lose this?

README.md
```diff
@@ -739,6 +740,40 @@ More details are in [its own README](./cmd/scanblocks/README.md). The repository
...
 
+## Front End Development
+
+Make sure you have a recent version of [node and npm](https://nodejs.org/en/download/) installed.
+You may want to use the [node version manager (nvm)](https://github.com/creationix/nvm) for managing
+your node download and installation.
+
+From the dcrdata root directory, run the following command to install the node modules.
+
+`npm install`
```
chappjc: What is the output of this command?  As a node noob, I was searching around to see what it spat out.

And after this command, is it ok to run dcrdata?  Or does this just get the depends to build the bundle?

README.md
```diff
@@ -739,6 +740,40 @@ More details are in [its own README](./cmd/scanblocks/README.md). The repository
...
+
+For development, there's a webpack script that watches for file changes and automatically compiles.
+To use it, run this command in a separate terminal and leave it running while you work.
+
+`npm run watch`
+
+For production, bundle assets via:
```
chappjc: Is it optional?  And I'm wondering the same thing as in my previous commit about where the outputs go.  It seems to be generating public/js/dist/app.bundle.js.  Is that all it makes?

README.md
buck54321: The primary output is a directory called `node_modules`. I think that's all.

README.md
buck54321: Only if you're editing javascript files.

chappjc:
Please see https://github.com/gozart1/dcrdata/pull/5 with buck's last two fixes but for your rebased branch

Issue #723, opened 2018-10-06
Add network hashrate chart


jzbz:
Currently https://explorer.dcrdata.org/charts has PoW difficulty but not network hashrate. Having the latter would be useful as well.

chappjc:
I started looking into this today.  We have a CalculateHashRate function that does the computation based on difficulty and target block time.  However, the implementation in dcrd's `getnetworkhashps` is very different, accumulating work over a variable period of blocks. The hander is defined here: https://github.com/decred/dcrd/blob/7e3dfb21974681090ad703c2350fe8146f14aa08/rpcserver.go#L3154

xaur:
Hashrate will be derived solely from chain data, not using anything that pools report -- correct?

chappjc:
Absolutely.  Only based on work done on the blockchain over a given period

oregonisaac:
Ping: suggesting we move forward with adding current hash rate based on the block difficulty and the default past 120 (or some other stated number) blocks using RPCs in dcrd.

chappjc:
@buck54321 has a proper solution underway

chappjc:
Resolved by https://github.com/decred/dcrdata/pull/808

PR #808, opened 2018-11-12
Chainwork DB upgrade. Hashrate and Chainwork charts. 

### LINKED ISSUE #744: include chainwork in relevant API responses
From @davecgh in Slack:
> Between the `getblockchaininfo` RPC and https://github.com/decred/dcrd/pull/1498 it's now possible to query the `chainwork` which is the cumulative total work of the chain up to that point (e.g. the total number of expected hashes needed to produce the chain up to that block).  I think it would be useful to add this to dcrdata.

Where:
1. The response for a block request.
2. Charts (chainwork over time).
3. Wherever else we return blockchain status, etc.

buck54321:
Resolves #744 and #723.

This PR performs a PostgreSQL DB upgrade, adding a `TEXT` column, `chainwork`. Relevant block insertion routines are updated to grab the [newly available chainwork value](https://github.com/decred/dcrd/pull/1498) from dcrd.  Having chainwork also enables quick calculation of network hashrate data. A network hashrate chart and "total work" chart have been added to `/charts`.

### DIFF
--- a/db/dcrpg/queries.go
+++ b/db/dcrpg/queries.go
@@ -10,7 +10,9 @@ import (
 	"database/sql"
 	"encoding/hex"
 	"fmt"
+	"math/big"
 	"strings"
+	"time"
 
 	"github.com/decred/dcrd/blockchain/stake"
 	"github.com/decred/dcrd/chaincfg"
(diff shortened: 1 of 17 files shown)


chappjc:
We're discussing in Slack if dcrd/rpcclient and dcrd/dcrjson are ready for a new version tag so we can bump the module without using a pseudo-version.

buck54321:
I kept all the changes in this PR, even though the title is a little outdated. This will add a new column to your PostgreSQL blocks table. I've just been dropping the column from the table on the command line as needed. 

I took a couple of liberties with the units. Since chainwork is a 32-byte hex string representing a number > 2^64, for use in JSON/charts I converted work to exahash and hashrate to terahash/s.

db/dcrpg/queries.go
```diff
@@ -2475,6 +2476,81 @@ func retrieveTicketByOutputCount(db *sql.DB, dataType outputCountType) (*dbtypes
...
+		if !exawork.IsUint64() {
+			log.Errorf("Failed to make uint64 from chainwork %s", workhex)
+			break
+		}
+		workdata.ChainWork = append(workdata.ChainWork, exawork.Uint64())
+		workdata.Time = append(workdata.Time, blocktime)
+
+		// Hashrates are binned into days
+		if blocktime > dayStamp {
+			if lastStamp > 0 {
```
buck54321: I suspect others might have opinions on this decision. Single-block hashrate data is worthless, and even up to many tens of blocks, it's pretty noisy. I chose to bin hashrate into days here, but binning into some number of blocks might also be appropriate. For reference, by default, the RPC call `getnetworkhashps` returns the most recent 120-block average.

db/dcrpg/upgrades.go
```diff
@@ -1195,3 +1227,74 @@ func versionAllTables(db *sql.DB, version TableVersion) error {
...
+	// Grab the blockhashes from rows that don't have chainwork.
+	rows, err := db.Query(`SELECT hash FROM blocks WHERE chainwork IS NULL;`)
+	if err != nil {
+		log.Error("Failed to query database for missing chainwork: %v", err)
+		return 0, err
+	}
+	defer rows.Close()
+
+	var hashStr string
+	var updated int64 = 0
```
chappjc: Please ommit `= 0` since that is the zero value that gets set anyway.

db/dcrpg/upgrades.go
```diff
@@ -1195,3 +1227,74 @@ func versionAllTables(db *sql.DB, version TableVersion) error {
...
+		}
+
+		_, err = stmt.Exec(chainWork, hashStr)
+		if err != nil {
+			log.Errorf("Failed to insert chainwork (%s) for block %s. Aborting chainwork sync: %v", chainWork, hashStr, err)
+			return updated, err
+		}
+		updated += 1
+		// Every two minutes, report the sync status.
+		if updated%100 == 0 && time.Now().Unix()-tReport > 120 {
```
chappjc: `time.Since(tReport) > 2*time.Minute`

db/dcrpg/upgrades.go
```diff
@@ -1195,3 +1227,74 @@ func versionAllTables(db *sql.DB, version TableVersion) error {
...
+			tReport = time.Now().Unix()
```
chappjc: `tReport = time.Now()`

db/dcrpg/upgrades.go
```diff
@@ -1195,3 +1227,74 @@ func versionAllTables(db *sql.DB, version TableVersion) error {
...
+	if err != nil {
+		log.Error("Failed to query database for missing chainwork: %v", err)
+		return 0, err
+	}
+	defer rows.Close()
+
+	var hashStr string
+	var updated int64 = 0
+	client := blockgate.Client()
+	tReport := time.Now().Unix()
```
chappjc: `tReport = time.Now()`

db/dcrpg/upgrades.go
```diff
@@ -1195,3 +1227,74 @@ func versionAllTables(db *sql.DB, version TableVersion) error {
...
+			log.Errorf("GetChainWork failed (%s). Aborting chainwork sync.: %v", hashStr, err)
+			return updated, err
+		}
+
+		_, err = stmt.Exec(chainWork, hashStr)
+		if err != nil {
+			log.Errorf("Failed to insert chainwork (%s) for block %s. Aborting chainwork sync: %v", chainWork, hashStr, err)
+			return updated, err
+		}
+		updated += 1
```
chappjc: `updated++`

db/dcrpg/upgrades.go
```diff
@@ -1195,3 +1227,74 @@ func versionAllTables(db *sql.DB, version TableVersion) error {
...
+
+	// Grab the blockhashes from rows that don't have chainwork.
+	rows, err := db.Query(`SELECT hash FROM blocks WHERE chainwork IS NULL;`)
+	if err != nil {
+		log.Error("Failed to query database for missing chainwork: %v", err)
+		return 0, err
+	}
+	defer rows.Close()
+
+	var hashStr string
```
chappjc: I suggest limiting the scope of `hashStr` to inside the `for` loop, but obv it's not very important.

db/dcrpg/upgrades.go
```diff
@@ -1195,3 +1227,74 @@ func versionAllTables(db *sql.DB, version TableVersion) error {
...
+	stmt, err := db.Prepare(`UPDATE blocks SET chainwork=$1 WHERE hash=$2;`)
+	if err != nil {
+		log.Error("Failed to prepare chainwork insertion statement: %v", err)
+		return 0, err
+	}
+	defer stmt.Close()
+
+	// Grab the blockhashes from rows that don't have chainwork.
+	rows, err := db.Query(`SELECT hash FROM blocks WHERE chainwork IS NULL;`)
+	if err != nil {
```
chappjc: You could avoid the `count` query by doing this query first, and then `if err == sql.ErrNoRows { return 0, nil }`, but since this code only runs once, it's no big deal.

chappjc:
BTW, I was testing a rebased version: https://github.com/chappjc/dcrdata/tree/chainwork-api

db/dcrpg/upgrades.go
buck54321: I use the count to calculate progress for the status reports also.

chappjc: Ah, I see that now.

Issue #809, opened 2018-11-12
[3.1] Fix table row height, and decimal spacing


chappjc:
In the `3.1-stable` branch, the last release showstopper is some formatting issues.

1. Height of rows in table on address page is usually too high.  e.g.:

![image](https://user-images.githubusercontent.com/9373513/48363843-83fc3300-e66c-11e8-9b50-c4cc43760f0c.png)

2. Decimal spacing is still inconsistent:

![image](https://user-images.githubusercontent.com/9373513/48363880-a3935b80-e66c-11e8-9d92-69a3f982db10.png)

chappjc:
resolved by https://github.com/decred/dcrdata/pull/810

Issue #782, opened 2018-10-31
nexthome.css is a mess, is overwriting core bootstrap classes


gozart1:
Right now nexthome.css is overwriting various bootstrap layout classes and this is a huge anti pattern.
https://github.com/decred/dcrdata/blob/master/public/css/nexthome.css#L509

Please, if you want to do front end work, learn what bootstrap is and how to use it. Especially the grid and positioning system.
https://css-tricks.com/ is a great resource for general css knowledge.
In general though, we should lean heavily on boostrap as it's a battle tested approach that solves countless cross browser issues.

I'm not sure who authored that nexthome.css code, I think @dmigwi just ported it from someone somewhere outside this repo. But it needs to be cleaned up.

chappjc:
If it is at all possible to keep nexthome.css from loading except on /nexthome, let's do that.

In order to have a stable 3.1, we need to have things almost the same as in 3.0, perhaps with an updated logo and some other tweaks, but no major redesign stuff.

@dmigwi I assigned you to have a look at addressing the issues addressed here.  I think this may be the time to step back from coding and look closer at bootstrap best practices so we are pointing in the right direction after this.

dmigwi:
@gozart1 Thanks for raising the above concerns, I had not given browser compatibility a serious thought till now. The design team(in @raedah's group) came up with the nexthome's page css so to as hasten the feature implementation.

@chappjc We shall do that ASAP.

chappjc:
Is there a repo to report such issues?  https://github.com/raedahgroup/design seems to be empty.

dmigwi:
The code is on a bitbucket repo https://bitbucket.org/vikkbehal/decred/src/master/

itswisdomagain:
> If it is at all possible to keep nexthome.css from loading except on /nexthome, let's do that.

I had this thought sometime ago, even the style name suggested the style should only apply to /nexthome. I tried to move the linking from extras.tmpl to nexthome.tmpl and other pages were affected so I left things as they were.

I strongly agree with linking nexthome.css to nexthome.tmpl only and then cleaning up nexthome.css to keep only styles that affect /nexthome; but how to fix the styling issues that will unearth in other pages after doing this is what I have little idea of. Before nexthome.css, did not every other page have the styles required to make them look the way they should?

chappjc:
I'm sure every other page had what it needed.  They should again.

chappjc:
[...]

PR #810, opened 2018-11-12
[3.1] backport formatting fixes

chappjc:
This addresses the UI formatting issues in the `3.1-stable` branch that are described in https://github.com/decred/dcrdata/issues/809.

### DIFF
--- a/public/css/nexthome.css
+++ b/public/css/nexthome.css
@@ -492,7 +492,7 @@ body {
     .top-nav .logo a {
         width: 90px !important;
     }
-    .col-md-6 {
+    .nh-col-md-6 {
         flex: 0 0 50%;
         max-width: 100%;
     }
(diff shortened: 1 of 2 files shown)


public/css/nexthome.css
```diff
@@ -492,7 +492,7 @@ body {
     .top-nav .logo a {
         width: 90px !important;
     }
-    .col-md-6 {
+    .nh-col-md-6 {
         flex: 0 0 50%;
         max-width: 100%;
     }
```
chappjc: Is it "bad" that we're still changing `.col-sm-4` and other bootstrap stuff?

dmigwi: Sure. Lesson learnt.

chappjc: OK, I'm not going to change all the others here.  But will ref issue https://github.com/decred/dcrdata/issues/782,  noting that there is still more to fix.

PR #811, opened 2018-11-12
css: trailing zeros should not have negative margin

chappjc:
When the `trailing-zeros` class is used with the `decimal` class, which adds
`margin-left: -2px;`, the zeros overlap with the other `decimal` digits to
the left. This overrides `margin-left` for elements with both classes.

Before:

image

After:

image

### DIFF
--- a/public/css/main.css
+++ b/public/css/main.css
@@ -316,6 +316,9 @@ body.darkBG .progress {
 .trailing-zeroes {
   opacity: .3
 }
+.decimal.trailing-zeroes {
+  margin-left: 0 !important;
+}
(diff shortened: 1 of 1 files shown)


public/css/main.css
```diff
@@ -316,6 +316,9 @@ body.darkBG .progress {
 .trailing-zeroes {
   opacity: .3
 }
+.decimal.trailing-zeroes {
+  margin-left: 0 !important;
+}
```
chappjc: @gozart1 Is this an OK way to avoid getting the -2px left margin from `.decimal`?

PR #815, opened 2018-11-13
js: replace history state when address page goes to list-view

chappjc:
When the address page is loaded with no URL #anchor, the default
anchor $list-view is used in a new URL history state. This caused an
unnecessary step in the brower history and made it nearly impossible
to navigate back to say the /tx page because of the pushState.  Instead
using replaceState resolves the issue.

**Note**: this is FOR DISCUSSION.  I think it is likely we will need use turbolinks
instead of the history API directly.

### DIFF
--- a/public/js/controllers/address.js
+++ b/public/js/controllers/address.js
@@ -142,7 +142,7 @@
                 var hashVal = window.location.hash.replace('#', '') || _this.defaultHash;
 
                 if (hashVal.length === 0 || hashVal === _this.defaultHash){
-                    history.pushState({},  this.addr, "#" + _this.defaultHash);
+                    history.replaceState({},  this.addr, "#" + _this.defaultHash);
                 } else {
                     var selectedVal = this.optionsTarget.namedItem(hashVal)
                     $(this.optionsTarget).val((selectedVal ? selectedVal.value : 'types'))


gozart1:
Seems like a fine fix for now. 

I think we ought bring in a dedicated client side routing lib to more easily and standardly back all state that is not possible to represent via the server rendered html.

So basically, if it's before the "#" in the url it's in the domain of golang + turbolinks.

If it's after the "#", https://visionmedia.github.io/page.js/ with the query string plugin looks good to me instead of using the history api directly. That would give us a convenient api for routing and parsing named parameters in the query string after the #, which should be all we need to comprehensively represent our client side state.

chappjc:
I'm not sure what to do with this because with broken back and forward arrows, this is no good.

gozart1:
Ah I see what you mean, we'll need to hook into the https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onpopstate event to make forward/back arrows work correctly

gozart1:
So basically, decouple the rendering from the routing. 

Whenever a user clicks a button or selects that chart type, that should update the history and then call a generic function (lets call it something like`handleHistoryChange`) that decides to changeView or drawGraph based on the current history state.

Likewise, the onpopstate handler would simply `handleHistoryChange` as well.

chappjc:
I came across `onpopstate` while looking into it... I was afraid of this.  Seems needlessly complicated. Can't we somehow use turbolinks.visit for this too?

chappjc:
A mitigation, not a solution, is to link to #list-view from the address links everywhere.

chappjc:
replaced by https://github.com/decred/dcrdata/pull/845

Issue #725, opened 2018-10-08
Pre-populating charts data runs at wrong times and more than one at a time


chappjc:
a rather confusing part of the way the new sync works is that 20 seconds is arbitrary, and while plenty long on a lot of machines, it can be too short on a meh VM.  It can just go back to the status page and redirect in another 20 seconds, which is super confusing.  It seems to have been stuck pre-populating the charts data for quite some time.  We should look at when this is run, and if we can optimize it.  I also noticed pre-populating can run multiple times when it should not be.

e.g. 

```
2018-10-08 17:16:27.693 [INF] DATD: dcrdata version 3.1.0-pre+322f9ea (Go version go1.11.1)                                                                                      
2018-10-08 17:16:27.693 [INF] DATD: Running in full-functionality mode with PostgreSQL backend enabled.                                                                          
2018-10-08 17:16:27.764 [DBG] RPCC: Attempting to connect to dcrd RPC localhost:9109 as user bdcrd (no TLS)                                                                      
2018-10-08 17:16:27.776 [INF] RPCC: Established connection to RPC server localhost:9109                                                                                          
2018-10-08 17:16:27.781 [INF] DATD: Connected to dcrd (JSON-RPC API v4.0.0) on MainNet                                                                                           
2018-10-08 17:16:27.785 [INF] SKDB: Loading ticket pool DB. This may take a minute...                                                                                            
2018-10-08 17:16:27.916 [INF] SKDB: Loading all ticket pool diffs...                                                                                                             
2018-10-08 17:16:47.052 [DBG] SKDB: len(poolDiffs)=281344                                               
2018-10-08 17:16:47.052 [DBG] SKDB: poolDiffs[0].Height=1, poolDiffs[end].Height=281344
2018-10-08 17:16:47.279 [DBG] SKDB: Opened existing stake db.                 
2018-10-08 17:16:47.279 [INF] SKDB: Advancing ticket pool DB to tip via diffs...
2018-10-08 17:16:47.647 [INF] SKDB: Pre-populating live ticket cache and computing pool value...
2018-10-08 17:17:03.713 [INF] DATD: SQLite DB successfully opened: dcrdata.sqlt.db
2018-10-08 17:17:03.732 [INF] PSQL: Pre-loading unspent ticket info for InsertVote optimization.
2018-10-08 17:17:04.404 [INF] PSQL: Storing data for 43500 unspent tickets in cache.
[...]

PR #816, opened 2018-11-13
pre-populate charts data tweaks

chappjc:
Do not pre-populate charts data in `explorer.New`. Wait for initial `Store`.

Give `explorer.chartDataCounter` a `Height` and `Update` function. `Height` returns -1 when data has never been assigned.
In `prePopulateChartsData`, return without update if cache is at same height as the `explorerUI` (as determined by `.pageData.BlockInfo.Height` via `(*explorerUI).Height`).
In `(*explorerUI).Store`, only run `prePopulateChartsData` AFTER storing the new `BlockInfo` in `pageData`.

Avoid launching two identical queries by holding the `chartDataCounter` cache locked when performing both the height check and update in `prePopulateChartsData`.
Add thread-unsafe `height` and `update` functions for `chartDataCounter` so that the cache can be locked for both operations manually.

This partially resolves https://github.com/decred/dcrdata/issues/725.

Previously:

```
<connect to DBs>
2018-11-13 17:05:16.015 [INF] EXPR: Mean Voting Blocks calculated: 7860
2018-11-13 17:05:16.024 [INF] EXPR: Pre-populating the charts data. This may take a minute...
2018-11-13 17:05:20.375 [INF] EXPR: Done Pre-populating the charts data
<run sync>
2018-11-13 17:05:23.439 [INF] DATD: All ready, at height 291870.
2018-11-13 17:05:23.442 [INF] EXPR: Pre-populating the charts data. This may take a minute...
2018-11-13 17:05:23.494 [DBG] EXPR: Got new block 291870 for the explorer.
2018-11-13 17:05:23.494 [INF] EXPR: Signaling new block to 0 websocket clients.
2018-11-13 17:05:23.496 [DBG] MEMP: mempoolDataCollector.Collect() completed in 1.491537ms
2018-11-13 17:05:23.496 [INF] EXPR: Pre-populating the charts data. This may take a minute...
2018-11-13 17:05:23.502 [DBG] EXPR: storeMempoolInfo() completed in 7.56529ms
2018-11-13 17:05:27.607 [INF] EXPR: Done Pre-populating the charts data
2018-11-13 17:05:27.827 [INF] EXPR: Done Pre-populating the charts data
```

The charts data pre-population was run three times: once before sync, once on initial explorer.Store, and again unconditionally/manually after everything is up and running and final ensureSync is done.

Now:

```
<no pre-populate before sync>
2018-11-13 16:58:28.984 [DBG] EXPR: Got new block 291869 for the explorer.
2018-11-13 16:58:28.984 [INF] EXPR: Pre-populating the charts data. This may take a minute...
2018-11-13 16:58:28.984 [INF] EXPR: Signaling new block to 0 websocket clients.
2018-11-13 16:58:28.986 [DBG] MEMP: mempoolDataCollector.Collect() completed in 1.720893ms
2018-11-13 16:58:28.991 [DBG] EXPR: storeMempoolInfo() completed in 7.001131ms
2018-11-13 16:58:33.255 [INF] EXPR: Done pre-populating the charts data.
2018-11-13 16:58:33.255 [DBG] EXPR: Not updating charts data again for height 291869.
```

There is no more pre-populate in explorer.New (when things wouldn't be syncd yet), and when `prePopulateChartsData` is called at the end of setup, it will return without querying the DB since the previous call to `prePopulateChartsData` in the initial `explore.Store` recorded the height and it is recognized as fresh data.

### DIFF
--- a/explorer/explorer.go
+++ b/explorer/explorer.go
@@ -107,16 +107,28 @@ var cacheChartsData chartDataCounter
 func (c *chartDataCounter) Height() int64 {
 	c.RLock()
 	defer c.RUnlock()
-	if c.Data == nil {
-		return -1
-	}
-	return c.updateHeight
+	return c.height()
 }
 
 // Update sets new data for the given height in the the charts data cache.
 func (c *chartDataCounter) Update(height int64, newData map[string]*dbtypes.ChartsData) {
 	c.Lock()
 	defer c.Unlock()
+	c.update(height, newData)
+}
+
+// height returns the last update height of the charts data cache. Use Height
+// instead for thread-safe access.
+func (c *chartDataCounter) height() int64 {
+	if c.Data == nil {
+		return -1
+	}
+	return c.updateHeight
+}
+
+// update sets new data for the given height in the the charts data cache. Use
+// Update instead for thread-safe access.
+func (c *chartDataCounter) update(height int64, newData map[string]*dbtypes.ChartsData) {
 	c.updateHeight = height
 	c.Data = newData
 }
@@ -403,9 +415,13 @@ func (exp *explorerUI) prePopulateChartsData() {
 		return
 	}
 
+	// Hold charts cache during check and update.
+	cacheChartsData.Lock()
+	defer cacheChartsData.Unlock()
+
 	// Avoid needlessly updating charts data.
 	expHeight := exp.Height()
-	if expHeight == cacheChartsData.Height() {
+	if expHeight == cacheChartsData.height() {
 		log.Debugf("Not updating charts data again for height %d.", expHeight)
 		return
 	}
@@ -428,7 +444,7 @@ func (exp *explorerUI) prePopulateChartsData() {
 		pgData[k] = v
 	}
 
-	cacheChartsData.Update(expHeight, pgData)
+	cacheChartsData.update(expHeight, pgData)
 
 	log.Info("Done pre-populating the charts data.")
 }
--- a/explorer/explorer.go
+++ b/explorer/explorer.go
@@ -96,17 +96,38 @@ type explorerDataSource interface {
 // chartDataCounter is a data cache for the historical charts.
 type chartDataCounter struct {
 	sync.RWMutex
-	Data map[string]*dbtypes.ChartsData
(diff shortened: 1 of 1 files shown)


chappjc:
Note that this means charts will not be available while still syncing.  I think this is the right idea.

explorer/explorer.go
```diff
@@ -513,6 +534,16 @@ func (exp *explorerUI) Store(blockData *blockdata.BlockData, msgBlock *wire.MsgB
 		go exp.updateDevFundBalance()
 	}
 
+	// Update the charts data after every five blocks or if no charts data
+	// exists yet. Do not update the charts data if blockchain sync is running.
```
dmigwi: This comment might require some editing unless when the charts data is empty then `cacheChartsData.Height() == -1`.

dmigwi: BTW, If the above is true then the first time 
`func (exp *explorerUI) Store(blockData *blockdata.BlockData, msgBlock *wire.MsgBlock) error {` 
runs 
`go exp.prePopulateChartsData()` 
will always run because the charts data is empty then.

dmigwi: Also how does the height become -1 unless you are setting it somewhere?

dmigwi: I have seen this function:
``` Golang
func (c *chartDataCounter) height() int64 {
	if c.Data == nil {
		return -1
	}
	return c.updateHeight
}
```
And in my opinion (not yet tested) the previous state (bug) might not yet be fixed.

dmigwi: It actually doesn't run after my test. Let me look at the code again to see what I missed.

chappjc: I just received this a.m.  maybe that broke something, but I think it's ok

chappjc: But it is correct that the first time Store runs it will run prepopulate. We have it explicitly run initially in main, along with another request near the end of main that ends up not running a duplicate query (most of the time).  But the design should consider consumers of the explorer package who may not do things like we do in main.  Also, the query might fail or timeout, so it's prudent to verify it's done.

PR #820, opened 2018-11-14
explorer: fix and test hashStart and hashEnd tmpl functions

chappjc:
The `hashStart` and `hashEnd` template functions must work with any length
input hash. This modifies these functions to work when the default indexing would
read outside the slice.

Add tests for `hashStart` and `hashEnd` that ensure they work with
different hash lengths, and that they work together to split the hash
without losing or repeating any characters.

While here, make `explorer.netName` work with a `nil` pointer input,
returning `"invalid"` as the network name instead of panicking.

### DIFF
--- a/explorer/explorerroutes.go
+++ b/explorer/explorerroutes.go
@@ -41,6 +41,9 @@ const homePageBlocksMaxCount = 30
 
 // netName returns the name used when referring to a decred network.
 func netName(chainParams *chaincfg.Params) string {
+	if chainParams == nil {
+		return "invalid"
+	}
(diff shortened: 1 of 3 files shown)


buck54321:
Yeah, that was a pretty egregious oversight on my part. My apologies. 

This looks good. The tests look right and are passing. The results are as expected on the transactions page.

chappjc:
Nah, it's all good.  I didn't bat an eye when I reviewed it previously, but when you think about a malfunctioning consumer of the function the issue is a lot more clear.  Anyway, I found writing the tests to be a good way to explain to myself (any anyone who reads them in the future) the intended semantics of the two functions, and to ensure they always complement each other properly.

Issue #823, opened 2018-11-16
reorg notification now comes after block connect ntfns


chappjc:
Our current reorganization handlers expect the following sequence of notifications from dcrd when there is a reorganization:
- Receive reorg notification from dcrd that includes the height and hash of the old and new chain tips.  dcrdata enters "reorg mode".
- Receive block connected ntfns, one at a time, for the new chain tip blocks, starting with the block following the common ancestor to both chains.
- When the new chain tip is received and stored, do the DB updates, and the reorg is done.  Reorg mode is turned off.
- Any more blocks from the block connected notifications are assumed to being added to the main chain.

With the merge of dcrd PR https://github.com/decred/dcrd/pull/1500, the notifications come in a different order: **the reorg ntfn comes AFTER the block connected notifications**.  Thus, the way we're doing it above no longer works.  Instead we should:
- For all blocks notified via the block connected ntfn, dcrdata must verify that the block connects on top of the current best block.  This is because block connected notifications come before the reorg notification and they cannot just be connected.
- Option 1: Register for disconnect ntfns and disconnect so that when connect ntfns come the connected blocks will be for the same tip block that dcrd has.
- Option 2: When connected blocks do not add to dcrdata's chain tip, ignore them.  A reorg notification should be coming.  When the reorg ntfn comes, do the disconnects and connects.

We will do Option 2.  However, a little work is needed when reorg mode starts since we won't have the `sideChain` slice containing the hashes of block we are going to reorg to main chain.  This is a minor issue since the code to find the common ancestor of the old and new main chains can gather this data.

Also, care must be taken with the "ignore them" bit of this change because a block connected notification for the tip may come while we are processing the reorg.  It might connect to the new tip, but if the reorg is in progress, it could be ignored. Either notification handling should be paused during reorg and/or the connect notifications should be queued for processing.

chappjc:
Ref: https://github.com/decred/dcrd/pull/1531

chappjc:
Resolved by https://github.com/decred/dcrdata/pull/828

PR #828, opened 2018-11-19
Work with the new dcrd reorg ntfn sequence

chappjc:
Resolving https://github.com/decred/dcrdata/issues/823

- [x] Modify `collectionQueue` to verify that new blocks are in sequence
- [x] Bump required dcrd JSON RPC version to 5.0.0
- [x] Create `rpcutils.CommonAncestor` function to provide full reorg details to each reorg handler and avoid redundant RPCs.
- [x] Rewrite each `ChainMonitor`'s reorg handling.  Each of `dcrsqlite`, `dcrpg`, and `stakedb` use a `switchToSideChain` function.  With `blockdata`, `ReorgHandler` simply stores the new chain tip's block data in the `reorgDataSavers`, which is presently just the explorer UI update.

There is no longer a need for `BlockConnectedHandler` in dcrpg and dcrsqlite, so it is removed.  The goroutines are no longer started of course, and the synchronous `collectionQueue` now includes only `sdbChainMonitor.BlockConnectedSync` and `wsChainMonitor.BlockConnectedSync` (the synchronous collection and storage functions for the stakedb and blockdata packages.).

In short, the roles of each data saver's `ReorgHandler` are:

1. blockdata: Collect and store in the special reorgDataSavers the data for just the new chain tip block.  This is presently used to update the explorer UI.
2. stakedb: Disconnect the old chain blocks from the StakeDatabase, then connect the new chain blocks.
3. dcrsqlite: Collect and store data for the new chain blocks.  The data for the old chain is not presently removed as it is not critical.
4. dcrpg: This is the most complex one.  All data related to old chain blocks are updated to reflect their new side chain status.  The blocks in the new chain (from a side chain) are stored in the DB, while updating existing records, which may be from the disconnected blocks.

The entire process is much simpler overall.  One source of complexity is the new reorg queue in the `notification` package, and the new and expanded `txhelpers.ReorgData` type that adds the entire lists of blocks in the old and new chains as well as the common ancestor.  The reorg queue is processed by `notification.ReorgSignaler`, which uses the new `rpcutils.CommonAncestor` to get the common ancestor and the full list of block in the old and new chains.

`rpcutils.CommonAncestor` finds the common ancestor and full list of block in the old and new chain by iteratively interrogating dcrd via RPC for blocks' previous block hash until the same hash is encountered for both chains at the same height.  This approach is used as opposed to two other ways:
1. Figuring out the common ancestor in each package's ReorgHandler using whatever data resources they have.  This is redundant and more error prone.
2. Going back from the old chain tip until dcrd indicates the chain has gone from side (-1 confirmations) to main (non-negative confirmations).  This is not robust as there may be data races if subsequent reorganization takes place during the handling of an earlier reorg.

### DIFF
--- a/blockdata/chainmonitor.go
+++ b/blockdata/chainmonitor.go
@@ -105,7 +105,8 @@ func (p *chainMonitor) collect(hash *chainhash.Hash) (*wire.MsgBlock, *BlockData
 	// relevant for the best block.
 	var blockData *BlockData
 	if chainHeight != height {
-		log.Infof("Behind on our collection...")
+		log.Debugf("Collecting data for block %v (%d), behind tip %d.",
+			hash, height, chainHeight)
 		blockData, _, err = p.collector.CollectHash(hash)
 		if err != nil {
 			return nil, nil, fmt.Errorf("blockdata.CollectHash(hash) failed: %v", err.Error())
@@ -199,7 +200,8 @@ out:
 			// Do not handle reorg and block connects simultaneously.
 			p.reorgLock.Lock()
 
-			log.Infof("Reorganize signaled to blockdata. Collecting data for NEW head block %v at height %d.",
+			log.Infof("Reorganize signaled to blockdata. "+
+				"Collecting data for NEW head block %v at height %d.",
 				newHash, newHeight)
 
 			// Collect data for the new best block.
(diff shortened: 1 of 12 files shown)


chappjc:
PR description updated with design and implementation details.  Will be ready testing by other team members shortly.  I will update the description when ready.

**Note this PR also bumps the required dcrd JSON RPC version to 5.0.0 (and only 5.0.0).  Previously dcrd JSON RPC versions are incompatible!**

PR #829, opened 2018-11-20
UTXOs on /tx matched against mempool

### LINKED ISSUE #825: transaction page, show Spent w/ mempool txs
Output has a Spent column specifying true or false. It is true if it is spent, but this only shows up accurately for transactions that are confirmed. If the output is spent by a transaction that is in the mempool, it does not properly show as spent or link to the spending transaction. This is particularly visible in the case of ticket purchases which happen as two transactions.

buck54321:
Resolves #825


chappjc:
Confirmed working for a ticket split transaction.

PR #831, opened 2018-11-21
/tx page Mempool inputs are grabbed from explorer

buck54321:
Improvement of #829

### DIFF
--- a/explorer/mempool.go
+++ b/explorer/mempool.go
@@ -101,6 +101,7 @@ func (exp *explorerUI) mempoolMonitor(txChan chan *NewMempoolTx) {
 			VoteInfo:  voteInfo,
 			VinCount:  len(msgTx.TxIn),
 			VoutCount: len(msgTx.TxOut),
+			Vin:       MsgTxMempoolInputs(msgTx),
 			Coinbase:  blockchain.IsCoinBaseTx(msgTx),
 		}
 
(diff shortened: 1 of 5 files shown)


explorer/mempool.go
```diff
@@ -386,3 +387,41 @@ func (votes byHeight) Len() int {
...
+	}
+	return
+}
+
+// GetTxMempoolInputs grabs very simple information about mempool transaction inputs that spend
+// a particular previous transaction's outputs. The returned slice has just enough information to
+// match an unspent transaction output.
+func (exp *explorerUI) GetTxMempoolInputs(txid string, txType string) []MempoolVin {
+	var vins []MempoolVin
+	exp.MempoolData.Lock()
```
chappjc: `exp.MempoolData.RLock()` and `defer exp.MempoolData.RUnlock()` since `exp.MempoolData` is not being modifies.  This will allow multiple readers without blocking.

explorer/mempool.go
```diff
@@ -386,3 +387,41 @@ func (votes byHeight) Len() int {
 func (votes byHeight) Swap(i, j int) {
 	votes[i], votes[j] = votes[j], votes[i]
 }
+
+// matchMempoolVins filters relevant mempool transaction inputs
+// whose previous outpoints match the specified transaction id.
+func matchMempoolVins(txid string, txsList []MempoolTx) (vins []MempoolVin) {
+	for idx := range txsList {
+		tx := txsList[idx]
```
chappjc: `MempoolTx`  is a large structure, and it wouldn't hurt to instead do `tx := &txsList[idx]`.

explorer/mempool.go
```diff
@@ -386,3 +387,41 @@ func (votes byHeight) Len() int {
 func (votes byHeight) Swap(i, j int) {
 	votes[i], votes[j] = votes[j], votes[i]
 }
+
+// matchMempoolVins filters relevant mempool transaction inputs
```
chappjc: For consistency, we always wrap after 80 columns.  I just use a wrapping extension (Rewrap for VS Code).

explorer/mempool.go
```diff
@@ -386,3 +387,41 @@ func (votes byHeight) Len() int {
...
+		vins = append(vins, MempoolVin{
+			TxId:   tx.TxID,
+			Inputs: inputs,
+		})
+	}
+	return
+}
+
+// GetTxMempoolInputs grabs very simple information about mempool transaction inputs that spend
+// a particular previous transaction's outputs. The returned slice has just enough information to
```
chappjc: same with the wrapping here

explorer/mempool.go
chappjc: But if we really wanted a copy, not a pointer, it would be better to do `for _, tx := range txsList {` since `idx` is used nowhere else.

explorer/mempool.go
```diff
@@ -386,3 +387,41 @@ func (votes byHeight) Len() int {
...
 }
+
+// matchMempoolVins filters relevant mempool transaction inputs
+// whose previous outpoints match the specified transaction id.
+func matchMempoolVins(txid string, txsList []MempoolTx) (vins []MempoolVin) {
+	for idx := range txsList {
+		tx := txsList[idx]
+		var inputs []MempoolInput
+		for vindex := range tx.Vin {
+			input := tx.Vin[vindex]
```
chappjc: Same comment about the copy and using `vindex` only here.

explorer/mempool.go
buck54321: For any Atom users that might read this, the equivalent package is `autoflow`, installed by default, and uses `shift+ctrl+q` to wrap the selection at 80 lines (configurable).

Issue #140, opened 2017-09-08
For each tx input, show vout index in source transaction


chappjc:
Perhaps replace the Sequence column, which is not very informative. (@RogueElement)

Instead of just the source transaction hash for each input, showing a prevout (txid:index) is more complete.

TODO (@chappjc): get the address of the pkScript of the referenced outpoint.

chappjc:
In progress (input addresses)

chappjc:
@raedah I think this addresses Issue https://github.com/dcrdata/dcrdata/issues/140.  Comments?

raedah:
Works for me.

PR #832, opened 2018-11-24
sample-nginx.conf updates

chappjc:
Remove "http2" from port 80 server since that does not work.
Add "gzip_vary on" to allow proxies to distinguish content types.
Tweak gzip compression settings, adding more types to match recently
added assets and more.
Add a number of security related headers.
Add "expires modified 2d;"  to the static assets location to set the
Cache-Control and Expires headers.

### DIFF
--- a/sample-nginx.conf
+++ b/sample-nginx.conf
@@ -62,7 +62,7 @@ http {
                 font/otf font/ttf image/bmp image/svg+xml image/x-icon application/octet-stream;
 
         add_header Referrer-Policy "no-referrer-when-downgrade" always;
-        add_header Content-Security-Policy "script-src 'self' data: 'unsafe-inline'; font-src 'self' data:" always;
+        add_header Content-Security-Policy "default-src 'self'; script-src 'self' 'unsafe-inline'; font-src 'self'; object-src 'none'; img-src 'self' data:; style-src 'self' 'unsafe-inline'; connect-src 'self' wss://*.dcrdata.org; manifest-src 'self'" always;
         add_header X-Content-Type-Options nosniff always;
         add_header X-Frame-Options SAMEORIGIN always;
         add_header X-XSS-Protection "1; mode=block" always;


degeri:
Sorry for the wall of text
 So I reviewed this.

### CSP:

1.having 'unsafe-inline' and data: for script will make it not useful its almost as same as not having CSP protection. Can we remove them ? 
2. can we set object-src to none ? 

The best rule would be:
`default-src 'none'; script-src 'self'; font-src 'self'; object-src 'none'; img-src 'self'; style-src 'self' 'unsafe-inline'; connect-src 'self' wss://*.dcrdata.org;manifest-src 'self'`

default-src -> set all to none
script-src -> only allow 'self' 
font-src -> only allow 'self'
object-src -> allow none 
img-src -> only allow 'self'
style-src -> allow 'self' and 'unsafe-inline'  (I gave  'unsafe-inline'  cause I thought it would be impractical to change all CSS as I saw alot of conflicts but if it can be done it will be awesome)
connect-src -> allow 'self' and wss://*.dcrdata.org (actually 'self' should have worked for "wss://*.dcrdata.org" but some browsers still cause issues here)
manifest-src -> only allow 'self'


https://csp-evaluator.withgoogle.com/?csp=https://alpha.dcrdata.org 

Can you spin up alpha.dcrdata.org with these new rules ? . I am sure CSP will cause errors but we can add whitelisting or fix specific codes like we did for politeiagui. I will help you with that. :) 



I already found some scripts code that might cause conflicts. 
"    <script data-turbolinks-eval="false"> " is an inline script. This should be moved to a JS file and not be served in source. 
image
image

Jquery: Jquery is notoriously CSP unfriendly. We might have to whitelist this. 

image

image


Note:

You can use "https://chrome.google.com/webstore/detail/caspr-enforcer/fekcdjkhlbjngkimekikebfegbijjafd" to quickly test CSP rules without messing with server configs. 
 


### Other headers we can add. 
**1. Implement HSTS.** 

That would be 
`add_header Strict-Transport-Security "max-age=31536000; includeSubdomains; preload";

**2. Implement "Referrer Policy" (this would prevent websites from finding the URL the user is coming from)** 

'add_header Referrer-Policy "no-referrer";'

**3. Implement "Feature Policy".**

"https://scotthelme.co.uk/a-new-security-header-feature-policy/" --> I think you might have better understanding if this is a good idea or not.

degeri:
Ok. I was not seeing the HSTS header. Can you point me to it ? 

Yes same-origin for referrer is safe no issues. 

For Feature Policy  `Feature-Policy: sync-xhr 'self' ` should be fine right ? (We can hold off on this until CSP is ready and in production) 


Yes. I think we can start working on CSP. Give me a couple of days I will find and comment here all the inline scripts that might cause issues. You can have a look and tell me which all can be fixed in code and which need whitelisting based on that I will make a new CSP rules :) . Does that sound ok ?

chappjc:
HSTS was just after the ssl settings: https://github.com/decred/dcrdata/pull/832/files#diff-b52dcb90f5406291989290b0780f0121R82

I'll tweak the settings on alpha.dcrdata.org later and we can try them out.

Thanks for reviewing the inline scripts!

chappjc:
@degeri I've updated some headers at alpha.dcrdata.org

Refs:
https://securityheaders.com/?q=https%3A%2F%2Falpha.dcrdata.org%2F&hide=on&followRedirects=on
https://github.com/WICG/feature-policy/blob/master/features.md
https://www.cspisawesome.com/content_security_policies

degeri:
1.
https://github.com/decred/dcrdata/blob/9fc6415620342ccc14b95396f8f298b6e6171ac2/views/extras.tmpl#L90

2. 

Not possible to whitelist as hash keeps changing.
https://github.com/decred/dcrdata/blob/master/public/js/vendor/turbolinks.min.js has inline scripts. 

3. 

https://github.com/decred/dcrdata/blob/master/public/js/vendor/jquery.min.js has inline scripts. 
4.

https://github.com/decred/dcrdata/blob/db8e600af08c210c88d0830001a71c2637247a93/views/charts.tmpl 

#140 and javascript:void(0);

5.
Also noticed some images when charting are loaded as "data:xxx"  which can be whitelisted using (img-src 'self' data:) ==> its broken right now 

So I read-up on turbolinks it seems its a known issue. Looks like a complicated fix :\  

https://github.com/turbolinks/turbolinks/issues/430
https://github.com/rails/rails/issues/31689
https://github.com/turbolinks/turbolinks/issues/330
https://github.com/turbolinks/turbolinks/issues/370

chappjc:
I knew this would bite us:

```html
    <script type="text/javascript">
        var ticketsPrice = {{.Data}};
    </script>
```

Using Go's html template engine to set JSON data in inline javascript.  We need to remove that anyway possible.

degeri:
ya , CSP is a bit of pain to implement properly ...  but it whips current code into shape and keeps it that way

chappjc:
I've added `data:` to the `img-src` rule on alpha.dcrdata.org, and pushed an update to the PR.

chappjc:
The issue https://github.com/decred/dcrdata/issues/843 tracks the inline script and data issues.

chappjc:
So, we can't solve the turbolinks CSP issue with nonces since the response header doesn't change on turbolinks navigation.  That leaves hash-source.  Both nonce and hash source are widely supported: https://caniuse.com/#feat=contentsecuritypolicy2

Issue #824, opened 2018-11-17
transaction page could show count of inputs and outputs


raedah:
Inputs (6)   Outputs (8)
This will be useful on pages with many inputs or outputs listed.

chappjc:
Resolved by https://github.com/decred/dcrdata/pull/833

PR #833, opened 2018-11-24
views/tx: show count of inputs and outputs

McEdward:
screenshot 126

### DIFF
--- a/explorer/explorertypes.go
+++ b/explorer/explorertypes.go
@@ -156,6 +156,8 @@ type TxInfo struct {
 	Type             string
 	Vin              []Vin
 	Vout             []Vout
+	VinLenght        int
+	VoutLenght       int
 	BlockHeight      int64
 	BlockIndex       uint32
 	BlockHash        string
--- a/explorer/explorertypes.go
+++ b/explorer/explorertypes.go
@@ -156,8 +156,6 @@ type TxInfo struct {
 	Type             string
 	Vin              []Vin
 	Vout             []Vout
-	VinLenght        int
-	VoutLenght       int
 	BlockHeight      int64
 	BlockIndex       uint32
 	BlockHash        string
--- a/explorer/explorerroutes.go
+++ b/explorer/explorerroutes.go
@@ -1100,6 +1100,8 @@ func (exp *explorerUI) TxPage(w http.ResponseWriter, r *http.Request) {
 			vin.Link = "/tx/" + vin.Txid + "/out/" + voutStr
 		}
 	}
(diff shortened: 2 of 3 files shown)


McEdward:
resolving issue #824

raedah:
Count column doesn't add any value. Sum at the top is all thats needed.

chappjc:
Let's consider this as a proposal until more people have had the opportunity to register their opinion on the numbering of each i/o row (as shown).  As I mentioned in issue #241 , the best way to convey this row-wise information is not clear.  This PR is one way that does eat a lot of horizontal space, and there may be tighter ways of conveying this info, even if that means it is not displayed except by mouse hover, etc.

buck54321:
Either but not both seems fine. Also, it's technically not the row count, but the vin/vout index, so it does have meaning beyond simply enumerating the rows.

chappjc:
Agreed, it's good to have.  When an output links to tx input where the output is spent, that index needs to be visible somewhere on the spending transaction's view as this PR does.
We can move forward with this and see what ideas pop out of the woodwork.

explorer/explorertypes.go
```diff
@@ -129,6 +129,8 @@ type TxInfo struct {
 	Type             string
 	Vin              []Vin
 	Vout             []Vout
+	VinLenght        int
```
buck54321: Lenght -> Length

chappjc:
There is also the possibility of giving the user the ability to select which columns to display.
It could be bulky and horrible, but if done in a modern and unobtrusive way, it could be pretty slick.

raedah:
I originally just thought it was row numbers that happened to start at zero, and that seemed a bit useless. But once I got that it was the vin/vout index I saw that it had meaning. Could put a help hover over on the column header which says 'vin/vout index'. 

Also seeing that instead of 
Input (1)    Outputs (3)
could just be
1 Input    3 Outputs

gozart1:
We could add verbs to help make the mechanics of a transaction a bit more explicit.

1 Input Consumed
2 Outputs Created

Inspired by Block Cypher, eg
https://live.blockcypher.com/btc/tx/0bde300106882ff2b98081419aa7465a64c00fbc4c2d573bf6593da8dc4cb489

chappjc:
@gozart1 I like that consumed/created language.

To be explicit about the value of the in/out index numbering, when discussing a transaction with more than a handful of inputs/output, it is hard to say which input or output you want to direct someone's attention to without having the index shown.

I also like the idea of putting hover text (`title`) for the column headers.

@McEdward Would you mind demoing the above changes?

Issue #784, opened 2018-10-31
Add Project Fund to the navigation menu


raedah:
It is a significant enough part of the project to make it easier to find. We can make this link, and the link to it from the home page, go the the merged debits view which makes the most sense for this particular data.

chappjc:
Yeah, agreed, merged debits is best for the link

xaur:
Treasury? It's getting more commonly used.

chappjc:
As long as it's Project and not Dev, I'm OK with it.  By my preference is Fund.

s-ben:
For reference, there was some [discussion](https://github.com/decred/dcrdocs/issues/582) of what to call it about a month ago on dcrdocs. There seemed to be a loose consensus around 'Decred Treasury', so I opened a PR that changes all references to 'Decred Treasury', hoping it would spark discussion (no comments yet). The PR description provides a summary of the discussion around this so far, at least what I've been able to gather. 
https://github.com/decred/dcrdocs/pull/690

s-ben:
FYI - dcrdocs went with `Decred Treasury`. Merged PR,
https://github.com/decred/dcrdocs/pull/690#event-1964084315

chappjc:
Ok. Treasury has a lot of baggage and connotations of centralized management, but ok.

xaur:
There was no decentralized management when this baggage formed. But it's a nice well known term to use. I think the effort of explaining how Decred Treasury is different is lower than introducing and explaining a new term.

s-ben:
It may be baggage, but it looks like gold plated, limited edition Gucci baggage.

PR #834, opened 2018-11-24
adding project fund to navigation menu as treasury same name change on the home page

McEdward:
screenshot 125

### DIFF
--- a/views/extras.tmpl
+++ b/views/extras.tmpl
@@ -77,7 +77,7 @@
                         <a data-keynav-skip href="/charts" title="Decred charts">Charts</a>
                         <a data-keynav-skip href="/agendas" title="Agendas">Agendas</a>
                         <a data-keynav-skip href="/parameters" title="Chain Parameters">Parameters</a>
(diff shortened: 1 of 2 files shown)


McEdward:
resolving issue #784

raedah:
We can make this link, and the link to it from the home page, go the the merged debits view.

explorer/explorerroutes.go
```diff
@@ -1856,6 +1856,7 @@ func (exp *explorerUI) commonData() *CommonPageData {
 	cd.Version = exp.Version
 	cd.ChainParams = exp.ChainParams
 	cd.BlockTimeUnix = int64(exp.ChainParams.TargetTimePerBlock.Seconds())
+	cd.DevAddress = exp.pageData.HomeInfo.DevAddress
```
chappjc: OK, but note that `exp.pageData` is generally to be protected by a `RWMutex`.  Since `HomeInfo` is a pointer that never changes, and `HomeInfo.DevAddress` never changes either, this is OK.  I might see about changing commonData() and explorerUI so that explorerUI has a CaommonPageData field that gets filled out on explorerUI construction, and only Tip is updated by this function.

PR #837, opened 2018-11-25
Webpack/ES6 follow up

### LINKED ISSUE #826: Block connected signal causes javascript error on /ticketpool
This is an issue in v3.1.0-rc1. When the `newblock` websocket signal is received by the ticketpool controller, an error occurs looking for the `ticket_pool_data.Mempool` property.

buck54321:
Just a couple of previously discussed issues to follow up #805. 

Resolves #826.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
     "eslint-plugin-promise": "^4.0.1",
     "eslint-plugin-standard": "^4.0.0",
     "mousetrap": "^1.6.2",
+    "mousetrap-pause": "^1.0.0",
     "notifyjs": "^3.0.0",
     "stimulus": "^1.1.0",
     "webpack": "^4.25.1",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -33,6 +33,7 @@
     "eslint-plugin-promise": "^4.0.1",
     "eslint-plugin-standard": "^4.0.0",
     "mousetrap": "^1.6.2",
+    "mousetrap-pause": "^1.0.0",
```
gozart1: Looks like this package is unused. Took a look at the source of that npm package and it is fancier than what we need. 

This version https://github.com/ccampbell/mousetrap/blob/master/plugins/pause/mousetrap-pause.js seems like the simplest approach, so we could just use that one by adding it to js/vendor instead of using npm.

chappjc: Tangentially, is this mousetrap-pause change related to the websocket json formatting fix?

PR #839, opened 2018-11-26
organize css into scss partials, add standard config stylelint, add babel-loader

gozart1:
* Convert all css (except for some libs) into scss into partials 
* Adds [stylelint](https://stylelint.io/) to lint scss using the standard configuration (160 built-in rules to catch errors, apply limits and enforce stylistic conventions)
* Moves index.js to the root of /public
* Moves dist folder to root of /public ( all production assets end up here, css, js ,etc ) 
* Imports only the used subset of bootstrap scss partials instead loading the entire library by itself
* Minifies css bundle for production
* Adds babel transpilation step

I had to re-organize nexthome.css around a bit to get it pass the linter. I've tested to make sure everything looks the same. One of the rules is that stylelint enforces is that rules must be defined in increasing order of specificity, and since nexthome has a lot of selectors that end in an element, this rule was violated quite a bit. It would be good to refactor nexthome so that it does not use long selector chains or base elements and instead keeps things flat by using specific classes. But that's out of scope for this PR.

Adding Babel to transpile ES2018 to ES5 gives us broad browser compatibility. It's being added as part of this PR because switching to bootstrap imports necessitated adding a custom minimizer for webpacks prod config, which in turn required a custom js minimizer (UglifyJS), which in turn chokes on `async await` functions. So adding babel was the missing piece and since I was planning to add this in my next PR it's included here for the sake of simplicity.


chappjc:
Looks like the Treasury link change caused a small conflict in extras.tmpl.

chappjc:
We're all fighting over extras.tmpl today it seems.  We'll need to add `rel="noopener noreferrer"`.  No worries though since I can do it with the github interface if you can't get to it.

gozart1:
Since mini-css-extract-plugin is listed as a dependency in package.json

Just doing `npm install` should work and that is already documented, but maybe it should be made clear that it needs to be run again when dependencies update.

chappjc:
So how do we use stylelint, eslinter, etc.?  We should enable them with Travis.

gozart1:
Good question, I've seen it done before and it's likely just a config.
But I'd need to do some research to see how it works with Travis.

Another option would be to use pre commit hooks and https://github.com/okonet/lint-staged looks great for doing it that way. 

Both approaches seem good to me, the 1st is more iron clad, the 2nd has a tighter feedback loop for committer.

chappjc:
I think hooking it up in Travis is fine since we go through PRs which allow iterations without much trouble.

But I'm actually wondering, how does one just run them manually?

chappjc:
Thanks for holding my hand on that.  Looks like you'd also have to install `stylelint-config-standard` globally too I think.  And use stylelint >= 0.9.7 or edit `"extends"` to point to the full path of the config.

So for CI, we can either run stylelint and check for a non-zero exit code, or just run `npm install` and `npm run build` and check those for non-zero error codes.  We'll want to do the latter anyway, so that might be the way to go, except that the error from `npm run build` is more cluttered than just the error from stylelint.  I'll toy with run_tests.sh, which is what Travis runs.

chappjc:
The following PR adds the npm commands to the CI script: https://github.com/decred/dcrdata/pull/850
I verified that linter complaints cause non-zero edit codes from npm run build, so this should be sufficient.  Note that the docker build image must be updated.  That's pending here: https://github.com/decred/dcrdocker/pull/43

gozart1:
Ah yea, I had mine installed globally. Don't think we need to worry about it here though since webpack handles running it. 

Having Travis simply run the build sounds like the right approach, I'll go ahead and move eslint to webpack.common.js instead of dev so that it always runs.

gozart1:
Have moved bootstrap to importing only the scss partials we actually use, which lead from one thing to another ending up with babel transpiler getting added as well.

This adds a bunch of polyfills to support futuristic js in older browsers, which does increase the JS payload a bit (80kb parsed size, 26kb gzipped)
polyfils
But has the benefit of letting us take advantage of all the latest features in Javascript without breaking the app for older browsers. Subsetting bootstrap and minifying the entire css bundle gives us a similar amount in css download savings so total page weight is not affected much. And since the reduced css complexity is the main factor in how fast we get a styled page to render, its a net win!

gozart1:
I could be wrong, but based on the examples I see out there and the docs, having index.js at the root makes sense because it acts as a manifest file of the entire front end application, and does things like:
```require('./scss/application.scss')```
So it's scope goes a bit beyond just the javascript.
I'm hoping we can refactor the current code in index.js (mainly move stuff out into dedicated modules) so all that's left is a high level initializer of the core application.

To keep things simple for the current ngnix gzipping I went ahead and moved the js and css bundles to `dist/js` and `dist/css`.

For additional web perf brownie points we could make a follow up PR to move asset compression out of the runtime and into the build step via [CompressionWebpackPlugin](https://webpack.js.org/plugins/compression-webpack-plugin/) and nginx [gzip_static](https://docs.nginx.com/nginx/admin-guide/web-server/compression/#sending-compressed-files).

**Comment by gozart1** [comment]
Thanks for pushing me on the perf opportunities.
I've now enabled code splitting so the app is loaded in 2 main chunks (app and vendor).
Furthermore, tippy is now lazy loaded via a dynamic import.
```
                   Asset      Size  Chunks             Chunk Names
           css/style.css   155 KiB       0  [emitted]  app
        js/app.bundle.js  55.6 KiB       0  [emitted]  app
      js/tippy.bundle.js  51.4 KiB       1  [emitted]  tippy
js/vendors~app.bundle.js  61.7 KiB       2  [emitted]  vendors~app
Entrypoint app [big] = js/vendors~app.bundle.js css/style.css js/app.bundle.js
...
WARNING in entrypoint size limit: The following entrypoint(s) combined asset size exceeds the recommended limit (244 KiB). This can impact web performance.
Entrypoints:
  app (273 KiB)
      js/vendors~app.bundle.js
      css/style.css
      js/app.bundle.js
```
It's still complaining 273kb is too fat, but it's not factoring in gzip, which helps a lot.
Gzipped JS bundles:
```
js/vendors~app.bundle.js (16.15 KB)
js/app.bundle.js (16.03 KB)
js/tippy.bundle.js (13.99 KB)
```
 So I think we are looking pretty dang good at this point!

bundles

I also discovered that we should be using babel-eslint for our eslint parser so that experimental features like dynamic imports are not flagged as errors. With that knowledge, I decided to reverse track and remove babel-polyfill added in the earlier commit because I actually initially added that to get past linter errors for the newish js syntax, and using the `babel-eslint` parser solves that a much better way.

If we ever get hordes of users in IE8 complaining the js doesn't work, we can always take the hit and add those 80kb of polyfills in for them. \s

chappjc:
Nice.  I'm happy to nag.

This is looking really tight.  RIP polyfill.

For everyone's reference, when in production, we don't serve the assets via dcrdata at all, instead using nginx with special cache and gzip rules for static assets.  Although I do like the idea of pre-gzipping and using gzip_static.  That will be particularly helpful when traffic really picks up.

chappjc:
The browser console is showing this error:

app.bundle.js:1 Uncaught ReferenceError: regeneratorRuntime is not defined
    at app.bundle.js:1
    at Module.<anonymous> (app.bundle.js:1)
    at d (app.bundle.js:1)
    at a (app.bundle.js:1)
    at vendors~app.bundle.js:1
    at vendors~app.bundle.js:1
    at Array.map (<anonymous>)
    at r (vendors~app.bundle.js:1)
    at Module.<anonymous> (app.bundle.js:1)
    at d (app.bundle.js:1)

gozart1:
Ah dang, that looks like it's due to removing the polyfill. Let me see if there's another way to fix this without adding the polyfill back.

gozart1:
After exploring various options, I've settled on a middle ground by importing only the polyfill for async functions instead of loading the whole babel-polyfill suite. This ends up adding just 2.32kb gzipped.

PR #785, opened 2018-11-01
Header footer fixup

gozart1:
This fixes the logo not showing.
It also removes the header and logo related css from nexthome.css.
So the heights of header and footer are back to their previous dimensions. 
The footer markup has been reworded a bit so that it reflows better.
The logo now collapses to just the symbol for narrow screens.
screen shot 2018-11-01 at 4 16 43 am
Also, there's now a white version of logo for night mode and sim/test net modes. 
screen shot 2018-11-01 at 4 02 53 am
screen shot 2018-11-01 at 4 02 42 am
screen shot 2018-11-01 at 4 05 04 am

There's definitely room for refinement here. But step one was just making it look not broken and I think this achieves that.

PR #840, opened 2018-11-27
Fixes reversetabnabbing (2)

degeri:
Looks like #785 broke the fix for the "JSON API Docs" link.


gozart1:
Good catch, looks like I didn't catch that link updating when rebasing that PR. Fixed now.

PR #844, opened 2018-11-27
start linting again with golangci-lint instead of gometalinter

chappjc:


api/insight/socket.io.go
```diff
@@ -124,7 +124,7 @@ func NewSocketServer(newTxChan chan *NewTx, params *chaincfg.Params) (*SocketSer
 		})
 	})
 
-	server.On("error", func(so socketio.Socket, err error) {
+	server.On("error", func(_ socketio.Socket, err error) {
```
chappjc: was unused parameter

api/types/apicache.go
```diff
@@ -476,7 +476,7 @@ func (pq *BlockPriorityQueue) Insert(summary *BlockDataBasic) (bool, *chainhash.
 	}
 
 	// At capacity
-	for int(pq.capacity) <= pq.Len() {
+	if int(pq.capacity) <= pq.Len() {
```
chappjc: Was meant to be if.  The return statements guarantee there cannot be multiple iterations.

db/dcrpg/indexing.go
```diff
@@ -357,7 +357,6 @@ func (pgb *ChainDB) DeindexAll() error {
 	for _, val := range allDeIndexes {
 		if err = val.DeIndexFunc(pgb.db); err != nil {
 			warnUnlessNotExists(err)
-			err = nil
```
chappjc: ineffectual assignment

db/dcrpg/queries.go
```diff
@@ -2597,7 +2597,7 @@ func retrieveChainWork(db *sql.DB) (*dbtypes.ChartsData, *dbtypes.ChartsData, er
 				log.Errorf("Failed to make uint64 from rate")
 				break
 			}
-			tDef := dbtypes.TimeDef{thisPt.time}
+			tDef := dbtypes.TimeDef{T: thisPt.time}
```
chappjc: Assignment with TimeDef uses keyed fields everywhere else.

db/dcrpg/upgrades.go
```diff
@@ -673,7 +675,9 @@ func (pgb *ChainDB) handleAgendasVotingMilestonesUpgrade() (int64, error) {
 			if val > 0 {
 				query := fmt.Sprintf("UPDATE agendas SET %v=true WHERE block_height=$1 AND agenda_id=$2", name)
 				_, err := pgb.db.Exec(query, val, id)
-				errorLog(name, id, err)
+				if errorLog(name, id, err) {
+					return rowsUpdated, err
```
chappjc: A non-nil error value returning from from `handleAgendasVotingMilestonesUpgrade` was impossible.  If an update statement fails, the error was supposed to stop this and trickle up.

db/dcrsqlite/sqlite.go
```diff
@@ -388,8 +388,8 @@ func (db *DB) StoreBlock(bd *apitypes.BlockDataBasic, isMainchain bool, isValid
 			db.dbSummaryHeight = height
 		}
 		if db.lastStoredBlock == nil || bd.Height >= db.lastStoredBlock.Height {
-			blockdata := apitypes.BlockDataBasic(*bd)
```
chappjc: superfluous conversion since `*bd` was already a `apitypes.BlockDataBasic`.  While here, change to new and copy to do the same thing.

explorer/explorerroutes.go
```diff
@@ -892,9 +892,9 @@ func (exp *explorerUI) TxPage(w http.ResponseWriter, r *http.Request) {
 	// Check for any transaction outputs that appear unspent.
 	unspents := UnspentOutputIndices(tx.Vout)
 	if len(unspents) > 0 {
-		// Grab the mempool transation inputs that match this transaction.
+		// Grab the mempool transaction inputs that match this transaction.
 		mempoolVins := exp.GetTxMempoolInputs(hash, tx.Type)
-		if mempoolVins != nil && len(mempoolVins) > 0 {
+		if len(mempoolVins) > 0 {
```
chappjc: `len` of a `nil` slice is 0.  The nil check was not necessary.

Issue #835, opened 2018-11-25
AddressMetrics runs too slowly


buck54321:
Very long load times (~10s) are observed on the `/address` page for the project treasury, and presumably any address with a lot of transactions. The majority of load time is spent on `dcrpg.AddressMetrics` in four queries of the following form

```
SELECT COUNT(DISTINCT date_trunc('year', block_time)) FROM addresses WHERE address=$1;
SELECT COUNT(DISTINCT date_trunc('month', block_time)) FROM addresses WHERE address=$1;
SELECT COUNT(DISTINCT date_trunc('week', block_time)) FROM addresses WHERE address=$1;
SELECT COUNT(DISTINCT date_trunc('day', block_time)) FROM addresses WHERE address=$1;
```

The values  collected here are used to determine whether the grouping selection button will be displayed on `/address`. See [here](https://github.com/decred/dcrdata/blob/9e23b9a8e6b5f93dacfe1de1822b29c2c81bde8a/public/js/controllers/address_controller.js#L334). 

It appears that the counts retrieved are used as essentially boolean values. I believe the complexity of the queries could be reduced substantially by simply looking at the first and last transaction times for a given address and disabling buttons based on the difference, e.g. if the difference is > 1 week, show the week button.

@dmigwi #798

dmigwi:
> I believe the complexity of the queries could be reduced substantially by simply looking at the first and last transaction times for a given address and disabling buttons based on the difference

This will not always work especially for addresses with very few transactions where the first transaction was made like years ago.
The above implementation also fixes the bug that you've highlighted there and I don't using approximations will even work better than the actual value, you will have introduced other bugs.

1. One of the solution would be to block the queries from running on consecutive times if the address remains constant over a certain period of time.
2. You could also create a cache with the data for the above queries for all the common addresses normally used.
3. And finally graphs db could be faster than pg but I don't think this is worth the hustle of moving dcrdata from psql to an appropriate graph db.

chappjc:
As @buck54321 points out, the JS only checks if the count<2, so it seems the result would be the same if you checked the oldest and newest times, and decided if they are in the same year/month/week/day.

buck54321:
[...]

Issue #814, opened 2018-11-13
impossible to go back when visiting address page


chappjc:
When clicking an address from a transaction view, you are sent to the address page *with no hash tag/#anchor*, but the address page immediately redirects to #list-view, making a new browser history entry.

That is,

Going to http://127.0.0.1:7777/tx/38a9c89aaa3445333c2ca1542c16d21ca33c946715a8e605e91f687df06e38e7 and clicking the first output's address:

http://127.0.0.1:7777/address/DsRZ8itu7XEkBNELzcJ4xKLapswhquKquUn

then takes you to:

http://127.0.0.1:7777/address/DsRZ8itu7XEkBNELzcJ4xKLapswhquKquUn#list-view

Then if you click back, you cannot get to the /tx page because it takes you to the /address page, which immediately redirects to the #list-view address page.

The issue seems to be with:

```js
history.pushState({},  this.addr, "#" + _this.defaultHash);
```

Perhaps we should be using `replaceState` instead: https://developer.mozilla.org/en-US/docs/Web/API/History_API#The_replaceState()_method

chappjc:
Another related issue is that using the browser's <- and -> buttons to go through the different address pages #anchors is not possible.

chappjc:
Partially resolved by PR https://github.com/decred/dcrdata/pull/815, but the general navigation issues mentioned [above](https://github.com/decred/dcrdata/issues/814#issuecomment-438143201) are not resolved.   Perhaps we should not be using the history API and instead use turbolinks?

ZeroASIC:
On dcrdata.org I never get forwarded to #list-view.  I tested it in Chrome and Firefox.  Is this a problem with the development version but not the live version?
https://explorer.dcrdata.org/tx/38a9c89aaa3445333c2ca1542c16d21ca33c946715a8e605e91f687df06e38e7

chappjc:
You'll need to try it on the dev version:
https://alpha.dcrdata.org/address/DsRZ8itu7XEkBNELzcJ4xKLapswhquKquUn
However, we are working out issues with the forward/back browser nav with
this, so it's not working properly anywhere yet.

ZeroASIC:
Yeah I noticed that on alpha.  I looked at a little bit yesterday and I think I could probably fix this.  Using the URL to contain information this way is the real culprit.  It should be based on the state of the html elements.

chappjc:
[...]

Issue #843, opened 2018-11-27
Inline JS in template files


buck54321:
The following template files have inline JS that should be remedied. 

- [x] **[address.tmpl](https://github.com/decred/dcrdata/blob/4b2a5db98ef2c5e0178d7824b0361894280d22bd/views/address.tmpl#L370)** -- Change QR code links to stimulus click handlers. Also page size selector `onchange` event. (#845)
- [x] **[agenda.tmpl](https://github.com/decred/dcrdata/blob/4b2a5db98ef2c5e0178d7824b0361894280d22bd/views/agenda.tmpl#L109)** -- Inline data move to AJAX (#847)
- [x] **[charts.tmpl](https://github.com/decred/dcrdata/blob/4b2a5db98ef2c5e0178d7824b0361894280d22bd/views/charts.tmpl#L141)** -- Inline data. Move to AJAX (#847)
- [x] **[explorer.tmpl](https://github.com/decred/dcrdata/blob/4b2a5db98ef2c5e0178d7824b0361894280d22bd/views/explorer.tmpl#L129)** -- `onchange` event handler to stimulus (#847)
- [x] **[extras.tmpl](https://github.com/decred/dcrdata/blob/4b2a5db98ef2c5e0178d7824b0361894280d22bd/views/extras.tmpl#L90)**  -- Inline data. Move to AJAX or websocket. (#873)
- [x] **[statistics.tmpl](https://github.com/decred/dcrdata/blob/4b2a5db98ef2c5e0178d7824b0361894280d22bd/views/statistics.tmpl#L225)** -- Move tooltip stuff to common file. `Maybe index.js`? (#873)
- [x] **[ticketpool.tmpl](https://github.com/decred/dcrdata/blob/4b2a5db98ef2c5e0178d7824b0361894280d22bd/views/ticketpool.tmpl#L54)** -- Inline data. Move to AJAX. (#851)
- [x] **[timelisting.tmpl](https://github.com/decred/dcrdata/blob/4b2a5db98ef2c5e0178d7824b0361894280d22bd/views/timelisting.tmpl#L132)** -- `onchange` event handler to stimulus (#847)
- [x] **[windows.tmpl](https://github.com/decred/dcrdata/blob/4b2a5db98ef2c5e0178d7824b0361894280d22bd/views/windows.tmpl#L132)** -- `onchange` event handler to stimulus (#873)

This PR encompasses #841.

buck54321:
[...]

Issue #643, opened 2018-08-28
address page, maintain page size setting


raedah:
When changing the view Type, the page size is switched back to its default. The expected behavior is for the page size selected to remain the same.

chappjc:
Seems to be addressed in https://github.com/decred/dcrdata/pull/633.  I started reviewing @dmigwi's PR today and thought I saw this fixed.

buck54321:
Resolved by #845

PR #845, opened 2018-11-28
Turbolinks-integrated query parameters and stimulus integration on address page

buck54321:
Addresses #814,  #815. 

The primary issue is that `history.replaceState` is not compatible with Turbolinks' hijacked browser navigation. This started as an attempt to simply replace the history in a way that Turbolinks could understand, but after seeing some deficiencies in the way the anchor text was being handled, it has expanded beyond that a little bit. I've tried to make the solution portable for inclusion in other pages as well.

What this solves

1. Forward and backward buttons will take you to and from a page as expected
2. More customization of url anchor strings allow more precise chart parameters. You can have a link to exactly the chart you want now, including a zoom level and bin size.

What's not working

1. Right now, the forward and back buttons take you to the page, but the view settings revert to the original settings (view, zoom, bin) from when you first navigated to the page (by link or omnibar), rather than the last state you left it in. I think this is related to Turbolinks caching, but haven't dug into it yet.
2. Probably other stuff. Needs some testing.

@gozart and @ZeroASIC. If you could find time to fiddle with this a little, I would appreciate it.


public/js/controllers/address_controller.js
```diff
@@ -80,36 +82,295 @@ function plotGraph (processedData, otherOptions) {
...
+
+  _appendHistory () {
+    // same as replaceHref, but creates a new entry in history for navigating
+    // with the browsers forward and back buttons
+    this.turbolinks.controller.pushHistoryWithLocationAndRestorationIdentifier(this.turbolinks.Location.wrap(this.url.href), this.turbolinks.uuid())
+    this.appendTimer = 0
+  }
+
+  replace (anchors) {
+    // anchors should be a list of strings
```
buck54321: Ignore this. Can also be an object.

ZeroASIC:
I'm still waiting for my dcrdata VM to sync.  Once it's done I'll take a closer look at these issues.

buck54321:
Yep. Removing it will be fine, though the url fragment will no longer change when you change the view or chart type. 

If you want to change the fragment, use can replace `pushState` with 

```
Turbolinks.controller.replaceHistoryWithLocationAndRestorationIdentifier(Turbolinks.Location.wrap(`#${chartName}`), Turbolinks.uuid())
```

ZeroASIC:
Yeah the basic navigation in the browser should always work.  I'm still waiting for my dcrdata VM to finish syncing.  Once it's done I'll take a closer look at these issues.

chappjc:
Thanks, @ZeroASIC.  Be sure to get postgres tuned up to save tons of time.  There's some tips in a conf file in db/dcrpg.

ZeroASIC:
Okay, looking this over you did get the navigation working correctly, but you've duplicated functionality that already exists in Javascripts URLSearchParams so you can use anchors.  Anchors weren't really meant to be used this way anyhow so I think it would be best if the anchors were removed from this page completely and replaced with URL Parameters.
https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/URLSearchParams

Also think about storing settings in the cookie and the url so they can be retrieved on page load if they're not in the URL already or the cookie can be rebuilt off of the URL.

buck54321:
I agree that things would be easier if we use the query parameters. Please note that we do have two different things happening though. Right now, the query parameters are used for list pagination, which causes a Turbolinks visit. Changing chart parameters does not cause a Turbolinks visit, so maybe that's why the chart parameter was put in the anchor to begin with. 

I'd very much prefer to use query parameters, though, so if nobody objects, I'll work in that direction.

ZeroASIC:
I get ya.  So `#types` stayes, but everything else goes into a query parameter and Turbolinks keeps working too.  Sounds good to me.

chappjc:
That plan sounds good @buck54321.  Please keep 3.1 in mind though since it has to backport smoothly.

buck54321:
Switched everything to query parameters and moved the inline JS to stimulus.

@chappjc This is a rather large change that would probably be difficult to cherry pick to 3.1-stable. Can I do a separate limited PR on that branch.

public/js/controllers/address_controller.js
```diff
@@ -80,36 +83,189 @@ function plotGraph (processedData, otherOptions) {
...
+  to (query) {
+    this.url.set('query', this.filteredQuery(query))
+    this.toHref()
+  }
+
+  filteredQuery (query) {
+    var filtered = {}
+    Object.keys(query).forEach(function (key) {
+      var v = query[key]
+      if (typeof v === 'undefined' || v === null) return
```
chappjc: I'm a bit out of my element here, but this struck me as a place for `continue` instead of `return`.  I assume this is processing all of the query key-values in a url.  And this is catching queries with no value set?

public/js/controllers/address_controller.js
```diff
@@ -80,36 +83,189 @@ function plotGraph (processedData, otherOptions) {
...
+      case 'null':
+        return null
+      case '':
+        return null
+      case 'undefined':
+        return null
+      case 'false':
+        return false
+      case 'true':
+        return true
```
chappjc: What happens with `?q=1`? Seems like that should be a valid way to say true as well.

buck54321: That will parse as the integer `1`, which it "truthy". Similarly `?q=0` would parse to the integer `0`, which is "falsy".

public/js/controllers/address_controller.js
buck54321: You actually give `forEach` a function to which each element is passed. Here, I am passing an anonymous function that is defined on the fly, so there is no loop to `continue` from. 

Even though I know that, I almost always put `continue` and encounter the error, so you're in good company.

chappjc: Ah, I see now.  Thanks.

chappjc:
When looking at this for 3.1, we should consider the simplest possible fix.  If that means not removing the pushState entirely and not introducing TurboQuery, then we should do that.  However, if there are small tweaks that will at least make the navigation arrows not broken in 3.1, then we should make those changes.

public/js/controllers/address_controller.js
```diff
@@ -80,36 +83,189 @@ function plotGraph (processedData, otherOptions) {
...
   static get targets () {
     return ['options', 'addr', 'btns', 'unspent',
-      'flow', 'zoom', 'interval', 'numUnconfirmed', 'formattedTime', 'txnCount']
+      'flow', 'zoom', 'interval', 'numUnconfirmed', 'formattedTime',
+      'pagesize', 'txntype', 'txnCount', 'qron', 'qroff']
   }
 
   initialize () {
-    var _this = this
+    var controller = this
```
gozart1: Instead of setting `this` onto a variable to pass it into new scope, we can just use fat arrow functions. ( since their function bodies have lexical scoping of `this`, see  https://toddmotto.com/es6-arrow-functions-syntaxes-and-lexical-scoping/#functionality-lexical-scoping-this ) 

So by using fat arrow function in the BLOCK_RECIEVED handler below, `this` would be the controller and the alias variable not necessary.

buck54321: I mostly do the `controller` thing for semantics (which is why `_this` just wasn't working for me), but arrow functions are definitely appropriate in many places. 

Just so I know, now that you have the ability, were we planning on using [a babel plugin to translate arrow functions](https://www.npmjs.com/package/@babel/plugin-transform-arrow-functions)?

gozart1: Production builds already transpile `=>` to `function` via https://babeljs.io/docs/en/babel-preset-env being set in our webpack prod configuration.
https://github.com/decred/dcrdata/blob/master/webpack.prod.js#L27

This preset essentially says to transpile everything to ES5 that can be transpiled without polyfills.

Some things can't be transpiled, like `async`/`await`, so we need to import a polyfill for those https://github.com/decred/dcrdata/blob/master/public/index.js#L1 ( regenerator-runtime is the polyfill for async await )

buck54321:
A set of optimizations for the address page. Page load times reduced from > 14 to ~ 2.25 seconds on my server.

1. Removes `AddressMetrics` collection. Instead, an initial set of chart data is collected, in the background if the user is on the list tab, and the bin and zoom buttons are selectively hidden based on the incoming data by looking at it's time range. In addition to the load time improvement, collecting the chart data is also faster than `AddressMetrics`.

2. Graph data cached. E.g. changing bin size from week to month and back to week doesn't incur an additional request. The cache is dumped on controller disconnect. 

3. List pagination is handled with ajax. While a Turbolinks visit is fast enought for list pagination alone, the chart component sometimes requires a substantial amount of data, so preventing additional requests is critical.

4. Full query parameter integration. Zoom, bin, chart type, sent/net/received checkboxes, and list pagination parameters can all be passed as query parameters. Additionally, they are updated as you change the associated inputs on the page. Sharing the url from the omnibar will bring a user to the exact same view currently displayed. 

5. Inline scripts removed. Complete stimulus integration. Very little jQuery remains. 

6. Dygraph script is only loaded when needed. Because we use Turbolinks, and `Dygraph` is global, there was no need to load it with the controller on every connection.

7. Forward and back browser navigation fixed.

Other than the project treasury, check out these addresses of varying activity level.

```
Dca3BGMZPYDBPVFMG66aopfgSzNxS2JxjV1 : 131 i/o
Dca71H3XmNkGgyuNacxaJLAnbxgAYU6g3JC : 449 i/o
Dca8bQ4UX3ZKFkNTbUkmHbrfdhur1sPSNfL : 3,314 i/o
DccJwfrmQYrUNjnbECoeWHNfuwxfFy39bNp : 27,218 i/o
```

Resolves #835, #643. Also part of #843.

explorer/explorerroutes.go
```diff
@@ -1165,6 +1164,13 @@ func (exp *explorerUI) AddressPage(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	pageKey, ok := r.Context().Value(ctxAddressPageKey).(string)
```
chappjc: Please doc that this will/should be one of the defined templates ("address" or "addresstable"), and the purpose of this is to select the template via configurable middleware (see explorerrouter.go).  Something along those lines please.

views/address.tmpl
```diff
@@ -174,8 +268,8 @@
                       >
                           <input type="button" class="chart-size btn btn_sm year" value="Year" name="year">
                           <input type="button" class="chart-size btn btn_sm month" value="Month" name="month">
-                          <input type="button" class="chart-size btn btn_sm week btn-active" value="Week" name="week">
-                          <input type="button" class="chart-size btn btn_sm day" value="Day" name="day">
+                          <input type="button" class="chart-size btn btn_sm week" value="Week" name="week">
+                          <input type="button" class="chart-size btn btn_sm day btn-active" value="Day" name="day">
                           <input type="button" class="chart-size btn btn_sm all" value="All Blocks" name="all">
```
chappjc: I was thinking about the "All Blocks" button for the bar/group size, and I think that "Block" is more consistent with "Year", "Month", etc. for describing the width of a bin.  Might as well make that change in this PR too, if you agree with that.

db/dbtypes/types.go
```diff
@@ -677,3 +680,198 @@ type SideChain struct {
 	Hashes  []string
 	Heights []int64
 }
+
+// AddressTx models data for transactions on the address page.
+type AddressTx struct {
```
chappjc: What is the motivation and justification for moving these types from `explorer` to `dbtypes`?  They really are designed to support the explorer rather than a db backend (AddressTx and AddressInfo are data for various pages, IOID does text formatting for display).

However, it always felt backwards that `dcrpg` should import `explorer`.

db/dcrpg/pgblockchain.go
```diff
@@ -711,11 +712,15 @@ func (pgb *ChainDB) HeightHashDB() (uint64, string, error) {
...
 func (pgb *ChainDB) Height() uint64 {
+	pgb.RLock()
+	defer pgb.RUnlock()
 	return uint64(pgb.bestBlock)
 }
 
 // HashStr uses the last stored block hash.
 func (pgb *ChainDB) HashStr() string {
+	pgb.RLock()
+	defer pgb.RUnlock()
```
chappjc: These are warranted, for sure, but please do audit for calls to these to make sure we won't introduce a deadlock somewhere.

db/dcrpg/pgblockchain.go
```diff
@@ -1341,6 +1346,202 @@ func (pgb *ChainDB) AddressHistory(address string, N, offset int64,
 	return addressRows, &balanceInfo, nil
 }
 
+// AddressData returns comprehensive, paginated information for an address.
+func (db *ChainDBRPC) AddressData(address string, limitN, offsetAddrOuts int64,
```
chappjc: Good move!  AddressPage had way too much going on.

db/dcrsqlite/apisource.go
```diff
@@ -1465,7 +1469,10 @@ func (db *wiredDB) GetExplorerAddress(address string, count, offset int64) (*exp
 		KnownTransactions: numberMaxOfTx,
 		KnownFundingTxns:  numReceiving,
 		KnownSpendingTxns: numSpending,
-	}, addrType, nil
+	}
+	addrData.PostProcess(uint32(db.GetHeight()))
```
chappjc: Please doc this line about sorting, etc.

explorer/explorerroutes.go
```diff
@@ -1586,6 +1420,8 @@ func (exp *explorerUI) Search(w http.ResponseWriter, r *http.Request) {
 
 	// Try aux DB if in full mode.
 	if !exp.liteMode {
+		// This is be unnecessarily duplicative and possible very
+		// slow for a very active addresss.
```
chappjc: I'd like to personally follow up on this PR because the call to `exp.blockData.GetExplorerAddress` several lines up is no longer appropriate above.  It was simply used to validate the address, but since adding `txhelpers.AddressValidation`, that's what should be done above (avoiding the expensive SearchRawTransactoinVerbose RPC).

The purpose of this "Try aux DB if in full mode" when the RPC was *just used* above was because transactions for orphaned blocks may not show up in the output of searchrawtransactions, but PostgreSQL may have recorded data for the transaction and associated addresses.  This is the same reasoning for checking PostgreSQL for transactions with `exp.explorerSource.Transaction` on [this line below](https://github.com/decred/dcrdata/pull/845/files#diff-022a8370d807e3c88802b53baa33af77R1463).

However, Search does not need to do the address queries at all.  That is the job of the address page.  Here we only need to do `txhelpers.AddressValidation` (as done in `GetExplorerAddress` and elsewhere) and redirect to the appropriate /address URL.

Anyway, I can pick this up after your PR since it's a Search page issue.

rpcutils/rpcclient.go
```diff
@@ -457,3 +457,56 @@ func GetChainWork(client *rpcclient.Client, hash *chainhash.Hash) (string, error
 	}
 	return header.ChainWork, nil
 }
+
+// UnconfirmedTxnsForAddress returns the chainhash.Hash of all transactions in
+// mempool that (1) pay to the given address, or (2) spend a previous outpoint
+// that paid to the address.
+func UnconfirmedTxnsForAddress(client *rpcclient.Client, address string, params *chaincfg.Params) (*txhelpers.AddressOutpoints, int64, error) {
```
chappjc: Another great move, thanks.

db/dbtypes/types.go
buck54321: Yeah. I sorta went out on a limb there. Initially, I wanted to use the code from `explorer.AddressPage` in a new function `explorer.AddressTable`. As I started to deconstruct it to move to what is now `pgblockchain.AddressData`, I saw that it was one of the few places in `db` where explorer types were being used, which as you mentioned felt backwards. 

I'm also thinking that there might be some additional optimization possible from a tighter integration with `pgblockchain`. For instance, there seems to be some code duplication here and here. 

It also seemed appropriate for the data source to do the brunt of the data manipulatio

explorer/explorerroutes.go
buck54321: This is outdated now. Now using [the same context with a different explorer function](https://github.com/decred/dcrdata/blob/54ed2a0d3cf9022f08c0fc81e1b9dc99ba7375d7/main.go#L529).

db/dbtypes/types.go
chappjc: `dbtypes` works for me.  We'll just want to put a bit about this design change in the commit messages.

BTW, for 4.0 and such API changes, we'll be bumping the module version from v3 to v4 shortly.

public/js/controllers/address_controller.js
```diff
@@ -73,312 +104,672 @@ function plotGraph (processedData, otherOptions) {
...
   static get targets () {
     return ['options', 'addr', 'btns', 'unspent',
-      'flow', 'zoom', 'interval', 'numUnconfirmed', 'formattedTime', 'txnCount']
+      'flow', 'zoom', 'interval', 'numUnconfirmed', 'formattedTime',
+      'pagesize', 'txntype', 'txnCount', 'qricon', 'qrimg', 'qrbox',
+      'paginator', 'pageplus', 'pageminus', 'listbox', 'table',
+      'range', 'chartbox', 'noconfirms', 'chart', 'pagebuttons']
   }
 
   initialize () {
```
gozart1: Most of the code inside `initialize` seems like it should be in `connect` instead so that for example if you navigate from one address to another via search bar, the state is wiped and rebuilt from scratch.
https://github.com/stimulusjs/stimulus/issues/75

db/dcrpg/pgblockchain.go
buck54321: Good call. Doesn't look like any chance of deadlocks, but there were a couple more reads that needed protection. 

The mutex is only used on the `bestBlock` and `bestBlockHash` members of `ChainDB`, so rather than embed the mutex in `ChainDB`, would it be better to have it as a named member?

chappjc: That's a good idea, although it might be cleaner to avoid a named mutex in this case.  I might make it similar to the changes in https://github.com/decred/dcrdata/pull/756.  It's a similar situation here (big struct where it's unclear what fields the mutex is supposed to guard).  i.e. something like:

```go
type BestBlock struct {
	sync.RWMutex
	Height int64
	Hash string
}
type ChainDB struct {
	...
	bestBlock          BestBlock
	...
}
// Height uses the last stored height.
func (pgb *ChainDB) Height() uint64 {
	pgb.bestBlock.RLock()
	defer pgb.bestBlock.RUnlock()
	return uint64(pgb.bestBlock.Height)
}
...
```

db/dcrpg/pgblockchain.go
```diff
@@ -711,23 +720,23 @@ func (pgb *ChainDB) HeightHashDB() (uint64, string, error) {
 }
 
 // Height uses the last stored height.
-func (pgb *ChainDB) Height() uint64 {
```
chappjc: I'm surprised losing (*ChainDB).Height/Hash didn't affect more code, namely callers from other packages.

To stabilize the `ChainDB` API a bit, we could add some wrappers for the removed functions, but this losing Height and Hash might be a good thing since it wasn't completely clear when to use Height vs. HeightDB, and Hash vs. HashDB.  So anyway, this looks good.

chappjc:
In testing, I noticed that the charts for /address/DsaHTRseceeRrCwcEDxkM51KWtpWgWXgv6h are empty, although list view shows two confirmed transactions.  The "Transaction Type" and "Send and Received" charts are empty, but there are axes shown.  However, the "Total Unspent" chart does not appear and instead says "There are no confirmed transactions to display.".

For this address, the zoom and group buttons also disappear and do not come back until reloading the page.

buck54321:
So in the following scenario, what is the expected behavior?

An address has a different data date range for say "All Transactions" and "Total Unspent". When they are zoomed on a date range, and switch charts, should the zoom level

1. Clamp onto the available data, keeping the same time range if possible
2. Always reset to "All Time"

buck54321:
This was more involved than you would think it would be, so I want to motivate some of the changes here. 

1. Added an appropriate amount of x padding around the data. The only way I can seem to do this with Dygraph is by adding additional dummy points at the beginning and end of the data set. This solves a couple of issues. First, the graph view will always contain full bars, where before, it always cut off in the middle of a bar. Second, this allows us to plot single points where previously charts would have been disabled for single points. Additionally, this allows an address with all transactions on the same day, but in multiple blocks, to group the transctions by day. Previously, the day grouping was disabled in this case.

2. Bar widths now correspond to bin size. Previously, there was an algorithm that iterated the data looking for the smallest gap, and based the bar width on that. That's clever, and sometimes might look better, but it makes it difficult to display things nicely, and I'm also fairly certain that for a bar graph, the expected behavior is that the bar width will correspond to the bin size. 

3. When changing views, zoom ranges are "validated". The zoom is "clamped" and any pre-selected zoom button (day, week, ...) are properly applied to the new data range. 

4. I made a few style changes that I should have put into a separate PR but didn't, so I'm hoping we can find agreement. Otherwise I'll roll them back. On the other hand, if we like the changes, I can apply the styling to /graphs, /ticketpool, ... as well. There are some additional, site-wide, graph styling changes I'd like to fiddle with, but I'll put them in a separate PR. 

Another potential issue related to @chappjc's comment is that the "unspent" data returned from the api (see `apiroutes.getAddressTxUnspentAmountData`) is always empty if the address currently has zero balance, even if there were times when the address had a balance. Shouldn't the "Total Unspent" chart essentially show the address balance over time/block? 

image

chappjc:
Yeah, I agree re: "for a bar graph, the expected behavior is that the bar width will correspond to the bin size."

I think it's OK to make style changes along with these changes, especially if they facilitate addressing the issues the PR is meant to address, but depending on the change (is it self-contained and potentially applicable to 3.1, will it be easier to rebase with a separate commit, etc.) it might be good to have those as a separate commit in this PR.  I'll leave that up to you.

buck54321:
The style changes only apply to the address chart, and are not necessary or important to back port. I'd love to leave them here.

chappjc:
Cool.  Works for me then.

buck54321:
I reworked the `unspent` graph into a balance over time graph. It uses the same API data as the "Sent and Received" graph. Both chart datasets are processed simultaneously, which saves another API call.

chappjc:
Would be awesome to get this in asap, so I'll give it another spin tonight.

chappjc:
The issue with disappearing elements that I mentioned in [this comment](https://github.com/decred/dcrdata/pull/845#issuecomment-446773535) still exists for me (production build with npm run build vs watch):

image

chappjc:
@gozart1 This is a g2g from my perspective.  Wanna have another look before merge?

buck54321:
*Final commit message* 

Front- and back-end fixes and tweaks for the address page.

Fix browser arrow navigation. Move list pagination to stimulus/ajax. Give buttons a more modern style. Store all list and chart settings as part of the url query parameters. Rework "Unspent Total" chart into "Balance" chart (balance v. time). Add padding to chart x range to improve data display. 

Load chart data silently to improve page load times. Cache chart data to reduce api calls. Moved all inline scripts from address.tmpl to a stimulus controller. Skipped repeated Dygraph script loads, i.e on second+ Turbolinks visit. Skip AddessMetrics collection in favor of on-the-fly button disabling in the browser, based on downloaded data range. Relocate a few associated explorer types and data processing to db. 

Expect some of these performance and style changes to be propagated to other pages, specifically charts.

PR #846, opened 2018-11-29
explorer/views: detect unsupported address types

chappjc:
rebirth of https://github.com/decred/dcrdata/pull/743
Resoves https://github.com/decred/dcrdata/issues/618.

This recognizes the following unsupported or indirectly-supported address types:
- P2PK: Converted to P2PKH and a link is given to the user.  This is not an automatic redirect so that the user can see that the address has changed.
- Bitcoin: Use btcutil to attempt decoding of an address if dcrutil fails to decode it.  If it is a bitcoin address, provide a link to blockcypher.
- Zero-pubkey/dummy address.

Notable code changes:
- Create txhelpers.AddressValidation().
- Create txhelpers types: AddressType and AddressError.
- Move IsZeroHashP2PHKAddress and ValidateNetworkAddress into txhelpers.
- Add Net to explorer.AddressInfo
- The expStatus types now have Is* functions to easily check their value in html templates.

Related fixes:
- Full mode will not query DB if address is the zero/dummy address.
- addrMetrics on the address page was a nil pointer when in lite mode. This is fixed.


chappjc:
Consider porting this to 3.1 for security reasons.

chappjc:
This recognizes additional address types, and validates all input address strings.

The following unsupported or indirectly-supported address types are now handled:

P2PK: Converted to P2PKH and a link is given to the user. This is not an 
automatic redirect so that the user can see that the address has 
changed.

Bitcoin: Use btcutil to attempt decoding of an address if dcrutil fails 
to decode it. If it is a bitcoin address, provide a link to blockcypher.

Zero-pubkey/dummy address.

Notable code changes:

Create txhelpers.AddressValidation().
Create txhelpers types: AddressType and AddressError.
Move IsZeroHashP2PHKAddress and ValidateNetworkAddress into txhelpers.
Add Net to explorer.AddressInfo
The expStatus types now have Is* functions to easily check their value 
in html templates.
Related fixes:

Full mode will not query DB if address is the zero/dummy address.
addrMetrics on the address page was a nil pointer when in lite mode. 
This is fixed.

PR #847, opened 2018-11-29
Agenda chart data: inline -> AJAX

buck54321:
Removes inline chart data from agenda.tmpl. Data is retrieved via AJAX instead.

#843

### DIFF
--- a/api/apirouter.go
+++ b/api/apirouter.go
@@ -173,6 +173,10 @@ func NewAPIRouter(app *appContext, useRealIP bool) apiMux {
 		})
 	})
 
+	mux.Route("/agenda", func(r chi.Router) {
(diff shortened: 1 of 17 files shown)


chappjc:
Looks nice.  Ready for testing?

buck54321:
Yep. Test away.

chappjc:
tACK.

Sorry, @buck54321, but you and gozart drew straws and his scss PR goes first, making your life a bit harder.  The index.js moved to the root of public, and there were css changes, so there will be conflicts.  But seriously, the bigger scss PR would be a pain to rebase I think.

buck54321:
No problem. I was anticipating that. Better to make the structural changes first.

chappjc:
Ready for that rebase now.

PR #850, opened 2018-11-29
Run npm install and run build in CI

chappjc:
The runs `npm install` and `npm run build` and checks the exit codes to ensure everything webpack does succeeds, including linting with eslinter and stylelint.

This requires https://github.com/decred/dcrdocker/pull/43 to update the docker build image.


chappjc:
This is now working with the new decred/dcrdata-golang-builder-1.11Docker Hub image added in https://github.com/decred/dcrdocker/pull/45.

PR #851, opened 2018-12-03
ticketpool chart data inline -> AJAX

buck54321:
Removes inline data from the ticketpool page. Initial chart data is retrieved instead via AJAX. 

There is a websocket call that retrieves the same data. I left the websocket source, though now that there is an API endpoint, there may be an argument against the duplication.


chappjc:
Got some more conflicts.  The one in explorerroutes.go is just because StatusPage got that extra input, but the call is gone so not really a conflict.  :)  The other conflicts are from your other PRs I think.

chappjc:
Code looks good, and everything behaves as expected in testing.  However, one think I just noticed now is that it looks like the ticketpool controller's "wrapper" target is not used.  If I'm just missing it, please let me know what it's for.

buck54321:
Oh yeah. I was planning on getting rid of some jQuery in favor of stimulus, but I guess I never finished. I've done that now.

I've also taken a queue from @gozart1 and switched to arrow functions where appropriate, though I left the more semantic `controller` variables in place.

PR #852, opened 2018-12-03
no-JS fixes

chappjc:
Set fallback fonts of the same class as the webfonts:
- monospace for inconsolata-v15-latin-regular
- sans-serif for source-sans-pro-v9-latin-regular
Hide all "Age" columns with jsonly class since all age values are set
via stimulus.js.
Also hide all ages in "(X ago)" spots.
Hide the sun/moon icon with jsonly since it is also a webfont. Ideally
both the Night Mode and Enable Hot Keys menu items would be hidden
entirely, but they are difficult to eliminate. `d-inline-block` uses
`display: inline-block !important`, which seems to neutralize jQuery's `show()`.
Hide the "since last block" and websocket status footer elements.
Prevent text like Coinbase and Stakebase from being hash-elided by
defining a new flag, TextIsHash, that is set in Go to avoid heuristics
in the frontend.

### DIFF
--- a/views/extras.tmpl
+++ b/views/extras.tmpl
@@ -77,7 +77,7 @@
                         {{else}}
                             <a data-keynav-skip href="https://explorer.dcrdata.org/" title="Home">Switch To Mainnet</a>
                         {{end}}
-                        <span data-keynav-skip data-action="click->menu#onSunClick" class="jsonly d-inline-block"><span id="sun-icon" class="dcricon-sun-fill no-underline pr-2"></span> Night Mode</span>
+                        <span data-keynav-skip data-action="click->menu#onSunClick" class="jsonly d-inline-block"><span id="sun-icon" class="dcricon-sun-fill no-underline pr-2 jsonly"></span> Night Mode</span>
                         <a data-keynav-skip data-turbolinks="false" href="#" class="jsonly" id="keynav-toggle">
                             <span class="text">Enable Hot Keys</span><span class="keys-hint">(&nbsp;&nbsp;<span class="arrows"> &#8592;<br>&#8594;</span>&nbsp;&nbsp;enter&nbsp;&nbsp;\&nbsp;&nbsp;=&nbsp;&nbsp;)</span>
                         </a>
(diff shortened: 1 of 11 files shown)


chappjc:
Before:

image
In the above image, not: font, "Coinbase" text jumbled because of hashElide, non-functional footer elements, and "( ago)" in the TIME.

After:

image

The other views show the removed "Age" columns, etc.

views/mempool.tmpl
```diff
@@ -27,7 +27,7 @@
                     <table class="">
                         <tr>
                             <td class="text-right pr-2 lh1rem nowrap p03rem0">LAST BLOCK</td>
-                            <td class="lh1rem"><a href="/block/{{.LastBlockHeight}}" data-target="mempool.bestBlock" data-hash="{{.LastBlockHash}}" data-keynav-priority>{{.LastBlockHeight}}</a> (<span data-target="mempool.bestBlockTime  main.age" data-age="{{.LastBlockTime}}"></span> ago)</td>
+                            <td class="lh1rem"><a href="/block/{{.LastBlockHeight}}" data-target="mempool.bestBlock" data-hash="{{.LastBlockHash}}" data-keynav-priority>{{.LastBlockHeight}}</a><span class="jsonly"> (<span data-target="mempool.bestBlockTime  main.age" data-age="{{.LastBlockTime}}"></span> ago)</span></td>
```
buck54321: Your fix is good, but the timer wasn't working because it was not referencing a valid controller. It needs 

`main.age` `->` `time.age`

views/extras.tmpl
```diff
@@ -77,7 +77,7 @@
                         {{else}}
                             <a data-keynav-skip href="https://explorer.dcrdata.org/" title="Home">Switch To Mainnet</a>
                         {{end}}
-                        <span data-keynav-skip data-action="click->menu#onSunClick" class="jsonly d-inline-block"><span id="sun-icon" class="dcricon-sun-fill no-underline pr-2"></span> Night Mode</span>
+                        <span data-keynav-skip data-action="click->menu#onSunClick" class="jsonly d-inline-block"><span id="sun-icon" class="dcricon-sun-fill no-underline pr-2 jsonly"></span> Night Mode</span>
```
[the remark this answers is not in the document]
chappjc: I discovered the d-inline-block issue, as I noted in the PR description, but I see it's quite a bit worse here.  I considered trying a `removeClass`, but decided not to do anything too drastic.  I'll leave this issue until after the bigger PRs go in.

views/mempool.tmpl
chappjc: Fixing that while I'm here.  Good catch.

chappjc:
@buck54321 Thanks for your PR.  Changes merged in.  Testing it out.

chappjc:
Seems all good now.  I had to put the `#menu span.jsonly, a.jsonly` into the new navigation.scss, but it's working again.

Issue #414, opened 2018-03-27
Dcrpg: all networks use the same database


RogueElement:
Currently all networks use the same default pg database, dcrdata and this will cause problems if checking out another network.

codemaestro64:
Can this be fixed by prefixing table names with current network within same database？

chappjc:
It could, but keeping the tables separate would make backup/restore much easier when just doing one network.  Also, changing the table names would unfortunately require a ton of edits.

Simply changing the default database to `dcrdata_{netname}` would be preferable.

ZeroASIC:
Well that was messy, I apologize for that.  I thought I had screwed something up (I actually made a white space mistake and deleted the commit only to redo it a few minutes later) but it looks like the master branch is a bit busted.

chappjc:
Resolved by https://github.com/decred/dcrdata/pull/862, although it is not the case by default that a dynamic DB name will be used.

PR #862, opened 2018-12-06
Update to use dynamic database name. #414

ZeroASIC:
This is for #414

### DIFF
--- a/main.go
+++ b/main.go
@@ -20,7 +20,6 @@ import (
 	"strings"
 	"sync"
 	"time"
-	"regexp"
 
 	"github.com/decred/dcrd/chaincfg/chainhash"
 	"github.com/decred/dcrd/rpcclient"
@@ -173,16 +172,14 @@ func _main(ctx context.Context) error {
 			DBName:       cfg.PGDBName,
 			QueryTimeout: cfg.PGQueryTimeout,
 		}
+
 		// Replace {netname} with curnet
-		if strings.Contains(dbi.DBName, "{netname}") {
-		  reg, err := regexp.Compile("[^a-zA-Z]+")
-		  if err != nil {
-		    return err
-		  }
-		  
-		  var networkName = reg.ReplaceAllString(curnet.String(), "");
-		  dbi.DBName = strings.Replace(dbi.DBName, "{netname}", networkName, 1);
+		const nettemplate = "{netname}"
+		if strings.Contains(dbi.DBName, nettemplate) {
+			networkname := strings.ToLower(curnet.String())
+			dbi.DBName = strings.Replace(dbi.DBName, nettemplate, networkname, -1)
 		}
+
 		chainDB, err := dcrpg.NewChainDBWithCancel(ctx, &dbi, activeChain, baseDB.GetStakeDB(), !cfg.NoDevPrefetch)
 		if chainDB != nil {
 			defer chainDB.Close()
--- a/main.go
+++ b/main.go
@@ -173,12 +173,8 @@ func _main(ctx context.Context) error {
 			QueryTimeout: cfg.PGQueryTimeout,
 		}
 
-		// Replace {netname} with curnet
-		const nettemplate = "{netname}"
-		if strings.Contains(dbi.DBName, nettemplate) {
-			networkname := strings.ToLower(curnet.String())
-			dbi.DBName = strings.Replace(dbi.DBName, nettemplate, networkname, -1)
-		}
+		// If using {netname} then replace it with netName(activeNet).
+		dbi.DBName = strings.Replace(dbi.DBName, "{netname}", netName(activeNet), -1)
 
 		chainDB, err := dcrpg.NewChainDBWithCancel(ctx, &dbi, activeChain, baseDB.GetStakeDB(), !cfg.NoDevPrefetch)
 		if chainDB != nil {
--- a/main.go
+++ b/main.go
@@ -20,6 +20,7 @@ import (
 	"strings"
 	"sync"
 	"time"
+	"regexp"
 
 	"github.com/decred/dcrd/chaincfg/chainhash"
 	"github.com/decred/dcrd/rpcclient"
@@ -171,6 +172,16 @@ func _main(ctx context.Context) error {
 			Pass:         cfg.PGPass,
 			DBName:       cfg.PGDBName,
 			QueryTimeout: cfg.PGQueryTimeout,
+		}
+		// Replace {netname} with curnet
+		if strings.Contains(dbi.DBName, "{netname}") {
+		  reg, err := regexp.Compile("[^a-zA-Z]+")
+		  if err != nil {
+		    return err
+		  }
+		  
+		  var networkName = reg.ReplaceAllString(curnet.String(), "");
+		  dbi.DBName = strings.Replace(dbi.DBName, "{netname}", networkName, 1);
 		}
 		chainDB, err := dcrpg.NewChainDBWithCancel(ctx, &dbi, activeChain, baseDB.GetStakeDB(), !cfg.NoDevPrefetch)
 		if chainDB != nil {
--- a/sample-dcrdata.conf
+++ b/sample-dcrdata.conf
@@ -50,7 +50,9 @@
 ;pg=true
 
 ; PostgreSQL database config (when pg=true)
-;pgdbname=dcrdata
+; Create a database for each network you plan to run.
+; {netname} will be replaced with the current network name. (e.g. MainNet, TestNet, SimNet)
+;pgdbname=dcrdata_{netname}
 ;pguser=dcrdata
 ;pgpass=
 ; Connect via TCP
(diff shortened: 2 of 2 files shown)


main.go
```diff
@@ -20,6 +20,7 @@ import (
 	"strings"
 	"sync"
 	"time"
+	"regexp"
```
chappjc: Before getting into the design (using a template with {netname}, etc.), I want to go over some general Go rules to follow.

Here we sort the imports, but keep the std lib ones separated from the remote ones by a new line.

main.go
```diff
@@ -172,6 +173,16 @@ func _main(ctx context.Context) error {
 			DBName:       cfg.PGDBName,
 			QueryTimeout: cfg.PGQueryTimeout,
 		}
+		// Replace {netname} with curnet
```
chappjc: For consistent style, please put a newline before the comment.  Not all blocks are separated by newlines of course, but chaining too many together gets confusing.  Plus it's almost always the case that if a chunk of code is commented like this, we will precede it with a newline.

main.go
```diff
@@ -172,6 +173,16 @@ func _main(ctx context.Context) error {
...
+		if strings.Contains(dbi.DBName, "{netname}") {
+		  reg, err := regexp.Compile("[^a-zA-Z]+")
+		  if err != nil {
+		    return err
+		  }
+		  
+		  var networkName = reg.ReplaceAllString(curnet.String(), "");
+		  dbi.DBName = strings.Replace(dbi.DBName, "{netname}", networkName, 1);
+		}
```
chappjc: And a newline after this block of code, before NewChanDB...

main.go
```diff
@@ -172,6 +173,16 @@ func _main(ctx context.Context) error {
 			DBName:       cfg.PGDBName,
 			QueryTimeout: cfg.PGQueryTimeout,
 		}
+		// Replace {netname} with curnet
+		if strings.Contains(dbi.DBName, "{netname}") {
```
chappjc: All .go code must be formatted consistently with `gofmt`.  Practically, this means running `go fmt ./...` at the root of the repo and/or using an IDE that will gofmt code on save.

main.go
```diff
@@ -172,6 +173,16 @@ func _main(ctx context.Context) error {
...
+		  reg, err := regexp.Compile("[^a-zA-Z]+")
```
chappjc: What is the reasoning for removing other valid DB name characters like numerals?

main.go
```diff
@@ -172,6 +173,16 @@ func _main(ctx context.Context) error {
...
+		  if err != nil {
+		    return err
+		  }
+		  
+		  var networkName = reg.ReplaceAllString(curnet.String(), "");
```
chappjc: For consistency, when defining a variable, use the syntax that omits var (`networkName := ...`).  Only when declaring with the intention of starting with the type's zero value do we do `var netName string`.

chappjc: The one modification to the network name that I would make is changing to all lower case.

sample-dcrdata.conf
```diff
@@ -50,7 +50,9 @@
 ;pg=true
 
 ; PostgreSQL database config (when pg=true)
-;pgdbname=dcrdata
+; Create a database for each network you plan to run.
+; {netname} will be replaced with the current network name. (e.g. MainNet, TestNet, SimNet)
+;pgdbname=dcrdata_{netname}
```
chappjc: Since the default is still `pgdbname=dcrdata` (and I think it should be), we should leave `;pgdbname=dcrdata` above these new lines and note that it is the default, then on these lines, we would explicitly mention the special template processing.

main.go
chappjc: Actually, please see the `netName` function at the end of config.go that gets the current lowercase network name out of the active network (chaincfg).

main.go
```diff
@@ -172,6 +173,16 @@ func _main(ctx context.Context) error {
...
+		  dbi.DBName = strings.Replace(dbi.DBName, "{netname}", networkName, 1);
```
chappjc: Might as well be `-1`.

chappjc: Best to define a single constant for `"{netname}"` since it is written as a literal twice.

main.go
chappjc: Note the CI (Travis) failure: https://travis-ci.org/decred/dcrdata/builds/464151405?utm_source=github_status&utm_medium=notification

main.go
ZeroASIC: I thought that if TestNet3 gets bumped to TestNet4 that it should use the same DB.  That way we can test any reorg that would get run on MainNet instead of renaming the DB.  This might not matter since most of the people that run this code are Devs and can do that pretty easily.

chappjc: The thing with testnet3 -> testnet4 is that they are entirely different blockchains with different genesis blocks.  Starting up on testnet4 for the first time using an existing testnet3 database won't work.

main.go
ZeroASIC: For the moment I updated this line to `networkname := strings.ToLower(curnet.String())`
I could add some logic to main.go and call 'netname` but that would be duplicating logic that already exists in config.go.  Alternatively I could add `cfg.NetworkName = netname(activeNet)` to config.go and eliminate this line completely.

chappjc: `netName` is in the same namespace (`package main`) as `_main`, so you can use it directly.  Unlike some other languages where the compilation unit is the source file (and whatever included headers), Go has all the source files for a package as part of one compilation unit.

sample-dcrdata.conf
```diff
@@ -50,9 +50,10 @@
 ;pg=true
 
 ; PostgreSQL database config (when pg=true)
-; Create a database for each network you plan to run.
-; {netname} will be replaced with the current network name. (e.g. MainNet, TestNet, SimNet)
-;pgdbname=dcrdata_{netname}
+; It's possible to have dcrdata switch between databases based on the network it's connected to.
```
chappjc: We're not particularly strict in conf files, but a general guideline is to wrap after about 80 columns.

sample-dcrdata.conf
```diff
@@ -50,9 +50,10 @@
 ;pg=true
 
 ; PostgreSQL database config (when pg=true)
-; Create a database for each network you plan to run.
-; {netname} will be replaced with the current network name. (e.g. MainNet, TestNet, SimNet)
-;pgdbname=dcrdata_{netname}
+; It's possible to have dcrdata switch between databases based on the network it's connected to. 
+; Create a database for each network you plan to run and set pgdbname=dcrdata_{netname}
```
chappjc: period (.)

sample-dcrdata.conf
```diff
@@ -50,9 +50,10 @@
...
+; {netname} will be replaced with the current network name. (e.g. mainnet, testnet, simnet)
```
chappjc: testnet -> testnet3

main.go
```diff
@@ -173,16 +172,14 @@ func _main(ctx context.Context) error {
 			DBName:       cfg.PGDBName,
 			QueryTimeout: cfg.PGQueryTimeout,
 		}
+
 		// Replace {netname} with curnet
```
chappjc: Comments in the form of sentences should have punctuation.

main.go
```diff
@@ -173,16 +172,14 @@ func _main(ctx context.Context) error {
...
-		if strings.Contains(dbi.DBName, "{netname}") {
-		  reg, err := regexp.Compile("[^a-zA-Z]+")
-		  if err != nil {
-		    return err
-		  }
-		  
-		  var networkName = reg.ReplaceAllString(curnet.String(), "");
-		  dbi.DBName = strings.Replace(dbi.DBName, "{netname}", networkName, 1);
+		const nettemplate = "{netname}"
```
chappjc: Idiomatic variable naming in Go uses snakeCase (`netTemplate`).

main.go
```diff
@@ -173,16 +172,14 @@ func _main(ctx context.Context) error {
...
+		if strings.Contains(dbi.DBName, nettemplate) {
+			networkname := strings.ToLower(curnet.String())
+			dbi.DBName = strings.Replace(dbi.DBName, nettemplate, networkname, -1)
```
chappjc: Basically something like this:

```go
		// Replace {netname} with active network name.
		const netTemplate = "{netname}"
		if strings.Contains(dbi.DBName, netTemplate) {
			dbi.DBName = strings.Replace(dbi.DBName, netTemplate,
				netName(activeChain), -1)
		}
```

chappjc: But now that I think about this, the `Contains` check can be omitted:

```go
dbi.DBName = strings.Replace(dbi.DBName, netTemplate, netName(activeChain), -1)
```

ZeroASIC: Okay, I'm about to commit an update for this.  It's a one line change with a comment now.  I removed the constant because it's only used in one place now, but I can easily add it back in.  I installed Atom and set it up with a linter and everything.  So far so good, except it changed a few files automatically so I reset them.  Thanks for all your feed back.

ZeroASIC:
Okay, I'm about to commit an update for this. It's a one line change with a comment now. I removed the constant because it's only used in one place now, but I can easily add it back in. I installed Atom and set it up with a linter and everything. So far so good, except it changed a few files automatically so I reset them. Thanks for all your feed back.

PR #865, opened 2018-12-06
.jsonly elements shown on every Turbolinks visit

buck54321:
Followup to #853. I'm not sure how I didn't see this in testing.


chappjc:
To clarify, is the 3.1 version (that PR) OK and this is just for master, or does 3.1 need a similar change?

buck54321:
Ah. I forgot it's inline on 3.1-stable, so it's evaluated on a Turbolinks visit anyway. No change needed.

chappjc:
OK, perfect.  That's what I though.  Merging...

PR #867, opened 2018-12-07
[3.1] add permalink icon on charts page

chappjc:
NOTE: This is a 3.1 workaround to the **charts** page navigation troubles in Issue https://github.com/decred/dcrdata/issues/814.  A proper resolution targeting dcrdata 3.2/4.0 is in https://github.com/decred/dcrdata/pull/845.

This removes the `pushState` call that broke the ability to navigate forward/back. As a result, the anchors no longer end up in the URL.  To at least provide a permalink for a given view of the **charts page**, this PR adds a link icon wrapped in an actual link that can be copied.

However, this may not be a desirable solution, even to get 3.1 out the door, if in 3.2/4.0 we use URL queries instead of anchors as this defeats the purpose of a permalink.  Perhaps we should eliminate the URL anchors entirely from the charts page and resolve it properly for the next release.  Thoughts, @buck54321 and @gozart1?


chappjc:
This would look like the following on the /charts page:

image

You could get at the permalink to the shown chart by copying the link on that image.

Note that I meant to say "charts page", not "address page", in the original PR description.

chappjc:
Obsoleted by https://github.com/decred/dcrdata/pull/870

PR #870, opened 2018-12-07
back and forward navigation fixed for charts page

buck54321:
Replaces pushState with a Turbolinks version of `replaceState` so that page navigation works with the browsers forward and back arrows.


chappjc:
tACK.  Really nice find.

PR #873, opened 2018-12-10
Last of the inline JS removed

buck54321:
1. Inline `DCRThings` moved to index.js
2. The `stats` page is a mystery to me (is there a link to it somewhere?), but it had some code for tooltips using a bootstrap jQuery plugin, which doesn't appear to be installed. I just removed the code. It does have a title that renders pretty nicely on dektop, though the page itself has some issues in firefox. I also removed some extraneous and duplicate `id` attributes I suspect of being copypasta.
3. New stimulus controller for the ticket price windows page.

### DIFF
--- a/public/index.js
+++ b/public/index.js
@@ -27,6 +27,13 @@ $.ajaxSetup({
   cache: true
 })
 
+{
+  let navBar = document.getElementById('navBar')
+  window.DCRThings = {}
+  window.DCRThings.targetBlockTime = navBar.dataset.blocktime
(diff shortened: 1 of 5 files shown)


chappjc:
Regarding the stats page, there are several pages that are not quite ready for primetime -- stats and nexthome are among them.

There's also /rejects (pos-invalidated blocks) and /side (side chains / orphaned blocks), which probably deserve links somewhere now.  I didn't want to put them on the nav menu since it's getting crowded already.  Suggestions are welcome from all on this.

PR #874, opened 2018-12-10
Source maps

gozart1:
* Add source maps for all js that webpack touches, to aid with debugging production builds.
* Remove source map reference from jquery-migrate.min.js


chappjc:
I think that jquery.com/download page is not up-to-date.  There seems to be a released v3.0.1 of jquery-migrate-plugin: https://github.com/jquery/jquery-migrate/#development-vs-production-versions

gozart1:
updated to jquery-migrate-3.0.1.min.js

PR #883, opened 2018-12-12
dcrsqlite: mitigate database is locked errors

chappjc:
This resolves an issue with go-sqlite3 where the following error may be encountered:

    database is locked

which corresponds to the [SQLITE_BUSY error](https://sqlite.org/rescode.html).

The issue is discussed at length in https://github.com/mattn/go-sqlite3/issues/274, but the gist is:

1. SQLite should be compiled in "sequential mode" to handle multithreaded access. This change was made for go-sqlite v1.10.0  in https://github.com/mattn/go-sqlite3/commit/acfa60124032040b9f5a9406f5a772ee16fe845e. This dependency update is made in https://github.com/decred/dcrdata/pull/877
2. Limiting the sql driver to just one active connection (the driver it maintains a connection pool) prevents concurrent access. 

   ```go
   db.SetMaxOpenConns(1)
   ```

3. Setting a non-zero value for "busy_timeout" instructs SQLite to retry the DB access for the specified amount of time before throwing the `SQLITE_BUSY` error.  This is a partial resolution.

   ```go
   dbPath = dbPath + "?cache=shared&_busy_timeout="  + DBBusyTimeout
   ```

### DIFF
--- a/db/dcrsqlite/sqlite.go
+++ b/db/dcrsqlite/sqlite.go
@@ -47,6 +47,10 @@ const (
 	// A couple database queries that are called before NewWiredDB
 	SetCacheSizeSQL       = "PRAGMA cache_size = 32768;"
 	SetSynchrounousOffSQL = "pragma synchronous = OFF;"
+
+	// DBBusyTimeout is the length of time in milliseconds for sqlite to retry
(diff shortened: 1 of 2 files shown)


chappjc:
This has been working without error on the 3.1-stable branch for several days.  Ready to merge to master.

PR #892, opened 2018-12-17
jQuery.getScript -> dynamic import. jQuery.ajax -> axios

buck54321:
Important step in dropping jQuery dependence. Adds dependency on axios module, so don't forget to `npm install`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
     "@babel/core": "^7.1.6",
     "@babel/plugin-syntax-dynamic-import": "^7.0.0",
     "@babel/preset-env": "^7.1.6",
+    "axios": "^0.18.0",
     "babel-eslint": "^10.0.1",
     "babel-loader": "^8.0.4",
     "babel-polyfill": "^6.26.0",
--- a/public/js/controllers/address_controller.js
+++ b/public/js/controllers/address_controller.js
@@ -1,13 +1,13 @@
-/* global Dygraph */
 /* global QRCode */
 /* global $ */
 import { Controller } from 'stimulus'
 import { isEmpty } from 'lodash-es'
-import { padPoints, sizedBarPlotter } from '../helpers/chart_helper'
+import { padPoints, sizedBarPlotter, ensureDygraph } from '../helpers/chart_helper'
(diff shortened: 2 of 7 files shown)


buck54321:
Just a couple of thoughts here for discussion.

Initially, I wanted to implement a drop in replacement for `jQuery.getScript`, but that becomes difficult because of webpack "magic strings" required for chunking. In fact, in your `importModule` function, the passed path is ignored and [the path to dygraphs is hard-coded](https://github.com/gozart1/dcrdata/blob/2a9ba74964e25bac3c5c5c8d45a0d713f32e7406/public/js/helpers/async_helper.js#L5), which is perhaps just an oversight but reveals the difficulty with a `getScript` replacement with webpack and dynamic imports. That's part of how I wound up with `ensureDygraph`. 

I might be wrong, but I believe that a dynamic import already "returns" a Promise, so we should be able to work with that Promise directly instead of wrapping it in another Promise. 

I guess I'm not seeing an issue in using callbacks. The Promise interface is designed specifically to use callbacks. That said, if there is some motivation and agreement to switch everything to await/async, that's fine, but we probably shouldn't apply them to the stimulus "lifecycle callbacks". I think you said on Slack that you already ran into an issue that I beleive is caused by doing that. Instead, a new async function could be spun up that is called at the end of stimulus `connect`. But if we are writing new and similar anync functions in every controller that imports Dygraph, that brings us right back to some common `ensureDygraph`-like function. 

I think this is an important conversation, so I'm looking forward to some back and forth. It'd be nice to get application standards established now while we're still in the process of moving to webpack/ES6.

gozart1:
Hey, good catch on the hardcoded import path, lol. I've rebased onto master and made a fresh PR here with some refinements to the approach.

The promise wrapper on the import call was simply to access the default module in a DRY way, and the new PR has restructured the code to reflect that intent more accurately.

It's a good question about stimulus lifecycle callbacks and if they are supported as async functions, I don't really see why not, and I haven't been able to find any issues, but it's worth digging into the codebase to tell for sure.

buck54321:
Deferring to @gozart via #900. I'll pick up ajax in another PR.

PR #893, opened 2018-12-18
CSS changes to decimalParts to allow copy paste.

### LINKED ISSUE #885: Amounts copied to clipboard include extra spaces
Reproduce:

* open Block page
* select DCR amount from Total DCR column
* Ctrl-C
* paste somewhere

The number is pasted like `102 . 78320917` instead of expected `102.78320917`.

Fee columns are not affected.

This also happens on Address page.

buck54321:
Resolves #885. 

Apparently, flexbox makes all children act as block elements in certain ways, one of which is that newlines are inserted on copy-paste between what would normally be inline elements like spans, regardless of the CSS `display` property and regardless of `!important` rules.


public/scss/mempool_vote_table.scss
```diff
@@ -7,67 +7,3 @@
 .matching-hash {
   background: #0078ff1c;
 }
-
-.mb-3 tbody tr td .fs14-decimal {
```
gozart1: I'm glad you are deleting these horribly brittle selectors.
But it turns out that this one in particular keeps the 'DCR' aligned to the left in the home page:
screen shot 2018-12-17 at 11 57 05 pm

But rather than author any styles to do this, you could just put d-flex on the wrapper div to align things

buck54321:
Removed excess margins around DCR unit and corrected a couple of alignment issues in the last commit. I avoided flexbox for alignment in places where standard inline behavior fit the bill. 

Switched to template literals for `humanize.DecimalParts`.

PR #896, opened 2018-12-19
A couple README updates.

buck54321:
Following up some recent changes, specifically the SCSS webpack integration and v4 bump.


README.md
```diff
@@ -727,6 +728,9 @@ modules.
 
 This will create and install into a directory named `node_modules`.
 
+You'll also want to run `npm install` after merging changes from upstream. It is
+run for you when you use the build script (`/dev/build.sh`).
```
chappjc: I think we should have this be a relative path.  Whatever conveys the repo root is fine by me, whether it's `./dev/build.sh' or a pseudo variable like `<REPO_ROOT>/dev/build.sh'

PR #900, opened 2018-12-21
Async dynamic imports

gozart1:
* Replace $.getScript with webpack dynamic imports
* Replace $.ajax for axios 
* Use async/await to handle xhr asynchronous control flow
* replace vendored qrcode lib with node-qrcode npm package which is module compatible and has active commits


buck54321:
I'd like to extend this just a touch with some [script caching](https://github.com/gozart1/dcrdata/pull/7) to prevent unnecessary script reloads on Turbolinks visits. This does force the dev to create an entry in `module_helper` for any scripts they want to import, but the benefits of caching are worth it avoids having "magic comments" all over the code.

gozart1:
The Webpack module loader already handles caching, so I don't see a strong need for a secondary caching layer.

I've gone ahead and removed the secondary cache for Dygraph in address controller to make this more clear. The webpack module cache check does have slightly more overhead than simply checking the presence of a variable in the controller, but the difference is under a millisecond and thus imperceptible to the end user. So in cases like these, we should always bias towards making the code as simple to understand as possible versus micro-optimizing the runtime. 

Although I like that you've made getDefault a one liner here https://github.com/gozart1/dcrdata/pull/7/files#diff-6d949a6ee81949ca3d9eea14d279f53dR18
It creates a poor separation of concerns, getDefault should not have any knowledge or opinions about what module the caller wants to use.

Also, I don't think we should abstract away the fact that these are webpack dynamic imports into a helper. Since dynamic imports are pretty edge and not commonly understood by devs ( as evidenced by us still learning about them ), the best thing to do is keep things very explicit.

This way, when a new dev comes in and looks at a controller, they'll see the "magic webpack strings" on know to educate themselves on them if they aren't familiar already.

buck54321:
My bad. I somehow got it in my head that the chunked vendor script was being re-evaluated at every import. I now see that's not true.

gozart1:
This last commit https://github.com/decred/dcrdata/pull/900/commits/5d0b62db04062eaef88309936e91548c6fd5cc17#diff-7a74ccfc9bf79b40b1e88f8aa91c81e9R8  standardizes the way libs are lazy loaded by scoping them to the root of the module, similar to @buck54321's initial spike on this. 

I think this makes the most sense, since that way glancing at the top of a file provides a comprehensive overview of its dependencies, regardless of how they are loaded.

gozart1:
I've updated the description to prepare for merge.

PR #902, opened 2018-12-22
Charts page URL query parameters

buck54321:
Integrates query parameters linked chart controls on charts page. Common code between address and chart controllers moved to chart_helper.


public/js/helpers/chart_helper.js
```diff
@@ -53,3 +53,101 @@ export function padPoints (pts, binSize, sustain) {
...
+      start = this.tryDecode(start)
+      end = start.end
+      start = start.start
+    }
+    return parseInt(start / 1000).toString(36) + '-' + parseInt(end / 1000).toString(36)
+  }
+
+  static decode (encoded, lims) {
+    // if lims are provided, encoded can be a zoomMap key
+    encoded = this.tryDecode(encoded)
```
gozart1: In general, I think it's easier to reason about a codebase where we pretend that the inputs to functions are immutable, and always return a new value. So instead of overwriting the input, do:
`let decoded = this.tryDecode(encoded)`

public/js/helpers/chart_helper.js
```diff
@@ -53,3 +53,101 @@ export function padPoints (pts, binSize, sustain) {
...
+    lims = this.tryDecode(lims)
```
gozart1: What does `lims` mean? I don't see this 2nd arg being passed anywhere and it's not clear to me what it's supposed to represent based in the name of the variable.

buck54321: Current x-axis limits. The second parameter is used in `Zoom.validate` and allows the keywords `day`, `week`, `month`, `year`, and `all`  to be used to set the range. Those values map the the last possible window of that size in the available chart range.  

Should I rename it or document it better?

chappjc: Good for now IMO until we deploy and get more feedback

chappjc:
@gozart1 Feel free to merge with an appropriate squash commit if you feel it's g2g. @buck54321, please confirm.

Issue #875, opened 2018-12-10
negative ages/times


chappjc:
Negative times show up in places (e.g. "-21442s since last block").  This seems to happen when the server is in a different time zone than the client.

buck54321:
Just to add a recent observation, because this bug is maybe weirder than we think. I've seen the block timer in the footer be an hour off on a local server when first firing up dcrdata, only to correct itself when the first new block came in.

ZeroASIC:
This was pretty easy to recreate I just sat there and waited for a new block.
https://alpha.dcrdata.org/address/Dsn86fav4tXZnXF1A6L7ngQyF8NWGbHgXqu#list-view

Notice that the Time UTC column changes to NaN-NaN-NaN 
Here's a screen shot
![screenshot_2018-12-18_16-03-11](https://user-images.githubusercontent.com/4195716/50185946-e507ce00-02de-11e9-9c52-ea46c1a0075f.png)

ZeroASIC:
Could this be related to issue #867?  That issue has a screenshot with NaN-NaN-NaN too.

chappjc:
Yes, that's the time issue in https://github.com/decred/dcrdata/issues/876.  It's not clear it is related to this issue, but it's possible.  I suspect Batman is to blame for that one.

ZeroASIC:
I just double checked and the Time Since Last Block was correct on page load, but went negative after about 30 seconds.  Time UTC was still correct this time.

chappjc:
A little debugging of alpha.dcrdata.org in FF:

![image](https://user-images.githubusercontent.com/9373513/50195832-5b6bf680-0306-11e9-97d6-e70ce0b7f250.png)

The client machine's time is about `1545187586`, whereas the `unixTime` fed to `timeSince` should be to the UTC block times (e.g. for block 301975, `"time": 1545187356`).

The input to `timeSince` is clearly incorrect:

`humanize.timeSince(Date.parse(el.dataset.age) / 1000)` in `setAges`.  The value of `el.dataset.age` in that is `"2018-12-19 02:42:36"`, which does not indicate time zone.

`Date.parse` needs to be given time zone information, or it should assume UTC, not local time.

PR #904, opened 2018-12-24
Index.js to stimulus controllers.

buck54321:
Nearly all of the code from index.js can be moved to stimulus controllers. Will also be translating jQuery on the way.


public/js/controllers/tx_controller.js
```diff
@@ -0,0 +1,92 @@
...
+import humanize from '../helpers/humanize_helper'
+
+export default class extends Controller {
+  static get targets () {
+    return ['unconfirmed', 'confirmations', 'formattedAge', 'age', 'progressBar',
+      'ticketStage', 'poolStatus', 'expiryBar']
+  }
+
+  connect () {
+    this.advanceTicketProgress = this._advanceTicketProgress.bind(this)
```
gozart1: Why is .bind(this) being used on controller methods?
Methods declared on the controller class are already scoped to the controller, so it seems redundant to do this. 

Also, what is the leading underscore supposed to indicate?

buck54321: > Methods declared on the controller class are already scoped to the controller

This is not true when used as event callbacks.

```
myFunc () {
    // do stuff
}

myCallback (e) {
    // do stuff
    this.myFunc()
}

connect () {
    globalEventBus.on('BLOCK_RECEIVED', this.myCallback)
}
```
won't work (`Cannot read property 'myFunc' of undefined`).

Typically, I'd just define is as an arrow function in the argument

```
globalEventBus.on('BLOCK_RECEIVED', (e) => {
    // do stuff
})
```

but I can't here because I need to retain a reference for unbinding on disconnect.

I could assign an arrow function to a property in `connect`/`initialize`.

```
connect () {
    this.myFunc = (e) => {
        //do stuff or call class method
    }
    globalEventBus.on('BLOCK_RECEIVED', this.myFunc)
}

disconnect () {
    globalEventBus.off('BLOCK_RECEIVED', this.myFunc)
}
```

But I used bind because I think it's cleaner.

gozart1: Ah right, that makes perfect sense, I think you're taking the right approach then. Thanks for the explanation.

public/js/controllers/time_controller.js
```diff
@@ -1,21 +1,36 @@
...
 
   disconnect () {
     this.stopAgeRefresh()
+    if (this.hasBlocktimeTarget) {
+      globalEventBus.off('BLOCK_RECEIVED', this.processBlock)
+    }
+  }
+
+  _processBlock (blockData) {
+    var block = blockData.block
```
ZeroASIC: block.unixStamp = (new Date(block.time)).getTime() / 1000

public/index.js
```diff
@@ -61,184 +48,9 @@ async function createWebSocket (loc) {
     var newBlock = JSON.parse(event)
     var b = newBlock.block
```
ZeroASIC: I think this line can be removed.

public/index.js
```diff
@@ -61,184 +48,9 @@ async function createWebSocket (loc) {
...
     b.unixStamp = (new Date(b.time)).getTime() / 1000
```
ZeroASIC: This line can either be removed or changed to:
`newBlock.block.unixStamp = (new Date(b.time)).getTime() / 1000`

I'm not completely certain how this object is scoped so if it doesn't carry over to the new functions you added then it needs to be added to those functions instead.  Alternatively it could be changed to:
`newBlock.block.time = (new Date(b.time)).getTime() / 1000`
and just overwrite the timestamp at this point because it doesn't seem to get used as anything but a Unix timestamp.  Then I could change the go daemon to return the Unix timestamp instead of the formatted date that doesn't get used.  In which case the format function would need that multiply by 1000 added back in.

public/js/helpers/humanize_helper.js
```diff
@@ -124,6 +124,12 @@ var humanize = {
       return result
     }
     return Math.floor(seconds) + 's'
+  },
+  formatTxDate: function (stamp, withTimezone) {
+    var d = new Date(stamp)
```
ZeroASIC: This might need to be changed to:
`var d = new Date(stamp) * 1000`

public/index.js
ZeroASIC: I'm testing my proposed changes and so far everything looks good.  The newBlock object is scoped as I expected and `unixStamp` can be replaced by `time`.

ZeroASIC:
Alternatively you could just merge this and I could handle these changes in a different pull request.  This might be slightly broken though.

public/index.js
buck54321: Good call.

public/index.js
buck54321: The formatted time string from the server's `TimeDef` type has direct compatibility with `Date`, so I'd like to leave `block.time` and `block.unixStamp` as separate representations. It doesn't add any appreciable bloat and they both have their uses right now. Further modification of time types would be far-reaching and outside the scope of this PR, but you could roll your own. I would suggest renaming `block.time` to `block.timeDef` or similar to add some clarity. 

There were also some big changes in the time types in #791 not long ago, so there seems to be a tug of war forming and I'm not picking sides yet.

public/js/helpers/humanize_helper.js
buck54321: It seems like this outdated argument name is a holdover from [here](https://github.com/decred/dcrdata/commit/de314ca52fec01026ae08dfef44a56ed838f4909). It should probably be changed to something more accurate for the data type being passed. Maybe something like `timeDef` to mirror the type name server-side?

public/index.js
ZeroASIC: > 
> There were also some big changes in the time types in #791 not long ago, so there seems to be a tug of war forming and I'm not taking a side yet.

Yeah, I'm trying to decide how to resolve it.  As long as you calculate the unix timestamp on this line, then most of the times should kick back in.

explorer/templates.go
```diff
@@ -315,7 +315,7 @@ func makeTemplateFuncMap(params *chaincfg.Params) template.FuncMap {
 		"toLowerCase": func(a string) string {
 			return strings.ToLower(a)
 		},
-		"fetchRowLinkURL": func(groupingStr string, start, end types.TimeDef) string {
+		"fetchRowLinkURL": func(groupingStr string, start, end time.Time) string {
```
buck54321: Note that this is a current bug on master that is preventing any of the time-based block pages (/weeks, /months, etc.) from loading.

ZeroASIC:
I've been testing this up date this afternoon and I think it fixes #875 too.  I don't see anymore negative times.

public/js/controllers/blocklist_controller.js
```diff
@@ -28,4 +35,70 @@ export default class extends Controller {
...
+          let link = document.createElement('a')
+          link.href = `/block/${block.height}`
+          link.textContent = block.height
+          link.classList.add(tr.dataset.linkClass)
+          newTd.appendChild(link)
+          break
+        case 'txs':
+          newTd.textContent = block.tx
+          break
+        case 'voters':
```
gozart1: Would be ideal to have just one label per datum. We already have an annoying discrepancy between the field names used in the go templates and the json api. So maybe these data attributes can stick to one of those field names?

Eg here `Voters` `votes` and `voters` are all labels used to represent the same datum.
If naming was more unified, then many cases of this switch could be handling via a default case of `newTd.textContent = block[dataType]`.

buck54321: That does clean it up a bit. Thank you.

chappjc:
tACK.  @buck54321 anything required to address comments by @ZeroASIC?

ZeroASIC:
> tACK. @buck54321 anything required to address comments by @ZeroASIC?

I think I'm going to make those changes on my own.  They can be marked as resolved and this can be merged in.

buck54321:
That sounds good. I addressed a couple of things that may have also have resolved issues that @ZeroASIC was seeing.

PR #905, opened 2018-12-26
db: GetHeight returns int64 and error

chappjc:
Modify the interfaces for both lite and full DBs so that
GetHeight returns (int64, error).
err == sql.ErrNoRows is expected with a fresh sync,
so do not log it.

### DIFF
--- a/explorer/explorerroutes.go
+++ b/explorer/explorerroutes.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2018, The Decred developers
+// Copyright (c) 2018-2019, The Decred developers
 // Copyright (c) 2017, The dcrdata developers
 // See LICENSE for details.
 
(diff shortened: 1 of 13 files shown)


explorer/explorerroutes.go
```diff
@@ -384,8 +399,10 @@ func (exp *explorerUI) StakeDiffWindows(w http.ResponseWriter, r *http.Request)
 		return
 	}
 	if err != nil {
-		log.Errorf("The specified windows are invalid. offset=%d&rows=%d: error: %v ", offsetWindow, rows, err)
-		exp.StatusPage(w, defaultErrorCode, "The specified windows could not found", "", ExpStatusNotFound)
+		log.Errorf("The specified windows are invalid. offset=%d&rows=%d: "+
+			"error: %v ", offsetWindow, rows, err)
+		exp.StatusPage(w, defaultErrorCode,
+			"The specified windows could not found", "", ExpStatusNotFound)
```
buck54321: be

PR #911, opened 2019-01-03
views/home: remove extra white space

chappjc:
This removes unwanted white space in the `"decimalFormat"` html template and in the formatting of the `NEXT TICKET PRICE ESTIMATE` values on the home page.

Before:

image

After:

image

This may have snuck in with https://github.com/decred/dcrdata/pull/893.  The amount selection fix in PR #893 is preserved.

### DIFF
--- a/views/extras.tmpl
+++ b/views/extras.tmpl
@@ -151,21 +151,16 @@
 </footer>
 {{end}}
 
-{{define "decimalParts"}}
-  <div class="decimal-parts d-inline-block">
-    {{if eq (len .) 4}}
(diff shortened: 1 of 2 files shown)


buck54321:
I noticed one other space that might have snuck in earlier. 

image

Otherwise, this tests well on all relevant pages.

chappjc:
Good catch.  I'll put a fix for that in here too.

chappjc:
@buck54321 Turns out that's just a regular space. I replaced it with a `&thinsp;`, but FF does not seem to change space widths correctly for some reason.   It look correct in Chrome though.

PR #916, opened 2019-01-07
Stylesheet cleanup

gozart1:
* customize base theme via bootstrap _.variables.scss when possible
* remove deprecated main.css
* stop importing unused dropdown component styles
* apply border-radius to progress bars via bootstrap "rounded" utility class


chappjc:
tACK

PR #918, opened 2019-01-07
dcrpg: Add block deletion/purge queries and functions.

chappjc:
Purging a block and all associated data from the DB involves removing
rows from several tables. This adds db/dcrpg/internal/rewind.go, which
defines the query strings to delete rows based on block hash, and
db/dcrpg/rewind.go, which defines the functions to execute the queries.
The added `DeleteBlockData` function performs the queries in the correct
order: vins, vouts, addresses, transactions, tickets, votes, misses,
blocks, and finally block_chain.

`DeleteBlockData` returns an instance of the new `dbtypes.DeletionSummary`
type that contains counts of the number of rows removed from each table,
and optionally timings for each query.

This also adds the helper functions `DeleteBestBlock` and `DeleteBlocks`.

`DeleteBestBlock` (and `DeleteBlocks`) returns the best block height _after_
attempting data removal.
Define `dcrpg.NotOneRowErrMsg`.
Do not return error from `DeleteBlockData` if `deleteBlockFromChain` has an
error with `NotOneRowErrMsg`, but _do warn_.
If `DeleteBestBlock` returns an empty string hash, that is a full rewind,
so do not return, and do add the last `DeletionSummary` to the result.

This adds the config option `--purge-n-blocks`, which is intended to rewind
the DBs from the highest best block across the DBs down by N blocks.
However, it presently only purges blocks from PostgreSQL (aux. DB).
`(*ChainDB).PurgeBestBlocks` wraps `DeleteBlocks` and a new best block
query.  Call it from `main` immediately after creating `auxDB`.

All rewind functions use a `sql.Tx`, where the new transaction is created in
`DeleteBlockData` for these functions.  `DeleteBlockData` is atomic with a single
`Tx` Commit/Rollback.

Define the `SqlExecutor` interface that is implemented by both `sql.DB` and
`sql.Tx`.  Use `SqlExecutor` to make functions like `sqlExec`,
`SetSpendingForFundingOP`, `UpdateBlockNext`, etc. more versatile.

Add `UpdateBlockNextByNextHash` to set `next_hash` in a row of the
`block_chain` table where `next_hash` is currently set to a given value.

**TODO**: Also purge blocks from `baseDB`. The `StakeDatabase` it manages is
already rewound automatically as this was needed for `auxDB`, but the SQLite
tables are unmodified.

---

e.g. with `--purge-n-blocks=666`:

```
2019-01-07 18:18:58.972 [INF] DATD: Sucessfully purged data for 666 blocks from the aux. DB (new height = 307043):
	  666 Blocks purged
	24501 Vins purged
	39210 Vouts purged
	47530 Addresses purged
	11166 Transactions purged
	 3217 Tickets purged
	 3318 Votes purged
	   12 Misses purged
2019-01-07 18:18:58.972 [INF] DATD: Rewinding StakeDatabase from 307709 to 307043.
2019-01-07 18:18:58.972 [INF] SQLT: Rewinding from 307709 to 307600
2019-01-07 18:18:59.170 [INF] SQLT: Rewinding from 307600 to 307400
2019-01-07 18:18:59.547 [INF] SQLT: Rewinding from 307400 to 307200
2019-01-07 18:18:59.899 [INF] SQLT: Rewinding from 307200 to 307043
```

**NOTE:** These changes are intended to support a thorough rewind (removal
of all data for the N best blocks in the dcrpg DB). Functionality exists to
remove N blocks from the main chain tip, but this moves the blocks to a side
chain, and does not actually remove any entries from the tables.  Completely
purging block data is helpful in a number of situations including: testing, and
recovery from DB corruption.

### DIFF
--- a/config.go
+++ b/config.go
@@ -102,6 +102,8 @@ type config struct {
 	DBFileName         string `long:"dbfile" description:"SQLite DB file name (default is dcrdata.sqlt.db)." env:"DCRDATA_SQLITE_DB_FILE_NAME"`
 	AgendaDBFileName   string `long:"agendadbfile" description:"Agenda DB file name (default is agendas.db)." env:"DCRDATA_AGENDA_DB_FILE_NAME"`
 
+	PurgeNBestBlocks int `long:"purge-n-blocks" description:"Purge all data for the N best blocks, using the best block across all DBs if they are out of sync."`
+
 	FullMode       bool          `long:"pg" description:"Run in \"Full Mode\" mode,  enables postgresql support" env:"DCRDATA_ENABLE_FULL_MODE"`
 	PGDBName       string        `long:"pgdbname" description:"PostgreSQL DB name." env:"DCRDATA_PG_DB_NAME"`
 	PGUser         string        `long:"pguser" description:"PostgreSQL DB user." env:"DCRDATA_POSTGRES_USER"`
@@ -499,6 +501,11 @@ func loadConfig() (*config, error) {
 			return nil, fmt.Errorf("sync-status-limit should not be set to "+
 				"a value less than 2 or more than %d", maxSyncStatusLimit)
 		}
+	}
+
+	// Validate block purge options.
+	if cfg.PurgeNBestBlocks < 0 {
+		return nil, fmt.Errorf("purge-n-blocks must be non-negative")
 	}
 
 	// Set the host names and ports to the default if the user does not specify
--- a/db/dbtypes/types.go
+++ b/db/dbtypes/types.go
@@ -362,6 +362,45 @@ func (p *VinTxPropertyARRAY) Scan(src interface{}) error {
 	return nil
 }
 
+// DeletionSummary provides the number of rows removed from the tables when a
+// block is removed.
+type DeletionSummary struct {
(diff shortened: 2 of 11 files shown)


config.go
```diff
@@ -102,6 +102,8 @@ type config struct {
 	DBFileName         string `long:"dbfile" description:"SQLite DB file name (default is dcrdata.sqlt.db)." env:"DCRDATA_SQLITE_DB_FILE_NAME"`
 	AgendaDBFileName   string `long:"agendadbfile" description:"Agenda DB file name (default is agendas.db)." env:"DCRDATA_AGENDA_DB_FILE_NAME"`
 
+	PurgeNBestBlocks int `long:"purge-n-blocks" description:"Purge all data for the N best blocks, using the best block across all DBs if they are out of sync."`
```
chappjc: To be clear, this means that if PG is at 1000 and SQLite is at 950, and `purge-n-blocks` is 50, then blocks 951 to 1000 in PG will be purged and SQLite will be untouched.

chappjc: But this PR only implements rewind of PG.

chappjc:
Putting the WIP label on as this PR still needs: context cancellation of deletion queries, handling of missing blocks, and handling of other missing data resulting in `sql.ErrNoRows`.

PR #921, opened 2019-01-09
Consolidate network overview design for /home [WIP]

gozart1:
* Bring the network overview design from /nexthome into /home


gozart1:
Switching tacts on this a bit. Since the network overview on nexthome is much nicer looking, I'm taking a shot at consolidating that with /home.

So far, I've rebuilt the layout in /home, but without any of the css ( using only bootstrap at this point for layout & styling ). Plan is to get all the content and a fully responsive layout in place before authoring any custom styles. ( Eg. I think we can still do 3 up on mobile views instead of stacking by hiding some of the non essential stuff )

I've incorporated all the existing stats from /home except for ticket pool size. But I'm wondering if we should whittle it down a bit more. Perhaps stuff like "Tickets in Mempool" & "Votes in Mempool" can be moved to a "Mempool" section together with the "Latest Transactions" component.

This is still a WIP, but would love to get peoples thoughts.

screen shot 2019-01-10 at 2 55 17 am

Also, if we can pull this off, then the new visual blocks page can get the full page experience it deserves.

buck54321:
I think this looks great. Just spitballin' a couple of random thoughts to spur discussion here. 

1. While it seems natural to put the "General" section first, it also seems like some of the least important information for a typical user of the site (whoever that guy is). I don't claim to know what is the most important information, but if I had to, I'd guess in no particular order: the ticket price, the POS and POW rewards, mining difficulty, network hashrate. 

2. I like the flat design, but I always think it looks weird when there is no distinction between the navbar/footer and main page content, i.e. white on white with no border. Its occurring in the screenshot you shared where the mining box is blending in with the footer with no clear boundary. I suppose the simple solution would be to put a light border or shadow on the navbar and footer.

chappjc:
"Votes in Mempool" has little to no value on the home page, but "Tickets in Mempool" is nice IMO.  Besides being useful information for helping you set your expectations for how long it might take to mine your ticket purchases, it also sort of brings the staking system to life.

gozart1:
Thanks for the feedback. Agree that ticket info should be prioritized.

How changing that card's title to "Voting"? 
Seems in line with the nomenclature change from Stakepool to VSP (Voting Service Provider).

Also renamed the "General" card to "Distribution", but not sure that's the right word either. General just seems so non-descriptive to me.

I've added a subtle border to the footer and header to help delineate from main content.
It's good for performance to be sparing with css box and text shadows because they have a tendency to be expensive to render which leads to fans cranking and draining battery life.

Here's the latest from riffing on the layout a bit:

screen shot 2019-01-17 at 12 45 10 am

Mobile, night mode
screen shot 2019-01-17 at 12 46 41 am

views/home.tmpl
```diff
@@ -205,7 +237,7 @@
                         <div class="d-flex flex-table-row mempool-row">
                             <a class="hash truncate-hash keyboard-target" style="flex: 1 1 auto" href="/tx/{{.Hash}}" title="{{.Hash}}">{{.Hash}}</a>
                             <span style="flex: 0 0 65px" class="mono text-right ml-1">{{.Type}}</span>
-                            <span style="flex: 0 0 105px" class="mono text-right ml-1 fs13">
+                            <span style="flex: 0 0 105px" class="mono text-right ml-1 fs13 text-secondary">
```
buck54321: Your screenshot is showing a weird alignment on this "column".

buck54321:
The shorter progress bars look great, though the block reward progress bar seems to change size from desktop to mobile in your screenshots.  On the desktop, it looks like it would fit in a 2 x 2 grid, but in the mobile view it takes the full row width. 

After this is merged, I wonder if we could also apply a similar style to the data at the top of the address and transactions pages?

views/home.tmpl
gozart1: fixed, was messing around with the layout there, but ended up sticking with the inline-block approach

gozart1:
@buck54321 I Initially gave that "Next Block Reward Reduction" progress bar some more space in the mobile via title wraps and that leads to some slight misalignment. But on second though I think letting it wrap is fine and worth it for the compactness.

Beyond that have shuffled things around a bit more
- Removed "Votes in Mempool" since that info will be displayed in mempool pane via https://github.com/decred/dcrdata/pull/961 
- Added back "Ticket Pool Size"

screen shot 2019-01-29 at 1 47 46 am

screen shot 2019-01-29 at 1 52 06 am

gozart1:
This is ready to merge, Here's the recap of changes:
* remove network overview controller, template & css
* rework /home to incorporate network overview design from /nexthome
* increase container max-width to 1260px
* shorten progress bar height

views/home.tmpl
```diff
@@ -10,186 +10,222 @@
...
-                        </tr>
-                        {{if .DevFund}}
-                        <tr class="align-baseline">
-                            <td class="text-right pr-2 lh1rem"><a href="/address/{{.DevAddress}}?txntype=merged_debit">TREASURY</a></td>
-                            <td>
-                                <div class="mono lh1rem p03rem0 fs14-decimal fs18">
-                                    <span data-target="homepage.devFund" class="d-inline-block">{{template "decimalParts" (amountAsDecimalParts .DevFund true)}}</span
-                                    ><span class="pl-1 unit lh15rem">DCR</span>
+                            </div>
+                            <div class="col-6 col-sm-5 col-md-6 col-lg-4 mb-3 mb-sm-2 mb-md-3 mb-lg-3">
```
buck54321: This one is `col-sm-5`, but all the others are `col-sm-4`. Is that intentional?

views/extras.tmpl
```diff
@@ -150,16 +150,20 @@
...
+<div class="decimal-parts d-inline-block">
+    {{if eq (len .) 4}}
+        <span class="int">{{ index . 0 }}.{{ index . 1 }}</span
+        >{{if gt (len (index . 2)) 0}}<span class="decimal">{{ index . 2 }}</span
+        ><span class="decimal trailing-zeroes">{{ index . 3 }}</span>{{end}}
+    {{else}}
+        <span class="int">{{index . 0}}</span
+        >{{if gt (len (index . 1)) 0}}<span class="decimal dot">.</span
+        ><span class="decimal">{{index . 1 }}</span
+        ><span class="decimal trailing-zeroes">{{index . 2}}</span>{{end}}
```
buck54321: There's a small difference in spacing here between your screenshot and what I'm seeing in chrome.

image

buck54321:
This looks really good at all screen sizes. Progress bars and values seem to be updating as expected. Nice improvement all around.

views/home.tmpl
gozart1: Yes, since there's free space to the right of it at that breakpoint, giving it a bit more width prevents the label from wrapping, which in turn uses less vertical space in the layout.

gozart1:
Changes made. @chappjc good spot on the annual return, I've added it back.

Not sure about encoding the time range in the linking url, maybe the default range of 1 month should be handled by that charts controller itself when no range is passed?

screen shot 2019-01-29 at 11 34 57 am

Issue #913, opened 2019-01-05
stress testing the APIs and WebSocket hub


chappjc:
This complements the Insight socket.io tests in https://github.com/decred/dcrdata/issues/912.

To support many non-browser client connections, stress testing needs to be performed on the following components:

1. The dcrdata API HTTP endpoints
2. The Insight API HTTP endpoints
3. The Insight WebSocket system (socket.io)

This issue refers to 1. and 2., while https://github.com/decred/dcrdata/issues/912 refers to 3.

Since @buck54321 is interested in this, the issue is tentatively assigned to him. The first step should be to get a proof-of-concept stress test client running that the devs can hack on.

Some ideas from a Slack disussion:

> buck54321 [4:57 AM]
> A topic that seems to be on people's minds is stress testing. https://github.com/decred/dcrdata/issues/60. 
> <snip>
> 
> chappjc [5:00 AM]
> We'll likely need to look beyond go test to do proper stress testing
> There's an Apache tool I start looking into, but we'll can also make a go http client and client pool on our own
> A standalone tool to start N connections per second
> Similar for websocket testing, but N persistent connections
> That's just my thought on the matter though.
> 
> buck54321 [5:04 AM]
> Do you cycle all endpoints or focus on certain ones?
> 
> chappjc [5:05 AM]
> I figure that would be a parameter of the testing tool
> There could be a couple different predefined cycles
> Or certain endpoints could be tested
> Perhaps a json file could list endpoints to hit
> My thought was to gather statistic on response times and response codes

szpasztor:
I've done a couple stress tests using [locust](https://github.com/locustio/locust) for REST APIs. With locust, endpoints are configured in a python script and the tool can be run locally or even in a master/slave setup (I think I used one m4.large EC2 instance as the master and several r5.large or Heroku slaves depending on use case for real world tests). The larger tests were simulating ~400K simultaneous logged-in users interacting every couple seconds with a statically defined endpoint list. 

I'm not sure what test size is needed here, but the bottleneck I'd eventually run into beyond 500K users was RAM usage as locust creates individual threads/processes for each user - hence the memory optimized EC2 instances. 

Locust doesn't natively support websocket as far as I know, but there seem to be solutions to do it by changing the source.

I can take this issue if needed.

chappjc:
Partially resolved by https://github.com/decred/dcrdata/pull/924, although the websocket stress test is needed.

PR #924, opened 2019-01-10
Load Testing Utility

buck54321:
This is a simple program to implement load testing. It uses [vegeta](https://github.com/tsenart/vegeta) for sending and parsing requests. Test (a.k.a. attack) parameters are defined in a JSON-formatted definition file, and then referenced as a command line argument or INI config file setting.  Results are output into JSON-formatted result files, as well as a summary on the command line. 

This is meant as a starting point for implementing comprehensive load testing/benchmarking. Ideas for expanding these tools might include... 

1. The load testing tool could potentially pull pprof profiles via HTTP, allowing snapshots to be generated without the need to synchronize time on the server side. 

2. Charts.

3. Further attack customization.

4. Synchronization of distributed attackers for simulating very high traffic. Probably won't need this one for a while.


testutil/apiload/config.go
```diff
@@ -0,0 +1,78 @@
...
+	DefaultResultDirectory = "results"
+	DefaultProfilesPath    = "profiles.json"
+	DefaultServer          = "http://localhost:7777"
+)
+
+type config struct {
+	ConfigPath      string `short:"c" long:"config" description:"Path to a custom configuration file."`
+	ResultDirectory string `short:"d" long:"directory" description:"Directory for the result files."`
+	ProfilesPath    string `short:"p" long:"profiles" description:"Path to custom attack profiles."`
+	Server          string `short:"s" long:"server" description:"Target base URL, with protocol."`
```
chappjc: It is worth mentioning that the trailing slash should be ommited from the server, or we should remove it in validation.

testutil/apiload/config.go
```diff
@@ -0,0 +1,78 @@
...
+	Attack          string `short:"a" long:"attack" description:"The attack to perform. The attack must match an entry in the profile definitions."`
+	ListAttacks     bool   `short:"l" long:"list" description:"List available attacks from attack profiles."`
+	FormatResults   bool   `short:"f" long:"format" description:"Pretty print the JSON result file."`
+	CPUs            int    `long:"cpus" description:"Maximum number of processors to use. Defaults to all available."`
```
chappjc: I suggest to point to the `"frequency"` parameter in profiles.json to adjust attack rate.  Also, if there was a specific use case for doing `runtime.GOMAXPROCS`, it is worth pointing it out.

testutil/apiload/main.go
```diff
@@ -0,0 +1,257 @@
...
+)
+
+// Attack specifies a duration and a series of attackers, each with their own
+// frequency and targets.
+type Attack struct {
+	Duration  int         `json:"duration"`
+	Attackers []*Attacker `json:"attackers"`
+}
+
+// Frequency is 1/second and endpoints are relative to root
```
chappjc: Please start with `// Attacker`.

testutil/apiload/main.go
```diff
@@ -0,0 +1,257 @@
...
+		log.Fatalf("%v", err)
+	}
+
+	if Config.ListAttacks {
+		listAttacks(profiles)
+		return
+	}
+
+	if Config.Attack == "" {
+		fmt.Println("No attack specified. Specify an attack with -a.\n")
```
chappjc: Linters compain about the "extra" newline since Println already inserts one.  Perhaps use fmt.Print or fmt.Printf and have two `\n`.

testutil/apiload/main.go
```diff
@@ -0,0 +1,257 @@
...
+	if Config.Duration > 0 {
+		seconds = Config.Duration
+	}
+
+	// Perform the load test.
+	log.Printf("Beginning %s. Duration: %d seconds.", Config.Attack, seconds)
+	duration := time.Duration(seconds*1000) * time.Millisecond
+	for idx, attacker := range attackProfile.Attackers {
+		if len(attacker.Endpoints) == 0 {
+			log.Fatalf("Empty enpoint list encountered for %s", Config.Attack)
```
chappjc: typo: endpoint

testutil/apiload/main.go
```diff
@@ -0,0 +1,257 @@
...
+	var problematic []string
+	addProblematic := func(result *vegeta.Result) {
+		_, endpoint := findAttacker(result)
+		for _, url := range problematic {
+			if url == endpoint {
+				return
+			}
+		}
+		problematic = append(problematic, endpoint)
+		log.Printf("Error code %d from %s: %s",
```
chappjc: When code is 0, the connection failed, but this is not necessarily obvious.  I think it is worth special casing a log message for code 0.

testutil/apiload/sample-apiload.conf
```diff
@@ -0,0 +1,15 @@
+; The path to the dcrdata server being tested.
+; server=http://localhost:7777
```
chappjc: Note here that trailing slash must be omitted (or strip it in config validation).

chappjc:
Resolving issue https://github.com/decred/dcrdata/issues/913

testutil/apiload/config.go
buck54321: Definitely needs a frequency argument. One small caveat. The frequency in the attack definitions are per attacker. I'll create a config entry that overrides all attacker frequencies, but finer control of multi-attackers will require editing/authoring attack definitions.

PR #925, opened 2019-01-10
dcrsqlite: have StoreBlock set existing blocks to side chain first

chappjc:
When storing data for a main chain block, `(db *DB).StoreBlock` will first
set `is_mainchain=false` for any other block at this height. Previously,
this was only being done in `Store`, which was not used during a reorg.
However, `Store` uses `StoreBlockSummary`->`StoreBlock`, so now this will
always happen.
Note that his is usually an unnecessary query since most of the time
`StoreBlock` is used outside of a reorg context.  But this solution is very
simple.  If performance of this query is an issue, it can be done
only from the `ChainMonitor`'s reorg handler.

NOTE: this should be backported for 3.1.2.

### DIFF
--- a/db/dcrsqlite/chainmonitor.go
+++ b/db/dcrsqlite/chainmonitor.go
@@ -59,9 +59,10 @@ func (p *ChainMonitor) switchToSideChain(reorgData *txhelpers.ReorgData) (int32,
 			reorgData.OldChainHeight, commonAncestorHeight)
 	}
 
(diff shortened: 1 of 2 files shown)


chappjc:
This does not need to be backported to 3.1 since dcrsqlite in 3.1 did not have side chain tracking.  The block summaries table did not even have hash or is_mainchain in 3.1.

PR #935, opened 2019-01-14
Account for front & backend timezone offsets when computing age

gozart1:
* fix negative numbers showing instead of age when timezone offfsets misalign 
(Builds upon https://github.com/decred/dcrdata/pull/927)

### DIFF
--- a/explorer/explorerroutes.go
+++ b/explorer/explorerroutes.go
@@ -32,11 +32,12 @@ import (
 // explorerUI.commonData returns an initialized instance or CommonPageData,
 // which itself should be used to initialize page data template structs.
 type CommonPageData struct {
-	Tip           *types.WebBasicBlock
-	Version       string
-	ChainParams   *chaincfg.Params
-	BlockTimeUnix int64
-	DevAddress    string
+	Tip            *types.WebBasicBlock
+	Version        string
+	ChainParams    *chaincfg.Params
+	BlockTimeUnix  int64
+	TimezoneOffset int
+	DevAddress     string
 }
 
 // Status page strings
@@ -1879,6 +1880,8 @@ func (exp *explorerUI) commonData() *CommonPageData {
 	cd.Version = exp.Version
 	cd.ChainParams = exp.ChainParams
 	cd.BlockTimeUnix = int64(exp.ChainParams.TargetTimePerBlock.Seconds())
+	_, offset := time.Now().Zone()
+	cd.TimezoneOffset = offset
 	cd.DevAddress = exp.pageData.HomeInfo.DevAddress
 	var err error
 	cd.Tip, err = exp.blockData.GetTip()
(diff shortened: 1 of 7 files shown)


ZeroASIC:
I'm just testing it a little bit and it looks like it works.  I did notice that the Address page has a column `Time UTC` but in this case should be `Time (CST)` or the appropriate timezone, which is how the homepage works.  I think line 221 of extras.tmpl needs to be updated from:
`<th>Time UTC</th>`
To:
`<th>Time ({{timezone}})</th>`

chappjc:
Note that we can also change the postgresql data type. Presently it is just `timestamp`, which does not include a time zone by default, but this could be made into `timestamptz` (alias for `timestamp with time zone`).

And we can redefine the `String` method of `TimeDef` to include the timestamp: https://play.golang.org/p/_LJPzm_D0it

```go
package main

import (
	"fmt"
	"time"
)

type TimeDef struct {
	T time.Time
}

func (t TimeDef) String() string {
	return t.T.Format("2006-01-02 15:04:05 -0700 MST")
}

func main() {
	t := TimeDef{T: time.Now()}
	fmt.Printf("Right now it is %v", t)
}
```

Which outputs `Right now it is 2009-11-10 23:00:00 +0000 UTC`

ZeroASIC:
Including the timezone in TimeDef was one of my suggestions in #927 as a better long term fix.  I was trying to fix the negative times without making fundamental changes to the timestamps.  Updating TimeDef and forcing the Unix timestamps into UTC might be a good idea.

ZeroASIC:
Yes it does.
`var thisDate = new Date('2009-11-10 23:00:00 +0000 UTC');
console.log(thisDate);`

That prints out a valid time string to the console.
https://jsfiddle.net/d6p4kc9q/

ZeroASIC:
> Yes it does.
> `var thisDate = new Date('2009-11-10 23:00:00 +0000 UTC');`
> `console.log(thisDate);`
> 
> That prints out a valid time string to the console.
> https://jsfiddle.net/d6p4kc9q/1/
> or an alert
> https://jsfiddle.net/d6p4kc9q/

Since that only works in Chrome, maybe the library gozart1 mentioned in the other PR would help.  https://momentjs.com/timezone/

gozart1:
I don't like including the offsets in the time values themselves since that creates repetition of information and we should minimize that to preserve precious screen real estate. 

Although one thing to consider is if we should rewrite the date to local time in js enabled situations. 

There's endless ways we can deal with handling time. So letting the non-js experience guide us seems like a good tie breaker.

gozart1:
If we did go with including timezone in the value, then the ideal format would be [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601), ideally the flavor that matches https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString output

ZeroASIC:
As long as we update `humanize.formatTxDate(stamp, withTimezone)` to return the correct format, then it shouldn't matter too much how it's handled.  I'll always vote for less data being passed from the server to the client instead of more.

PR #938, opened 2019-01-15
dcrpg: remove several unused columns from blocks table.

chappjc:
Remove `extra_data`, `final_state`, `merkle_root`, and `stake_root` from `blocks`.
Remove the corresponding fields from `dbtypes.Block`.
Bump dcrpg tables version from 3.7.0 to 3.8.0.
Implement upgrade of table versions that removes the columns.
Add `deleteColumnsIfFound` to upgrades.go for future use.

### DIFF
--- a/db/dcrpg/internal/blockstmts.go
+++ b/db/dcrpg/internal/blockstmts.go
@@ -97,7 +97,6 @@ const (
 	SelectBlockHashByHeight = `SELECT hash FROM blocks WHERE height = $1 AND is_mainchain = true;`
 	SelectBlockHeightByHash = `SELECT height FROM blocks WHERE hash = $1;`
 
-	//RetrieveBestBlock          = `SELECT * FROM blocks ORDER BY height DESC LIMIT 0, 1;`
 	RetrieveBestBlockHeightAny = `SELECT id, hash, height FROM blocks
 		ORDER BY height DESC LIMIT 1;`
 	RetrieveBestBlockHeight = `SELECT id, hash, height FROM blocks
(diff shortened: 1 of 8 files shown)


chappjc:
@buck54321 Do you have a feeling for if these columns would every be useful (considering we can pull them one at a time via RPC from dcrd)?

buck54321:
I'm not familiar with the details of `final_state`, but the Merkle roots and `extra_data` can certainly be restricted to the verbose block API call. No reason to store them from what I can tell.

chappjc:
Thanks for the second opinion.  It doesn't really make sense to have these data in a SQL tables.

db/dcrpg/internal/blockstmts.go
```diff
@@ -101,7 +97,7 @@ const (
 	SelectBlockHashByHeight = `SELECT hash FROM blocks WHERE height = $1 AND is_mainchain = true;`
 	SelectBlockHeightByHash = `SELECT height FROM blocks WHERE hash = $1;`
 
-	RetrieveBestBlock          = `SELECT * FROM blocks ORDER BY height DESC LIMIT 0, 1;`
+	//RetrieveBestBlock          = `SELECT * FROM blocks ORDER BY height DESC LIMIT 0, 1;`
```
buck54321: Any reason to keep this around?

chappjc: Nope.  I meant to remove that after verifying it was unused.  Thanks.

chappjc:
`VACUUM FULL blocks` took about 5 seconds, and worked without error.

```
2019-01-15 11:42:25.068 [DBG] PSQL: Table transactions: v3.7.0
2019-01-15 11:42:25.070 [INF] PSQL: Reclaiming disk space from removed columns...
2019-01-15 11:42:29.449 [INF] PSQL: Modified the vins table version to 3.8.0
```

PR #939, opened 2019-01-15
dcrsqlite: Only move blocks to side chain during reorg.

chappjc:
`StoreBlock` previously set all blocks at a given height as side chain
(`is_mainchain=false`) before adding the new block data, but this was
unnecessary unless the new block was being added as part of a reorg,
where there are actually such blocks that are being orphaned/demoted.
This change moves the `setHeightToSideChain` call from `StoreBlock` to
`switchToSideChain` where it belongs.

This resolves a significant performance regression with initial sync.

### DIFF
--- a/db/dcrsqlite/chainmonitor.go
+++ b/db/dcrsqlite/chainmonitor.go
@@ -59,12 +59,7 @@ func (p *ChainMonitor) switchToSideChain(reorgData *txhelpers.ReorgData) (int32,
 			reorgData.OldChainHeight, commonAncestorHeight)
 	}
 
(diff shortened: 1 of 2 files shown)


chappjc:
For reference,

```
[INF] SQLT: Rescan finished successfully at height 309782.
[DBG] SQLT: New best block (chain server):     309782
[DBG] SQLT: New best block (sqlite block DB):  309782
[DBG] SQLT: New best block (stakedb):          309782
[INF] SQLT: resyncDBWithPoolValue completed in 1h21m15.735684642s
...
[INF] PSQL: Avg. speed: 1124 tx/s, 4152 vout/s
```

For my reference:
**--> 1hr 21 minutes <--**

PR #940, opened 2019-01-15
api cache: Cache stake info too.

chappjc:
Add `StakeInfoExtended`, `height`, and `hash` to `CachedBlock`.
Create `(*APICache).StoreStakeInfo`.
Create `(*APICache).GetStakeInfo` and `(*APICache).GetStakeInfoByHash`, and
use them in `(*DB).RetrieveStakeInfoExtended[ByHash]`.
Do not enable API cache by default; enable it in `main` after sync.

### DIFF
--- a/api/types/apicache.go
+++ b/api/types/apicache.go
@@ -1,3 +1,4 @@
+// Copyright (c) 2018-2019, The Decred developers
 // Copyright (c) 2017, Jonathan Chappelow
 // See LICENSE for details.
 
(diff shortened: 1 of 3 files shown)


chappjc:
Warm cache:

```
$  ./apiload -a rotation --frequency 500
2019/01/15 08:46:07 Beginning rotation. Duration: 45 seconds.
Requests      [total, rate]            22500, 500.02
Duration      [total, attack, wait]    45.123965384s, 44.99801748s, 125.947904ms
Latencies     [mean, 50, 95, 99, max]  20.304529ms, 841.271µs, 100.052046ms, 203.023132ms, 258.288393ms
Bytes In      [total, mean]            1946746, 86.52
Bytes Out     [total, mean]            0, 0.00
Success       [ratio]                  100.00%
Status Codes  [code:count]             200:22500  
Error Set:
```

PR #941, opened 2019-01-15
handle chainwork errors

buck54321:
During chart data population, if invalid values of chainwork are found, the process loop will no longer end. Instead, invalid values are counted and skipped. Error message displayed if count > zero.


buck54321:
For testing, I used an random number generator to randomly change the chainwork to "0"  on a few hundred blocks in `dcprg.retrieveChainWork`. I'd like to see what it does on alpha, though.

chappjc:
I manually patched the one bad row on alpha yesterday.  I'll see about fudging the issue on my local workstation.

PR #942, opened 2019-01-16
link from orphaned block to sidechain page

buck54321:


public/scss/typography.scss
```diff
@@ -142,3 +142,7 @@ a.grayed {
   color: red;
   font-size: 14px;
 }
+
+.attention:hover {
```
buck54321: This is to override https://github.com/decred/dcrdata/blob/7e7f38e54bb129ca2626669d4b8af2a415d1e005/public/scss/themes.scss#L58, which caused the red link to turn blue on hover.

PR #945, opened 2019-01-17
add merged and merged credit views

### LINKED ISSUE #742: Address view merged tx mode can support multiple type filters
Currently we have a Merged Debit view. Merging all inputs for a txid can apply to all filter types. Rather than just having a Merged view for debits, merging could be toggled on separately for all filter options, All, Credits, Debits.

buck54321:
Makes a few changes to the  `AddressInfo` type. `TxCount` is an `int` instead of a getter, and removed the `NumMergedSpent` member. Instead, merged type rows are only counted on relevant requests, instead of
every time (see `RetrieveAddressSpentUnspent` changes). For simple debit and credit views, this provides a substantial performance boost.

Changes the response from `explorer.AddressTable` to a JSON object, with a `tx_count` and `html` field. The count is needed since the merged rows are not being counted every time anymore. This is possible because if there are credits and debits available, there are merged rows available, an vice versa.

Resolves #742

### DIFF
--- a/db/dcrpg/queries.go
+++ b/db/dcrpg/queries.go
@@ -41,6 +41,15 @@ const (
 	NotOneRowErrMsg = "failed to update exactly 1 row"
 )
 
+const (
+	creditDebitQuery = iota
+	creditQuery
+	debitQuery
+	mergedCreditQuery
+	mergedDebitQuery
+	mergedQuery
+)
+
 // Maintenance functions
 
 // closeRows closes the input sql.Rows, logging any error.
@@ -1099,7 +1108,7 @@ func retrieveAddressTxsCount(ctx context.Context, db *sql.DB, address, interval 
 // for the given address, the total amounts spent and unspent, and the the
 // number of distinct spending transactions.
 func RetrieveAddressSpentUnspent(ctx context.Context, db *sql.DB, address string) (numSpent, numUnspent,
-	amtSpent, amtUnspent, numMergedSpent int64, err error) {
+	amtSpent, amtUnspent int64, err error) {
 	// The sql.Tx does not have a timeout, as the individial queries will.
 	var dbtx *sql.Tx
 	dbtx, err = db.BeginTx(context.Background(), &sql.TxOptions{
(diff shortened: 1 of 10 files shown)


db/dcrpg/queries.go
```diff
@@ -1347,37 +1395,74 @@ func retrieveAddressIoCsv(ctx context.Context, db *sql.DB, address string) (csvR
...
+			err = rows.Scan(&addr.TxHash, &addr.ValidMainChain, &blockTime.T,
+				&value, &addr.MergedCount)
+		case mergedQuery:
+			err = rows.Scan(&addr.TxHash, &addr.ValidMainChain, &blockTime.T,
+				&addr.AtomsCredit, &addr.AtomsDebit, &addr.MergedCount)
+			value = int64(addr.AtomsCredit - addr.AtomsDebit)
+			if value >= 0 {
+				addr.IsFunding = true
+			} else {
+				value = -1 * value
```
chappjc: Is the idea to treat this transaction as a "net" spend/fund for the address?

buck54321: It is, though I definitely had some reservations about this approach. Separate in/out figures would be nice, but I definitely want a sum as well. The address page table is already short on page space for smaller screens, and three columns with 8-point decimal precision take up a lot of width.

public/js/controllers/address_controller.js
```diff
@@ -185,11 +185,8 @@ export default class extends Controller {
     var cdata = ctrl.data
     ctrl.dcrAddress = cdata.get('dcraddress')
     ctrl.paginationParams = {
-      'offset': parseInt(cdata.get('offset')),
-      'all': parseInt(cdata.get('fundingCount')) + parseInt(cdata.get('spendingCount')),
-      'credit': parseInt(cdata.get('fundingCount')),
-      'debit': parseInt(cdata.get('spendingCount')),
-      'merged_debit': parseInt(cdata.get('mergedCount'))
+      offset: parseInt(cdata.get('offset')),
+      count: parseInt(cdata.get('txnCount')),
```
chappjc: Trailing comma.

views/extras.tmpl
```diff
@@ -201,19 +201,24 @@
...
-          {{else}}
-              <th class="text-right">Credit DCR</th>
-          {{end}}
-          {{if eq $txType "merged_debit"}}
+              <th class="text-center">Debit DCR</th>
+          {{else if eq $txType "merged_credit" }}
+              <th># of Outputs(s)</th>
+              <th class="text-center">Credit DCR</th>
+          {{else if eq $txType "merged" }}
+              <th>I/O&nbsp;Count</th>
```
chappjc: I suggest we put a `title` element on this to explain this is the number of inputs + outputs for this transaction that correspond to the address.

PR #950, opened 2019-01-21
dcrsqlite: provide advice for recovery on unknown ticket error

chappjc:
In `(*WiredDB).resyncDB`, recognize the unknown ticket spent in block
error, parse the ticket tx hash, get the block height of the ticket, and
append `purge-n-blocks` recovery instructions to the error message.

Add `(*WiredDB).supplementUnknownTicketError` to perform this error
message check/supplement.

Also clean up waitForSync.

### DIFF
--- a/db/dcrsqlite/dcrsqlite_test.go
+++ b/db/dcrsqlite/dcrsqlite_test.go
@@ -4,10 +4,54 @@ package dcrsqlite
 This file contains package-related test-setup utils
 */
 import (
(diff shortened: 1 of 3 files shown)


chappjc:
This is up on alpha.dcrdata.org (not that there's anything to see).

For anyone who runs into the "unknown ticket" error, *please* run with this PR to verify the message is sensible, and that the suggested recovery switches do the trick.

Issue #575, opened 2018-07-27
unknown ticket ... spent in block


JFixby:
This error randomly occurs killing the `dcrdata`.

Log:

```
...
2018-07-27 12:13:18.239 [INF] SQLT: resyncDBWithPoolValue() completed in 1m59.2488589s
2018-07-27 12:13:18.239 [INF] DATD: SQLite sync ended at height 4095
2018-07-27 12:13:18.239 [INF] DATD: PostgreSQL sync ended at height -1
2018-07-27 12:13:18.239 [ERR] DATD: dcrsqlite.SyncDBAsync failed at height 4095.
2018-07-27 12:13:18.239 [INF] EXPR: Stopping mempool monitor
2018-07-27 12:13:18.239 [INF] EXPR: Stopping websocket hub.
2018-07-27 12:13:18.239 [ERR] EXPR: Do not send on stopPing channel, only close it.
2018-07-27 12:13:18.771 [INF] DATD: Closing connection to dcrd.
2018-07-27 12:13:18.771 [INF] DATD: Bye!
2018-07-27 12:13:19.032 [ERR] DATD: unknown ticket a10547a4bcb59914e85d747a1acb971f2f989b9a0600ec5c4d60d6a58a679dc8 spent in block
```

chappjc:
There are some easy ways to accidentally misconfigure dcrdata (or rebuilddb2) to get that error:

These are:

1. the data directory (`datadir`) has data for the wrong network
2. you forgot `--testnet` and it's using mainnet params for a testnet `datadir`
3. you are connecting to a testnet/mainnet node with with the (opposite) mainnet/testnet switch

I don't think there is an unknown bug, but rather a need to check things to prevent the above.

chappjc:
Code from @JFixby: https://github.com/decred/dcrdata/compare/master...JFixby:utsib

JFixby:
Happening here: https://github.com/decred/dcrdata/blob/master/db/dcrsqlite/sync.go#L88

JFixby:
Test reproducing the issue: https://github.com/decred/dcrdata/commit/b750e60b3cca688ae5ec9efe4673147ecf179c34

JFixby:
The test reproduces behavior of the `stakedb.go/ConnectBlock()` function: 
https://github.com/decred/dcrdata/blob/2614a4e3c0e65120f3746672e8cbfe6d15469d5d/stakedb/stakedb.go#L481
at
[...]

PR #952, opened 2019-01-23
Update dcrutil module dependency for Block cache fix.

chappjc:
Update to the latest master of dcrutil to get the Block bytes cache fix.


chappjc:
Tests pass, and I've been running this for a few hours.  Going to merge this now since this has the potential to mitigate some issues web seen. Ref: https://github.com/decred/dcrdata/issues/575

PR #954, opened 2019-01-25
[travis] remove cache

dajohi:


run_tests.sh
```diff
@@ -1,4 +1,13 @@
 #!/usr/bin/env bash
+
+# usage:
+# ./run_tests.sh                         # local, go 1.11
```
chappjc: Only "local" works for dcrdata at present.  Is that going to change?

run_tests.sh
```diff
@@ -10,77 +19,58 @@ set -ex
...
   env GORACE='halt_on_error=1' go test -v -race ./...
-  if [ $? != 0 ]; then
-    echo 'go tests failed'
-    exit 1
-  fi
+
+  # check linters
+  golangci-lint run --deadline=10m --disable-all --enable govet --enable staticcheck \
+    --enable gosimple --enable unconvert --enable ineffassign --enable structcheck \
+    --enable goimports --enable misspell
```
chappjc: What was the error with `unparam`?  This PR works on my laptop even when I add it back in here.

.travis.yml
```diff
@@ -21,4 +21,4 @@ cache:
     - ~/.cache
 
 script:
-  - ./run_tests.sh $GOVERSION
+  - ./run_tests.sh docker
```
chappjc: Perhaps it's my laptop, but I can only get ./run_tests.sh to work with `./run_tests.sh 1.11`, not `./run_tests.sh docker` or `./run_tests.sh podman`.

run_tests.sh
chappjc: The unparam version must be newer, and it finds legit unparam candidates.  Will address in a future PR.

PR #958, opened 2019-01-26
dcrpg/insight: fix get block summary by time

chappjc:
This function was still querying with `int64` times instead of `time.Time`
values, which is necessary since the table column became a `TIMSTAMP`.

Note: This is not a backport candidate for 3.1 since 3.1 used int64 for time columns.


chappjc:
The right blocks get returned.  However I noticed that the date formats are not in UTC (ugh), so that will be in a issue to address for the future.

PR #961, opened 2019-01-28
Add live data summaries to the homepage

buck54321:
Add live data summaries to the homepage

This is meant to complement the work being done in PR #921 and be a little experimental, so don't hold back on the criticism. 
    
Adds some additional data summary and display styles to the homepage. Introduces the `threeSigFigs` number formatting functions in both golang template functions and javascript `humanize` helpers, which allows short but precise representations of values from 1e-8 up, using k,M,B prefixes.

In the blocks section, the values of the most recent "proposed" block are displayed prominently, along with an accounting of the current votes for that block. In the mempool section, the count and total DCR for "likely" transactions are displayed, along with a small graphic that illustrates their proportions in the mempool total. "likely" transactions do not include votes for any but the most recently proposed block, or for duplicate votes that can pop up when a stakeholder votes multiple times with a different agenda choice. The "likely" total will count more than 20 new tickets, though.

All new displayed values are shown identically with or without JS enabled in the browser. If JS is enabled, the values will update live. Instead of simply showing the age column when JS is enabled, the age column now replaces the timestamp column, which is redundant. The size column is sacrificially hidden at certain breakpoints so that the tables always display nicely. The vote type column headers are also shortened to a single letter. A new pure-CSS tooltip is used on the headers and elsewhere to add additional information when hovered on desktop or tapped on mobile. Values are shown with the new `threeSigFigs` format, which does mean that users would have to click into the /block or /tx page to get the full 8-point precision.

image

Javascript `Mempool` class keeps a very shallow copy of the mempool. No new websocket or API calls are necessary. `Mempool` has methods for retreiving some statistical data about the mempool and tracking votes for recent block(s).

image

image

image


explorer/explorer_test.go
```diff
@@ -26,3 +31,18 @@ func TestSimNetName(t *testing.T) {
 		t.Errorf(`Net name not "Simnet": %s`, netName)
 	}
 }
+
+// func TestThreeSigFigs(t *testing.T) {
```
chappjc: Looks like a really nice change in the screenshots.  Is it a WIP still though?  Maybe just this test?

views/home.tmpl
```diff
@@ -186,73 +186,185 @@
...
-                            <th class="jsonly">Age</th>
-                            <th>Time ({{timezone}})</th>
+                        <tr class="card-theader">
+                            <th class="text-left pl-3">Height</th>
+                            <th><div class="position-relative text-center" data-tooltip="regular transactions">R</div></th>
+                            <th><div class="position-relative text-center" data-tooltip="votes">V</div></th>
+                            <th><div class="position-relative text-center" data-tooltip="tickets">T</div></th>
+                            <th><div class="position-relative text-center" data-tooltip="revocations">X</div></th>
+                            <th><div class="position-relative text-center" data-tooltip="total DCR sent">DCR</div></th>
+                            <th class="d-none d-sm-table-cell d-md-none d-lg-table-cell"><div class="position-relative text-center" data-tooltip="total DCR sent">Size</div></th>
```
buck54321: oops. wrong tooltip.

views/home.tmpl
```diff
@@ -186,73 +186,185 @@
...
+                    <span class="h6 position-relative" data-tooltip="total DCR sent in this block">value</span><br>
+                    <span class="h4" data-target="homepage.blockTotal">{{threeSigFigs .Total}}</span><span class="h6"> DCR</span>
+                  </div>
                 </div>
 
-                <div class="d-flex align-items-center">
-                    <h4>Latest Blocks</h4> <a href="/blocks" class="pl-2 keyboard-target"><small>see more ...</small></a>
+                <div class="d-flex justify-content-between align-items-center px-2 mx-auto">
+                    <div class="text-left mt-2 pr-3 d-inline-block">
+                      <span class="h6 d-inline-block mb-2 position-relative" data-tooltip="stakeholder's votes for the proposed block.">votes</span><br>
```
chappjc: "stakeholder's votes for the proposed block"  -> my understanding of what is shown here is that these votes are the ones in mempool that are set to approve/disapprove the current best block (the "proposed" block).  This makes sense since until stakeholders approve the block, the transactions are not considered permanent although they are spendable.

chappjc: NVM, I just saw "proposed" and the block number above it. Very clear now.

views/home.tmpl
```diff
@@ -186,73 +186,185 @@
...
+                      <div class="d-inline-block p1 h4" data-target="homepage.blockVotes" data-hash="{{.Hash}}">
+                        {{range $index, $vote := $.BlockTally}}
+                          <span class="d-inline-block dcricon-{{if eq $vote 0}}reject{{else if eq $vote 1}}affirm{{else}}missing{{end}}"></span>
```
chappjc: How did you decide to handle votes with the same validator index (same ticket being spent), but with different vote bits set, as is often the case with misconfigured redundant wallets?

chappjc: I think each of `dcricon-{reject,affirm,missing}` deserve tooltips.  Especially in the missing case.  Perhaps for missing it could be something like "this selected ticket has yet to vote" or "waiting for this vote".  The affirm and reject options are much clearer, but I personally think it's worth reiterating whenever we get the chance that the votes is to validate/invalidate the regular transaction tree for this block.

views/home.tmpl
```diff
@@ -186,73 +186,185 @@
...
+                            <td class="text-center" data-type="votes">{{.Voters}}</td>
+                            <td class="text-center" data-type="tickets">{{.FreshStake}}</td>
+                            <td class="text-center" data-type="revocations">{{.Revocations}}</td>
+                            <td class="text-center" data-type="value">{{threeSigFigs .Total}}</td>
+                            <td class="text-center d-none d-sm-table-cell d-md-none d-lg-table-cell" data-type="value">{{.FormattedBytes}}</td>
+                            <td class="text-right pr-2" data-type="age" data-target="time.age" data-age="{{.BlockTime}}">{{.BlockTime}}</td>
                         </tr>
                         {{end}}
                     </tbody>
                 </table>
```
chappjc: I may be in the minority here, but I think that below both the "Latest Blocks" and "Mempool" tables it is worth having a "more" or "see all" kind of link, even though these links exist on the "Latest Blocks" and "Mempool" headers.

public/scss/home.scss
```diff
@@ -0,0 +1,73 @@
...
+.mempool-card {
+  background-color: #e3eaed;
+}
+
+.darkBG .mempool-card {
+  background-color: #1c1512;
+}
+
+.block-card {
+  background-color: #efece8;
```
chappjc: Not sure I love this peach/rust look to the Latest Blocks card.  I don't feel strongly about it though.

explorer/templates.go
```diff
@@ -194,6 +194,50 @@ func amountAsDecimalPartsTrimmed(v, numPlaces int64, useCommas bool) []string {
...
+		return fmt.Sprintf("%.1fM", math.Round(v/1e5)/10)
+	} else if v >= 1e6 {
+		return fmt.Sprintf("%.2fM", math.Round(v/1e4)/1e2)
+	} else if v >= 1e5 {
+		return fmt.Sprintf("%dk", int(math.Round(v/1e3)))
+	} else if v >= 1e4 {
+		return fmt.Sprintf("%.1fk", math.Round(v/100)/10)
+	} else if v >= 1e3 {
+		return fmt.Sprintf("%.2fk", math.Round(v/10)/1e2)
+	} else if v >= 100 {
```
chappjc: All the cases from here to the end can be covered by the magic of `"g"` where precision applies to the count of all digits, not just the ones to the right of the decimal:

```go
// either of:
strconv.FormatFloat(v, 'g', 3, 64)
fmt.Sprintf("%.3g", v)
```

Note that these functions round rather than truncate so the result should be as intended. https://play.golang.org/p/bgLV1WXNaxZ

explorer/types/explorertypes.go
```diff
@@ -529,6 +550,80 @@ type VotingInfo struct {
...
+
+// Affirmations counts the number of selected ticket holders who have voted
+// in favor of the block for the given hash.
+func (tally *VoteTally) Affirmations() (c int) {
+	for _, affirmed := range tally.Marks {
+		if affirmed {
+			c++
+		}
+	}
+	return c
```
chappjc: Just `return` should be good here.

explorer/types/explorertypes.go
```diff
@@ -529,6 +550,80 @@ type VotingInfo struct {
...
+}
+
+// VoteCount is the number of votes received.
+func (tally *VoteTally) VoteCount() (c int) {
```
chappjc: Might as well be `VoteCount() int`.

mempool/collector.go
```diff
@@ -300,11 +300,15 @@ func ParseTxns(txs []exptypes.MempoolTx, params *chaincfg.Params, lastBlock *Blo
 	revs := make([]exptypes.MempoolTx, 0)
 	regular := make([]exptypes.MempoolTx, 0)
 
-	var totalOut float64
+	var regularTotal, ticketTotal, voteTotal, revTotal float64
+	var likely bool
```
chappjc: Seems like the meaning is "likely to be mined".  Is that the intended semantics?

mempool/collector.go
```diff
@@ -300,11 +300,15 @@ func ParseTxns(txs []exptypes.MempoolTx, params *chaincfg.Params, lastBlock *Blo
...
+
+	var totalOut, likelyTotal float64
 	var totalSize int32
 	votingInfo := exptypes.VotingInfo{
```
chappjc: Seems like a `NewVotingInfo` c'tor would be good to have to ensure setting the max votes correctly and making the maps.  BTW, it could set the initial capacity of the maps to params.TicketsPerBlock, but I have a feeling the initial capacity of a default map is above that anyway.

public/js/helpers/mempool_helper.js
```diff
@@ -0,0 +1,159 @@
...
+  isQuestionableVote (tx) {
+    if (tx.Type !== 'Vote') return false
+    if (!tx.vote_info.last_block) return true
+    for (var i = 0; i < this.mempool.length; i++) {
+      let v = this.mempool[i]
+      if (v.type === 'Vote' && ticketSpent(v) === ticketSpent(tx)) return true
+    }
+  }
+
+  votes (hash) {
```
chappjc: Would it be fair to describe this function as `tallyVotesOnBlock`?

public/scss/utils.scss
```diff
@@ -163,3 +163,32 @@
   width: 1%;
   white-space: nowrap;
 }
+
+[data-tooltip]:hover::before {
```
chappjc: I love these `data-tooltip`s!

mempool/collector.go
buck54321: More like "likely available to be mined" because it doesn't ever exclude tickets regardless of their count. I'm thinking `likely` -> `mineable` or `likelyMineable`

explorer/templates.go
buck54321: There seems to be some important differences. `'%g` will switch to scientific notation if is uses fewer characters. Also, `%g` removes trailing zeros, which can be undesirable.

mempool/collector.go
chappjc: Yeah, either way is good.

explorer/templates.go
chappjc: Good point about removing trailing zeros.  However, for the numbers below 100, and sig digits, it won't ever go to scientific notation.

buck54321: Hmm. Maybe I'm not understanding. https://play.golang.org/p/HUNZjKF-Tui

chappjc: Ah, got it.  Any smaller than 0.0001 and it switches to sci notation.

buck54321:
Works in changes from #921, which included a full rework of the network info column. Switches the columns based on feedback on slack and matrix. Adds a little approved/rejected message when the necessary votes have been received.

Switches the semantics of the vote tallying a little so that a vote result (approved, rejected, not enough votes) is calculated and returned with the slice of votes. Also changed the enum for [no, yes, missing] from 0, 1, 2 to -1, 1, 0, which seemed appropriate for what they represent. Added contructor for `explorertypes.VotingInfo`.

Uses a simple two-tone grey color scheme for now. Changes the `text-success` boostrap class elements to a custom `text-green` because the former had poor contrast on lighter backgrounds and small font size.

image

image

PR #975, opened 2019-02-01
stop the floating point math for fee summation

chappjc:
Summing up fees as floating point numbers was inaccurate (duh). Instead,
add as `int64` and then get the decimal representation for display.

Two big offenders:
- `(*MempoolInfo)Trim` added fees in float.
- `ParseTxns` did lots of summation in float instead of int, leading to loss of precision all over.  This fixes it by summing in `int64` and then converting to coin/DCR at the end.

Bad:

```
"Fees": 0.042270350000000005
"Total": 8727.786064009997
```

Good:

```
"Fees": 0.04227035
"Total": 8727.78606401
```

### DIFF
--- a/explorer/types/explorertypes.go
+++ b/explorer/types/explorertypes.go
@@ -390,15 +390,17 @@ func (mpi *MempoolInfo) Trim() *TrimmedMempoolInfo {
 	mpi.RUnlock()
 
 	// Calculate total fees for all mempool transactions.
-	getTotalFee := func(txs []*TrimmedTxInfo) (total float64) {
+	getTotalFee := func(txs []*TrimmedTxInfo) dcrutil.Amount {
+		var sum dcrutil.Amount
 		for _, tx := range txs {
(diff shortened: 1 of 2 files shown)


explorer/types/explorertypes.go
```diff
@@ -390,15 +390,17 @@ func (mpi *MempoolInfo) Trim() *TrimmedMempoolInfo {
 	mpi.RUnlock()
 
 	// Calculate total fees for all mempool transactions.
-	getTotalFee := func(txs []*TrimmedTxInfo) (total float64) {
+	getTotalFee := func(txs []*TrimmedTxInfo) dcrutil.Amount {
+		var sum dcrutil.Amount
 		for _, tx := range txs {
-			total += tx.Fees
+			sum += tx.TxBasic.Fee
```
chappjc: NOTE: We can do this as a consequence of https://github.com/decred/dcrdata/pull/966

PR #980, opened 2019-02-02
Fix incorrect agenda page data display

dmigwi:
Fixes issues on the `/agenda/{agenda_id}` page.
- [x] Orders the second chart's data by block height.
    - "Vote Choices By Block" chart query used to order data by [blockheight without the need](https://github.com/decred/dcrdata/blob/c536907948a9a64358a70c084ac88060db5ec65a/db/dcrpg/internal/stakestmts.go#L379) of `ORDER BY block_height` sql clause 
- [x] Tidy up the agenda details table to only show user relevant information.
  - Removed` isAbstain` and `isNot` columns from the table on the said page.
  - Ensure `count` column holds the correct count per vote choice.
  - Hide `progress` column till when its entry values are greater than 0 since its 0 most times even when the agenda voting was completed long time ago. More explanation is found [here](https://github.com/decred/dcrdata/pull/980#discussion_r253580452)

**Before**:
screenshot from 2019-02-02 07-15-22

**After**:
screenshot from 2019-02-02 07-07-57


dmigwi:
Just discovered that an implementation to replace the temporary hard coded way to determine when an agenda id voting starts and ends is tightly coupled to this thus I will be adding the implementation rather than create a new branch.

dmigwi:
PR is ready for review now.

db/dcrpg/pgblockchain.go
```diff
@@ -1827,6 +1839,23 @@ func (pgb *ChainDB) AddressTransactionRawDetails(addr string, count, skip int64,
...
+// have not yet activated. Activated is the height when the agenda Rule Change
+// is implemented. Find more details on Rule Change Interval (RCI) here:
+// https://docs.decred.org/glossary/#rule-change-interval-rci.
+func updateVotesMilestones(blockChainInfo *dcrjson.GetBlockChainInfoResult) {
+	for agendaID, entry := range blockChainInfo.Deployments {
+		// Add missing agenda ID information.
+		if _, ok := VotingMilestones[agendaID]; !ok && entry.Since > 0 {
+			VotingMilestones[agendaID] = dbtypes.MileStone{
+				LockedIn:  entry.Since,
+				Activated: entry.Since + 8064,
```
chappjc: What if this is the future?

db/dcrpg/pgblockchain.go
```diff
@@ -1827,6 +1839,23 @@ func (pgb *ChainDB) AddressTransactionRawDetails(addr string, count, skip int64,
...
+			}
+		}
+	}
```
chappjc: What if the agenda item is actually "failed" rather than lockedin or active?

explorer/explorerroutes.go
```diff
@@ -1631,6 +1631,24 @@ func (exp *explorerUI) AgendaPage(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	yes, abstain, no, err := exp.explorerSource.AgendaCummulativeVoteChoices(agendaId)
+	if err != nil {
+		log.Errorf("fetching cummulative votes choices count failed: %v", err)
+	}
+
+	// Overrides the default count entry which is always zero especially when the
```
chappjc: "always" and "especially" are a confusing combination.  Can you expand on what exactly agendaInfo has set at this point (from GetAgendaInfo).

Also, please rename all `GetAgendaInfo` to just `AgendaInfo`.  We're trying to stop using a "get" prefix unless there's a good reason.

views/agenda.tmpl
```diff
@@ -70,10 +70,10 @@
                     <th>ID</th>
                     <th>Description</th>
                     <th>Bits</th>
-                    <th>IsAbstain</th>
-                    <th>IsNo</th>
                     <th>Count</th>
-                    <th>Progress</th>
+                    {{ if ne (index .Choices 0).Progress 0.0}}
+                        <th>Progress</th>
```
chappjc: Why not show progress 0?

dmigwi: Once a given agenda has been voted in, especially for the old agendas like lnfeatures, sdiffalgorithm and lnsupport the progress value remains to be zero while voting on the said agendas was done long time ago.
Since some data e.g progress is going to available only at certain statuses, we saw it wise with raedah to hide the progress value if its equal to 0.

explorer/explorerroutes.go
dmigwi: Noted.

db/dcrpg/pgblockchain.go
dmigwi: This makes sense. I will check if the status is "failed" before assigning that activated height value.

dmigwi:
I have update the PR description.

views/agenda.tmpl
chappjc: For a vote that is done, I agree progress 0 is confusing, but I'm thinking of agenda items that are pre-voting or just started voting.  I'm OK for now with hiding it like this when progress is 0, but I suspect it there may be states where showing progress 0 does make sense.

db/dcrpg/internal/stakestmts.go
```diff
@@ -358,25 +358,22 @@ const (
...
+	selectAgendasQuery = `
 			count(CASE WHEN agenda_vote_choice = $1 THEN 1 ELSE NULL END) AS yes,
 			count(CASE WHEN agenda_vote_choice = $2 THEN 1 ELSE NULL END) AS abstain,
 			count(CASE WHEN agenda_vote_choice = $3 THEN 1 ELSE NULL END) AS no,
 			count(*) AS total
-		 FROM agendas
+		FROM agendas
 		WHERE agenda_id = $4
-		  AND block_height <= (` + agendaLockinBlock + `)
-		GROUP BY block_height;`
```
chappjc: I can see why `GROUP BY block_height ORDER BY block_height` is needed, but it's strange that it ordered properly before.  Just by luck I suppose?

dmigwi: Still not sure why it worked before and can't work now.

db/dcrpg/pgblockchain.go
```diff
@@ -880,6 +902,17 @@ func (pgb *ChainDB) AgendaVotes(agendaID string, chartType int) (*dbtypes.Agenda
 	return avc, pgb.replaceCancelError(err)
 }
 
+// AgendaCummulativeVoteChoices fetches the total vote choices count for the
+// provided agenda.
+func (pgb *ChainDB) AgendaCummulativeVoteChoices(agendaID string) (yes,
```
chappjc: Cummulative -> Cumulative

views/agenda.tmpl
dmigwi: I could add a check for "defined" agenda state to address that concern.

db/dcrpg/pgblockchain.go
```diff
@@ -33,6 +34,27 @@ import (
...
+	// attain the minimum threshold set. Activation height is not set for such an
+	// agenda.
+	FailedAgendaState = "failed"
+
+	// PassedAgendaState is the agenda state when the agenda is considered to
+	// have passed after attaining the minimum set threshold. This agenda can
+	// it activation height set.
+	PassedAgendaState = "lockedin"
+
+	// ActivatedAgendaState is the agenda state set after 8064 blocks since when
```
chappjc: I suggest referring to `chaincfg.RuleChangeActivationInterval`(e.g. 8064 blocks = 2016 * 4 for 4 weeks on mainnet).

db/dcrpg/pgblockchain.go
```diff
@@ -231,6 +234,7 @@ type ChainDB struct {
 	InReorg            bool
 	tpUpdatePermission map[dbtypes.TimeBasedGrouping]*trylock.Mutex
 	utxoCache          utxoStore
+	chainInfo          *dbtypes.BlockChainData
```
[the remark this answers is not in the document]
chappjc: Actually, having `locked_in, activated, hard_forked` in the agendas table didn't sit well with me since there it implied we were tracking the agenda item state for each tx_hash or over time or something.

Now that I look back at the agendas table, it looks almost completely redundant with the votes table.  The votes table even has vote version and vote_bits.  So it seems that all the agendas table needs is a agenda id, agenda choice, and a link to the row id in the votes table.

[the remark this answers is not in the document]
dmigwi: Sure you make some valid points. I think we settled on a new table then to avoid too much data duplication like where a given vote tx has voted for more than one agenda. ie
```
test_dcrdata=# select tx_hash from votes where height = 173456;
                             tx_hash                              
------------------------------------------------------------------
 9e5715b7632b3bc0d12e6e301fa03ea4f6ae19a818ac6e2195b3f97dfac430a7
 f6cae37b23c6d920b506909e56a8cd8196e969cefa095acfa19bdafa53c5e26d
 f36334ccbc2a9ccb2aa2037ef0cccc49e692075b6980e5c0519b4ac36841b49f
 a8877f81af4ca83cce6dbe055a99ee9a1fdfcb88e49fa997a1ebfea172c4c279
 953a74c6c5d6f7acb207b6458da733c2c728de1e038605f250250fb19a8c09bc
(5 rows)

test_dcrdata=# select agenda_id, tx_hash from agendas where block_height = 173456;
   agenda_id    |                             tx_hash                              
----------------+------------------------------------------------------------------
 lnfeatures     | 9e5715b7632b3bc0d12e6e301fa03ea4f6ae19a818ac6e2195b3f97dfac430a7
 lnfeatures     | f6cae37b23c6d920b506909e56a8cd8196e969cefa095acfa19bdafa53c5e26d
 lnfeatures     | f36334ccbc2a9ccb2aa2037ef0cccc49e692075b6980e5c0519b4ac36841b49f
 sdiffalgorithm | a8877f81af4ca83cce6dbe055a99ee9a1fdfcb88e49fa997a1ebfea172c4c279
 lnsupport      | a8877f81af4ca83cce6dbe055a99ee9a1fdfcb88e49fa997a1ebfea172c4c279
 lnfeatures     | 953a74c6c5d6f7acb207b6458da733c2c728de1e038605f250250fb19a8c09bc
(6 rows)

```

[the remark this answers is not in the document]
dmigwi: `lnsupport` and `sdiffalgorithm` have the same vote version.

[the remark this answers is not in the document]
chappjc: Maybe a table called `agenda_votes`?  I imaging there could be a high level `agendas` table for someother purpose like describing the agendas themselves, their state/outcome, etc.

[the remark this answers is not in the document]
chappjc: Ah, if the `agenda_votes` table has an `agenda_id` column of an integer type, the `agendas` table could map the id to the text name (e.g. "lnfeatures") along with the other high level agenda stuff.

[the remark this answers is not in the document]
dmigwi: That high level table is in bolt db.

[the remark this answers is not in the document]
dmigwi: Its [here](https://github.com/decred/dcrdata/tree/master/db/agendadb) infact data on `/agendas` page is got from that high level table.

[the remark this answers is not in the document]
chappjc: So perhaps:

```
junk=# create table agenda_votes (id SERIAL PRIMARY KEY, vote_row_id INT8, agenda_id INT8, agenda_vote_choice INT2);
CREATE TABLE

junk=# create table agendas (id SERIAL PRIMARY KEY, name TEXT, locked_in BOOLEAN, activated BOOLEAN, hard_forked BOOLEAN, description TEXT);
CREATE TABLE

junk=# \d agenda_votes
                                   Table "public.agenda_votes"
       Column       |   Type   | Collation | Nullable |                 Default                  
--------------------+----------+-----------+----------+------------------------------------------
 id                 | integer  |           | not null | nextval('agenda_votes_id_seq'::regclass)
 vote_row_id        | bigint   |           |          | 
 agenda_id          | bigint   |           |          | 
 agenda_vote_choice | smallint |           |          | 
Indexes:
    "agenda_votes_pkey" PRIMARY KEY, btree (id)

junk=# \d agendas
                               Table "public.agenda"
   Column    |  Type   | Collation | Nullable |              Default               
-------------+---------+-----------+----------+------------------------------------
 id          | integer |           | not null | nextval('agenda_id_seq'::regclass)
 name        | text    |           |          | 
 locked_in   | boolean |           |          | 
 activated   | boolean |           |          | 
 hard_forked | boolean |           |          | 
 description | text    |           |          | 
Indexes:
    "agenda_pkey" PRIMARY KEY, btree (id)

```

dmigwi:
The agendas table reorganization is complete. Two new tables "agendas" and "agenda_votes" were created. The 3.9.0 db upgrade has been tested and completes in approximately 30 minutes (on my machine).

chappjc:
Nice.  Looking forward to checking it out.

db/dcrpg/queries.go
```diff
@@ -59,6 +59,11 @@ func closeRows(rows *sql.Rows) {
 	}
 }
 
+// storedAgendas holds the current state of agenda data already saved in the
+// db. This helps track changes in the lockedIn and activated heights when they
+// happen.
+var storedAgendas map[string]dbtypes.MileStone
```
chappjc: `storedAgendas` does not seem to have any purpose.

Also, queries.go is not the best spot for a package-level variable, and `InsertVotes` should probably not have the responsibility of managing this map.

db/dcrpg/queries.go
```diff
@@ -309,8 +314,9 @@ func InsertTickets(db *sql.DB, dbTxns []*dbtypes.Tx, txDbIDs []uint64, checked,
 //
 // Outputs are slices of DB row IDs for the votes and misses, and an error.
```
chappjc: What is `votesMilestones` used for, and is it modified?

db/dcrpg/queries.go
```diff
@@ -436,35 +490,47 @@ func InsertVotes(db *sql.DB, dbTxns []*dbtypes.Tx, _ /*txDbIDs*/ []uint64, fTx *
...
 
-			// THIS IS A TEMPORARY SOLUTION till activated, lockedIn and hardforked
-			// height values can be sent via an rpc method.
-			progress, ok := VotingMilestones[val.ID]
-			if ok {
-				lockedIn = (progress.LockedIn == tx.BlockHeight)
-				activated = (progress.Activated == tx.BlockHeight)
-				hardForked = (progress.HardForked == tx.BlockHeight)
+			if p.ID == 0 {
+				p.ID = s.ID
```
chappjc: From what I can tell, this is the only update to `p`.  But it's not clear why InsertVotes is modifying it or generally what is going on with `votesMilestones` or `storedAgendas`.

db/dcrpg/queries.go
```diff
@@ -535,6 +601,31 @@ func RetrieveMissedVotesInBlock(ctx context.Context, db *sql.DB, blockHash strin
...
+	if err != nil {
+		return nil, err
+	}
+
+	currentMilestones := make(map[string]dbtypes.MileStone)
+	defer closeRows(rows)
+
+	for rows.Next() {
+		var name string
+		m := dbtypes.MileStone{}
```
chappjc: When doing the zero value init, just do `var m dbtypes.MileStone`, and only when assigning a non-default/zero value, please do `m := `.

db/dcrpg/upgrades.go
```diff
@@ -623,6 +661,29 @@ func (pgb *ChainDB) handleUpgrades(client *rpcutils.BlockGate,
 		if err = pgb.ReindexAddressesBlockTime(); err != nil {
 			return false, fmt.Errorf("failed to reindex addresses table: %v", err)
 		}
+
+	case agendasTableReorgUpdate:
```
chappjc: "`Reorg`" is to suggestive of a chain reorg.  Perhaps something more along the lines of "agendasTableV2" or "agendasTablePruning".

db/dcrpg/upgrades.go
```diff
@@ -455,6 +482,16 @@ func (pgb *ChainDB) handleUpgrades(client *rpcutils.BlockGate,
 	case blocksRemoveRootColumns:
 		tableReady, err = deleteRootsColumns(pgb.db)
 		tableName, upgradeTypeStr = "blocks", "remove columns: extra_data, merkle_root, stake_root, final_state"
+	case agendasTableReorgUpdate:
+		i = 4000
```
chappjc: Please document any magic number choices.

db/dcrpg/upgrades.go
```diff
@@ -1271,6 +1420,19 @@ func deleteRootsColumns(db *sql.DB) (bool, error) {
 	return deleteColumnsIfFound(db, table, cols)
 }
 
+func deleteVoteMilestoneColumns(db *sql.DB) (bool, error) {
+	table := "agendas"
```
chappjc: There's no `table` variable in `addVotesBlockTimeColumn`, and for consistency I suggest eliminating the variable and passing a string literal.

main.go
```diff
@@ -510,6 +510,16 @@ func _main(ctx context.Context) error {
 		// and signal the websocket hub to send progress updates to clients.
 		barLoad = make(chan *dbtypes.ProgressBarLoad, 2)
 		explore.BeginSyncStatusUpdates(barLoad)
+
+		// Fetch the latest blockdata which needed to update the agendas db while
+		// db sync is in progress.
+		d, _, err := collector.Collect()
```
chappjc: `bci, err :=  baseDB.BlockchainInfo()` would be preferable here since `Collect` does lots of unnecessary stuff here.

db/dcrpg/pgblockchain.go
```diff
@@ -1827,6 +1847,68 @@ func (pgb *ChainDB) AddressTransactionRawDetails(addr string, count, skip int64,
...
+			Status:     dbtypes.AgendaStateFromStr(entry.Status),
+			StartTime:  time.Unix(int64(entry.StartTime), 0),
+			ExpireTime: time.Unix(int64(entry.ExpireTime), 0),
+		}
+
+		// state "defined" is not considered since voting hasn't started.
+		switch agendaInfo.Status {
+		case dbtypes.StartedAgendaState:
+			// Since the vote is in progress current best block height is set as
+			// the vote end.
```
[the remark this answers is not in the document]
dmigwi: Everytime there is a status change, data is updated in the agendas db.

db/dcrpg/queries.go
dmigwi: I was looking for a simple way to have the data stored in agendas table i.e `lockedIn`, `Hardforked` and `Activated` heights as up to date as possible without making too many db accesses everytime we are updating the agenda_votes table.
`votesMilestones` holds the updated blockchain info always while `storedAgendas` holds a mirror copy of what in the agendas db. If there is a difference between the two while blockchain sync is in progress, an update is made to agendas table. The update is made only when needed i.e. where there is a status change.

db/dcrpg/queries.go
dmigwi: This is has the `deployment` map found in the blockchain info map. It hold the updated version of the blockchain info.

chappjc: I mean please describe in the docs

db/dcrpg/queries.go
chappjc: "without making too many db accesses everytime we are updating the agenda_votes table" <- that clarifies it.  Please comment code gratuitously to explain what's going on.

db/dcrpg/queries.go
```diff
@@ -356,15 +362,63 @@ func InsertVotes(db *sql.DB, dbTxns []*dbtypes.Tx, _ /*txDbIDs*/ []uint64, fTx *
...
 		if errRoll := dbtx.Rollback(); errRoll != nil {
 			log.Errorf("Rollback failed: %v", errRoll)
 		}
 	}
 
+	// Attempt to update the storedAgenda map is help keep track of the
+	// lockedIn, hardForked and activated heights.
+	if len(storedAgendas) == 0 {
+		var id int64
+		storedAgendas, err = retrieveAllAgendas(db)
```
chappjc: All this needs commenting to express the intend.  This line would be like

```go
// Attempt to retrieve agendas from the database.
```

db/dcrpg/queries.go
```diff
@@ -356,15 +362,63 @@ func InsertVotes(db *sql.DB, dbTxns []*dbtypes.Tx, _ /*txDbIDs*/ []uint64, fTx *
...
+		if err != nil {
+			bail()
+			return nil, nil, nil, nil, nil,
+				fmt.Errorf("retrieveAllAgendas failed: : %v", err)
+		}
+
+		if storedAgendas == nil {
```
chappjc:
```go
// If there are no agendas in the database, create the `storedAgendas` cache from `votesMilestones.AgendaMileStones` (from getblockchaininfo RPC), then insert the agenda information into the database.
```

db/dcrpg/queries.go
```diff
@@ -356,15 +362,63 @@ func InsertVotes(db *sql.DB, dbTxns []*dbtypes.Tx, _ /*txDbIDs*/ []uint64, fTx *
...
+			storedAgendas = make(map[string]dbtypes.MileStone)
+		}
+
+		for name, d := range votesMilestones.AgendaMileStones {
+			m, ok := storedAgendas[name]
+			if !ok || (d.Status != m.Status) {
```
chappjc: docs for the status check are helpful (e.g. `if there is a status difference between the milestone info from RPC and the cached milestone data,...`).  Just writing the comment makes me wonder how m.Status would be anything but the empty string since we only got to this part of the code because `storedAgendas == nil` and we just initialized the map.

db/dcrpg/upgrades.go
chappjc: Is that correct for all chains?

dmigwi:
@chappjc Adding the `startheight` has already bore some fruits on current charts. check `sdiffalgorithm` chart:
**Before**
screenshot from 2019-01-31 19-17-03

**After**
screenshot from 2019-02-08 20-03-51

cc @raedah

dmigwi:
I have added `/api/agendas` endpoint on request by raedah.

screenshot from 2019-02-09 00-13-19

dmigwi:
https://github.com/decred/dcrdata/issues/994,

db/dcrpg/queries.go
```diff
@@ -308,9 +308,12 @@ func InsertTickets(db *sql.DB, dbTxns []*dbtypes.Tx, txDbIDs []uint64, checked,
 // subsequent cache lookups by other consumers will succeed.
 //
 // Outputs are slices of DB row IDs for the votes and misses, and an error.
+// votesMilestones holds upto date blockchain info deployment data needed
+// to update changes in agendas table.
```
chappjc: Sorry to be a stickler, but the reason I brought up this bit of docs is because the docs for InsertVotes didn't mention that it updated the agendas table too.  Note that it describes it's updates to votes and misses.

So, this added comment is good, but you've not yet described what's being updated in the agendas table or that InsertVotes is now responsible for keeping the agendas table current.

db/dcrpg/queries.go
chappjc: I phrased it this way because at this stage it's acceptable to have *no* agendas stored in the DB.  Part of what goes on below is to store missing agendas in the DB, and update the storedAgendas cache.  I feel strongly that the intent of the code needs to be documented more here.

db/dcrpg/queries.go
```diff
@@ -436,35 +486,51 @@ func InsertVotes(db *sql.DB, dbTxns []*dbtypes.Tx, _ /*txDbIDs*/ []uint64, fTx *
...
-			index, err := dbtypes.ChoiceIndexFromStr(val.Choice.Id)
-			if err != nil {
-				bail()
-				return nil, nil, nil, nil, nil, err
-			}
+			// As of here, storedAgendas should not be empty and
+			// votesMilestones.AgendaMileStones should have cached the latest
+			// blockchain deployment info. The change in status is detected as
+			// the change between respective agendas statuses stored in the two
+			// maps. Its then updated in storedAgendas and agendas table.
```
chappjc: "It's" because this is a contraction of "It is" rather than a possessive.

db/dcrpg/queries.go
```diff
@@ -1282,6 +1373,16 @@ func RetrieveAddressTxnsOrdered(ctx context.Context, db *sql.DB, addresses []str
 	return
 }
 
+// retrieveRCIWindowStartHeight helps in obtaining the RCI startheight for the
+// current active voting session when the agenda status is "started". By obtaining
+// the accurate startheight the votes cast in the previous voting sessions are ignored.
+func retrieveRCIWindowStartHeight(ctx context.Context, db *sql.DB,
+	starttime time.Time, chainRCI int64) (startheight int64, err error) {
```
chappjc: define `chainRCI` in the docs too.

db/dcrpg/upgrades.go
```diff
@@ -65,23 +68,23 @@ type VinVoutTypeUpdateData struct {
 	TxType     stake.TxType
 }
 
-// VotingMilestones defines the various milestones phases have taken place when
+// votingMilestones defines the various milestones phases have taken place when
 // the agendas have been up for voting. Only sdiffalgorithm, lnsupport and
 // lnfeatures agenda ids exist since appropriate voting mechanisms on the dcrd
 // level are not yet fully implemented.
-var VotingMilestones = map[string]dbtypes.MileStone{
+var votingMilestones = map[string]dbtypes.MileStone{
```
chappjc: This should work if we just declared `var votingMilestones = map[string]dbtypes.MileStone{}`.  I'm not a fan of hard coding data that we can and do request from the authoritative source.

db/dcrpg/upgrades.go
chappjc: I think this should be set to stakevalidationheight so it works on testnet also.

db/dcrpg/upgrades.go
```diff
@@ -602,10 +646,6 @@ func (pgb *ChainDB) handleUpgrades(client *rpcutils.BlockGate,
 			return false, fmt.Errorf("failed to index agendas table: %v", err)
 		}
 
-		log.Infof("Index the agendas table on Block Time...")
```
chappjc: Might as well remove the above `IndexAgendasTableOnAgendaID` too since the agendas table is just going to be dropped and recreated.

db/dcrpg/upgrades.go
```diff
@@ -1271,6 +1422,13 @@ func deleteRootsColumns(db *sql.DB) (bool, error) {
 	return deleteColumnsIfFound(db, table, cols)
 }
 
+func addVotesBlockTimeColumn(db *sql.DB) (bool, error) {
+	newColumns := []newColumn{
+		{"block_time", "TIMESTAMP", ""},
```
chappjc: `TIMESTAMPTZ`

db/dcrpg/pgblockchain.go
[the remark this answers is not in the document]
dmigwi: Noted. I will do away with it.

db/dcrpg/upgrades.go
dmigwi: `IndexAgendasTableOnAgendaID` is still in use to create the unique constraint for the agendas table.

chappjc: I meant this particular call to `IndexAgendasTableOnAgendaID`.  A subsequent upgrade just drops the agendas table, correct?  Then after that it is indexed.

dmigwi:
Please allow me to run a couple of more tests on this PR upgrade before you merge. Haven't done so after I completed the full db sync.

dmigwi:
@chappjc Just fixed a bug with the TableVersions regex expression and all is well now.
It supports the following version formats.
screenshot from 2019-02-14 05-20-19

db/dcrpg/tables.go
```diff
@@ -297,7 +298,15 @@ func TableVersions(db *sql.DB) map[string]TableVersion {
...
-			re := regexp.MustCompile(`^v(\d+)\.?(\d?)\.?(\d?)$`)
+
+			// This regex expression should detect the following versions format:
+			// v3
+			// v3.0
+			// v3.0.0
+			// v3.6.0
+			// v3.10.0
+			// v10.10.10
+			re := regexp.MustCompile(`^v(\d+)\.?(\d+)?\.?(\d+)?$`)
```
chappjc: Kind of thing that would be good for a unit test.  For another time though.

PR #982, opened 2019-02-04
Rework mempool page

buck54321:
Adjustments to the mempool page layout. Compare with [current master](https://alpha.dcrdata.org/mempool) and [this mockup](https://user-images.githubusercontent.com/33739891/39491479-8fbb9a38-4d5a-11e8-9860-12b2a65ef5cc.png) from #451.

image

image

Incorporates client-side mempool from #961 for live updates.

Moves `likelyMineable` totals and counts to their own explorer type. Creates a `mempoolDump` template to pass mempool stats to stimulus via data attributes.

WIP until at least the following are approved/discussed.

1. Vote table columns. More or less? Compare to mockup. We also have the vote icons now that might come in handy. Tables get a little tight at low screen width. With the current set of columns, I needed to set the vote table to allow scrolling at the smallest widths. 

2. I have left the entire hash in place, though that has some unfortunate consequences at smaller widths. 

3. This PR removes "HODL" watermark. I think its an improvement. Other might not. 

4. I did not use the blue color from the mockup for sections headers, as that color already designates links. Should we use it anyway, or a different color, or leave them.


lte13:
Nice! What other sites need fixing? Also the mobile version is prettyt broken.

buck54321:
@lte13 The `/block` and `/tx` pages will need similar work done. You could look at #451 to see where things are headed. 

> Also the mobile version is prettyt broken.

Are you talking about this PR? Screenshots and suggestions are certainly welcome.

chappjc:
The "HODL" watermark can certainly go.  It was a gag that was so subtle it didn't really bug anyone.  But still, it was a gag that ran it's course.  :)

Some ideas for the votes table on mobile:
 - The Voting On and Best Block can perhaps be merged.  It could stay called Voting On, but if it's the best block it could be like "312343 (best)".
- Perhaps Validator ID could be removed.  On mobile I'd argue that's fine.
- Since width is so limited, both Total DCR and Time in Mempool could go on mobile too.

buck54321:
Rebased and tweaked vote table for mobile view.

image

buck54321:
@chappjc I just saw that you were thinking of hiding "Total DCR", and I hid the "Size" column instead. I left "Validator ID" because it might help explain why there are more than 5 votes on the best block, which is common (different agenda choices from different VSP).

buck54321:
Switched hidden mobile votes table column from *Size* to *Total DCR*. Removed WIP.

chappjc:
Thanks.  Size is strangely useful because it gives you a hint if it is a regular solo, pool, or weird split ticket being spent.

gozart1:
Using the "https://getbootstrap.com/docs/4.2/content/tables/#responsive-tables" utility class will make the table horizontally scrollable when screen real estate is tight. Not ideal, but probably the least bad solution. Best to scope it to the breakpoint where its actually needed.

public/scss/mempool.scss
```diff
@@ -0,0 +1,74 @@
...
+}
+
+.pointer-bottom {
+  left: 50%;
+  bottom: -40px;
+  border-color: white transparent transparent transparent;
+  transform: translateX(-50%);
+}
+
+.mempool-tables {
```
gozart1: Instead of creating mempool specific table styles, is there a way we can keep this generic and author in https://github.com/decred/dcrdata/blob/master/public/scss/table.scss ?

Since there are a lot of tables throughout the explorer, we should try to maintain and perfect a core set of styles that we use everywhere instead of subtle differences with associated custom css between the views. 

There's probably even some opportunities to tune the bootstrap variables and get rid of some of the stuff we currently have layered on top of that in table.scss

buck54321: Agree. I'd like to consolidate the table styling in two steps though. I'll move the mempool table CSS to tables.scss and cards.scss here and rename things to be more general, but leave the scoping so that this PR is still just the mempool page. In a follow up PR, I'll hit the table styling and see if we can't get it down to just a few declarations, but would be site-wide.

buck54321:
I did use `table-responsive-md` back in https://github.com/decred/dcrdata/pull/982/commits/a15cfa490e145232d5f0e51ebf6ff8c96a049e90, and there was some discussion above of what columns might be able to be sacrificed to avoid scrolling altogether. Right now, it works down to 375px without a need for scrolling, but with the loss of a couple pieces of data, so it's definitely a trade-off. 

I don't have a strong opinion either way.

public/scss/mempool.scss
chappjc: The PR is good for me now. @gozart1?

gozart1: Yep, looking good.

PR #983, opened 2019-02-05
Purge orphaned tip blocks on startup

### LINKED ISSUE #601: startup should not assume DBs on main chain

buck54321:
Uses the existing purge-n-blocks loop. `rpcutils.OrphanedTipLength` compares a database to dcrd from the tip height down until a common mainchain ancestor is found, and returns the number of sidechain blocks to be purged.

Resolves #601

### DIFF
--- a/main.go
+++ b/main.go
@@ -286,6 +286,11 @@ func _main(ctx context.Context) error {
 		}
 	}
 
+	// Give a chance to abort a purge.
+	if shutdownRequested(ctx) {
+		return nil
+	}
+
 	if blocksToPurge > 0 {
 		// The number of blocks to purge for each DB is computed so that the DBs
 		// will end on the same height.
--- a/main.go
+++ b/main.go
@@ -255,8 +255,38 @@ func _main(ctx context.Context) error {
 		return
 	}
 
-	// Optionally purge best blocks according to config.
-	if cfg.PurgeNBestBlocks > 0 {
+	// Check for database tip blocks that have been orphaned. If any are found,
+	// purge blocks to get to a common ancestor. Only message when purging more
+	// than requested in the configuration settings.
+	blocksToPurge := cfg.PurgeNBestBlocks
+	_, _, baseHeight, auxHeight, err := Heights()
+	if err != nil {
+		return fmt.Errorf("Failed to get Heights for tip check: %v", err)
+	}
+
+	if baseHeight > -1 {
+		orphaned, err := rpcutils.OrphanedTipLength(ctx, dcrdClient, baseHeight, baseDB.DB.RetrieveBlockHash)
+		if err != nil {
+			return fmt.Errorf("Failed to compare tip blocks for the base DB: %v", err)
+		}
+		if int(orphaned) > blocksToPurge {
+			blocksToPurge = int(orphaned)
+			log.Infof("Orphaned tip detected on base DB. Purging %d blocks", blocksToPurge)
+		}
+	}
+
+	if usePG && auxHeight > -1 {
+		orphaned, err := rpcutils.OrphanedTipLength(ctx, dcrdClient, auxHeight, auxDB.BlockHash)
+		if err != nil {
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -255,8 +255,38 @@ func _main(ctx context.Context) error {
 		return
 	}
 
-	// Optionally purge best blocks according to config.
-	if cfg.PurgeNBestBlocks > 0 {
+	// Check for database tip blocks that have been orphaned. If any are found,
+	// purge blocks to get to a common ancestor. Only message when purging more
+	// than requested in the configuration settings.
+	blocksToPurge := cfg.PurgeNBestBlocks
```
chappjc: good solution.

rpcutils/rpcclient.go
```diff
@@ -448,6 +449,53 @@ func CommonAncestor(client *rpcclient.Client, hashA, hashB chainhash.Hash) (*cha
 	return &hashA, chainA, chainB, nil
 }
 
+// OrphanedTipLength finds a common ancestor by iterating block heights
+// backwards until a common block hash is found. Unlike CommonAncestor, an
+// orphaned DB tip whose corresponding block is not known to dcrd will not cause
+// cause an error. The number of blocks that have been orphaned is returned.
+// Realistically, this should rarely be anything but 0 or 1, but no limits are
+// placed here on the number of blocks checked.
+func OrphanedTipLength(ctx context.Context, client *rpcclient.Client,
```
chappjc: Trying to think how this could bet tested via go test.  One idea is to replace `*rpcclient.Client` with `BlockHashGetter` so you can stub out the client.  I think it would be worth testing all the edge cases by manipulating what the `BlockHashGetter` and the `hashFunc` return.

rpcutils/rpcclient.go
```diff
@@ -448,6 +449,53 @@ func CommonAncestor(client *rpcclient.Client, hashA, hashB chainhash.Hash) (*cha
...
+	commonHeight := tipHeight
+	var dbHash string
+	var err error
+	var dcrdHash *chainhash.Hash
+	for {
+		// Since there are no limits on the number of blocks scanned, allow
+		// cancellation for a clean exit.
+		select {
+		case <-ctx.Done():
+			return 0, nil
```
chappjc: If there is an interrupt, and this returns `0, nil` then I feel like there needs recognize this in main.  It could either be one of these:

```go
// Allow Ctrl-C to halt startup here.
	if shutdownRequested(ctx) {
return nil
```

or `OrphanedTipLength` could return -1 or a special non-nil error.

rpcutils/rpcclient.go
buck54321: Thanks for the hint here. I already had the test written, but with a `rpcclient.Client` with my local dcrd credentials, so I didn't include it.

rpcutils/rpcclient_test.go
```diff
@@ -114,3 +117,101 @@ func TestReverseStringSlice(t *testing.T) {
...
+	hashes = map[int64]string{
+		5: "something else",
+		4: "side block",
+		3: "blah",
+		2: "blue",
+		1: "flu",
+		0: "fly",
+	}
+
+	uncommon, err = OrphanedTipLength(ctx, client, 5, hashFunc)
```
chappjc: Just need to `_` out `uncommon`. (ineffassign)

ZeroASIC:
I just tested this on a testnet database that was on the wrong chain and stuck and it seems to have updated everything correctly.  It was stuck on block 122934 so I checked blocks 122924-122934 after updating and everything looks the same as on testnet.dcrdata.org.

PR #984, opened 2019-02-05
build.sh: use gnu parallel if available

chappjc:
If the GNU `parallel` command is available, use it for gzipping the static assets.

Also check if `7za` is installed and fall back on `gzip`.


chappjc:
Tested all four code paths (gzip, 7za-gzip, parallel+gzip, parallel+7za-gzip).

PR #985, opened 2019-02-05
dcrsqlite: add best block queries with no is_mainchain constraints

chappjc:
Before the sqlite table upgade is performed, it is still necessary to
query the best block height in the block and stake info tables. This
adds alternate queries that do not reference the `is_mainchain` column.
Note that the upgrade worked before 9a5780a3dd2c75570d7b3fdf6119aa3c5e82736d, which added the `is_mainchain` constraint to the queries.

Add `RetrieveHighestBlockHeight` and `RetrieveHighestStakeHeight` that
perform the query without `is_mainchain`, and add the wrappers
`getBlockSummaryHeightAnyChain` and `getStakeInfoHeightAnyChain`.

`NewDB` is modified to recognize `"no such column: is_"` errors, and
to fallback on the `"AnyChain"` functions to set the heights.

At the end of a successful table upgrade, query for the stake info best
block height using the block summary table join where `is_mainchain=true`.

### DIFF
--- a/db/dcrsqlite/sqlite.go
+++ b/db/dcrsqlite/sqlite.go
@@ -88,6 +88,7 @@ type DB struct {
 	getBlockHashSQL, getBlockHeightSQL                           string
 	getBlockSizeSQL, getBlockSizeRangeSQL                        string
 	getBestBlockHashSQL, getBestBlockHeightSQL                   string
+	getHighestBlockHashSQL, getHighestBlockHeightSQL             string
 	getMainchainStatusSQL, invalidateBlockSQL                    string
 	setHeightToSideChainSQL                                      string
 	deleteBlockByHeightMainChainSQL, deleteBlockByHashSQL        string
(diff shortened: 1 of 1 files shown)


db/dcrsqlite/sqlite.go
```diff
@@ -256,12 +266,41 @@ func NewDB(db *sql.DB) (*DB, error) {
...
+	// Attempt to retrieve the best mainchain block heights from both tables,
+	// but fallback to the highest block if the tables have yet to be upgraded
+	// with the is_mainchain column. Log the latter case.
 	var err error
+	noSuchColPrefix := "no such column: "
 	if d.dbSummaryHeight, err = d.getBlockSummaryHeight(); err != nil {
-		return nil, err
+		// If the table scheme lacks the is_mainchain column, and upgrade is
+		// required. Fall back to querying without the is_mainchain condition.
+		ind := strings.LastIndex(err.Error(), noSuchColPrefix+"is_")
```
chappjc: This is somewhat inelegant compared to a `PRAGMA table_info` query, but it's simple and requires no extra query, just inspection of the error.

chappjc:
Successful upgrade of https://beta.dcrdata.org/ from 3.1 to 4.0-pre:

```
[INF] SQLT: Block summary table missing column "is_mainchain". Table upgrade required.
[INF] SQLT: Block summary table missing column "dcrdata_block_summary.is_mainchain". Table upgrade required.
[INF] SQLT: Detected old SQLite table structure. Updating.
```

Note: The PostgreSQL scheme is still upgrading from 3.5.5 to 3.8, so it's it's still not up as of right now.

Issue #962, opened 2019-01-28
The time format needs to include time zone


chappjc:
Even with the fixes to https://github.com/decred/dcrdata/issues/875, the DB tables that have a TIMESTAMP type column should have this type changed to TIMESTAMPTZ to make reading/writing less error prone.  One such issue is with the /insight/api/blocks endpoint where the blocks' times do not show the time zone, and worse, are not necessarily in UTC (they are only UTC if the machine clock is in UTC).

chappjc:
Setting 4.0 milestone and pinning it since this needs to be done for 4.0.

xaur:
How aborting startup if the machine clock is not UTC?

chappjc:
@xaur That's not good enough unfortunately.  The system clock would also need to be UTC (in addition to the hardware/machine/RTC clock).

For example, on an Arch Linux desktop, my hardware clock is in UTC, but my system has a timezone.  So we'd really need to abort startup if the system/OS clock is not in UTC, which I don't think is reasonable for devs.  This is good on a cloud server though.

```
$ sudo timedatectl
               Local time: Tue 2019-02-05 10:47:53 CST
           Universal time: Tue 2019-02-05 16:47:53 UTC
                 RTC time: Tue 2019-02-05 16:47:53
                Time zone: America/Chicago (CST, -0600)
System clock synchronized: yes
              NTP service: active
          RTC in local TZ: no
```

chappjc:
I'm addressing this.

chappjc:
Resolved by PR #1002 and https://github.com/decred/dcrdata/pull/988

PR #988, opened 2019-02-06
time: make a RFC3339 formatter for machines/JS

chappjc:
Would be a workaround for https://github.com/decred/dcrdata/issues/999, but for DBs in various states between the 3.1 and 4.0 releases, it is needed to rebuild fresh once PR https://github.com/decred/dcrdata/pull/1002 is merged.

Our time string "2006-01-02 15:04:05" was not parseable by multiple
browsers, and it lacked a time zone. This adds a second format for both
`TimeDef` types in db/dbtypes and explorer/types to format the string in
RFC3339 (`"2006-01-02T15:04:05Z07:00"`). TODO: eliminate this
duplicate type.
This format is generated with the new `(TimeDef).RFC3339()` functions.

Call `RFC3339()` from `MarshalJSON` use the `time.RFC3339` format in
`UnmarshalJSON`.  This changes the date format in all of the API endpoints
so that the times may be parsed by javascript in dumb browsers like
Safari and Epiphany.

Fix the age controller's time parsing on several pages by using the new
`RFC3339` method in the HTML templates to generate a more JS friendly
time string.

Ref. for RFC3339 and ISO8601: https://tools.ietf.org/html/rfc3339#section-5.8

This is related to https://github.com/decred/dcrdata/issues/962, but a the time columns in the DB tables should still
be changed from `TIMESTAMP` to `TIMESTAMPTZ`.

### DIFF
--- a/db/dbtypes/types.go
+++ b/db/dbtypes/types.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2018, The Decred developers
+// Copyright (c) 2018-2019, The Decred developers
 // See LICENSE for details.
 
 package dbtypes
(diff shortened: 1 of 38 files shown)


chappjc:
This has some unintended time zone consequences.  Will need to investigate

chappjc:
@gozart1 and @buck54321 This did the trick so that Safari and Epiphany can parse the time strings, but having the bare "Z" trailing in the string makes it explicit that it's UTC time and some js controllers are mad.

chappjc:
I have a commit yet to push for the (TimeDef).Scan (sql.Scanner) so that it matches the Stringer.  It's in my desktop

views/block.tmpl
```diff
@@ -132,7 +132,7 @@
                     <tr>
                         <td class="text-right lh1rem pr-2 xs-w117">TIME</td>
                         <td class="lh1rem" style="min-width: 178px;">
-                            ({{timezone}}) {{.BlockTime}}<span class="op60 fs12 nowrap jsonly"> (<span data-target="time.age" data-age="{{.BlockTime.RFC3339}}"></span> ago)</span>
+                            (UTC) {{.BlockTime}}<span class="op60 fs12 nowrap jsonly"> (<span data-target="time.age" data-age="{{.BlockTime.RFC3339}}"></span> ago)</span>
```
gozart1: Why switch back to hardcoding UTC here?
Seems like we can simply keep `({{timezone}}) {{.BlockTime}}` since current setup does support the backend running in any timezone.

chappjc: `{{timezone}}` is the timezone of the server, which is not relevant to the client.  We want to show a UTC time regardless of the server time zone and the client timezone.

I have yet to push another commit to the go code to make that happen correctly.  The commit I have in preparation (it's very short, I just haven't gotten back to it) is making the sql date-time reader (`sql.Scanner`) work the exact reverse to the writer (`sql.Valuer`).  The essence of the remaining issue is that the writer takes a UTC time and saves it to the DB table without a anything indicating the timezone, but the reader processes the time and assumes it is local time.

chappjc: Actually, for backward compatibility with existing table data, I'm not going to change what's written to the table until we do an actual scheme upgrade.  I'll leave the original data in the table, which has local times using the default valuer for time.Time when no timeone is permitted in the DB datatype, and also leave the default time.Time scanner, which loads the time as local time.  This at least is consistent.  The frontend solution in this case is simply to convert the time string to UTC when constructing the TimeDef for the explorer frontend.

chappjc:
This has been adapted following merge of https://github.com/decred/dcrdata/pull/1002.  Please have another look, thanks!

db/dbtypes/types.go
```diff
@@ -48,15 +48,104 @@ type TimeDef struct {
 	T time.Time
 }
 
+const (
+	timeDefFmtHuman = "2006-01-02 15:04:05 (MST)"
```
chappjc: Can be anything since we don't rely on this format to be machine-parsed.

db/dbtypes/types.go
```diff
@@ -48,15 +48,104 @@ type TimeDef struct {
...
+	timeDefFmtJS    = time.RFC3339
```
chappjc: This one is machine readable though.

db/dbtypes/types.go
```diff
@@ -48,15 +48,104 @@ type TimeDef struct {
...
+	return t.T.Format(timeDefFmtHuman)
+}
+
+// RFC3339 formats the time in a machine-friendly layout.
+func (t TimeDef) RFC3339() string {
+	return t.T.Format(timeDefFmtJS)
+}
+
+// UNIX returns the UNIX epoch time stamp.
+func (t TimeDef) UNIX() int64 {
```
chappjc: And of course, this is the most friendly to machines.

db/dcrpg/queries.go
```diff
@@ -454,7 +454,7 @@ func InsertVotes(db *sql.DB, dbTxns []*dbtypes.Tx, _ /*txDbIDs*/ []uint64, fTx *
 			}
 
 			err = agendaStmt.QueryRow(val.ID, index, tx.TxID, tx.BlockHeight,
-				tx.BlockTime.T, lockedIn, activated, hardForked).Scan(&rowID)
+				tx.BlockTime, lockedIn, activated, hardForked).Scan(&rowID)
```
chappjc: Instead of passing a `time.Time`, we pass a `dbtypes.TimeDef` thanks to it being a `sql.Valuer` now.

public/index.js
```diff
@@ -49,7 +46,7 @@ async function createWebSocket (loc) {
   var updateBlockData = function (event) {
     console.log('Received newblock message', event)
     var newBlock = JSON.parse(event)
-    newBlock.block.unixStamp = (new Date(newBlock.block.time).getTime() / 1000) - window.DCRThings.netTimezoneOffset
```
chappjc: the block time now includes time zone

views/block.tmpl
```diff
@@ -132,7 +132,7 @@
                     <tr>
                         <td class="text-right lh1rem pr-2 xs-w117">TIME</td>
                         <td class="lh1rem" style="min-width: 178px;">
-                            ({{timezone}}) {{.BlockTime}}<span class="op60 fs12 nowrap jsonly"> (<span data-target="time.age" data-age="{{.BlockTime}}"></span> ago)</span>
+                            {{.BlockTime}}<span class="op60 fs12 nowrap jsonly"> (<span data-target="time.age" data-age="{{.BlockTime.UNIX}}"></span> ago)</span>
```
chappjc: Time zone shown in String now.

public/index.js
chappjc: Note this comes from Json, so it's format by marshaljson

chappjc:
Rebased on agenda tables upgrade.

chappjc:
Seems to be working on Epiphany (a.k.a. Gnome Web) too:

image

Merging so we can iterate on the frontend bits.

Issue #993, opened 2019-02-10
default zoom on /ticketpool would be better set as All Time instead of Day


raedah:

chappjc:
Resolved by https://github.com/decred/dcrdata/pull/997

PR #997, opened 2019-02-10
Make ticketpool chart all time zoom button default

dmigwi:
Set "All Time" as the default Zoom button then ticketpool chart

### DIFF
--- a/public/js/controllers/ticketpool_controller.js
+++ b/public/js/controllers/ticketpool_controller.js
@@ -123,7 +123,7 @@ export default class extends Controller {
       'price_chart': null,
       'donut_chart': null
     }
-    this.zoom = 'day'
+    this.zoom = 'all'
     this.bars = 'all'
 
     Dygraph = await getDefault(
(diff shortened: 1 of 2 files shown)


dmigwi:
https://github.com/decred/dcrdata/issues/993

chappjc:
I'm not sure if this has anything to do with this PR, but I seem to occasionally get a hang with the spinning update graphic.  Looking at the console I see this:

image

@buck54321 What do you thing this might be related to?

dmigwi:
I haven't encountered such a case scenario but my initial analysis, show that some part payload expected for the price graph might be missing at those specific instances.
Possible cause being that the queries for the above chart take too long than expected. If I may ask, Can you be able to replicate the error above since I haven't seen such on my end? Do that bug happen when your system is under heavy load?

chappjc:
I can replicate it just by refreshing a few times.  webpack all fresh and browser hard-refreshed.

chappjc:
I found two bugs with the ticketpool js controller when looking into the unrelated issue.  Proposed fixes.: https://github.com/decred/dcrdata/pull/1003

PR #1000, opened 2019-02-11
Add hidepgconfig flag

dmigwi:
`hidepgconfig` blocks logging of the postgresql db configuration which happens by default.

To startup dcrdata now without displaying the tabular configuration settings can be achieved by running:
```shell
go build ./ && ./v4 --hidepgconfig
```

### DIFF
--- a/cmd/rebuilddb2/config.go
+++ b/cmd/rebuilddb2/config.go
@@ -55,7 +55,7 @@ type config struct {
 	HTTPProfile  bool   `long:"httpprof" short:"p" description:"Start HTTP profiler."`
 	CPUProfile   string `long:"cpuprofile" description:"File for CPU profiling."`
 	MemProfile   string `long:"memprofile" description:"File for mempry profiling."`
-	HidePGConfig bool   `long:"hidepgconfig" description:"If set to true it blocks logging of the PostgreSQL db configuration on system start up. Defaults to false."`
+	HidePGConfig bool   `long:"hidepgconfig" description:"Blocks logging of the PostgreSQL db configuration on system start up."`
 
 	// DB
 	DBHostPort             string `long:"dbhost" description:"DB host"`
--- a/cmd/rebuilddb2/config.go
+++ b/cmd/rebuilddb2/config.go
@@ -45,16 +45,17 @@ var (
 
 type config struct {
 	// General application behavior
-	ConfigFile  string `short:"C" long:"configfile" description:"Path to configuration file"`
-	ShowVersion bool   `short:"V" long:"version" description:"Display version information and exit"`
-	TestNet     bool   `long:"testnet" description:"Use the test network (default mainnet)"`
-	SimNet      bool   `long:"simnet" description:"Use the simulation test network (default mainnet)"`
-	DebugLevel  string `short:"d" long:"debuglevel" description:"Logging level {trace, debug, info, warn, error, critical}"`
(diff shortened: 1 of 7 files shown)


cmd/rebuilddb2/config.go
```diff
@@ -45,16 +45,17 @@ var (
...
+	ShowVersion  bool   `short:"V" long:"version" description:"Display version information and exit"`
+	TestNet      bool   `long:"testnet" description:"Use the test network (default mainnet)"`
+	SimNet       bool   `long:"simnet" description:"Use the simulation test network (default mainnet)"`
+	DebugLevel   string `short:"d" long:"debuglevel" description:"Logging level {trace, debug, info, warn, error, critical}"`
+	Quiet        bool   `short:"q" long:"quiet" description:"Easy way to set debuglevel to error"`
+	LogDir       string `long:"logdir" description:"Directory to log output"`
+	HTTPProfile  bool   `long:"httpprof" short:"p" description:"Start HTTP profiler."`
+	CPUProfile   string `long:"cpuprofile" description:"File for CPU profiling."`
+	MemProfile   string `long:"memprofile" description:"File for mempry profiling."`
+	HidePGConfig bool   `long:"hidepgconfig" description:"If set to true it blocks logging of the PostgreSQL db configuration on system start up. Defaults to false."`
```
chappjc: It's OK to omit mention of the default for boolean switches (on the command line it's run like `--hidepgconfig` with no value).  So you can word it to describe what it means if the switch is provided ("Block the logging...").

db/dcrpg/pgblockchain.go
```diff
@@ -427,9 +427,10 @@ type DBInfo struct {
 // parameters. By default, duplicate row checks on insertion are enabled. See
 // NewChainDBWithCancel to enable context cancellation of running queries.
 func NewChainDB(dbi *DBInfo, params *chaincfg.Params,
-	stakeDB *stakedb.StakeDatabase, devPrefetch bool) (*ChainDB, error) {
+	stakeDB *stakedb.StakeDatabase, devPrefetch, isHidePGConfig bool) (*ChainDB, error) {
```
chappjc: Just `hidePGConfig` is fine since `hide` is a verb indicating the action/state when it is true.

db/dcrpg/pgblockchain.go
```diff
@@ -457,17 +458,21 @@ func NewChainDBWithCancel(ctx context.Context, dbi *DBInfo, params *chaincfg.Par
 	}
 	log.Info(pgVersion)
 
-	perfSettings, err := RetrieveSysSettingsPerformance(db)
-	if err != nil {
-		return nil, err
-	}
-	log.Infof("postgres performance settings:\n%v", perfSettings)
+	// if the hidepgconfig flag was set to true then the PostgreSQL configuration
+	// settings will not be logged by default on system start up.
```
chappjc: Since the notion of a default does not exist _inside_ `NewChainDBWithCancel`, let's just say "Optionally log the PostgreSQL configuration."

sample-dcrdata.conf
```diff
@@ -42,6 +42,10 @@
 ; syncing is done.
 ; sync-status-limit=1000
 
+; Blocks logging of the PostgreSQL db configuration on system start up. Defaults
```
chappjc: Just "Block logging of the PostgreSQL db configuration on system start up." Default for booleans are an implied false when the switch is omited.

PR #1001, opened 2019-02-11
Fix windows view drop down navigation bug

dmigwi:
On `/ticketpricewindows` page the dropdown menu did not work because the blocklist `data-controller` was missing.

### DIFF
--- a/views/windows.tmpl
+++ b/views/windows.tmpl
@@ -3,7 +3,7 @@
 <html lang="en">
 
 {{template "html-head" "Decred Windows List"}}
-<body class="{{ theme }}" data-controller="ticketwindow">
+<body class="{{ theme }}" data-controller="ticketwindow  blocklist">
     {{template "navbar" . }}
     <div class="container main">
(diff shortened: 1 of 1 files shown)


views/windows.tmpl
```diff
@@ -3,7 +3,7 @@
 <html lang="en">
 
 {{template "html-head" "Decred Windows List"}}
-<body class="{{ theme }}" data-controller="ticketwindow">
+<body class="{{ theme }}" data-controller="ticketwindow  blocklist">
```
chappjc: Pleaes remove extra space.

views/windows.tmpl
```diff
@@ -3,7 +3,7 @@
 <html lang="en">
 
 {{template "html-head" "Decred Windows List"}}
-<body class="{{ theme }}" data-controller="ticketwindow">
+<body class="{{ theme }}" data-controller="ticketwindow blocklist">
     {{template "navbar" . }}
     <div class="container main">
```
chappjc: The `data-controller="ticketwindow blocklist"` bit seems to be an attribute of the `div` rather than `body` in other templates.  Is this correct and OK?

PR #1010, opened 2019-02-15
build.sh: remove old v3

chappjc:
Remove the old v3 binary to avoid confusion.
Silence gnu parallel.
Remove old views and public folders when installing.

### DIFF
--- a/dev/build.sh
+++ b/dev/build.sh
@@ -21,7 +21,7 @@
 #   utility is not used since 7za compression rates are slightly better even for
 #   the gz format.
 #
-# Copyright (c) 2018, The Decred developers.
+# Copyright (c) 2018-2019, The Decred developers.
 # See LICENSE for details.
 
 REPO=`git rev-parse --show-toplevel 2> /dev/null`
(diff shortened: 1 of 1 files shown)


dev/build.sh
```diff
@@ -39,6 +39,10 @@ fi
 set -e
 
 pushd $ROOT > /dev/null
+
+# Remove the old v3 binary to avoid confusion.
+if [[ -f "v3" ]]; then rm v3; fi
+
```
dmigwi: Since you will be building fresh binaries below why don't you delete all previous binaries like below:
```shell
# deletes all previous executable binary versions.
rm v[0-9]*
```

dev/build.sh
```diff
@@ -70,6 +73,7 @@ DEST=$2
 
 if [[ -n "$DEST" ]]; then
     sudo install -m 754 -o dcrdata -g decred ./v3 ${DEST}/dcrdata
```
dmigwi: Since all the previous binary versions including `v3` have been deleted, you could use a regex expression to select the latest executable binary built. i.e.
```shell
sudo install -m 754 -o dcrdata -g decred ./v[0-9]* ${DEST}/dcrdata
```

dev/build.sh
chappjc: ah, great suggestion, thanks

Issue #1005, opened 2019-02-13
Database sync bug (updateExplorer chan *chainhash.Hash closed prematurely)


dmigwi:
@chappjc I was almost done with full db sync from scratch on master branch since yesterday when the following error occurred. Got clues on the possible causes?

In the meantime I will be looking into it.
```
2019-02-13 19:46:43.820 [INF] PSQL: Updating from vins 13100000 to 13199999...
2019-02-13 19:54:06.144 [INF] PSQL: Updating from vins 13200000 to 13299999...
2019-02-13 20:01:23.843 [INF] PSQL: Updating from vins 13300000 to 13399999...
2019-02-13 20:08:46.034 [INF] PSQL: Updating from vins 13400000 to 13499999...
2019-02-13 20:15:33.692 [INF] PSQL: Updating from vins 13500000 to 13599999...
2019-02-13 20:22:41.710 [INF] PSQL: Updating from vins 13600000 to 13699999...
2019-02-13 20:30:04.509 [INF] PSQL: Updating from vins 13700000 to 13799999...
2019-02-13 20:37:27.069 [INF] PSQL: Updating from vins 13800000 to 13899999...
2019-02-13 20:44:46.908 [INF] PSQL: Updating from vins 13900000 to 13999999...
2019-02-13 20:52:13.922 [INF] PSQL: Updating from vins 14000000 to 14009507...
2019-02-13 20:52:55.496 [INF] PSQL: Updated 12145709 rows of address table
2019-02-13 20:52:55.497 [INF] PSQL: Indexing addresses table on address...
2019-02-13 20:54:02.934 [INF] PSQL: Indexing addresses table on matching tx hash...
2019-02-13 20:55:04.378 [INF] PSQL: Indexing addresses table on block time...
2019-02-13 20:55:20.370 [INF] PSQL: Indexing addresses table on vout Db ID...
2019-02-13 20:55:48.440 [INF] PSQL: Sync finished at height 318141. Delta: 318142 blocks, 5625488 transactions, 14009523 ins, 20759111 outs, 15058099 addresses
2019-02-13 20:55:48.452 [INF] DATD: PostgreSQL sync ended at height 318141
2019-02-13 20:55:48.489 [INF] SQLT: Current best block (chain server):     318329
2019-02-13 20:55:48.489 [INF] SQLT: Current best block (sqlite block DB):  318141
2019-02-13 20:55:48.489 [INF] SQLT: Current best block (stakedb):          318141
[...]

PR #1011, opened 2019-02-15
Open the barload channel always

dmigwi:
Premature closing of the `barLoad` channel resulted to a panic when trying to send data through a closed `barLoad` channel. 
Sending `nil` via the channel results to `barLoad` being assigned `nil` rather than closing it thereby avoiding the panic.
#1005

### DIFF
--- a/db/dcrpg/sync.go
+++ b/db/dcrpg/sync.go
@@ -187,7 +187,7 @@ func (db *ChainDB) SyncChainDB(ctx context.Context, client rpcutils.MasterBlockG
 	}
 
 	// Safely send sync status updates on barLoad channel, and set the channel
-	// to nil if it was closed or the buffer is full.
+	// to nil if the buffer is full.
 	sendProgressUpdate := func(p *dbtypes.ProgressBarLoad) {
 		if barLoad == nil {
 			return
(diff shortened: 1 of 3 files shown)


db/dcrpg/sync.go
```diff
@@ -201,7 +201,7 @@ func (db *ChainDB) SyncChainDB(ctx context.Context, client rpcutils.MasterBlockG
 	}
 
 	// Safely send new block hash on updateExplorer channel, and set the channel
-	// to nil if it was closed or the buffer is full.
+	// to nil if the buffer is full.
```
chappjc: `(*WiredDB).resyncDB` has similar incorrect language.  Would you update there too please?

main.go
```diff
@@ -951,9 +951,9 @@ func _main(ctx context.Context) error {
 	// Enable new blocks being stored into the base DB's cache.
 	baseDB.EnableCache()
 
-	// Deactivate displaying the sync status page after the db sync was completed.
+	// Block further usage of the barLoad by sending a nil value
 	if barLoad != nil {
-		close(barLoad)
+		barLoad <- nil
```
chappjc: This will hang if: (1) sync hit an error and bailed out, or (2) `barLoad` is a nil channel (when `!displaySyncStatusPage`).  Please do one of these so it won't block:

```go
select {
	case barLoad <- nil:
	default:
}
```

PR #1012, opened 2019-02-15
Central server for exchange rate data

buck54321:
For load-balanced backends, a central server for exchange data is needed. Dcrrates runs an `ExchangeBot` and takes subscriptions from gRPC clients, pushing updates when they are received. The Dcrrates server requires TLS, and will auto-generate a self-signed certificate if one is neither provided nor already exists in the application directory.

There is a script to generate the gRPC definition files from the proto file using `protoc`. This is only needed if the proto file is modified.

ExchangeBot itself has been updated with an option to get exchange data from dcrrates instead of directly from the APIs. Note that the `ExchangeBot.IsUpdated` method has been removed. It was only used in `exchanges_test.go`, and the tests have been updated to accomodate both modes of operation.


chappjc:
Ref: https://github.com/decred/dcrdata/pull/951#issuecomment-456457421

dcrrates/server/config.go
```diff
@@ -0,0 +1,97 @@
...
+	defaultExchangeRefresh = "20m"
+	defaultExchangeExpiry  = "60m"
+	defaultListen          = ":7778"
+	defaultBtcIndex        = "USD"
+)
+
+var (
+	defaultConfigPath = filepath.Join(dcrutil.AppDataDir("dcrdata", false), "dcrrates-server.conf")
+	defaultCertPath   = filepath.Join(dcrutil.AppDataDir("dcrdata", false), "rpc.cert")
+	defaultKeyPath    = filepath.Join(dcrutil.AppDataDir("dcrdata", false), "rpc.key")
```
buck54321: I guess I don't need three identical function calls here.

dcrrates/server/log.go
```diff
@@ -0,0 +1,51 @@
...
+// Write writes the data in p to standard out and the log rotator.
+func (logWriter) Write(p []byte) (n int, err error) {
+	os.Stdout.Write(p)
+	return logRotator.Write(p)
+}
+
+// inititializeLogging initializes the logging rotater to write logs to logFile
+// and create roll files in the same directory.  It must be called before the
+// package-global log rotater variables are used.
+func inititializeLogging(logFile string) {
```
buck54321: whoops

dcrrates/server/types.go
```diff
@@ -0,0 +1,178 @@
...
+
+	server.DeleteClient(sid)
+	if peerFound {
+		log.Infof("Client at %s has disconnected", peerInfo.Addr.String())
+	}
+	return
+}
+
+// AddClient adds a client to the map and advance the streamCounter.
+func (server *RateServer) AddClient(stream GRPCStream, hello *dcrrates.ExchangeSubscription) (RateClient, StreamID) {
```
chappjc: Seems like AddClient and DeleteClient have no use outside of `dcrrates/server`.  Should/could they be unexported?

dcrrates/server/types.go
```diff
@@ -0,0 +1,178 @@
...
+			Indices: server.xcBot.Indices(token),
+		})
+	}
+
+	// Get the address for an Infof. Seems like peerInfo is always found, but
+	// am checking anyway.
+	peerInfo, peerFound := grpcPeer.FromContext(client.Stream().Context())
+	if peerFound {
+		log.Infof("Client has connected from %s", peerInfo.Addr.String())
+	}
```
chappjc: What does it mean if `!peerFound`?  Should we log that too or even delete the client and return?

dcrrates/server/config.go
```diff
@@ -0,0 +1,97 @@
...
+const (
+	defaultLogPath         = "logs"
+	defaultExchangeRefresh = "20m"
+	defaultExchangeExpiry  = "60m"
+	defaultListen          = ":7778"
+	defaultBtcIndex        = "USD"
+)
+
+var (
+	defaultConfigPath = filepath.Join(dcrutil.AppDataDir("dcrdata", false), "dcrrates-server.conf")
```
chappjc: I think it would be more intuitive if this were dcrrates.conf since the binary is dcrrates.

dcrrates/server/types.go
buck54321: I'm not really certain the conditions under which `FromContext` might not find the context variable,  [see here](https://github.com/grpc/grpc-go/blob/2773c7bbcf81cf358d3f0038b1469b2d44062acb/peer/peer.go#L48), but since it's returned, I was just checking for good measure. Should I reject connections where an address can't be found?

dcrrates/server/tls.go
```diff
@@ -0,0 +1,101 @@
...
+)
+
+// openRPCKeyPair creates or loads the RPC TLS keypair specified by the
+// application config.
+func openRPCKeyPair(certPath, keyPath string) (credentials.TransportCredentials, error) {
+	// Generate a new keypair when the key is missing.
+	_, e := os.Stat(keyPath)
+	keyExists := !os.IsNotExist(e)
+	if !keyExists {
+		cert, err := generateRPCKeyPair(certPath, keyPath, certWriter{})
```
chappjc: One thing that would be helpful in this function (either way `cert` is created) is to show the hostnames that the cert is valid for.

dcrrates/server/main.go
```diff
@@ -0,0 +1,136 @@
...
+		return
+	}
+
+	inititializeLogging(filepath.Join(cfg.LogPath, "dcrrate-server.log"))
+
+	if cfg.CertificatePath == "" || cfg.KeyPath == "" {
+		log.Errorf("TLS certificate and key files must be provided")
+		return
+	}
+	creds, err := openRPCKeyPair(cfg.CertificatePath, cfg.KeyPath)
```
chappjc: This may not be straightforward, but a way to specify additional host names for which the cert should be valid would be very helpful.  I had to use `dcrd/cmd/gencerts` with `-H` to make a cert I could use on the rates server host.

sample-dcrdata.conf
```diff
@@ -74,3 +74,8 @@
 ; comma-separated list. Currently available: coinbase, coindesk, binance,
 ; bittrex, dragonex, huobi, poloniex
 ; disable-exchange=dragonex,huobi
+
+; Pull exchange data from a dcrrates server at the network address given by
+; ratemaster. Requires the server's TLS certificate. 
+;ratemaster=
+;ratecert=
```
chappjc: ratecert=rates.cert and was expecting for it to look in ~/.dcrdata/ for that, but it didn't find it. Seems it was looking in the wording directory, or maybe the path of the binary(?). If that's consistent with how dcrdcert works, that's OK, but I was surprised.

dcrrates/server/sample-dcrrates-server.conf
```diff
@@ -0,0 +1,26 @@
...
+;listen=:7778
+
+; A directory to put log files. The directory will be created if it doesn't
+; exist.
+;logpath=logs
+
+; Disable individual exchanges. Multiple exchanges can be disabled with a
+; comma-separated list. Currently available: coinbase, coindesk, binance,
+; bittrex, dragonex, huobi, poloniex.
+;disable-exchange=
```
chappjc: does not seem to have an effect

sample-dcrdata.conf
```diff
@@ -74,3 +74,8 @@
 ; comma-separated list. Currently available: coinbase, coindesk, binance,
 ; bittrex, dragonex, huobi, poloniex
 ; disable-exchange=dragonex,huobi
+
+; Pull exchange data from a dcrrates server at the network address given by
+; ratemaster. Requires the server's TLS certificate. 
+;ratemaster=
```
chappjc: If their is only a host and no port, it should default to 7778

dcrrates/server/sample-dcrrates-server.conf
buck54321: I wasn't able to reproduce this. Could you check again with the new commit?

dcrrates/rateserver/config.go
```diff
@@ -0,0 +1,184 @@
...
+	} else {
+		cfg.LogPath = cleanAndExpandPath(cfg.LogPath)
+	}
+
+	return &cfg, nil
+}
+
+// cleanAndExpandPath expands environment variables and leading ~ in the passed
+// path, cleans the result, and returns it.
+func cleanAndExpandPath(path string) string {
```
buck54321: This was copied from the root `config.go`. I wasn't certain how to document / copyright.

dcrrates/server/types.go
chappjc: Note sure.  I'll keep an eye on this when it's deployed to alpha.

PR #1015, opened 2019-02-15
dcrpg: upgrades do not need a blockgate

chappjc:
Just use a rpcclient.Client.


chappjc:
Upgrade succeeded as before:

```
[INF] PSQL:  - 1137598 rows in agendas table (prune agendas table upgrade) were successfully upgraded in 740.64 sec
```

PR #1016, opened 2019-02-15
gov: Restructure decred's Agendas and Proposals governance implementations

dmigwi:
This introduces two pages under `/proposals` and `/propasal/{id}` url paths. The proposals data is sourced from [politeia's](https://proposals.decred.org/api/v1) API endpoints and stored in the storm db as the other high level on-chain agenda details

**Proposal Page**
screenshot from 2019-02-16 01-36-16

**Proposals Page**
screenshot from 2019-02-16 01-36-06


buck54321:
This is great @dmigwi . I've been looking forward to seeing Politeia data incorporated. This is important information that demonstrates one of Decred's strongest features. A couple of suggestions from the first impression.

1. The proposals page should say "Politeia" somewhere. It might even be good to have a one or two sentence description of what Politeia is. "Off-chain" is also maybe a little misleading because Politeia is anchored to the blockchain, and that's one of its most exciting features. 

2. Let's pull the vote counts out of the table and give them some visual prominence. Maybe even move them towards the top. Other's might say otherwise, but I don't think we need all three columns that basically say the same thing. Maybe just "Approvals" and "Rejections" or "Ayes" and "Nays". Some variation of this 

image

3. The front-end is starting to move through a redesign with inspiration from #451. /home and /mempool pages have iterated, so any efforts towards trying to find a common style would be good. If not, that's okay too. @gozart1 and I like to pick at that stuff. But at some point, we'll need to stop allowing public pages with the older style.

4. Eventually some kind of filtering will be needed. Are you just sorting by last update right now?

5. Not certain how this would work, but maybe some kind of key or legend that explains what the different phases of the proposal are, i.e. "vetted",  "active" etc.

dmigwi:
Great Notes @buck54321;
- Concerning the new #451 inspired theme, I will be modifying some more stuff to get it closer to where is supposed to be.
- The filtering of all proposals is done by the last Update field.
- You just gave me an idea on how to the use space created after adding pagination. This which will create space for the various legends.

raedah:
Proposal items should likely be imported by oldest first so that the order stays consistent on subsequent syncs.

main.go
```diff
@@ -517,9 +509,35 @@ func _main(ctx context.Context) error {
...
+
+	// NewProposalsDB creates an off-chain db instance that helps to store and
+	// retrieve off-chain proposals data.
+	offChainInstance, err := offchaindb.NewProposalsDB(cfg.PoliteiaAPIURL,
+		filepath.Join(cfg.DataDir, cfg.OffChainDBFileName))
+	if err != nil {
+		return fmt.Errorf("failed to create new off-chain db instance: %v", err)
+	}
+
+	// CheckOffChainUpdates checks for offchain db updates.
```
chappjc: The comments in the above block of code are written like function docs.  Here at the call sites in main they need to explain what they are doing and why.

e.g.

```go
// Load or create a new agenda DB.
```

```go
// Refresh agenda data from the latest information from dcrd.  This is also done when... .
```

```go
// Load or create a new Politeia proposal DB.
```

etc.

db/offchaindb/db.go
```diff
@@ -309,6 +311,10 @@ func NewProposalsDB(dbPath, politeiaURL string) (*ProposalDB, error) {
 		return nil, fmt.Errorf("missing politeia API URL")
 	}
 
+	if dbPath == "" {
```
chappjc: Might as well check this before `os.Stat`.

db/offchaindb/db.go
```diff
@@ -91,12 +90,15 @@ type Votes struct {
...
-		Description string `json:"description"`
-		Bits        int32  `json:"bits"`
-	} `json:"option"`
-	VotesReceived int64 `json:"votesreceived"`
+	Option        VoteResults `json:"option"`
+	VotesReceived int64       `json:"votesreceived"`
+}
+
+// VoteResults defines the actual high level vote results for the specific agenda.
+type VoteResults struct {
```
chappjc: `VoteOption`

api/apiroutes.go
```diff
@@ -21,11 +21,11 @@ import (
 	"time"
 
 	"github.com/decred/dcrd/chaincfg"
-	"github.com/decred/dcrd/dcrjson/v2"
-	"github.com/decred/dcrd/rpcclient/v2"
+	dcrjson "github.com/decred/dcrd/dcrjson/v2"
+	rpcclient "github.com/decred/dcrd/rpcclient/v2"
```
chappjc: Was a linter complaining?  The `/vN` suffix is an exception when modules are enabled.  No need explicitly name them.

db/offchaindb/db.go
```diff
@@ -0,0 +1,483 @@
+// Copyright (c) 2019, The Decred developers
+// See LICENSE for details.
+
+package offchaindb
```
chappjc: Please document the new packages with `// package offchaindb ...` (or whatever package name it turns into).

db/offchaindb/db.go
```diff
@@ -0,0 +1,483 @@
...
+	"os"
+	"time"
+
+	"github.com/asdine/storm"
+)
+
+const (
+	// vettedProposalsRoute is the URL path that returns the public/vetted
+	// proposals.
+	vettedProposalsRoute = "/proposals/vetted"
```
chappjc: Please use the constants in https://github.com/decred/politeia/blob/master/politeiawww/api/v1/v1.go

```go
import piapi "github.com/decred/politeia/blob/master/politeiawww/api/v1"
```

Then you just use `piapi.RouteAllVetted` and `piapi.RouteAllVoteStatus` (or any other routes named there).

db/offchaindb/db.go
```diff
@@ -0,0 +1,483 @@
...
+		return "unreviewed changes"
+	case AbandonedProposalStatus:
+		return "abandoned"
+	default:
+		return "unknown"
+	}
+}
+
+// ProposalStatusFrmStr converts the string into ProposalStatusType value.
+func ProposalStatusFrmStr(val string) ProposalStatusType {
```
chappjc: All this stuff is in `piapi` too. `PropStatus`, etc.

 https://github.com/decred/politeia/blob/73ba15aa6d3e8ba590f5cf5fc1120938997283d6/politeiawww/api/v1/v1.go#L321

db/offchaindb/db.go
```diff
@@ -0,0 +1,483 @@
...
+		return nil, err
+	}
+
+	defer response.Body.Close()
+
+	return ioutil.ReadAll(response.Body)
+}
+
+// saveProposals adds the proposals data to the db.
+func (db *ProposalDB) saveProposals(URLParams string) (int, error) {
```
chappjc: This muddles Pi client functionality and DB functionality.  When this package is split up, it will force us to keep functions will be kept smaller, more testable, and more logically organize.

README.md
```diff
@@ -381,7 +381,8 @@ List of recognized environment variables:
 | The number minimum number of new tickets that must be seen to trigger a new mempool report.                              | DCRDATA_MP_TRIGGER_TICKETS     |
 | Dump to file the fees of all the tickets in mempool.                                                                     | DCRDATA_ENABLE_DUMP_ALL_MP_TIX |
 | SQLite DB file name (default is dcrdata.sqlt.db)                                                                         | DCRDATA_SQLITE_DB_FILE_NAME    |
-| Voting agendas DB file name (default is agendas.db)                                                                      | DCRDATA_AGENDA_DB_FILE_NAME    |
+| On-Chain Voting agendas DB file name (default is agendas.db)                                                             | DCRDATA_ON_CHAIN_DB_FILE_NAME  |
+| Off-Chain Voting proposals DB file name (default is proposals.db)                                                        | DCRDATA_OFF_CHAIN_DB_FILE_NAME |
```
chappjc: Please remove the on/off chain wording here too.

api/apiroutes.go
```diff
@@ -140,6 +142,7 @@ func NewContext(client *rpcclient.Client, params *chaincfg.Params, dataSource Da
 		AuxDataSource: auxDataSource,
 		LiteMode:      liteMode,
 		xcBot:         xcBot,
+		OnChainDB:     onchainInstance,
```
chappjc: I'd prefer to see `AgendaDB` (or similar) instead of `OnChainDB`

config.go
```diff
@@ -94,13 +95,15 @@ type config struct {
...
 	MempoolMinInterval int    `long:"mp-min-interval" description:"The minimum time in seconds between mempool reports, regarless of number of new tickets seen." env:"DCRDATA_MEMPOOL_MIN_INTERVAL"`
 	MempoolMaxInterval int    `long:"mp-max-interval" description:"The maximum time in seconds between mempool reports (within a couple seconds), regarless of number of new tickets seen." env:"DCRDATA_MEMPOOL_MAX_INTERVAL"`
 	MPTriggerTickets   int    `long:"mp-ticket-trigger" description:"The number minimum number of new tickets that must be seen to trigger a new mempool report." env:"DCRDATA_MP_TRIGGER_TICKETS"`
 	DBFileName         string `long:"dbfile" description:"SQLite DB file name (default is dcrdata.sqlt.db)." env:"DCRDATA_SQLITE_DB_FILE_NAME"`
-	AgendaDBFileName   string `long:"agendadbfile" description:"Agenda DB file name (default is agendas.db)." env:"DCRDATA_AGENDA_DB_FILE_NAME"`
+	OnChainDBFileName  string `long:"onchaindbfile" description:"On-Chain DB file name (default is agendas.db)." env:"DCRDATA_ON_CHAIN_DB_FILE_NAME"`
+	OffChainDBFileName string `long:"offchaindbfile" description:"Off-Chain DB file name (default is proposals.db)." env:"DCRDATA_OFF_CHAIN_DB_FILE_NAME"`
```
chappjc: Please change on/off here too.

explorer/explorer.go
```diff
@@ -19,13 +19,15 @@ import (
 	"time"
 
 	"github.com/decred/dcrd/chaincfg"
-	"github.com/decred/dcrd/dcrjson/v2"
+	dcrjson "github.com/decred/dcrd/dcrjson/v2"
```
chappjc: Is a linter complaining about this import without the name?  With modules enabled, you don't need to name it as the v2 is ignored.

explorer/explorer.go
```diff
@@ -218,6 +236,8 @@ type explorerUI struct {
 	Mux              *chi.Mux
 	blockData        explorerDataSourceLite
 	explorerSource   explorerDataSource
+	onChainSource    agendaBackend
+	offChainSource   politeiaBackend
```
chappjc: Please update on/off everywhere.  It just does not have enough meaning to be clear.  Someone reading the code has to look back to the `explorerUI` to see what they are really for (agendas and Pi).

gov/agendas/deployments.go
```diff
@@ -1,142 +1,115 @@
-// Copyright (c) 2018, The Decred developers
+// Copyright (c) 2018-2019, The Decred developers
 // Copyright (c) 2017, Jonathan Chappelow
 // See LICENSE for details.
 
-package agendadb
+// agendas package manages the various deployment agendas that are directly voted
+// upon with the vote bits in vote transactions.
+
```
chappjc: For godoc to work properly with the package docs, I believe there cannot be a blank line between the comments and `package agendas`: https://blog.golang.org/godoc-documenting-go-code

gov/agendas/deployments.go
```diff
@@ -1,142 +1,115 @@
...
-		}
-	}
+	return &AgendaDB{sdb: db}, nil
+}
 
-	agendaDB := &AgendaDB{
-		sdb:        db,
-		NumAgendas: numAgendas,
-		NumChoices: numChoices,
+// countProperties fetches the count of Agendas and Choices. It the appends them
```
chappjc: just counts number of Agendas now.

gov/politeia/piclient/piclient.go
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) 2019, The Decred developers
+// See LICENSE for details.
+
+// piclient handles the http requests made to API URL paths.
```
chappjc: ...made to Politeia API URLs.

gov/politeia/piclient/piclient.go
```diff
@@ -0,0 +1,34 @@
...
+import (
+	"fmt"
+	"io/ioutil"
+	"net/http"
+)
+
+// HandleGetRequests accepts a http client and API URL path as arguments. If the
+// parameters are valid, a GET request is made to the API URL path passed. The
+// body returned is then decoded into []byte and returned.
+func HandleGetRequests(client *http.Client, _URLPath string) ([]byte, error) {
```
chappjc: Variables with an underscore prefix are typically for reserved (e.g. language internals).  Unless we really need it this way, let's stick with `URLPath` or similar.

run_tests.sh
```diff
@@ -53,6 +54,9 @@ testrepo () {
 
   echo "------------------------------------------"
   echo "Tests completed successfully!"
+
+  # Remove test-data-repo data
+  rm -rf $TMPDIR/test-data-repo* $TMPFILE
```
chappjc: Should clean up the temp dir too:

```
rm -rf $TMPDIR $TMPFILE
```

dmigwi:
I've some extra minor features to add (proposals pagination and a legend for the various `Status` and `State` keywords). Was supposed to be done by now but the corrupt golang installation is messing me up.

dmigwi:
Latest Change:
- The UI on the two pages "/proposals" and "/proposal/{ID}" has been reorganized.
- Proposal Update is divided into two major sections:
      1. **Update for the In progress Proposals** - This is update for current proposals with vote statuses `PropVoteStatusNotAuthorized`, `PropVoteStatusAuthorized` or `PropVoteStatusStarted`. All the in new in progress proposals data is updated apart from from proposals whose vote status is still   `PropVoteStatusNotAuthorized`. Happens first always.
      2. **Update for new proposals** - This is for proposals not yet in the db. It fetches the list of proposals from the Politeia API since the last proposal was updated in the Proposals db.
- Apart from the initial updates check on system start up, proposal db update is run in intervals of 20 blocks when the block chain is not running.

Here are the Modifies UI pages: 
**Proposals Page:***
screenshot from 2019-02-21 12-02-40


**Proposal Page:**
screenshot from 2019-02-21 11-58-14

public/js/controllers/proposal_controller.js
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  connect () {
+    this.setChartData()
+
+    this.chart = new Chart(
+      document.getElementById('donutgraph').getContext('2d'), {
+        options: {
+          width: 300,
+          height: 300,
```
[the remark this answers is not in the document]
dmigwi: This is not the chart that is going here. Another Chart in a follow up PR will be sharing the real estate so I hope to resize the current chart then.

gov/agendas/deployments.go
```diff
@@ -178,73 +145,66 @@ func (db *AgendaDB) updatedb(voteVersion uint32, client *rpcclient.Client) {
...
+
 	// voteVersion is vote version as of when lnsupport and sdiffalgorithm votes
 	// casting was activated. More information can be found here
 	// https://docs.decred.org/getting-started/user-guides/agenda-voting/#voting-archive
 	// Also at the moment all agenda version information available in the rpc
 	// starts from version 4 by default.
-
 	var voteVersion uint32 = 4
-	for adb.IsAgendasAvailable(voteVersion) {
+	for db.isAgendasAvailable(voteVersion) {
```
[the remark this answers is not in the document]
dmigwi: Based on how the update works this error is expected. I think not logging it makes sense.
Doing this `for db.isAgendasAvailable(voteVersion + 1) {` means that we shall not be fetching data when the voteVersion is 4.

gov/politeia/proposals.go
```diff
@@ -0,0 +1,286 @@
...
+
+		proposal, err := piclient.RetrieveProposalByToken(db.client, db.APIURLpath,
+			val.Censorship.Token)
+		if err != nil {
+			return 0, fmt.Errorf("RetrieveProposalByToken failed: %v ", err)
+		}
+
+		proposal.ID = val.ID
+
+		err = db.dbP.Update(proposal)
```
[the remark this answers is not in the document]
dmigwi: Sure

gov/politeia/types/types.go
```diff
@@ -0,0 +1,187 @@
...
+	Timestamp       uint64             `json:"timestamp"`
+	UserID          string             `json:"userid"`
+	Username        string             `json:"username"`
+	PublicKey       string             `json:"publickey"`
+	Signature       string             `json:"signature"`
+	Version         string             `json:"version"`
+	Censorship      CensorshipRecord   `json:"censorshiprecord"`
+	NumComments     int32              `json:"numcomments"`
+	StatusChangeMsg string             `json:"statuschangemessage"`
+	PublishedDate   uint64             `json:"publishedat" storm:"index"`
```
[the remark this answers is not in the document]
dmigwi: The spelling mistake is on Politieia end. If my mind serves me right `TimeDef` did not prevent duplication of any sort unless there is something I am missing out on. 
On top of helping to store data, the above struct also helps to un marshal the API response returned by the Politeia API.

gov/politeia/types/types.go
```diff
@@ -0,0 +1,187 @@
...
+		return "unvetted"
+	case VettedState:
+		return "vetted"
+	default:
+		return "unknown"
+	}
+}
+
+// ProposalStateFromStr converts the string into ProposalStateType value.
+func ProposalStateFromStr(val string) ProposalStateType {
```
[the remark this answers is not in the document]
dmigwi: This could be used later.

views/proposal.tmpl
```diff
@@ -0,0 +1,213 @@
...
+                                    {{if .TotalVotes}}
+                                        {{.TotalVotes}}
+                                    {{else}}
+                                        N/A
+                                    {{end}}
+                                </span>
+                            </td>
+                        </tr>
+                         <tr>
+                            <td class="text-right pr-2 lh1rem vam nowrap xs-w117">Eligible Votes Count</td>
```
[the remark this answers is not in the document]
dmigwi: I think there is another field called "Total Votes". I think having just "Vote Count" might not be as clear as possible and its not instinctual that will check if a tool tip exists to get more information on the same.

gov/politeia/types/types.go
```diff
@@ -0,0 +1,187 @@
...
+
+package types
+
+import piapi "github.com/decred/politeia/politeiawww/api/v1"
+
+// ProposalInfo holds the proposal details as document here
+// https://github.com/decred/politeia/blob/master/politeiawww/api/v1/api.md#proposal.
+// It also holds the votes status details.
+type ProposalInfo struct {
+	ID              int                `json:"id" storm:"id,increment"`
```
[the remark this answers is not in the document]
dmigwi: I found it problematic to query a nested field so that is why I used an ID field to help query the proposals. Also the field is managed by the db, so very few chances of a duplicate db entry for the same proposal being updated exist.

views/proposal.tmpl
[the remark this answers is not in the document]
buck54321: I see. Sounds good.

gov/politeia/types/types.go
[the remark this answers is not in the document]
buck54321: To be clear, I'm not talking about querying a nested field. The ID field itself can be a string set to the token, and you'd only have one unique id for the proposal instead of two. It likely would never be an issue unless there was a problem coming from the Politeia API though, so if you're committed to the auto-increment, it's probably okay.

config.go
```diff
@@ -571,3 +588,24 @@ func netName(chainParams *netparams.Params) string {
...
+// i.e. with input http://www.mydomain.com/xxxxx, http://www.mydomain.com should
+// be returned.
+func retrieveRootPath(path string) (string, error) {
+	r, err := url.Parse(path)
+	if err != nil {
+		return "", fmt.Errorf("invalid '%s' url used. error: %v", path, err)
+	}
+
+	// If the url scheme was not found, a regex expression can be used to
+	// eliminate the unwanted part.
```
dmigwi: Why the regex expression evaluation makes sense here: https://github.com/golang/go/commit/ea0d2c14f8010f31dc7b892e6a546cfdb4d8ef21

config.go
```diff
@@ -571,3 +588,24 @@ func netName(chainParams *netparams.Params) string {
...
+	if r.Scheme == "" {
+		exp := regexp.MustCompile(`([\/?]\S*)`)
```
chappjc: Generally we should make tests for our regexps since they are so full of gotchas.  This is ok for here though.

chappjc:
@dmigwi Tentatively approved, although needs a rebase/resolve, and a final test on my part.

chappjc:
(and approval by @buck54321)

dmigwi:
I am rebasing this PR.

config.go
buck54321: Here are a couple of tests that may or may not display unexpected behavior when pointing to a local API server.

https://play.golang.org/p/GqrdcQE85H5

dmigwi:
@buck54321 I am not sure of what the said test should return. Is `:1234` supposed to be a valid url?
If not, I think all is well https://play.golang.org/p/uJdgixbOhsa

config_test.go
```diff
@@ -238,28 +238,36 @@ func TestRetrieveRootPath(t *testing.T) {
...
-		{"mydomain.com?id=1", "mydomain.com"},
-		{"192.168.10.12/xxxxx", "192.168.10.12"},
-		{"mydomain.com/xxxxx?id=1", "mydomain.com"},
-		{"www.mydomain.com/xxxxx", "www.mydomain.com"},
-		{"http://www.mydomain.com/xxxxx", "http://www.mydomain.com"},
-		{"https://www.mydomain.com/xxxxx", "https://www.mydomain.com"},
-		{"https://www.mydomain.com/xxxxx?id=1", "https://www.mydomain.com"},
+		// :1234 is consided invalid url since the root url is also used as a
+		// hyperlink reference on the frontend.
+		{":1234", "", true},
```
dmigwi: This should not be a valid URL since politeia is an external system. For use on other applications, add a host/scheme would work.

buck54321:
Expected

`retrieveRootPath(localhost:1234/api/) -> localhost:1234`
`retrieveRootPath(192.168.0.2:1234/api/) -> 192.168.0.2:1234`

Actual

`retrieveRootPath(localhost:1234/api/) -> localhost:1234/api/`
`retrieveRootPath(192.168.0.2:1234/api/) -> error`



https://play.golang.org/p/JIJbMLCfxLb

dmigwi:
Taking this URL example above: 
screenshot 2019-02-26 at 06 09 31

I envisioned a case scenario where `func retrieveRootPath(path string) (string, error)` would return parts 1-5 of a URL only at all times. IMO, allowing special subpaths like `/api/` complicates things like when someone wants to use `localhost:1234` instead of `localhost:1234/api/`.  Adding `/api/` when needed is easier than removing it when not needed.

Here is the `retrieveRootPath` function doc https://github.com/decred/dcrdata/pull/1016/files#diff-b4bda758a2aef091432646c354b4dc59R597.

for `retrieveRootPath(192.168.0.2:1234/api/) -> error`
As documented here https://github.com/golang/go/issues/21415#issuecomment-321966574 and [here](https://github.com/decred/dcrdata/pull/1016/files#diff-fb6a7899d06670e0e85669d2652ba911R248). I don't think making this exception is okay because this error is returned by the official golang's `net/url` package. Making this exception could come to haunt us later and that is why I think we should not allow it.

dmigwi:
@buck54321 Here is why trying to reverse this `retrieveRootPath(192.168.0.2:1234/api/) -> error` is a bug-in-waiting.
https://play.golang.org/p/VTZu41i1nG-

buck54321:
All addresses need the protocol.

https://play.golang.org/p/DRVTc3efkdb

We certainly want to support connecting by IP:Port, but I guess adding the protocol is just required for IP addresses then.

main.go
```diff
@@ -506,9 +499,43 @@ func _main(ctx context.Context) error {
...
+	// creates a new http client needed to query Politeia API endpoints.
+	proposalsInstance, err := politeia.NewProposalsDB(cfg.PoliteiaAPIURL,
+		filepath.Join(cfg.DataDir, cfg.ProposalsFileName))
+	if err != nil {
+		return fmt.Errorf("failed to create new proposals db instance: %v", err)
+	}
+
+	// Retrieves newly added proposals and addes them to the proposals db.
+	if err = proposalsInstance.CheckProposalsUpdates(); err != nil {
+		return fmt.Errorf("updating proposals db failed: %v", err)
```
chappjc: This is a show-stopper.  dcrdata has to continue working when Politeia is down.

main.go
```diff
@@ -506,9 +499,43 @@ func _main(ctx context.Context) error {
...
+	// store and retrieve proposals data. Proposals votes is Off-Chain
+	// data stored in github repositories away from the decred blockchain. It also
+	// creates a new http client needed to query Politeia API endpoints.
+	proposalsInstance, err := politeia.NewProposalsDB(cfg.PoliteiaAPIURL,
+		filepath.Join(cfg.DataDir, cfg.ProposalsFileName))
+	if err != nil {
+		return fmt.Errorf("failed to create new proposals db instance: %v", err)
+	}
+
+	// Retrieves newly added proposals and addes them to the proposals db.
```
chappjc: Retrieves -> Retrieve (similarly for the above comments), addes (adds) -> add

```go
// Retrieve newly added proposals and add them to the proposals db.
```

explorer/explorer.go
```diff
@@ -558,13 +581,28 @@ func (exp *explorerUI) Store(blockData *blockdata.BlockData, msgBlock *wire.MsgB
...
+		// Update the charts data after every five blocks or if no charts data
+		// exists yet.
+		if newBlockData.Height%5 == 0 || cacheChartsData.Height() == -1 {
+			// This must be done after storing BlockInfo since that provides the
+			// explorer's best block height, which is used by prePopulateChartsData
+			// to decide if an update is needed.
+			go exp.prePopulateChartsData()
+		}
+
+		// Update the proposals db at intervals of 20 blocks as 5 blocks is too often.
```
chappjc: Don't need to say "as 5 blocks is too often" because 5 was ad hoc for the charts data anyway, and "too often" is not qualified (how is 20 chosen?).

explorer/explorer.go
```diff
@@ -558,13 +581,28 @@ func (exp *explorerUI) Store(blockData *blockdata.BlockData, msgBlock *wire.MsgB
...
+		// Also to update the Proposal db the politeia API is queried and can be
```
chappjc: Please rephrase this sentence:

```go
// Update the proposal DB.  This is run asynchronously since it involves a query to Politeia (a remote system) and we do not want to block execution.
```

xaur:
Don't forget to right-align numerical columns :)

dmigwi:
screenshot from 2019-02-27 01-21-03

gov/agendas/deployments.go
```diff
@@ -101,26 +102,31 @@ func (db *AgendaDB) loadAgenda(agendaID string) (*AgendaTagged, error) {
...
-				StartTime:      v.StartTime,
-				ExpireTime:     v.ExpireTime,
-				Status:         v.Status,
-				QuorumProgress: v.QuorumProgress,
-				Choices:        v.Choices,
-				VoteVersion:    voteInfo.VoteVersion,
-			})
-		}
+	if err != nil {
+		return
```
chappjc: Ah, crud, I think we should log err to help debug network connectivity.  Can you make the logging addition?

dmigwi: I had an error being logged there initially but because it was always logged when vote version scanning took place I thought doing away with it would be better.

dmigwi: The error made buck54321 think something was wrong https://github.com/decred/dcrdata/pull/1016/files/270b8a8bfa4fb616be3ef8e14d2631e1d7158be0#diff-4fe4b07854ee46957bec741a6797b39d

dmigwi: The link above seems broken, here is the screenshot though.
screenshot 2019-02-28 at 21 16 20

chappjc: gotcha, I was thinking of the http.Client.  This is OK for the RPC client.

dmigwi:
- This add two new pages under `/proposals` and `/propasal/{id}` url paths. Proposals listing data is queried from Politeia API endpoints.
- Restructured the agendas listing implementation so that It can be in  a shared 'gov' directory with proposal listing implementation.
- Tests for both proposals and agendas listing implementations were added.

dmigwi:
@chappjc 👆

PR #1017, opened 2019-02-15
rpcutils: fix BlockGate

chappjc:
Handle a cache miss in Block() if the waiter was too slow to retrieve
the block when notified.

Both signalHash and signalHeight require both height and hash inputs
so that they can avoid extra map lookups in blockWithHash and
hashAtHeight, respectively.

Make signalHash and signalHeight lock-free.  They are not thread-
safe -- no longer run them as goroutines. Instead have them
retrieve the wait channel, remove the wait channel from the list,
and launch at goroutine just to send the height/hash to each waiter.

WaitForHeight now checks if the block is cached, in addition to
checking fetchToHeight.

### DIFF
--- a/rpcutils/blockgate.go
+++ b/rpcutils/blockgate.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2017, The dcrdata developers
+// Copyright (c) 2017-2019, The dcrdata developers
 // See LICENSE for details.
 
 package rpcutils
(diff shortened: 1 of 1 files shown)


chappjc:
Tested on upgrade from dcrpg 3.5.5 to 3.10.0 (includes the agendas upgrade):

```
[INF] PSQL:  - 1137598 rows in agendas table (prune agendas table upgrade) were successfully upgraded in 739.25 sec
```

If all is still well on a fresh sync, this is good to go.

chappjc:
Also tested with:
- fresh sync
- catch up on startup
- catch up on startup where sqlite is at a lower height than psql (when a `fetchToHeight` is set for the blockgate that causes sqlite to pull blocks on demand)

PR #1019, opened 2019-02-16
Refine mempool page layout

gozart1:
* Make top margin the same as other pages ( If there's interest in increasing the amount of top margin for page headings, it should be done globally, so the heading is always at the same height between the various pages )
* Increase font size of overview stat values
* Rework layout of secondary card to same labeling style as primary card
* Reduce spacing between tables

Current:
screen shot 2019-02-16 at 10 17 59 pm

Proposed Changes:
proposed

proposed changes (mobile):
screen shot 2019-02-16 at 10 15 17 pm

### DIFF
--- a/views/mempool.tmpl
+++ b/views/mempool.tmpl
@@ -9,126 +9,111 @@
         data-target="mempool.mempool"
         {{template "mempoolDump" .}}
         >
-            <div class="py-4 h4">Mempool</div>
+            <div class="pb-2 h4">Mempool</div>
 
-            <div class="row mx-1 my-2">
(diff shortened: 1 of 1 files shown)


chappjc:
@buck54321 So move forward as is and iterate or do you feel it's best if @gozart1 does some tweaking now?

buck54321:
It's trivial to tweak it and not a big deal, so let's merge it. We should try to come together on some things before major release, though. I've been trying to read more about effective use of negative space. [This](https://usabilitygeek.com/effective-use-negative-space-web-design/) and [this](https://speckyboy.com/negative-space-web-design/) are a couple of quick reads. I'm not all that good at it yet, but I think that DCRData has both room for growth and unique challenges in this area.

PR #1026, opened 2019-02-18
rework block page

buck54321:
Restyles /block page. Compare to prototype from #451.

Comparisons with prototype: Instead of the coinbase reward field, which is listed below, I've included the size field on the left, and added the fees field on the details pane. I've removed the "Block Details" header and stole the description for the right pane, as "Block Hashes" was not very accurate. This left the page a little out of sync with the recently worked mempool page, so I'm certainly open to adding it back if we can find the right description. Where the prototype had a single table in which the transaction type was
selectable, the separate tables are left here and the table style from #982 has been applied. This keeps everything no-js friendly.

Adds a couple of methods for custom time display to `explorertypes.TimeDef`. The types are not mirrored in `dbtypes.TimeDef`.

Tweaks the way live confirmation messaages are calculated so that each `newblock.confirmations` controller target can have a custom text.

image


chappjc:
Lookin' nice.  Just a minor critique.

I'm not totally in love with the background for the block hash.  I like that it's giving the hash a feel of a variable or monospace data, but perhaps some tweaking on the background color or maybe the feel of the box.

Some thoughts about order of the Block Details items:

- Merkle and Stake Root probably have the least meaning to the average user.
- Most useful: Ticket Price, Pool Size, maybe PoW Diff
- Then Blk Version and Stk Version if there are upgrades happening
- Then Fees and Vote Bits (says how this block voted on the previous)
- Lastly the nonce and roots.

If the extra data decoded to anything meaningful, that would be interesting too, but that's for another day.

Also, I didn't look at the code yet, is the USD value the current rate (same for every block shown at that at given time)?

buck54321:
Yeah, I guess the fiat conversion is only useful for the most recent few blocks. It's changed to only show for those blocks within the last hour. I wonder if we should show something different for older blocks.

Good call on the table order. Things are now switched up a bit.

For the hash, I've switched to the lightest color in the palette and added a little vertical padding, which I think helped. Let me know what you think. 

I've updated the original image above with the new screenshots.

buck54321:
Trying block hash without background color. A few other little tweaks.

image

chappjc:
I'm 100% for this change.  @gozart1, any feedback before merging?

gozart1:
Suggest reducing the line height of values like Block Time & Total Sent, so that when their text wraps due to lack of horizontal space, the value is visually grouped together in the vertical spacing of the layout.  Can use the `lh1rem` class to achieve that.

gozart1:
Likewise for Merkle Root & Stake Root, using the `lh1rem` will improve the look when the text wraps.

buck54321:
image

PR #1032, opened 2019-02-19
api: add /block/.../raw endpoints

chappjc:
This adds the following endpoints:

- "/block/best/raw"
- "/block/{ind}/raw"
- "/block/hash/{hash}/raw"

These return a new JSON type defined by `api/types.BlockRaw`.


api/apiroutes.go
```diff
@@ -441,6 +443,37 @@ func (c *appContext) getBlockHeader(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	msgBlock, err := c.BlockData.GetBlockByHash(hash)
+	if err != nil {
+		apiLog.Errorf("Unable to get block %s: %v", hash, err)
+		http.Error(w, http.StatusText(422), 422)
+		return
+	}
+
+	var buf bytes.Buffer
```
jrick: I would use `strings.Builder` to avoid a `[]byte` to `string` copy

PR #1040, opened 2019-02-23
Order chainwork query by block height

### LINKED ISSUE #1004: chainwork query should order by height instead of time
Time does not have to be monotonically increasing with time, but chainwork must be with chain height.

https://github.com/decred/dcrdata/blob/7e7f38e54bb129ca2626669d4b8af2a415d1e005/db/dcrpg/internal/blockstmts.go#L201

buck54321:
Resolves #1004. I did verify that dygraphs handles out of order timestamps without issue.

### DIFF
--- a/db/dcrpg/internal/blockstmts.go
+++ b/db/dcrpg/internal/blockstmts.go
@@ -198,7 +198,7 @@ const (
 	UpdateBlockNextByNextHash = `UPDATE block_chain SET next_hash = $2 WHERE next_hash = $1;`
 
 	// Grab the timestamp and chainwork.
-	SelectChainWork = `SELECT time, chainwork FROM blocks WHERE is_mainchain = true ORDER BY time;`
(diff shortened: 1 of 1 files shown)


db/dcrpg/internal/blockstmts.go
```diff
@@ -198,7 +198,7 @@ const (
 	UpdateBlockNextByNextHash = `UPDATE block_chain SET next_hash = $2 WHERE next_hash = $1;`
 
 	// Grab the timestamp and chainwork.
-	SelectChainWork = `SELECT time, chainwork FROM blocks WHERE is_mainchain = true ORDER BY time;`
+	SelectChainWork = `SELECT time, chainwork FROM blocks WHERE is_mainchain = true ORDER BY height;`
```
buck54321: I actually wasn't certain under what conditions an ORDER BY column can be absent from the SELECT list, but PostgreSQL seems to be okay with it in this simple query.

chappjc: Unlike a group by or distinct, it's ok to order by anything even if not selected in the output.

PR #1042, opened 2019-02-25
rework tx page

buck54321:
Restyles /tx page. Compare to single-transaction, vote, and ticket prototypes from #451.

Fiat values are shown and tagged with `(today)` to indicate the conversion uses the current rate. This is different than what was done in #1026, where conversions are only shown for blocks < 1 hour old. Not sure which is better.

Adds a number of methods to the `explorer.TxInfo` to help clean up the template.

Adds a new icon for regular transactions.

Refactors `tx_controller`. Drops the `DCRThings` global object. Using actual number of tickets in ticket pool for calculating expiration probability rather than network parameter.

image

image


xaur:
How about using title "Transaction" instead of "Regular"?

PR #1048, opened 2019-02-25
view/home: display exchange rate with all big digits and two decimals

chappjc:
Show exchange rate like this:

image

instead of this (nevermind the number):

image

This is based on feedback from jy-p, who raises a good point. 90 cents shouldn't be a much smaller font.  Also, fractions of cents don't have any real significance.


buck54321:
Penny precision is more than enough considering our confidence in these numbers, but may want a workaround for `ExchangeCurrency = "BTC"`, which is possible, and would display as `0.00`.

http://localhost:7777/api/exchanges?indent=1&code=BTC

buck54321:
Alternative solution in #1049

chappjc:
Closing in favor of https://github.com/decred/dcrdata/pull/1049

PR #1049, opened 2019-02-26
more fiat conversions

buck54321:
Adds fiat conversions to the homepage for ticket price, coin supply, miner block reward, treasury block reward, and treasury balance. 

`exchanges` package gets `Conversion` type which has the `TwoDecimal` method for displaying its value, with a minimum of two decimal places and a minimum of three significant figures.

Values are updated live with websocket `newblock` updates.


chappjc:
Got a minor conflict now, @buck54321

chappjc:
Looks good.
image

PR #1051, opened 2019-02-27
cache: add cache package and AddressCache for dcrpg

chappjc:
Add a new `cache` package, starting with addresscache.go, which includes
`AddressCache` and `AddressCacheItem`.
Use `AddressCache` in `ChainDB`.
Remove `DevFundBalance` since it benefits from the general `AddressCache`.
Remove `addressCounts` since `AddressCache` includes balance.
To update an address' data, add `RetrieveAllMainchainAddressTxns` and
`RetrieveAllMainchainAddressMergedTxns`, which do not use `offset` or `N`.
Also cache transaction view type row counts via the new
`(*ChainDB).CountTransactions` function, which use cache or DB queries to
get the row counts for any of the supported txn view types.

Also:
- Use `vouts` table primary key in `SelectAddressUnspentWithTxn`.
- Avoid height query in `AddressUTXO`, using latest stored height instead.
- fix: `RetrieveAddressUTXOs` forgot the block time
- Rename `AddrTxnType` to `AddrTxnViewType`.
- Remove unnecessary `TimeDefs` in `scanAddressMergedRows` and
  `scanAddressQueryRows`.
- Add `Height() int64` to `explorer/explorerDataSource` interface to use the
  no-query getter. This speeds up `BlockHashPathOrIndexCtx` middleware.

### DIFF
--- a/db/dbtypes/types.go
+++ b/db/dbtypes/types.go
@@ -227,19 +227,6 @@ func (a AgendaStatusType) MarshalJSON() ([]byte, error) {
 	return json.Marshal(a.String())
 }
 
-// IsMerged indicates if the address transactions view type is a merged view. If
-// the type is invalid, a non-nil error is returned.
-func (a AddrTxnType) IsMerged() (bool, error) {
-	switch a {
-	case AddrTxnAll, AddrTxnCredit, AddrTxnDebit:
-		return false, nil
-	case AddrMergedTxn, AddrMergedTxnCredit, AddrMergedTxnDebit:
-		return true, nil
-	default:
-		return false, fmt.Errorf("unrecognized address transaction view: %v", a)
-	}
-}
-
 // AgendaStatusFromStr creates an agenda status from a string.
 func AgendaStatusFromStr(status string) AgendaStatusType {
 	switch strings.ToLower(status) {
--- a/db/dbtypes/types.go
+++ b/db/dbtypes/types.go
@@ -1068,15 +1068,14 @@ func ReduceAddressHistory(addrHist []*AddressRow) *AddressInfo {
 		}
 		coin := dcrutil.Amount(addrOut.Value).ToCoin()
 		tx := AddressTx{
-			Time:      addrOut.TxBlockTime,
-			InOutID:   addrOut.TxVinVoutIndex,
-			TxID:      addrOut.TxHash,
-			TxType:    txhelpers.TxTypeToString(int(addrOut.TxType)),
-			MatchedTx: addrOut.MatchingTxHash,
-			IsFunding: addrOut.IsFunding,
+			Time:           addrOut.TxBlockTime,
+			InOutID:        addrOut.TxVinVoutIndex,
+			TxID:           addrOut.TxHash,
+			TxType:         txhelpers.TxTypeToString(int(addrOut.TxType)),
+			MatchedTx:      addrOut.MatchingTxHash,
+			IsFunding:      addrOut.IsFunding,
+			MergedTxnCount: addrOut.MergedCount,
 		}
-
-		tx.MergedTxnCount = addrOut.MergedCount
 
 		if addrOut.IsFunding {
 			// Funding transaction
--- a/db/dbtypes/types.go
+++ b/db/dbtypes/types.go
@@ -258,12 +258,12 @@ func AgendaStatusFromStr(status string) AgendaStatusType {
 	}
 }
 
-// AddrTxnType enumerates the different transaction types as displayed by the
-// address page.
-type AddrTxnType int
+// AddrTxnViewType enumerates the different address transaction view types as
+// displayed by the address page.
+type AddrTxnViewType int
 
 const (
-	AddrTxnAll AddrTxnType = iota
+	AddrTxnAll AddrTxnViewType = iota
 	AddrTxnCredit
 	AddrTxnDebit
 	AddrMergedTxnDebit
@@ -272,8 +272,8 @@ const (
 	AddrTxnUnknown
 )
 
-// AddrTxnTypes is the canonical mapping from AddrTxnType to string.
-var AddrTxnTypes = map[AddrTxnType]string{
+// AddrTxnViewTypes is the canonical mapping from AddrTxnViewType to string.
+var AddrTxnViewTypes = map[AddrTxnViewType]string{
 	AddrTxnAll:          "all",
 	AddrTxnCredit:       "credit",
 	AddrTxnDebit:        "debit",
@@ -283,12 +283,25 @@ var AddrTxnTypes = map[AddrTxnType]string{
 	AddrTxnUnknown:      "unknown",
 }
 
-func (a AddrTxnType) String() string {
-	return AddrTxnTypes[a]
+func (a AddrTxnViewType) String() string {
+	return AddrTxnViewTypes[a]
 }
 
-// AddrTxnTypeFromStr attempts to decode a string into an AddrTxnType.
-func AddrTxnTypeFromStr(txnType string) AddrTxnType {
+// IsMerged indicates if the address transactions view type is a merged view. If
+// the type is invalid, a non-nil error is returned.
+func (a AddrTxnViewType) IsMerged() (bool, error) {
+	switch a {
+	case AddrTxnAll, AddrTxnCredit, AddrTxnDebit:
+		return false, nil
+	case AddrMergedTxn, AddrMergedTxnCredit, AddrMergedTxnDebit:
+		return true, nil
+	default:
+		return false, fmt.Errorf("unrecognized address transaction view: %v", a)
+	}
+}
+
+// AddrTxnViewTypeFromStr attempts to decode a string into an AddrTxnViewType.
+func AddrTxnViewTypeFromStr(txnType string) AddrTxnViewType {
 	txnType = strings.ToLower(txnType)
 	switch txnType {
 	case "all":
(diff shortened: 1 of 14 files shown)


db/dbtypes/types.go
```diff
@@ -258,12 +245,12 @@ func AgendaStatusFromStr(status string) AgendaStatusType {
 	}
 }
 
-// AddrTxnType enumerates the different transaction types as displayed by the
-// address page.
-type AddrTxnType int
+// AddrTxnViewType enumerates the different address transaction view types as
+// displayed by the address page.
+type AddrTxnViewType int
```
chappjc: The previous name implied it was the *transaction's* type (e.g. vote, revocation, etc), but it's better described as the page view.

db/dcrpg/insightapi.go
```diff
@@ -155,16 +148,6 @@ func (pgb *ChainDB) GetBlockHash(idx int64) (string, error) {
 	return hash, nil
 }
 
-// AddressBalance returns a AddressBalance for the specified address,
-// transaction count limit, and transaction number offset.
-func (pgb *ChainDB) AddressBalance(address string, N, offset int64) (*dbtypes.AddressBalance, error) {
-	_, balance, err := pgb.AddressHistoryAll(address, N, offset)
```
chappjc: This was a hack to get the balance via AddressHistoryAll.  The N and offset args did nothing.

db/dcrpg/internal/addrstmts.go
```diff
@@ -155,13 +160,13 @@ const (
...
-			tx_vin_vout_index,
-			pkscript
+			addresses.tx_vin_vout_index,
+			vouts.pkscript
 		FROM addresses
 		JOIN transactions ON
 			addresses.tx_hash = transactions.tx_hash
-		JOIN vouts ON addresses.tx_hash = vouts.tx_hash AND addresses.tx_vin_vout_index=vouts.tx_index
-		WHERE addresses.address=$1 AND addresses.is_funding = TRUE AND addresses.matching_tx_hash = '' AND valid_mainchain = TRUE
+		JOIN vouts ON addresses.tx_vin_vout_row_id = vouts.id
```
chappjc: There was no reason to join on tx_hash AND tx_index when `tx_vin_vout_row_id` is the primary key (`id`) in the `vouts` table when `is_funding=true` (and it is in the WHERE below).

db/dcrpg/pgblockchain.go
```diff
@@ -45,37 +46,6 @@ var (
 // agenda_votes table.
 var storedAgendas map[string]dbtypes.MileStone
 
-// DevFundBalance is a block-stamped wrapper for dbtypes.AddressBalance. It is
-// intended to be used for the project address.
-type DevFundBalance struct {
```
chappjc: Obsoleted

db/dcrpg/pgblockchain.go
```diff
@@ -365,20 +342,6 @@ func (db *ChainDBRPC) MissingSideChainBlocks() ([]dbtypes.SideChain, int, error)
 	return blocksToStore, nSideChainBlocks, nil
 }
 
-// addressCounter provides a cache for address balances.
-type addressCounter struct {
```
chappjc: obsoleted

db/dcrpg/pgblockchain.go
```diff
@@ -560,6 +523,9 @@ func NewChainDBWithCancel(ctx context.Context, dbi *DBInfo, params *chaincfg.Par
 		hash:   bestHash,
 	}
 
+	addrCache := cache.NewAddressCache(500)
+	addrCache.DevAddress = devSubsidyAddress
```
chappjc: project fund address is set to prevent purging its data from cache when cache reaches capacity.

db/dcrpg/pgblockchain.go
```diff
@@ -820,6 +786,29 @@ func (block *BestBlock) Hash() *chainhash.Hash {
...
+	return pgb.bestBlock.hash, pgb.bestBlock.height
+}
+
+// BestBlockHash is a getter for ChainDB.bestBlock.hash.
+func (pgb *ChainDB) BestBlockHash() *chainhash.Hash {
+	return pgb.bestBlock.Hash()
+}
+
+// BestBlockHashStr is a getter for ChainDB.bestBlock.hash.
+func (pgb *ChainDB) BestBlockHashStr() string {
```
chappjc: We were querying the DB too much when the best block data is saved on Store.

db/dcrpg/pgblockchain.go
```diff
@@ -1277,236 +1287,356 @@ func (pgb *ChainDB) ticketPoolVisualization(interval dbtypes.TimeBasedGrouping)
 	return
 }
 
-// retrieveDevBalance retrieves a new DevFundBalance without regard to the cache
-func (pgb *ChainDB) retrieveDevBalance() (*DevFundBalance, error) {
-	bb, hash, err := pgb.HeightHashDB()
+func (pgb *ChainDB) updateProjectFundCache() error {
+	// Update non-merged rows.
+	_, _, err := pgb.AddressHistory(pgb.devAddress, 1, 0, dbtypes.AddrTxnAll)
```
chappjc: AddressHistory is used to update both balance and rows.

db/dcrpg/pgblockchain.go
```diff
@@ -1277,236 +1287,356 @@ func (pgb *ChainDB) ticketPoolVisualization(interval dbtypes.TimeBasedGrouping)
...
-			go func() {
-				runtime.Gosched()
-				if err := updateBalance(); err != nil {
-					log.Error(err)
-				}
-			}()
-			return nil
-		}
-		return updateBalance()
+func (pgb *ChainDB) FreshenAddressCaches(lazyProjectFund bool, expireAddresses []string) error {
```
chappjc: I should document `expireAddresses`

db/dcrpg/pgblockchain.go
```diff
@@ -2540,6 +2658,17 @@ func (pgb *ChainDB) StoreBlock(msgBlock *wire.MsgBlock, winningTickets []string,
 	numVouts = errStk.numVouts + errReg.numVouts
 	numAddresses = errStk.numAddresses + errReg.numAddresses
 
+	// Merge the affected addresses.
```
chappjc: should explain this is for cache expiration

db/dcrpg/pgblockchain.go
```diff
@@ -2578,9 +2707,9 @@ func (pgb *ChainDB) StoreBlock(msgBlock *wire.MsgBlock, winningTickets []string,
 		return
 	}
 
-	// If not in batch sync, lazy update the dev fund balance
+	// If not in batch sync, lazy update the dev fund balance.
```
chappjc: should update this comment too.

db/dcrpg/queries.go
```diff
@@ -1585,12 +1616,11 @@ func scanAddressQueryRows(rows *sql.Rows, queryType int) (ids []uint64, addressR
 	for rows.Next() {
 		var id uint64
 		var addr dbtypes.AddressRow
-		var txHash sql.NullString
-		var blockTime dbtypes.TimeDef
+		var matchingTxHash sql.NullString
```
chappjc: `txHash` is never nul or empty.  `matchingTxHash` is often empty, but I'm not even sure it is allowed to be null.  This is just to be safe.

db/dcrpg/queries.go
```diff
@@ -1599,8 +1629,10 @@ func scanAddressQueryRows(rows *sql.Rows, queryType int) (ids []uint64, addressR
 
 		switch queryType {
 		case creditQuery:
+			// addr.IsFunding == true
 			addr.AtomsCredit = addr.Value
 		case debitQuery:
+			// addr.IsFunding == false
```
chappjc: Can remove these two commented lines.  IsFunding is set directly by rows.Scan

explorer/explorermiddleware.go
```diff
@@ -63,14 +63,7 @@ func (exp *explorerUI) BlockHashPathOrIndexCtx(next http.Handler) http.Handler {
...
 			} else {
-				maxHeight, err = exp.explorerSource.HeightDB()
-				if err != nil {
-					log.Errorf("HeightDB() failed: %v", err)
-					exp.StatusPage(w, defaultErrorCode,
-						"an unexpected error had occured while retrieving the best block",
-						"", ExpStatusError)
-					return
-				}
+				maxHeight = exp.explorerSource.Height()
```
chappjc: Another unnecessary DB query.

db/dcrpg/pgblockchain.go
chappjc: In #1053, these calls to `AddressHistory` should be replaced by `AddressBalance` and two `AddressRows` calls.

db/dcrpg/pgblockchain.go
```diff
@@ -1277,236 +1289,358 @@ func (pgb *ChainDB) ticketPoolVisualization(interval dbtypes.TimeBasedGrouping)
...
+	if err != nil {
+		return 0, err
+	}
+	return count, nil
 }
 
 // AddressHistory queries the database for rows of the addresses table
 // containing values for a certain type of transaction (all, credits, or debits)
 // for the given address.
 func (pgb *ChainDB) AddressHistory(address string, N, offset int64,
```
buck54321: Function description needs updating.

db/dcrpg/pgblockchain.go
```diff
@@ -1582,35 +1711,31 @@ func (db *ChainDBRPC) AddressData(address string, limitN, offsetAddrOuts int64,
...
-			case dbtypes.AddrMergedTxn:
-				addrData.TxnCount, err = CountMergedTxns(ctx, db.db, address)
-			}
+		// Obtain the TxnCount, which pertains to the number of table rows.
+		addrData.IsMerged, err = txnType.IsMerged()
+		if err != nil {
+			return nil, err
+		}
+		if addrData.IsMerged {
+			// For merged views, either query the DB or check the process the
```
buck54321: extra words

PR #1053, opened 2019-02-27
caching and query tuning for InsightAddressTransactions

chappjc:
This accelerates InsightAddressTransactions, which is used for several
Insight API endpoints.

`InsightAddressTransactions` is modified to use the `rowsMerged` cache,
first checking that each address is in the cache, and then merging the
results as would be done by the DB query.

Add `(*ChainDB).AddressRows` to get the cached address rows, or query the
DB and update the cache.
Use cache-enabled `AddressRows` for `AddressTxIoCsv`.
Also add `dcrpg.MakeCsvAddressRows` to convert from `[]*dbtypes.AddressRow`
to [][]string for a CSV output.

Use AddressRows and AddressBalance in updateProjectFundCache instead
of the AddressHistory cludge.

Etc.:
- Remove the `JOIN` from `SelectAddressesAllTxn` since `block_time` is
  already in the addresses table and sufficient for getting a "recent"
  transactions slice for mempool purposes.
- Add `SelectBlockTimeByHeight` query for getting the main chain block time
  by height, which is then used in `InsightAddressTransactions`.
- Add `RetrieveBlockTimeByHeight`, using `SelectBlockTimeByHeight`.
- `RetrieveAddressTxnsOrdered` now accepts a recent block time instead of
  height.  As such, `InsightAddressTransactions` first queries for the time
  of the block at a recent block height (callers choice).

### DIFF
--- a/db/dcrpg/insightapi.go
+++ b/db/dcrpg/insightapi.go
@@ -56,12 +56,46 @@ func (pgb *ChainDBRPC) SendRawTransaction(txhex string) (string, error) {
 }
 
 // InsightAddressTransactions performs a db query to pull all txids for the
-// specified addresses ordered desc by time.
-func (pgb *ChainDB) InsightAddressTransactions(addr []string, recentBlockHeight int64) ([]string, []string, error) {
+// specified addresses ordered desc by time. It also returns a list of recently
+// (defined as greater than recentBlockHeight) confirmed transactions that can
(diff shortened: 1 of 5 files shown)


chappjc:
There's a nasty stack overflow in master right now that I think this resolves.  Merging and deploying on alpha.

PR #1063, opened 2019-03-01
db/dcrpg: Remove non-determinist VersionCheck implementation

dmigwi:
- Adds drop duplicates implementation in agendas and agenda_votes tables
- Adds a more deterministic way to check if the auxiliary db requires an upgrade, re index or rebuild.

### DIFF
--- a/db/dcrpg/pgblockchain.go
+++ b/db/dcrpg/pgblockchain.go
@@ -612,7 +612,7 @@ func (pgb *ChainDB) VersionCheck(client *rpcclient.Client) error {
 
 	for _, val := range tableUpgrades {
 		switch val.UpgradeType {
-		case Upgrade, ReIndex:
+		case Upgrade, Reindex:
 			// Select the all tables that need an upgrade or reindex.
 			needsUpgrade = append(needsUpgrade, val)
 
@@ -628,8 +628,8 @@ func (pgb *ChainDB) VersionCheck(client *rpcclient.Client) error {
 		return nil
 	}
 
-	// AddAuxPendingDBUpgrades adds the pending db upgrades and reindexes.
-	_, err := pgb.AddAuxPendingDBUpgrades(client, needsUpgrade[0].CurrentVer,
+	// UpgradeTables adds the pending db upgrades and reindexes.
+	_, err := pgb.UpgradeTables(client, needsUpgrade[0].CurrentVer,
 		needsUpgrade[0].RequiredVer)
 	if err != nil {
 		return err
--- a/db/dcrpg/pgblockchain.go
+++ b/db/dcrpg/pgblockchain.go
@@ -607,34 +607,35 @@ func (pgb *ChainDB) VersionCheck(client *rpcclient.Client) error {
 		log.Debugf("Table %s: v%s", tab, ver)
 	}
 
-	if tableUpgrades := TableUpgradesRequired(vers); len(tableUpgrades) > 0 {
-		if tableUpgrades[0].UpgradeType == "upgrade" || tableUpgrades[0].UpgradeType == "reindex" {
-			// CheckForAuxDBUpgrade makes db upgrades that are currently supported.
-			isSuccess, err := pgb.CheckForAuxDBUpgrade(client)
-			if err != nil {
-				return err
-			}
-			// Upgrade was successful, no need to proceed.
-			if isSuccess {
-				return nil
-			}
-		}
+	needsUpgrade := make([]TableUpgrade, 0)
+	tableUpgrades := TableUpgradesRequired(vers)
 
-		// ensure all tables have "ok" status
-		OK := true
-		for _, u := range tableUpgrades {
-			if u.UpgradeType != "ok" {
-				log.Warnf(u.String())
-				OK = false
-			}
-		}
-		if OK {
-			log.Debugf("All tables at correct version (%v)", tableUpgrades[0].RequiredVer)
-			return nil
-		}
+	for _, val := range tableUpgrades {
+		switch val.UpgradeType {
+		case Upgrade, ReIndex:
+			// Select the all tables that need an upgrade or reindex.
+			needsUpgrade = append(needsUpgrade, val)
 
-		return fmt.Errorf("rebuild of PostgreSQL tables required (drop with rebuilddb2 -D)")
+		case Unknown, Rebuild:
+			// All the tables require rebuilding.
+			return fmt.Errorf("rebuild of PostgreSQL tables required (drop with rebuilddb2 -D)")
+		}
 	}
+
+	if len(needsUpgrade) == 0 {
+		// All tables have the correct version.
+		log.Debugf("All tables at correct version (%v)", tableUpgrades[0].RequiredVer)
+		return nil
+	}
+
+	// AddAuxPendingDBUpgrades adds the pending db upgrades and reindexes.
+	_, err := pgb.AddAuxPendingDBUpgrades(client, needsUpgrade[0].CurrentVer,
+		needsUpgrade[0].RequiredVer)
+	if err != nil {
+		return err
+	}
+
+	// Upgrade was successful.
 	return nil
 }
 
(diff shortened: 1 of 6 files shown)


db/dcrpg/pgblockchain.go
```diff
@@ -606,34 +606,35 @@ func (pgb *ChainDB) VersionCheck(client *rpcclient.Client) error {
...
-			isSuccess, err := pgb.CheckForAuxDBUpgrade(client)
-			if err != nil {
-				return err
-			}
-			// Upgrade was successful, no need to proceed.
-			if isSuccess {
-				return nil
-			}
-		}
+	needsUpgrade := make([]TableUpgrade, 0)
```
chappjc: Why `make([]TableUpgrade, 0)`?  in terms of appending, it's equally valid to append to a nil slice and a non-nil zero length slice.  Did you mean to do `make([]TableUpgrade, 0, len(vers))`

db/dcrpg/queries.go
```diff
@@ -212,10 +212,34 @@ func DeleteDuplicateMisses(db *sql.DB) (int64, error) {
 	} else if isuniq {
 		return 0, nil
 	}
-	execErrPrefix := "failed to delete duplicate misses: "
+	execErrPrefix := "failed to delete duplicate in misses: "
```
chappjc: It would read fine as "duplicate misses" or "duplicates in misses"

db/dcrpg/tables.go
```diff
@@ -74,34 +74,61 @@ type TableVersion struct {
...
-		return "ok"
+		return OK
 	}
 }
 
 func (s TableVersion) String() string {
 	return fmt.Sprintf("%d.%d.%d", s.major, s.minor, s.patch)
 }
 
+// CompatibilityAction default stringer
```
chappjc: will make linters mad if it does not start with `// String `, so maybe `// String implements Stringer for CompatibilityAction` (the `type Stringer interface` in the `fmt` package)/

db/dcrpg/pgblockchain.go
dmigwi: I always forget this.

db/dcrpg/queries.go
```diff
@@ -216,6 +216,30 @@ func DeleteDuplicateMisses(db *sql.DB) (int64, error) {
...
+// DeleteDuplicateAgendaVotes deletes rows in misses with duplicate votes-row-id,
+// and agendas-row-id leaving the one row with the lowest id.
+func DeleteDuplicateAgendaVotes(db *sql.DB) (int64, error) {
+	if isuniq, err := IsUniqueIndex(db, "uix_agenda_votes"); err != nil && err != sql.ErrNoRows {
+		return 0, err
+	} else if isuniq {
+		return 0, nil
+	}
+	execErrPrefix := "failed to delete duplicate agenda_votes: "
+	return sqlExec(db, internal.DeleteAgendasDuplicateRows, execErrPrefix)
```
chappjc: Should be using `internal.DeleteAgendaVotesDuplicateRows`.

db/dcrpg/queries.go
```diff
@@ -216,6 +216,30 @@ func DeleteDuplicateMisses(db *sql.DB) (int64, error) {
 	return sqlExec(db, internal.DeleteMissesDuplicateRows, execErrPrefix)
 }
 
+// DeleteDuplicateAgendas deletes rows in misses with duplicate names leaving
```
chappjc: Needs to refer to "agendas" instead of "misses"

db/dcrpg/queries.go
```diff
@@ -216,6 +216,30 @@ func DeleteDuplicateMisses(db *sql.DB) (int64, error) {
...
+	if isuniq, err := IsUniqueIndex(db, "uix_agendas_name"); err != nil && err != sql.ErrNoRows {
+		return 0, err
+	} else if isuniq {
+		return 0, nil
+	}
+	execErrPrefix := "failed to delete duplicate agendas: "
+	return sqlExec(db, internal.DeleteAgendasDuplicateRows, execErrPrefix)
+}
+
+// DeleteDuplicateAgendaVotes deletes rows in misses with duplicate votes-row-id,
```
chappjc: Needs to refer to "agenda_votes" instead of "misses"

db/dcrpg/tables.go
```diff
@@ -74,34 +74,61 @@ type TableVersion struct {
...
 }
 
+// CompatibilityAction defines the action to be taken once the current and the
+// required pg table versions are compared.
+type CompatibilityAction int8
+
+const (
+	Rebuild CompatibilityAction = iota
+	Upgrade
+	ReIndex
```
chappjc: Just `Reindex` please.

db/dcrpg/upgrades.go
```diff
@@ -78,20 +78,11 @@ var votingMilestones = map[string]dbtypes.MileStone{}
...
-	upgradeInfo := TableUpgradesRequired(TableVersions(pgb.db))
-
-	if len(upgradeInfo) > 0 {
-		version = upgradeInfo[0].CurrentVer
-		needVersion = upgradeInfo[0].RequiredVer
-	} else {
-		return false, nil
-	}
-
+// AddAuxPendingDBUpgrades upgrades all the tables with the pending updates from
```
chappjc: The name of the function is confusing to me.  Since it's a method of `ChainDB`, no need for the "Aux" part.  Also, I'm confused about the "Add" part.  What does that imply?  From the description, I think `UpgradeTables` would be appropriate.

db/dcrpg/upgrades.go
```diff
@@ -401,8 +389,8 @@ func (pgb *ChainDB) CheckForAuxDBUpgrade(dcrdClient *rpcclient.Client) (bool, er
 	}
 
 	// Ensure the required version was reached.
-	upgradeInfo = TableUpgradesRequired(TableVersions(pgb.db))
-	if len(upgradeInfo) > 0 && upgradeInfo[0].UpgradeType != "ok" {
+	upgradeInfo := TableUpgradesRequired(TableVersions(pgb.db))
+	if len(upgradeInfo) > 0 && upgradeInfo[0].UpgradeType != OK {
```
chappjc: This is still randomly ordered because `TableUpgradesRequired` loops over the map returned by `TableVersions`. If the first table in the loop is OK, then there will still be no upgrade. This needs to check if any of the tables require an upgrade.

db/dcrpg/pgblockchain.go
```diff
@@ -628,8 +628,8 @@ func (pgb *ChainDB) VersionCheck(client *rpcclient.Client) error {
 		return nil
 	}
 
-	// AddAuxPendingDBUpgrades adds the pending db upgrades and reindexes.
-	_, err := pgb.AddAuxPendingDBUpgrades(client, needsUpgrade[0].CurrentVer,
+	// UpgradeTables adds the pending db upgrades and reindexes.
+	_, err := pgb.UpgradeTables(client, needsUpgrade[0].CurrentVer,
```
chappjc: The ordering issue is here too.

dmigwi: On this one I might use some more clarification. `needsUpgrade` only holds table version that require `Upgrade` or `Reindex`. So if the code got here then an upgrade or reindexing must happen.

What could possibly be affected is where that upgrade starts from but from our earlier discussion only a newly created table(with upgrade type `OK`) has different current and required table version since the rest have similar table versions.

Unless how we fetch the table versions is buggy (which could make it harder to determine the current and required version) IMO I think this should be okay.

chappjc: Yes, you are correct.  Thanks for going over that.

chappjc:
Did you have a chance to test with a db at 3.5.5?  It's not easy for me to test this now.

dmigwi:
I made the mistake of deleting all my old db backups sometimes last month. I don't have a 3.5.5 db backup thus did not test it.

chappjc:
That's OK.  I'll sync dcrdata 3.x tonight and try it in the a.m.

PR #1064, opened 2019-03-01
deps: remove btcsuite deps

dajohi:
- switch to jessevdk/go-flags to match other dcr repos.
- drop bitcoin support.


chappjc:
Why drop bitcoin address detection?

txhelpers/txhelpers.go
```diff
@@ -979,27 +977,12 @@ const (
...
-// decoding fails, AddressErrorDecodeFailed is returned with AddressTypeUnknown.
-// If the address is a Bitcoin address, AddressErrorBitcoin is returned with
-// AddressTypeBitcoin. If the address decoded successfully as a Decred address,
-// it is checked against the specified network. If it is the wrong network,
-// AddressErrorWrongNet is returned with AddressTypeUnknown. If the address is
-// the correct network, the address type is obtained. A final check is performed
-// to determine if the address is the zero pubkey hash address, in which case
-// AddressErrorZeroAddress is returned with the determined address type. If it
-// is another address, AddressErrorNoError is returned with the determined
-// address type.
```
chappjc: The other comments about `AddressErrorWrongNet`, etc. are still valid

dajohi: Right, but do we need to list all of the checks in the comment?  If you pass an invalid address, the error will tell you why.

dajohi: We are pulling a lot of code (btcsuite) to just tell the user they entered a Bitcoin address instead of a Decred address.   We don't care about bitcoin.

chappjc: Ok about bitcoin, but I think the comments describe the expected behavior and I don't want to remove them.

chappjc:
@raedah Thoughts about removing Bitcoin address detection?  I think it might have been your suggestion.

xaur:
The only scenario that comes to mind is if people start doing more swaps and accidently enter BTC address. But even then, a general "bad address" error should be enough for them to figure they did something wrong.

PR #1065, opened 2019-03-01
Remove testfile output

dajohi:


chappjc:
This is the test data that is used by `TestTicketPoolPersistent`.  The test fail without it.

PR #1066, opened 2019-03-01
stakedb: close db before returning error.

dajohi:


jrick:
There's a cleaner way:

```
func Foo() (err error) {
    db, err := openDB()
    if err != nil {
        return err
    }
    defer func() {
        if err != nil {
            db.Close()
        }
    }()

    // ...
    if err != nil {
        return err // will close db
    }

    return nil // won't close db
}
```

dajohi:
updated to @jrick code

stakedb/ticketpool.go
```diff
@@ -151,6 +151,15 @@ func NewTicketPool(dataDir, dbSubDir string) (*TicketPool, error) {
 	if err != nil {
 		return nil, fmt.Errorf("failed badger.Open: %v", err)
 	}
+	defer func() {
+		if r := recover(); r != nil {
+			db.Close()
+			panic(r)
+		}
+		if err != nil {
```
jrick: not the same error. you need a named error return for this trick to work.

PR #1069, opened 2019-03-02
multi: do not export the mutex

dajohi:
It is the responsibility of the pkg to protect internal data, not the caller.


api/types/apicache.go
```diff
@@ -468,7 +468,7 @@ type blockHeap []*CachedBlock
 
 // BlockPriorityQueue implements heap.Interface and holds CachedBlocks
 type BlockPriorityQueue struct {
-	*sync.RWMutex
+	mtx                  sync.RWMutex
```
chappjc: Must be a pointer because the Sort interface copies the mutex.

dajohi: changed func instead.

chappjc: Changing to pointer receivers works too.  Is this fundamentally better though?

PR #1070, opened 2019-03-02
remove unused file

dajohi:


chappjc:
Some trivia - blockmemdb.go was the very first "db" so I could test an interface idea before implementing sqlite.  All done with that I'd say!

PR #1071, opened 2019-03-02
mempool: prevent double rlock

dajohi:
The mempool monitor is already read locked, so just access directly.

### DIFF
--- a/mempool/monitor.go
+++ b/mempool/monitor.go
@@ -183,7 +183,7 @@ func (p *MempoolMonitor) TxHandler(client *rpcclient.Client) {
 						Choices:     choices,
 						TicketSpent: msgTx.TxIn[1].PreviousOutPoint.Hash.String(),
 					}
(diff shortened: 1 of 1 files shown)


dajohi:
```POTENTIAL DEADLOCK: Duplicate locking, saw callers this locks in one goroutine:                                     
current goroutine 149791 lock &{{{0 0} 0 0 1 0}}
all callers to this lock in the goroutine
git/dcrdata/mempool/monitor.go:153 mempool.(*MempoolMonitor).TxHandler { p.mtx.RLock()      // do not allow p.inventory to be reset } <<<<<

git/dcrdata/mempool/monitor.go:92 mempool.(*MempoolMonitor).LastBlockHash { p.mtx.RLock() } <<<<<                   
git/dcrdata/mempool/monitor.go:187 mempool.(*MempoolMonitor).TxHandler { voteInfo.ForLastBlock = voteInfo.VotesOnBlock(p.LastBlockHash().String()) }
```

PR #1072, opened 2019-03-02
share and protect API Status struct

dajohi:
With appContext, protect with its own mutex.
With insightApiContext, no need for a mutex since it is never
modified after creation.

### DIFF
--- a/api/apiroutes.go
+++ b/api/apiroutes.go
@@ -120,6 +120,7 @@ type appContext struct {
 	BlockData     DataSourceLite
 	AuxDataSource DataSourceAux
 	LiteMode      bool
+	statusMtx     sync.RWMutex
 	Status        apitypes.Status
 	JSONIndent    string
 	xcBot         *exchanges.ExchangeBot
@@ -171,7 +172,7 @@ out:
 				break out
 			}
 
-			c.Status.Lock()
+			c.statusMtx.Lock()
 			c.Status.Height = height
 			// If DB height agrees with node height, then we're ready.
 			c.Status.Ready = c.Status.Height == c.Status.DBHeight
@@ -180,11 +181,11 @@ out:
 			c.Status.NodeConnections, err = c.nodeClient.GetConnectionCount()
 			if err != nil {
 				c.Status.Ready = false
-				c.Status.Unlock()
+				c.statusMtx.Unlock()
 				log.Warn("Failed to get connection count: ", err)
 				break keepon
 			}
-			c.Status.Unlock()
+			c.statusMtx.Unlock()
 
 		case height, ok := <-notify.NtfnChans.UpdateStatusDBHeight:
 			if !ok {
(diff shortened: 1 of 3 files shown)


api/apiroutes.go
```diff
@@ -1454,8 +1459,12 @@ func (c *appContext) addressIoCsv(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	c.statusMtx.RLock()
+	statusHeight := c.Status.Height
+	c.statusMtx.RUnlock()
+
 	filename := fmt.Sprintf("address-io-%s-%d-%s.csv", address,
-		c.Status.GetHeight(), strconv.FormatInt(time.Now().Unix(), 10))
+		statusHeight, strconv.FormatInt(time.Now().Unix(), 10))
```
chappjc: This is actually why I avoid adding named mutexes that apply to a specific field of the struct.  Having the Status type have it's own mutex makes ensures the mutex is only used for Status accesses.  Plus the `GetHeight()` method does the locking for you, and there's no temp variable.  All around it's cleaner IMO.

api/insight/apiroutes.go
```diff
@@ -472,7 +472,7 @@ func (c *insightApiContext) getTransactions(w http.ResponseWriter, r *http.Reque
 		}
 		addresses := []string{address}
 		rawTxs, recentTxs, err :=
-			c.BlockData.ChainDB.InsightAddressTransactions(addresses, int64(c.Status.GetHeight()-2))
+			c.BlockData.ChainDB.InsightAddressTransactions(addresses, int64(c.Status.Height-2))
```
chappjc: It's definitely a bug with `insightApiContext` never updates any of the status fields.  The idea is probably to have the `insightApiContext` have a pointer to the same Status in `appContext`.

api/apiroutes.go
```diff
@@ -338,7 +339,11 @@ func (c *appContext) coinSupply(w http.ResponseWriter, r *http.Request) {
 
 func (c *appContext) currentHeight(w http.ResponseWriter, _ *http.Request) {
 	w.Header().Set("Content-Type", "text/plain; charset=utf-8")
-	if _, err := io.WriteString(w, strconv.Itoa(int(c.Status.GetHeight()))); err != nil {
+	c.statusMtx.RLock()
+	statusHeight := c.Status.Height
+	c.statusMtx.RUnlock()
+
+	if _, err := io.WriteString(w, strconv.Itoa(int(statusHeight))); err != nil {
```
chappjc: Like the other example, this is far less clean.

api/apiroutes.go
dajohi: The fields are exported so the API allows you to bypass the function calls that keep data integrity.  Perhaps a third solution is needed :)

chappjc: True, there is ample opportunity for messing with the data the way it is set up.  Unexporting the fields and making a setter method is probably the third method then.

chappjc:
As per the [review discussion](https://github.com/decred/dcrdata/pull/1072#discussion_r261841262), the plan of action is now to unexport the status fields, and make thread safe accessors.  There was already a getter, leaving a needed setter to be implemented.  Finally, sharing the status struct between dcrdata api and insight api is needed.

chappjc:
Closing in favor of a future PR addressing the related issue https://github.com/decred/dcrdata/issues/1094

Issue #1009, opened 2019-02-15
add block version to /blocks list


raedah:

xaur:
That's useful on individual block page, but wouldn't it be too verbose for blocks list most of the time (except consensus vote times)?

buck54321:
Viewing them collectively gives you a sense of how many nodes have upgraded. Is that the idea?

chappjc:
I think that's the idea.  Although once the upgrade happens (threshold met on the 1000 block rolling window), the info will become far less interesting.

xaur:
If the idea is to give a sense of how many nodes have upgraded then it overlaps with voting.decred.org.

raedah:
Its true it wouldnt always be so useful. Would be useful during an upgrade. In general would also increase awareness that blocks have a version and what the current version is. Its also a very thin field requiring 2 chars width, maybe a bit more to say Version in the header, or could abbreviate to Ver. Regarding the voting website, there is in general a lot of overlap with it and dcrdata and the voting site still leaves a lot to be desired.

buck54321:
We can definitely get the block version in there. Those tables need some attention anyway. Here's the current mobile view.

![image](https://user-images.githubusercontent.com/6109680/53560280-9f068c00-3b11-11e9-8057-e8a115099660.png)


The information on voting.decred.org should also be available through dcrdata. We should also be displaying some distillation on the homepage. Even casual visitors should know that right now a vote has been triggered and voting commences in ~14 days. Some form of the same information could also be displayed at the top of the blocks page, in which case I think we could remove the block version from the table again.

xaur:
> Even casual visitors should know that right now a vote has been triggered and voting commences in ~14 days

This triggered decred/dcrweb#579 as an extra place to remind about CVs.

chappjc:
resolved by https://github.com/decred/dcrdata/pull/1081

PR #1081, opened 2019-03-03
explorer: adjust blocks tables

### LINKED ISSUE #1030: Right-align numbers on blocks page
Please right-align all numbers on https://explorer.dcrdata.org/blocks , basically all columns except Time.

dcrdata v3.2.0-release

buck54321:
Makes adjustments to the tables on /blocks, /ticketpricewindows, and the time-based block groupings pages.

Adds block version column to /blocks (#1009).  /blocks has `explorertypes.BlockBasic`, so `Version` and `AllTxs` fields of `explorertypes.BlockInfo` graduated to fields in `BlockBasic`, which it already embeds. Appropriate changes were made in `makeExplorerBlockBasic`, including a change to remove an extraneous revocation counting loop. A `TxCount` field is added to `dbtypes.BlocksGroupedInfo` as well, for the grouped blocks listings.

JS: `humanize.formatTxDate` updated to follow up changes in #988, and renamed to `date`. `blocklist_controller` updated for new "version" field in websocket block.

Style changes to tables and pagination controls, mostly for mobile-friendliness. On smallest screen sizes, individual counts for 4 transaction types are hidden in favor of a single tx count column.

Also resolves #1030


explorer/explorerroutes.go
```diff
@@ -477,6 +477,12 @@ func (exp *explorerUI) timeBasedBlocksListing(val string, w http.ResponseWriter,
 		return
 	}
 
+	// If the view is "years" and the top row is this year, modify the formatted
+	// time string to indicate its a partial result.
+	if val == "Years" && len(data) > 0 && data[0].EndTime.T.Year() == time.Now().Year() {
+		data[0].FormattedStartTime = fmt.Sprintf("%s so far", time.Now().Format("2006"))
```
buck54321: Doing it this way prevents the need for a wide "end date" column on /years.

chappjc:
#1077 is in now.

explorer/explorerroutes.go
chappjc: Makes a lot of sense.  I'm just wondering if YTD or "year to date" might be clearer than say "2019 so far".  I'm OK with this though.

buck54321: Year to date is much better. Should it be hyphenated, "year-to-date" or not, "year to date"?

chappjc: Good with YTD as you set it.  Let's see if we get comments when live.

PR #1082, opened 2019-03-03
pubsub/explorer: mempool data cleanup

dajohi:


pubsub/pubsubhub.go
```diff
@@ -89,32 +76,24 @@ type connection struct {
...
+	invsMtx    sync.RWMutex
+	invs       *exptypes.MempoolInfo
 }
 
 // NewPubSubHub constructs a PubSubHub given a primary and auxiliary data
 // source. The primary data source is required, while the aux. source may be
 // nil, which indicates a "lite" mode of operation. The WebSocketHub is
 // automatically started.
-func NewPubSubHub(dataSource wsDataSource, auxDataSource wsDataSourceAux) *PubSubHub {
+func NewPubSubHub(dataSource wsDataSource, liteMode bool) (*PubSubHub, error) {
```
chappjc: We can get rid of the `liteMode` field and this input argument too.  If we need the aux DB in the future, we can a bring it back.

dajohi: removed.

pubsub/pubsubhub.go
```diff
@@ -123,7 +100,7 @@ func NewPubSubHub(dataSource wsDataSource, auxDataSource wsDataSourceAux) *PubSu
 	// Development subsidy address of the current network
 	devSubsidyAddress, err := dbtypes.DevSubsidyAddress(params)
 	if err != nil {
-		log.Warnf("NewPubSubHub: bad project fund address (%v)", err)
```
chappjc: So it was actually just a warning because in testnet2 there was no dev subsidy address, but we didn't want to break everything in dcrdata.  Does it hurt to ignore this error?  I don't expect a future testnet will have this problem, but who knows.

davecgh: There have been tests added to prevent that issue in the future, so I highly doubt it will be an issue.  Also, I'd rather just fix testnet out of the gate if there is some issue like that in the future instead of adding hacks in other software like dcrdata.

chappjc: Music to my ears.

explorer/explorer.go
```diff
@@ -254,6 +247,9 @@ type explorerUI struct {
 	// page that should be accessible during DB synchronization.
 	displaySyncStatusPage atomic.Value
 	politeiaAPIURL        string
+
+	invsMtx sync.RWMutex
+	invs    *types.MempoolInfo
```
chappjc: Well, I approved but Travis did not.  The Mempool type and it's mutex are unused.

Also, I generally don't like the pattern of a named mutex for a certain field.  I see no issue with embedding a mutex in the type you are locking access on.  Either way it's the user's responsibility to use the lock, but with it in the data type, it's clearer what it's for.  I don't push back though.

PR #1085, opened 2019-03-04
explorer: rework address page

buck54321:
Restyles address page in line with #451. There was no prototype for a general address page, but hints are taken from the prototype for the treasury fund page.

View selector is removed and charts are always displayed if JS is enabled. Implements new compact style chart controls. Tables updated with v2 style tweaks and made mobile friendly.

`dbtypes.AddressTx` receives two new fields, `ToStake` and `FromStake`, which indicate the fraction of input and output DCR respectively that is stake-related. These values are displayed as percentages on the address' page if they are non-zero.

Adds i/o counts and fiat conversion (if available) to overview data. A new SVG-based loading spinner is used for chart and list loading.

image


db/dcrpg/pgblockchain.go
```diff
@@ -1397,31 +1397,23 @@ func (pgb *ChainDB) DevBalance() (*dbtypes.AddressBalance, error) {
...
-	}
 
 	return balance, nil
 }
 
 // AddressSpentUnspent attempts to retrieve balance information for a specific
 // address from cache, and if cache is stale or missing data for the address, a
 // DB query is used. A successful DB query will freshen the cache.
-func (pgb *ChainDB) AddressSpentUnspent(address string) (int64, int64, int64, int64, error) {
+func (pgb *ChainDB) AddressSpentUnspent(address string) (*dbtypes.AddressBalance, error) {
```
buck54321: The return type is changed to `dbtypes.AddressBalance` here and `RetrieveAddressSpentUnspent` to prevent adding two more return fields. Seems to make the code cleaner anyways, but maybe should rename of the `*SpentUnspent` functions.

views/address.tmpl
```diff
@@ -15,287 +15,302 @@
...
-                          <option {{if eq $txType "merged_debit"}}selected{{end}} value="merged_debit">Merged Debits</option>
-                      </select>
-                    </div>
-                  </div>
+      </div>
+    </div>
+    {{if not .IsDummyAddress}}
+    <div class="position-relative" data-target="address.listbox">
+        <div class="row align-items-center">
+          <div class="mr-auto mb-0 h4 col-24 col-sm-6">Transactions</div>
```
buck54321: I guess the table rows aren't actually "Transactions", but "Transaction inputs and outputs, optionally merged into blocks", which doesn't quite ring. 

It's "HIstory" on master.

db/dbtypes/types.go
```diff
@@ -1030,35 +1043,50 @@ type AddressInfo struct {
...
+// stake-related.
+func (balance *AddressBalance) HasStakeInputs() bool {
+	return balance.ToStake > 0
 }
 
 // ReduceAddressHistory generates a template AddressInfo from a slice of
 // AddressRow. All fields except NumUnconfirmed and Transactions are set
 // completely. Transactions is partially set, with each transaction having only
 // the TxID and ReceivedTotal set. The rest of the data should be filled in by
 // other means, such as RPC calls or database queries.
```
chappjc: Let's supplement these docs with the new to/from stake (fraction) outputs.

db/dcrpg/pgblockchain.go
chappjc: Seems like we should remove one of AddressBalance and AddressSpentUnspent.

db/dcrpg/pgblockchain.go
```diff
@@ -1397,31 +1397,23 @@ func (pgb *ChainDB) DevBalance() (*dbtypes.AddressBalance, error) {
 // and if cache is stale or missing data for the address, a DB query is used. A
 // successful DB query will freshen the cache.
 func (pgb *ChainDB) AddressBalance(address string) (*dbtypes.AddressBalance, error) {
-	numSpent, numUnspent, amtSpent, amtUnspent, err :=
-		pgb.AddressSpentUnspent(address)
+	balance, err := pgb.AddressSpentUnspent(address)
```
chappjc: Seems like the whole function now reduces to `return pgb.AddressSpentUnspent(address)`

chappjc:
Looks and works great to me.  @gozart1 Did you wanna have a look at this one?

chappjc:
@gozart1  Bookmark this one if you wanna have a look.  Gonna merge since it's working well.

PR #1089, opened 2019-03-05
Do not allocate a big.Int for each row.

dajohi:
Both variables are set each iteration, so just reuse them.


dajohi:
Fixed.  only exawork can be allocated once.

PR #1093, opened 2019-03-05
dcrpg/internal: SelectTxsPerDay using blocks table

chappjc:
This changes `SelectTxsPerDay` to query from the blocks table instead of the transactions table. This is much faster.

Before:

> Execution Time: 2071.276 ms (cost=297206.48..297257.65)

After:

> Execution Time: 198.559 ms (cost=60490.22..60541.39)

The results are identical.

Also remove "= TRUE" from some queries since it's already boolean.

### DIFF
--- a/db/dcrpg/internal/blockstmts.go
+++ b/db/dcrpg/internal/blockstmts.go
@@ -174,6 +174,10 @@ const (
 		WHERE is_valid = FALSE
 		ORDER BY height DESC;`
 
+	SelectTxsPerDay = `SELECT date_trunc('day',time) AS date, sum(numtx)
+		FROM blocks
+		GROUP BY date ORDER BY date;`
+
 	// blocks table updates
 
 	UpdateLastBlockValid = `UPDATE blocks SET is_valid = $2 WHERE id = $1;`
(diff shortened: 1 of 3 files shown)


db/dcrpg/internal/txstmts.go
```diff
@@ -108,9 +108,6 @@ const (
 		ORDER BY is_mainchain DESC, is_valid DESC, block_time DESC
 		LIMIT 1;`
 
-	SelectTxsPerDay = `SELECT date_trunc('day',time) AS date, count(*) FROM transactions
-		GROUP BY date ORDER BY date;`
```
chappjc: This is the old query on `transactions` using `count(*)`.

db/dcrpg/internal/blockstmts.go
```diff
@@ -174,6 +174,10 @@ const (
 		WHERE is_valid = FALSE
 		ORDER BY height DESC;`
 
+	SelectTxsPerDay = `SELECT date_trunc('day',time) AS date, sum(numtx)
+		FROM blocks
+		GROUP BY date ORDER BY date;`
```
chappjc: This is the new query on `blocks`, using `sum(numtx)`.

dmigwi: This is far much better.

PR #1096, opened 2019-03-05
apicache: merge time restriction into doesNeedReheap

dajohi:
this saves an RLock/RUnlock every 250ms.


api/types/apicache.go
```diff
@@ -760,7 +759,8 @@ func (pq *BlockPriorityQueue) setAccessTime(t time.Time) {
 func (pq *BlockPriorityQueue) doesNeedReheap() bool {
 	pq.mtx.RLock()
 	defer pq.mtx.RUnlock()
-	return pq.needsReheap
+	return pq.needsReheap &&
+		time.Since(lastAccessTime) > 7*time.Second
```
chappjc: `time.Since(lastAccess)`?

dajohi: copied wrong window.  fixed.

PR #1097, opened 2019-03-05
dbtypes/dcrpg: add MergeRows

chappjc:
Add `dbtypes.MergeRows` to converts a slice of non-merged
(regular addresses table row data) into a slice of merged
address rows. This involves merging rows with the same
transaction hash into a single entry by combining the signed
values. The `IsFunding` field of a merged transaction indicates
if the net value is positive or not, although the `Value` field is
an absolute value (always positive). A map of transaction hash
to `*AddressRow` is also returned.  `MergedRows` will return a
non-nil error of a merged row is detected in the input since
only non-merged rows are expected.

Add a test for `MergeRows` in pgblockchain_test since it is best
tested with a live DB using the project fund address.

Also change `AddressTransactionsAllMerged` so it only gets
`valid_mainchain` rows.
Change `RetrieveAllMainchainAddressMergedTxns`, which
actually got all rows including invalidated and side chain rows,
so that it can be restricted to `valid_mainchain=true` rows.
Ensure that `retrieveAddressTxns` only requests valid mainchain results
when requesting merged results to match the non-merged request.

**NEXT:** Never fetch merged rows and instead use `dbtypes.MergedRows`.

### DIFF
--- a/db/dcrpg/pgblockchain_test.go
+++ b/db/dcrpg/pgblockchain_test.go
@@ -109,6 +109,46 @@ func TestChainDB_AddressTransactionsAll(t *testing.T) {
 	}
 }
 
+func TestMergeRows(t *testing.T) {
+	address := "Dcur2mcGjmENx4DhNqDctW5wJCVyT3Qeqkx"
+
+	rows, err := db.AddressTransactionsAll(address)
+	if err != nil {
+		t.Errorf("err should have been nil, was: %v", err)
+	}
+	if rows == nil {
+		t.Fatalf("should have rows, got none")
+	}
+
+	tStart := time.Now()
+	mergedRows, mrMap, err := dbtypes.MergeRows(rows)
+	if err != nil {
+		t.Fatalf("MergeRows failed: %v", err)
+	}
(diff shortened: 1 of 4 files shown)


db/dcrpg/pgblockchain_test.go
```diff
@@ -109,6 +109,46 @@ func TestChainDB_AddressTransactionsAll(t *testing.T) {
...
+
+	mergedRows0, err := db.AddressTransactionsAllMerged(address)
+	if err != nil {
+		t.Errorf("err should have been nil, was: %v", err)
+	}
+	if mergedRows0 == nil {
+		t.Fatalf("should have rows, got none")
+	}
+
+	if len(mergedRows) != len(mergedRows0) {
```
chappjc: Will expand this test to validate the fields

chappjc: This test is now completed with full validation of the merged rows.

chappjc:
This is required for changes to prevent excessive memory use by the address cache.  Merging now that the test function is completed.

PR #1101, opened 2019-03-07
gov/agendas: Remove hard-coded votes version

dmigwi:
This PR replaced the hard-coded initial votes version with data obtained from `chaincfg.Params.Deployments` map.

### DIFF
--- a/gov/agendas/deployments.go
+++ b/gov/agendas/deployments.go
@@ -107,7 +107,7 @@ func agendasForVoteVersion(ver uint32, client DeploymentSource) ([]AgendaTagged,
 		return nil, err
 	}
 
-	// Set the agendas slice size
+	// Set the agendas slice capacity.
 	agendas := make([]AgendaTagged, 0, len(voteInfo.Agendas))
 	for i := range voteInfo.Agendas {
 		v := &voteInfo.Agendas[i]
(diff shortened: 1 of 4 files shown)


gov/agendas/deployments.go
```diff
@@ -122,42 +124,37 @@ func agendasForVoteVersion(ver uint32, client DeploymentSource) (agendas []Agend
...
-
-	return true
+	return agendas, nil
 }
 
-// updatedb used when needed to keep the saved db up-to-date.
-func (db *AgendaDB) updatedb(voteVersion uint32, client DeploymentSource) int {
+// updatedb checks if vote versions available in chaincfg.ConsensusDeployment
+// are already updated in the agendas db, if not yet their data is updated.
+// dcrjson.GetVoteInfoResult and chaincfg.ConsensusDeployment hold almost similar
```
dmigwi: `chaincfg.Params.Deployments` was a great find but It is missing some important data like votes `status` that is easily returned in `dcrjson.GetVoteInfoResult` by `GetVotesVersion` rpc endpoint.

main.go
```diff
@@ -512,7 +512,10 @@ func _main(ctx context.Context) error {
 	var _ agendas.DeploymentSource = dcrdClient
```
chappjc: Please remove this line while here.  Passing dcrdClient to CheckAgendasUpdates accomplishes the same complile time check.

PR #1102, opened 2019-03-07
db/internal: use bytes.Buffer

dajohi:
```
benchmark                         old ns/op     new ns/op     delta
BenchmarkMakeARRAYOfBIGINTs-2     26671         4559          -82.91%

benchmark                         old allocs     new allocs     delta
BenchmarkMakeARRAYOfBIGINTs-2     300            9              -97.00%

benchmark                         old bytes     new bytes     delta
BenchmarkMakeARRAYOfBIGINTs-2     22770         2040          -91.04%
```

### DIFF
--- a/db/dcrpg/internal/common.go
+++ b/db/dcrpg/internal/common.go
@@ -1,6 +1,9 @@
 package internal
 
-import "fmt"
+import (
+	"bytes"
+	"strconv"
+)
 
 const (
 	IndexExists = `SELECT 1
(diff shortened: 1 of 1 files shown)


chappjc:
Nice.  Just a couple comments.  strings.Builder would be even better about allocs.  Also, I'm not certain we're using all these functions.

chappjc:
Wanna check in those benchmarks?  Would be helpful.

Issue #1103, opened 2019-03-07
data race


dajohi:
commit fcd8c3488fd393703274587dd2ca1cf8582dd599 
```
==================                                                                                                   
WARNING: DATA RACE                                                                                                  
Write at 0x00c0239d6078 by goroutine 100:                                                                            
  github.com/decred/dcrdata/v4/explorer.(*explorerUI).BeginSyncStatusUpdates.func2.1()                              
      /home/dcrdata/git/dcrdata/explorer/syncstatus.go:92 +0xca                                                      
  github.com/decred/dcrdata/v4/explorer.(*explorerUI).BeginSyncStatusUpdates.func2()                                
      /home/dcrdata/git/dcrdata/explorer/syncstatus.go:103 +0x926                                                    
                                                                                                                    
Previous read at 0x00c0239d6078 by goroutine 58:                                                                     
  github.com/decred/dcrdata/v4/explorer.(*explorerUI).BeginSyncStatusUpdates.func1()                                
      /home/dcrdata/git/dcrdata/explorer/syncstatus.go:74 +0x171                                                     
                                                                                                                    
Goroutine 100 (running) created at:                                                                                  
  github.com/decred/dcrdata/v4/explorer.(*explorerUI).BeginSyncStatusUpdates()                                      
      /home/dcrdata/git/dcrdata/explorer/syncstatus.go:86 +0xea                                                      
  main._main()                                    
      /home/dcrdata/git/dcrdata/main.go:595 +0x2797                                                                  
  github.com/decred/dcrdata/v4/db/dcrpg.NewChainDBWithCancel()                                                      
      /home/dcrdata/git/dcrdata/db/dcrpg/pgblockchain.go:446 +0x18e                                                 
  main._main()                                                                                                      
      /home/dcrdata/git/dcrdata/main.go:187 +0xabc7                                                                 
  fmt.Sscanf()                                                                                                      
      /home/dcrdata/go/src/fmt/scan.go:114 +0xd0                                                                     
  github.com/btcsuite/goleveldb/leveldb/storage.fsParseName()                                                       
      /home/dcrdata/go/pkg/mod/github.com/btcsuite/goleveldb@v1.0.0/leveldb/storage/file_storage.go:643 +0x144       
  github.com/btcsuite/goleveldb/leveldb/storage.(*fileStorage).List()
[...]

PR #1106, opened 2019-03-08
explorer/syncstatus.go: Fix barload data race

dmigwi:
Fixes: #1103

Adds a `stopTimer` channel shared between the goroutines launched in `BeginSyncStatusUpdates(barLoad chan *dbtypes.ProgressBarLoad) {`

### DIFF
--- a/explorer/syncstatus.go
+++ b/explorer/syncstatus.go
@@ -67,13 +67,15 @@ func (exp *explorerUI) BeginSyncStatusUpdates(barLoad chan *dbtypes.ProgressBarL
 
 	// stopTimer allows safe exit of the goroutine that triggers periodic
 	// websocket progress update.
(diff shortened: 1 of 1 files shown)


explorer/syncstatus.go
```diff
@@ -65,19 +65,28 @@ func (exp *explorerUI) BeginSyncStatusUpdates(barLoad chan *dbtypes.ProgressBarL
 		return
 	}
 
+	// stopTimer allows safe exit of the goroutine that triggers periodic
+	// websocket progress update.
+	stopTimer := make(chan bool)
```
chappjc: Works for me, but generally want to use `chan struct{}` if the value is unused since it is optimized to a zero byte payload in the channel send.

PR #1107, opened 2019-03-08
api: sanity check tx input

dajohi:
Verify input is a valid hash before sending further up the code
chain.


dajohi:
```
2019-03-08 01:50:50.802 [ERR] PSQL: GetRawTransactionVerbose failed for: TsThUjrXRzaDZqgk7D7rTVf19uL2Zb4BBR5  
```

chappjc:
Also a great change because we shouldn't be logging an error because the user input was incorrect.

PR #1113, opened 2019-03-09
api: sanity check addresses (round 1)

dajohi:


middleware/apimiddleware.go
```diff
@@ -229,13 +231,22 @@ func GetBlockHashCtx(r *http.Request) (string, error) {
...
-func GetAddressCtx(r *http.Request) string {
-	address, ok := r.Context().Value(CtxAddress).(string)
+func GetAddressCtx(r *http.Request, activeNetParams *chaincfg.Params) (string, error) {
+	addressStr, ok := r.Context().Value(CtxAddress).(string)
 	if !ok {
 		apiLog.Trace("address not set")
-		return ""
+		return "", fmt.Errorf("address not set")
+	}
+	address, err := dcrutil.DecodeAddress(addressStr)
```
chappjc: Seems OK, but as you see in `getAddressInfo` it is run again.

dajohi: yes, future diff will be returning a diff type, if we go that way.

middleware/apimiddleware.go
```diff
@@ -229,13 +231,22 @@ func GetBlockHashCtx(r *http.Request) (string, error) {
...
+	if err != nil {
+		return "", fmt.Errorf("invalid address '%v': %v",
+			addressStr, err)
+	}
+	if !address.IsForNet(activeNetParams) {
+		return "", fmt.Errorf("%v is invalid for this network",
+			address)
 	}
-	return address
+	return address.String(), nil
```
chappjc: might as well return `addressStr` since we have it.

dajohi: addressStr it is.

PR #1116, opened 2019-03-09
explorer: create stub when address validates as other network

buck54321:
Avoid a panic at https://github.com/decred/dcrdata/blob/e402297f5a80421b4b1308525cf2ef2faf03ce0d/explorer/explorerroutes.go#L1554


buck54321:
Looks like 3.2-stable is good as is. https://github.com/decred/dcrdata/blob/12a221ed7f3a01f1a38749ac5698e267a59e8007/explorer/explorerroutes.go#L1393

PR #1120, opened 2019-03-09
explorer: tweak navbar and footer layouts

buck54321:
Uses a fixed height for the navbar and footer, and sets the html element top and bottom padding to match, which makes the body element exactly the visible portion between the navbar and footer. Selectively hides footer elements for mobile-friendliness.

Search bar given a button and a maximum length that is enough for any block hash. Instead of the blue outline on focus, uses a light border instead.

image


chappjc:
Note sure I like removing the version.  The whole version string isn't needed, but the 4.0.0 part should be somewhere I think.

buck54321:
It's only hidden on the mobile view. I think I can squeeze it in though.

chappjc:
Oh, that's cool then.  If it's on the desktop site then that's good enough.

PR #1125, opened 2019-03-12
gov: implement live vote tracking

buck54321:
Adds `VoteTracker`, which refreshes with new blocks to update a cache of voting-related data. VoteTracker produces a `VoteSummary`, which has fields customized for use in templates. There might be some space to integrate `VoteTracker` with `AgendaDB`.

Front end: The information is incorporated at the top of the /agendas page here, with intention to use the same data in some way on the homepage and more. To display a graphic without using the `charts.js` package (which is huge), a new `VoteMeter` class displays a simple graphic with a central value displayed using the Canvas API. It'd be good to drop `charts.js`, which is dcrdata's largest vendor package and we barely use it. `VoteMeter` is about 6 kB unminified, compared to 160 kB for a minified `charts.js`.

image

### DIFF
--- a/gov/agendas/tracker.go
+++ b/gov/agendas/tracker.go
@@ -459,7 +459,7 @@ func spoof(summary *VoteSummary) {
 	agenda := &summary.Agendas[0]
 	summary.VotingTriggered = false
 	agenda.VotingTriggered = false
-	agenda.IsVoting = false
+	agenda.IsVoting = true
 	agenda.IsDefined = false
 
 	agenda.Aye = agenda.Quorum * 9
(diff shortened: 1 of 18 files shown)


chappjc:
That screen shot looks great.  Regarding chartsjs, I've been thinking we should remove the doughnutGraph on the ticket pool age, and all the solo/pool/other related stuff.  Identifying tickets as solo or pool is not possible, and it's misleading at best to show it.

buck54321:
Changes to the meters.

image

buck54321:
image

gov/agendas/tracker.go
```diff
@@ -0,0 +1,489 @@
...
+			PassThreshold: tracker.passThreshold,
+			// LockCount:     tracker.lockCount,
+		}
+		status := agenda.Status
+		agendaSummary.IsLocked = status == statusLocked
+		agendaSummary.IsFailed = status == statusFailed
+		agendaSummary.IsActive = status == statusActive
+		agendaSummary.IsVoting = status == statusStarted
+		agendaSummary.IsDefined = status == statusDefined
+		for idx := range agenda.Choices {
```
chappjc: This is actually fine since `idx` here is shadowing `idx` of the outer loop, but it's slightly alarming to see.  I noticed you used `idy` in other such cases, so I figured that might have been your intention.

PR #1126, opened 2019-03-12
charts cache: Incremental charts data update (on /charts page)

### LINKED ISSUE #1034: Need incremental handling of charts data
`pgblockchain.GetPgChartsData` is used to pull all charts data from the database every time a block is received, which takes a long time. Most if not all of those datasets should be maintained incrementally, probably with their own `BlockDataSaver` as well as some reorg handling. 

Optionally, rather than use `dbtypes.ChartsData` for everything, we may want to create custom types for each dataset and have the `BlockDataSaver` create JSON-formatted byte arrays for faster retrieval. 

`GetPgChartsData` can still be called once on startup to initially populate the datasets.

dmigwi:
fixes #1034 

- It reduces the average charts data load and update time from 20 secs to 3 sec.
- A dump file(*.gob) is created in the dcrdata home directory once the system is shutdown. On restarting the system the cache data in the dump file is scanned and set as the current cache entries.
- The dump file is deleted once its reading is complete.


dmigwi:
**Update:**

This PR loads test Pg and Sqlite test data into respective dbs for testing.  The tests dbs are later wiped out after the tests are done.

- To run all the newly added charts test, execute:
 ```Bash
     bash run_tests_with_chartests.sh
```
- The charts tests confirm that the incremental charts change added results to the same data just like when  fresh db queries would have been run for all charts.

- Instructions on how to modify the dbs test data can be found [here](https://github.com/dcrlabs/bug-free-happiness/pull/2/files)

**Comment by chappjc** [comment]
Thanks, Migwi.  Would you please rebase and resolve?  OK to squash "fix" commits in the process.

File: db/dcrsqlite/apisource.go [hunk 1]
```diff
@@ -992,25 +992,40 @@ func (db *WiredDB) GetPoolValAndSizeRange(idx0, idx1 int) ([]float64, []float64)
...
 
-// GetSqliteChartsData fetches the charts data from the sqlite db.
-func (db *WiredDB) GetSqliteChartsData() (map[string]*dbtypes.ChartsData, error) {
-	poolData, err := db.RetrieveAllPoolValAndSize()
-	if err != nil {
-		return nil, err
+// SqliteChartsData fetches the charts data from the sqlite db.
+func (db *WiredDB) SqliteChartsData(data []*dbtypes.ChartsData) (err error) {
+	if len(data) != dbtypes.SqliteChartsCount {
+		data = make([]*dbtypes.ChartsData, dbtypes.SqliteChartsCount)
```
chappjc: How does the caller get this?

db/dcrsqlite/apisource.go
```diff
@@ -992,25 +992,40 @@ func (db *WiredDB) GetPoolValAndSizeRange(idx0, idx1 int) ([]float64, []float64)
...
 	}
 
-	var data = map[string]*dbtypes.ChartsData{
-		"ticket-pool-size":  {Time: poolData.Time, SizeF: poolData.SizeF},
-		"ticket-pool-value": {Time: poolData.Time, ValueF: poolData.ValueF},
-		"fee-per-block":     feeData,
+	poolData := data[dbtypes.TicketPoolSize.SqlitePos()]
+	poolValue := data[dbtypes.TicketPoolValue.SqlitePos()]
+
+	var wCopy *dbtypes.ChartsData
```
chappjc: What is the reason for a working copy?

db/dcrsqlite/apisource.go
```diff
@@ -992,25 +992,40 @@ func (db *WiredDB) GetPoolValAndSizeRange(idx0, idx1 int) ([]float64, []float64)
...
+	//  Append the all data to the working Copy.
+	if poolData != nil && poolValue != nil {
+		wCopy = new(dbtypes.ChartsData)
+		*wCopy = *poolData
+		wCopy.ValueF = poolValue.ValueF
```
chappjc: None of the underlying slices' data buffers are copies.  They still share the same memory.

db/dcrsqlite/sqlite.go
```diff
@@ -982,27 +984,34 @@ func (db *DB) RetrievePoolValAndSizeRange(ind0, ind1 int64) ([]float64, []float6
 
 // RetrieveAllPoolValAndSize returns all the pool values and sizes stored since
 // the first value was recorded up current height.
-func (db *DB) RetrieveAllPoolValAndSize() (*dbtypes.ChartsData, error) {
-	chartsData := new(dbtypes.ChartsData)
+func (db *DB) RetrieveAllPoolValAndSize(chartsData *dbtypes.ChartsData) (*dbtypes.ChartsData, error) {
```
chappjc: This is confusing since this chartsData is a merger of two different charts data sets.  The caller has to know this and prepare the inputs, and tease apart the output.  I think this would be better taking the three slices as inputs, and returning the modified (or new) slices as outputs.

explorer/explorer.go
```diff
@@ -413,17 +362,35 @@ func New(cfg *ExplorerConfig) *explorerUI {
...
+	t := time.Now()
 	if !exp.liteMode {
-		exp.prePopulateChartsData()
+		var defHeight int64 = 1
+
+		if err := ReadCacheFile(cacheDumpPath, defHeight); err != nil {
+			log.Errorf("Cache dump data loading failed: %v", err)
+
+			// If the cache data loading fails, the db querying will be
+			// triggered in the next explorer.Store call. Since this method is
```
chappjc: Needs to be triggered now.

explorer/explorer.go
```diff
@@ -477,33 +444,69 @@ func (exp *explorerUI) prePopulateChartsData() {
...
-	if err != nil {
+
+	sqliteData := make([]*dbtypes.ChartsData, sqliteChartsCount)
+
+	// Sqlite charts count is defined by thus data is stored between index
+	// (dbtypes.PgChartsCount) to the end of the cache.
+	// Make the data copy if the charts count matches the chartsCount value.
+	if len(chartsData) == chartsCount {
+		copy(sqliteData, chartsData[pgChartsCount:])
+	}
```
chappjc: else?

explorer/explorer.go
```diff
@@ -477,33 +444,69 @@ func (exp *explorerUI) prePopulateChartsData() {
...
+	sqliteChartsCount := dbtypes.SqliteChartsCount
+	chartsCount := pgChartsCount + sqliteChartsCount
+	pgData := make([]*dbtypes.ChartsData, pgChartsCount)
+
+	// Since Pg charts count defined by dbtypes.PgChartsCount, their data is
+	// stored from index 0-(dbtypes.PgChartsCount -1) in the charts cache.
+	// Make the data copy if the charts count matches the chartsCount value.
+	if len(chartsData) == chartsCount {
+		copy(pgData, chartsData[:pgChartsCount])
 	}
```
chappjc: else?

main.go
```diff
@@ -1168,16 +1178,24 @@ func _main(ctx context.Context) error {
...
+	// This dumps the cache charts data into a file for future use on system
+	// exit.
+	defer func() {
+		er := explorer.WriteCacheFile(dumpPath)
+		if er != nil {
+			log.Errorf("WriteCacheFile failed: %v", er)
+		} else {
+			log.Debug("Dumping the charts cache data was successful")
+		}
+	}()
```
chappjc: Why here instead of up after explore.PrepareCharts?

db/dcrpg/pgblockchain.go
```diff
@@ -2572,96 +2572,117 @@ func (pgb *ChainDB) chainWork(chainData [2]*dbtypes.ChartsData) ([2]*dbtypes.Cha
 }
 
 // PgChartsData retrieves the data for the different types of charts.
-func (pgb *ChainDB) PgChartsData(oldData []*dbtypes.ChartsData) error {
-	txRate, err := pgb.txPerDay(oldData[dbtypes.TxPerDay.Pos()])
+func (pgb *ChainDB) PgChartsData(data []*dbtypes.ChartsData) error {
+	// CAUTION: if dbtypes.PgChartsCount is out-of-date, it will always trigger
+	// a fresh charts data update instead of an incremental charts data update.
+	// Other unexpected bugs may arise.
```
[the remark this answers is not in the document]
dmigwi: Would returning an error message be better?

db/dcrsqlite/apisource.go
dmigwi: I found it to affect the underlying array thus the working copy was the solution.

db/dcrpg/pgblockchain.go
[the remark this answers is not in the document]
dmigwi: An error message feels like an overkill to a situation that could happen often during the normal functionality of the code. How about?
```Go
        if len(data) != dbtypes.PgChartsCount {
		data = make([]*dbtypes.ChartsData, dbtypes.PgChartsCount)
                log.Debug("Found some charts data missing thus initiated a fresh data query")
	}
```

db/dbtypes/types.go
```diff
@@ -498,6 +498,103 @@ func ChoiceIndexFromStr(choice string) (VoteChoice, error) {
...
+const (
+	// SqliteChartsCount defines the count of charts that use the sqlite db as
+	// data source and their data is held in cacheChartsData cache. When new
+	// sqlite charts are added, this value should be updated to avoid errors.
+	SqliteChartsCount = 3
+
+	// PgChartsCount defines the count of charts that use the auxiliary db as
+	// data source and their data is held in cacheChartsData cache. When new
+	// pg charts are added, this value should be updated to avoid errors.
+	PgChartsCount = 13
```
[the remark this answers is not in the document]
dmigwi: Even with an array we might still need to have a such a value defined some where so as to keep its length and capacity as optimal as possible.

dmigwi:
With this PR in, debugging inaccurate charts display values would require one to delete `*.gob` file in the `datadir` otherwise restarting the system will not fix the issue.

db/dbtypes/types.go
```diff
@@ -498,6 +498,103 @@ func ChoiceIndexFromStr(choice string) (VoteChoice, error) {
...
+	POWDifficulty:   "pow-difficulty",
+	TicketByWindows: "ticket-by-outputs-windows",
+	TicketPoolSize:  "ticket-pool-size",
+	TicketPoolValue: "ticket-pool-value",
+	TicketPrice:     "ticket-price",
+	TicketsByBlocks: "ticket-by-outputs-blocks",
+	TicketSpendT:    "ticket-spend-type",
+	TxPerBlock:      "tx-per-block",
+	TxPerDay:        "tx-per-day",
+}
```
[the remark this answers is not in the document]
dmigwi: More charts will added after this PR but you raise a valid point.

explorer/explorer.go
```diff
@@ -616,14 +621,16 @@ func (exp *explorerUI) Store(blockData *blockdata.BlockData, msgBlock *wire.MsgB
...
 
-	// Do not run updates if blockchain sync is running.
-	if !exp.AreDBsSyncing() {
+	// Do not run updates if blockchain sync is running or lite mode is on.
+	if !exp.AreDBsSyncing() && !exp.liteMode {
 		// Update the charts data after every five blocks or if no charts data
 		// exists yet.
+
 		if newBlockData.Height%5 == 0 || cacheChartsData.Height() == -1 {
```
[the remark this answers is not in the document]
dmigwi: Changes made after every block are too small to be visible on the charts since most of them hold data running into thousands of entries.
With this PR it takes 1-3 sec to fetch the incremental updates. Imagine a case scenario where you spend 2 sec on average per block to fetch data that a user might not easily notice. An interval of 5 blocks ensures that the other 8 sec are spent doing other things instead of querying the charts data.

db/dbtypes/types.go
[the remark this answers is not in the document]
dmigwi: >I'd much rather have `ChartType` be `string` itself, and the data cache be a `map[ChartType]*ChartsData` similar to the way it was implemented before.

Using a `string` to reference charts IMO introduces a possible bug in case one references a non-existent chart while assigning or fetching data.

explorer/explorer.go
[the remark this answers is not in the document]
buck54321: I'm okay with 5 block increments. Charts can be zoomed though, so there aren't always going to be thousands of data points displayed.

db/dbtypes/types.go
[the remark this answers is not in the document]
chappjc: I agree with @buck54321 about the map.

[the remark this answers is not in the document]
dmigwi: Okay. I will revert back to the map implementation.

db/dbtypes/types.go
```diff
@@ -498,6 +498,37 @@ func ChoiceIndexFromStr(choice string) (VoteChoice, error) {
...
+)
+
+var (
+	// SqliteCharts helps select all the charts that use the SQlite db as a data
+	// source from the charts cache data.
+	SqliteCharts = []string{FeePerBlock, TicketPoolSize, TicketPoolValue}
+
+	// PgCharts helps select all the charts that use the auxiliary db as a data
+	// source from the charts cache data.
+	PgCharts = []string{AvgBlockSize, BlockChainSize, ChainWork, CoinSupply,
```
chappjc: Might as well have these be arrays instead of slices. `[...]string{`

explorer/explorer.go
```diff
@@ -479,33 +454,57 @@ func (exp *explorerUI) prePopulateChartsData() {
...
+
+		// Pick all the sqliteCharts
+		for _, chartName := range dbtypes.SqliteCharts {
+			dataset := chartsData[chartName]
+			sqliteData[chartName] = dataset
+		}
 	}
+
+	var err error
+	pgData, err = exp.explorerSource.PgChartsData(pgData)
```
chappjc: The `pgData` as a return value is not needs.  The changes to the input map will be visible here in the caller.  This works because `PgChartsData` is not in charge of making the map, just updating it.

explorer/explorer.go
```diff
@@ -479,33 +454,57 @@ func (exp *explorerUI) prePopulateChartsData() {
...
+
+	chartsData := cacheChartsData.get()
+	pgData := make(map[string]*dbtypes.ChartsData)
+	sqliteData := make(map[string]*dbtypes.ChartsData)
+
+	if chartsData != nil {
+		// Pick all the PgCharts
+		for _, chartName := range dbtypes.PgCharts {
+			dataset := chartsData[chartName]
+			pgData[chartName] = dataset
```
chappjc: `pgData[chartName] = chartsData[chartName]`

explorer/explorer.go
```diff
@@ -479,33 +454,57 @@ func (exp *explorerUI) prePopulateChartsData() {
...
+		}
+
+		// Pick all the sqliteCharts
+		for _, chartName := range dbtypes.SqliteCharts {
+			dataset := chartsData[chartName]
+			sqliteData[chartName] = dataset
```
chappjc: `sqliteData[chartName] = chartsData[chartName]`

explorer/explorer.go
```diff
@@ -479,33 +454,57 @@ func (exp *explorerUI) prePopulateChartsData() {
...
+		}
+
 		log.Errorf("Invalid PG data found: %v", err)
 		return
 	}
 
 	log.Debugf("Retrieving charts data from base DB.")
-	sqliteData, err := exp.blockData.GetSqliteChartsData()
+
+	sqliteData, err = exp.blockData.SqliteChartsData(sqliteData)
```
chappjc: Same deal here with the modification of the map

explorer/explorer.go
```diff
@@ -415,24 +364,50 @@ func New(cfg *ExplorerConfig) *explorerUI {
...
+		var defHeight int64 = 1
+
+		if err := ReadCacheFile(cacheDumpPath, defHeight); err != nil {
+			log.Debugf("Cache dump data loading failed: %v", err)
+
+			// If the cache data loading fails, the db querying will be triggered.
+			exp.prePopulateChartsData()
+		}
+
+		log.Debugf("completed the initial charts cache scanning in %v", time.Since(t))
```
chappjc: `"Completed ..."`

explorer/chartscache.go
```diff
@@ -124,3 +126,278 @@ func ReadCacheFile(filePath string, height int64) error {
...
+	data, ok = cacheData[dbtypes.BlockChainSize]
+	if ok {
+		var index = len(data.Time) - 1
+		for ; index > 0; index-- {
+			if data.Time[index].UNIX() < timeAfterAncestor {
+				break
+			}
+		}
+		cacheData[dbtypes.BlockChainSize].Time = cacheData[dbtypes.BlockChainSize].Time[:index]
+		cacheData[dbtypes.BlockChainSize].ChainSize = cacheData[dbtypes.BlockChainSize].ChainSize[:index]
```
chappjc: Given the indexing and number of edge case, the above loop and slice resize should be testable.  Having it be reusable would also benefit a number of the other data sets.

explorer/chartscache.go
```diff
@@ -124,3 +126,278 @@ func ReadCacheFile(filePath string, height int64) error {
...
+		cacheData[dbtypes.CoinSupply].Time = cacheData[dbtypes.CoinSupply].Time[:index]
+		cacheData[dbtypes.CoinSupply].ValueF = cacheData[dbtypes.CoinSupply].ValueF[:index]
+	}
+
+	// "duration-btw-blocks" chart data.
+	data, ok = cacheData[dbtypes.DurationBTW]
+	if ok {
+		var index = len(data.Height) - 1
+		for ; index > 0; index-- {
+			if data.Height[index] < heightAfterAncestor {
```
chappjc: Based on these conditions, the loop will break when `data.Height[index]` is equal to ancestor height.  I would think it would be easier to reason about the code if it just used `<=` with the appropriate height condition.  When the loop breaks, the index of the ancestor is used as the end slice index (`Height[:index]`), so it indeed is removing the ancestor.  That is OK, although theoretically unnecessary if it is the _common_ ancestor, but it does contradict the wording "drops all the chart data entries since the common ancestor" in the function docs.

explorer/chartscache.go
```diff
@@ -0,0 +1,403 @@
...
+}
+
+// ChartTypeData is a thread-safe way to access chart data of the given type.
+func ChartTypeData(chartType string) (data *dbtypes.ChartsData, ok bool) {
+	cacheChartsData.RLock()
+	defer cacheChartsData.RUnlock()
+
+	// Data updates replace the entire map rather than modifying the data to
+	// which the pointers refer, so the pointer can safely be returned here.
+	data, ok = cacheChartsData.Data[chartType]
```
chappjc: Omitting the bool is better in this case IMO.  The zero value is a `nil` `*dbtypes.ChartsData`.

db/dcrsqlite/blocksretrieval_test.go
```diff
@@ -31,7 +33,7 @@ func TestEmptyDBRetrieveBlockFeeInfo(t *testing.T) {
 		t.Fatalf("Failed to obtain test DB: %v", err)
 	}
 
-	result, err := db.RetrieveBlockFeeInfo()
+	result, err := db.RetrieveBlockFeeInfo(nil)
```
chappjc: 2 inputs needed

explorer/chartscache.go
```diff
@@ -124,3 +126,278 @@ func ReadCacheFile(filePath string, height int64) error {
...
+	for {
+		select {
+		case data, ok := <-m.cacheChan:
+			if !ok {
+				return
+			}
+
+			// Drop all entries since since the ancestor block after reorganization
+			// of all PG and Sqlite DB is complete.
+			if err := m.explorer.DropToAncestor(uint64(data.NewChainHeight)); err != nil {
```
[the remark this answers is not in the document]
dmigwi: Yes. And this is invoked after the other chain reorganizations are completed.

explorer/chartscache.go
```diff
@@ -124,3 +126,278 @@ func ReadCacheFile(filePath string, height int64) error {
...
+
+		case <-m.ctx.Done():
+			return
+		}
+	}
+}
+
+// DropToAncestor drops all the chart data entries since the common ancestor
+// block. If the ancestor block is in the last entry added, then the last chart
+// data entry is the only one that is dropped.
```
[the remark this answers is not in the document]
dmigwi: Yes. This happens for the charts whose data is grouped in days or via TP window. I will change the language used there.

explorer/chartscache.go
```diff
@@ -0,0 +1,403 @@
...
+}
+
+// cacheChartsData holds the prepopulated data that is used to draw the charts.
+var cacheChartsData chartDataCache
+
+// get provides a thread-safe way to access the all cache charts data.
+func (c *chartDataCache) get() map[string]*dbtypes.ChartsData {
+	c.RLock()
+	defer c.RUnlock()
+	return c.Data
```
chappjc: The map may still be accessed concurrently.  e.g. `prePopulateChartsData` reads from the map while `DropToAncestor` reads and writes.  `get()` just controls access to the `Data` field of `chartDataCache`, not the map itself.

explorer/chartscache.go
```diff
@@ -124,3 +126,278 @@ func ReadCacheFile(filePath string, height int64) error {
...
+	if ok {
+		var index = len(data.Time) - 1
+		for ; index > 0; index-- {
+			if data.Time[index].UNIX() < timeAfterAncestor {
+				break
+			}
+		}
+		cacheData[dbtypes.AvgBlockSize].Time = cacheData[dbtypes.AvgBlockSize].Time[:index]
+		cacheData[dbtypes.AvgBlockSize].Size = cacheData[dbtypes.AvgBlockSize].Size[:index]
+	}
```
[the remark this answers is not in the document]
dmigwi: That was just a precautionary measure since the previous data fetch could have failed for any chart.

explorer/chartscache.go
dmigwi: > above loop and slice resize

Since `cacheData` is now a map are you referring to the individual chart data sets which are the only slices here?

explorer/chartscache.go
```diff
@@ -0,0 +1,403 @@
...
+			if data.Time[index].UNIX() < timeAfterAncestor {
+				break
+			}
+		}
+		cacheData[dbtypes.TicketPoolValue].Time = cacheData[dbtypes.TicketPoolValue].Time[:index]
+		cacheData[dbtypes.TicketPoolValue].ValueF = cacheData[dbtypes.TicketPoolValue].ValueF[:index]
+	}
+
+	// Update the modified cache data.
+	cacheChartsData.Update(exp.Height(), cacheData)
```
chappjc: All the above modifications to the charts data already affect the stored `Data` map. Note that in this function above the map itself is not modified at all, just the fields of the stored structs, so it's not clear why reassigning `cacheData` to the `Data` field is done.  But even if the `cacheData` map were modified (insert/delete), those changes would still be visible in the already-stored map in `Data` because an initialized `map` is a reference type in golang.  In fact, internally, a map is a pointer.

explorer/chartscache.go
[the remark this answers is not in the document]
chappjc: But [common ancestor height + 1] != data.NewChainHeight

explorer/chartscache.go
chappjc: More abstract than charts data an the map.
Two things: (1) computing `index` given a slice of times and another reference time, and (2) reslicing an arbitrary slice with the index check the slice length to avoid panic and then reslice.

explorer/chartscache_test.go
```diff
@@ -118,3 +120,103 @@ func TestChartsCache(t *testing.T) {
...
+			index:          2,
+		},
+		{
+			HeightsArr:     []uint64{279156, 279157, 279158, 279159, 279160, 279161},
+			AncestorHeight: 279157,
+			index:          1,
+		},
+		{
+			HeightsArr:     []uint64{279156, 279157, 279158, 279159, 279160, 279161},
+			AncestorHeight: 279161,
```
chappjc: Thanks for adding these!  Please also add a test for when `AncestorHeight` is greater than all of the elements in `HeightsArr`.

explorer/chartscache_test.go
```diff
@@ -118,3 +120,103 @@ func TestChartsCache(t *testing.T) {
...
+		},
+		{
+			HeightsArr:     []uint64{279156, 279157, 279158, 279159, 279160, 279161},
+			AncestorHeight: 279158,
+			index:          2,
+		},
+		{
+			HeightsArr:     []uint64{279156, 279157, 279158, 279159, 279160, 279161},
+			AncestorHeight: 279157,
+			index:          1,
```
chappjc: As I mentioned previously, it's not a big deal to remove the data for the common ancestor, but it's not necessary.

dmigwi: Actually let me change something that I have seen because of this.

dmigwi: Since `a[:len(a)]` returns the whole array it makes sense to return the whole array length if the ancestor data point is the last in the dataset or is greater than the last dataset entry.

dmigwi:
https://github.com/decred/dcrdata/pull/1219

chappjc:
I didn't test with a simnet reorg, so we'll have to follow up on this.

PR #1127, opened 2019-03-12
explorer: only signal mempool update once

chappjc:
The `mempool` package was designed to trigger the mempool update events,
but the `sigMempoolUpdate` sends in `StoreMPData` for both `explorerUI` and
`PubSubHub` were not removed. This eliminates these duplicate sends.


chappjc:
Tested working with the psclient democlient and the web browser console.

Issue #1128, opened 2019-03-12
illegible text in dark mode


chappjc:
![image](https://user-images.githubusercontent.com/9373513/54237672-7fde0600-44e4-11e9-9267-99352ce8b26e.png)

chappjc:
Turns out that this was Bootstrap being upgraded too liberally from 4.1.3 to 4.3.x.  With the incompatible Bootstrap 4.3.1, there is a `color` property set for the `.table` class, which is taking priority over the `body.darkBG` style:

![image](https://user-images.githubusercontent.com/9373513/54255751-48dd1400-4527-11e9-91fc-0d9e0ec6ce73.png)

chappjc:
We should check in a working `package-lock.json`.

chappjc:
Resolved by https://github.com/decred/dcrdata/pull/1129 and https://github.com/decred/dcrdata/pull/1121

PR #1129, opened 2019-03-13
tighten npm dep version specs

chappjc:
Resolving https://github.com/decred/dcrdata/issues/1128.

This unbreaks formatting issues with Bootstrap 4.3 and possibly 4.2.
Run npm install after applying these changes.

The bug in issue #1128 is manifested when bootstrap is allowed to upgrade too liberally:

```sh
$ npm list |grep bootstrap     
├── bootstrap@4.3.1 invalid                        
```

With this change:

```sh
$ npm list |grep bootstrap     
├── bootstrap@4.1.3
```

Also pins down qrcode more tightly.

TODO: We should consider checking in a package-lock.json with our stamp of approval.  Otherwise we aren't sufficiently controlling what goes into the asset bundle.

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,9 +59,9 @@
     "webpack-merge": "^4.1.4"
   },
   "dependencies": {
-    "bootstrap": "^4.1.3",
+    "bootstrap": "~4.1.3",
     "dompurify": "^1.0.8",
     "lodash-es": "^4.17.11",
-    "qrcode": "^1.3.2"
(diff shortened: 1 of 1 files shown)


chappjc:
If we wanna upgrade Bootstrap, we can make an issue for that.  The devs have mostly been running with bootstrap 4.1.3, so 4.2/4.3 are largely untested.

PR #1130, opened 2019-03-13
rpcutils/main: disable rpcclient autoreconnect

chappjc:
The RPC client autoreconnect option should be turned OFF.
The assumption is that once dcrdata starts up, there will never be a missed notification.


chappjc:
This is not wonderful, but it needs to be this way until the client put in a disabled state, the consumers in a waiting state, and then perform the appropriate sync before turning things back in.  That's a fair bit of refactoring.

PR #1136, opened 2019-03-14
cache/dcrpg: per-address History/charts data cache

chappjc:
This adds to the address cache all of the history that was obtained via
DB query by `(*ChainDB).TxHistoryData`. This corresponds to the data
retrieved from the API via AJAX for the charts on the address page. e.g.:
`/api/address/{addr}/amountflow/month`, `/api/address/{addr}/types/week`, etc.

The `AddressCacheItem` `metrics` placeholder is changed to `history`, of
a new type `TxHistory` that contains all types of address history used by
the charts on the address page.
`(*AddressCache).HistoryChart` gets the specified chart from cache.
`(*AddressCache).StoreHistoryChart` stores the given chart of the
specified type in cache.

`(*ChainDB).TxHistoryData` is modified similar to other cache-wired
functions to first check the cache. For a cache miss, the cache is held
locked while the actual DB queries are run, and the retrieved history
data is stored in cache.

dbtypes: `TimeIntervals` is changed to a fixed-length array, and
`AllGrouping` is added to the array.  `const NumIntervals = 5` is set
to guarantee the expected length, and provide a const value for
`cache.TxHistory` type in defining its `ChartsData` arrays.

Also, the cache reports logged by the address cache now make a table
of hits/misses for type of cached data.

### DIFF
--- a/db/cache/addresscache.go
+++ b/db/cache/addresscache.go
@@ -358,6 +358,24 @@ func AllDebitAddressRows(rows []*dbtypes.AddressRow) []*dbtypes.AddressRow {
 	return out
 }
 
+// TxHistory contains ChartsData for different chart types (tx type, amount
+// flow, and unspent amount), each with data at known time intervals
+// (TimeBasedGrouping).
+type TxHistory struct {
+	TypeByInterval       [dbtypes.NumIntervals]*dbtypes.ChartsData
+	AmtFlowByInterval    [dbtypes.NumIntervals]*dbtypes.ChartsData
+	UnspentAmtByInterval [dbtypes.NumIntervals]*dbtypes.ChartsData
+}
+
+// Clear sets each *ChartsData to nil, effectively clearing the TxHistory.
+func (th *TxHistory) Clear() {
+	for i := 0; i < dbtypes.NumIntervals; i++ {
+		th.TypeByInterval[i] = nil
+		th.AmtFlowByInterval[i] = nil
+		th.UnspentAmtByInterval[i] = nil
+	}
+}
+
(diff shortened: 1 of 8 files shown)


chappjc:
@dmigwi Note that this is not a cache for the general ChartsData, but rather the per-address data needed by the address page charts.

db/cache/addresscache.go
```diff
@@ -426,14 +444,30 @@ func (d *AddressCacheItem) UTXOs() ([]apitypes.AddressTxnOutput, *BlockID) {
...
 	}
-	return d.metrics, d.blockID()
+
+	var cd *dbtypes.ChartsData
+	switch addrChart {
+	case dbtypes.TxsType:
+		cd = d.history.TypeByInterval[chartGrouping]
+	case dbtypes.AmountFlow:
+		cd = d.history.AmtFlowByInterval[chartGrouping]
+	case dbtypes.TotalUnspent:
```
buck54321: This endpoint is no longer used and misleading at the very least. See https://github.com/decred/dcrdata/pull/845#issuecomment-447893775. Might want to drop it altogether.

chappjc: Great.  I noticed it wasn't used by the address page, but hadn't checked the API thoroughly.

dmigwi:
This looks above board to me.

chappjc:
Periodic cache report looks like:

```
[DBG] PSQL: ADDRESS CACHE: addresses = 12, rows = 511875, utxos = 260024
[DBG] PSQL: ADDRESS CACHE:
		            HITS | MISSES
		rows         245 |     11
		balance       58 |      4
		utxos        158 |      1
		hist           0 |      0
```

PR #1137, opened 2019-03-14
stakedb: drop dep and do FIFO by hand.

dajohi:


chappjc:
However I'm concerned this is going to have a ton more allocs compared to a "container/list" based implementation like Lane did

jrick:
This will be far more optimized than using a (doubly!) linked list.

Another option to avoid reallocs of the backing array would be to copy the items down the queue:

```
copy(q, q[1:])
q[len(q)-1] = item
```

but I wouldn't recommend this here because the values are chainhash.Hash (32 bytes each) in contiguous memory and that is a lot of copying.

Depending on how much the append overallocates, there there won't be a new backing array allocation.  The more it overallocates capacity, the fewer times it will need to reallocate.

jrick:
https://play.golang.org/p/oQgqW9iABBl

chappjc:
There's no question this is more optimal in terms of CPU operations, but it's not at all clear to me this is better memory management.  That said, I think this is a good change to make.  I just want to do some profiling to understand the trade offs.

jrick:
https://play.golang.org/p/58qgU0uIrkF

linked lists are extremely inefficient.. in any language on modern computer architectures...

jrick:
This is another look at the in use objects after a garbage collection runs: https://play.golang.org/p/iEAbaMePpE2.  Not only are the total allocs lower, but there's only a single allocation live at any point in time, instead of two allocations per every container/list.List element.

chappjc:
@jrick Thanks for taking the time to demonstrate that.  It's quite clear now.  I admit that I made some incorrect assumptions about how it was implemented.

PR #1141, opened 2019-03-15
apis: make compression off by default and add --compress-api

chappjc:
GZ compression with a middleware is not always useful, especially when
a reverse proxy like nginx is in use.  This makes compression a config
setting (`--compress-api`, off by default).

This enables compression for ALL Insight API endpoints, and just verbose
and raw dcrdata API endpoints.

chi is updated to include the compression middleware rewrite.


chappjc:
This is now updates the chi/middleware package, which had a broken compression middleware, probably at v4. The compress middleware now has tests, and is rewritten to reduce allocations.

PR #1144, opened 2019-03-15
txhelpers: use async getrawtransaction verbose

chappjc:
The accelerates `txhelpers.UnconfirmedTransactionsForAddress` by making async
getrawtransaction (verbose) RPCs to facilitate greater concurrency.

Also tighten up `MsgTxFromHex`.

### DIFF
--- a/txhelpers/txhelpers.go
+++ b/txhelpers/txhelpers.go
@@ -247,32 +247,30 @@ func TxConsumesOutpointWithAddress(msgTx *wire.MsgTx, addr string,
 	// Send all the raw transaction requests
 	type promiseGetRawTransaction struct {
 		result rpcclient.FutureGetRawTransactionVerboseResult
-		prevtx chainhash.Hash
+		inIdx  int
 	}
 	numPrevOut := len(msgTx.TxIn)
-	promisesGetRawTransaction := make([]promiseGetRawTransaction, numPrevOut)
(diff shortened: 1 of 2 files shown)


chappjc:
Tested on address page and API endpoints.  MsgTxFromHex test added.

txhelpers/txhelpers.go
```diff
@@ -241,30 +244,55 @@ func TxInvolvesAddress(msgTx *wire.MsgTx, addr string, c VerboseTransactionGette
...
+			continue // coinbase or stakebase
+		}
+		promisesGetRawTransaction[inIdx] = promiseGetRawTransaction{
+			result: c.GetRawTransactionVerboseAsync(hash),
+			prevtx: *hash,
+		}
+	}
+
 	// For each TxIn of this transaction, inspect the previous outpoint.
 	for inIdx, txIn := range msgTx.TxIn {
```
dajohi: range over promisesGetRawTransaction and I you can ditch the zeroHash.IsEqual check below since it is done above.

chappjc: I suppose we can do that, but we're checking `prevOut := &txIn.PreviousOutPoint` so we'll need to either store those in the first loop is eliminate the checks in the second.

chappjc: Good suggestion.  I had another shot and came up with https://github.com/decred/dcrdata/pull/1144/commits/f5de2c6109544281ab946b62f31a2bcbb4515302

PR #1146, opened 2019-03-15
explorer: add steps to ticket price chart

### LINKED ISSUE #991: Ticket price chart must have steps
Linear interpolation on ticket price chart is misleading: ticket price does not linearly change between adjustments. https://dcred.eu/posStats shows accurate steps.

buck54321:
Resolves #991

### DIFF
--- a/public/js/controllers/charts_controller.js
+++ b/public/js/controllers/charts_controller.js
@@ -87,12 +87,15 @@ function nightModeOptions (nightModeOn) {
 }
 
 function ticketsFunc (gData) {
-  return map(gData.time, (n, i) => {
-    return [
-      new Date(n),
-      gData.valuef[i]
-    ]
-  })
+  var data = []
+  var lastIdx = gData.time.length - 1
+  for (let i = 0; i < lastIdx; i++) {
+    data.push([new Date(gData.time[i]), gData.valuef[i]])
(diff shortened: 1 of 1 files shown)


public/js/controllers/charts_controller.js
```diff
@@ -87,12 +87,15 @@ function nightModeOptions (nightModeOn) {
...
-    return [
-      new Date(n),
-      gData.valuef[i]
-    ]
-  })
+  var data = []
+  var lastIdx = gData.time.length - 1
+  for (let i = 0; i < lastIdx; i++) {
+    data.push([new Date(gData.time[i]), gData.valuef[i]])
+    data.push([new Date(gData.time[i + 1]), gData.valuef[i]])
```
chappjc: It's somewhat unfortunate that this doubles the data length.  I was hoping the graph lib would have a plot option like "stairs" or something.

chappjc:
Looks good:

image

public/js/controllers/charts_controller.js
buck54321: Actually it does. I didn't see it before, but dygraphs has a `stepPlot` options. I'll fix it now.

chappjc: Excellent!  Thanks.

chappjc:
Still looking good with the `stepPlot` option on.

PR #1147, opened 2019-03-15
explorer: improve dark mode for visual blocks

### LINKED ISSUE #1006: visual blocks need dark mode color scheme

buck54321:
Resolves #1006. I'm thinking is just needed these small tweaks. @raedah can you confirm?

image


raedah:
Black border is an improvement and probably good enough for now. I am sure there are other interesting creative things that can be done with this representation. At one point we talked about creating different themes.

PR #1149, opened 2019-03-15
explorer: strip leading/trailing whitespace from search

chappjc:
This strips leading and trailing Unicode-recognized whitespaces from the search term.
Removing these characters allows pasting data that might have tabs, spaces, etc.
that were accidentally picked up when copied from somewhere else.
Return `http.StatusBadRequest` (400) for an empty search term.


chappjc:
@buck54321 That's great.  Would you put up a second PR?

PR #1152, opened 2019-03-16
politeia/piclient: Check for valid http status code

### LINKED ISSUE #1132: proposals db: invalid character
```
[ERR] DATD: updating proposals db failed: invalid character '<' looking for beginning of value
```

This was a transient error, but there may be a parsing issue we should look into.

dmigwi:
fixes #1132

Since Politeia [guarantees](https://github.com/decred/politeia/blob/master/politeiawww/api/v1/api.md#http-status-codes-and-errors) that most requests should return http status code `200 OK`. 
Any of the Politeia API endpoints request returned with another other status code is considered to be a failed request.

### DIFF
--- a/gov/politeia/piclient/piclient.go
+++ b/gov/politeia/piclient/piclient.go
@@ -28,8 +28,14 @@ func HandleGetRequests(client *http.Client, URLPath string) ([]byte, error) {
 	}
 
 	response, err := client.Get(URLPath)
-	if err != nil {
-		return nil, err
+	if err != nil || response == nil {
+		return nil, fmt.Errorf("request failed: %v", err)
+	}
+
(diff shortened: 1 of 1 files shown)


gov/politeia/piclient/piclient.go
```diff
@@ -28,8 +28,14 @@ func HandleGetRequests(client *http.Client, URLPath string) ([]byte, error) {
 	}
 
 	response, err := client.Get(URLPath)
-	if err != nil {
-		return nil, err
+	if err != nil || response == nil {
+		return nil, fmt.Errorf("Request failed: %v", err)
```
chappjc: "request failed: %v" (starting with lower case as per Effective Go)

PR #1155, opened 2019-03-18
dcrrates: add readme to rateserver

buck54321:


dcrrates/rateserver/README.md
```diff
@@ -0,0 +1,49 @@
...
+```
+
+See [DCRData's notes on Go](https://github.com/decred/dcrdata#building-dcrdata-with-go-111)
+for additional configuration details.
+
+### Using RateServer with DCRData
+
+To point DCRData to RateServer, you must specify DCRData's `ratemaster` and
+`ratecert` configuration options. `exchange-monitor` must also be enabled. Set
+`ratemaster` to the same value as RateServer's `listen` parameter (defaults to
```
chappjc: This is the only part that's a little confusing.  I might set rateserver's `listen` to `0.0.0.0:7778`, but with dcrdata the `ratemaster` would have to be the public IP or fqdn.

PR #1161, opened 2019-03-20
Use updated full-size favicon.

jholdstock:
A new favicon was generated which fills the maximum available space, rather than having a large border. This makes the DCR logo more visible, as can be seen in the following screenshot comparing dcrdata and dcrweb:

Screenshot from 2019-03-20 07-12-12


public/images/favicon/manifest.json
```diff
@@ -1,18 +1,20 @@
...
         {
-            "src": "/images/favicon/android-chrome-512x512.png?v=fi5jKKtbwv",
+            "src": "/images/favicon/android-chrome-512x512.png?v=vMddPLeYWy",
             "sizes": "512x512",
             "type": "image/png"
         }
     ],
     "theme_color": "#ffffff",
     "background_color": "#ffffff",
+    "start_url": "https://decred.org",
```
dajohi: not sure this is correct.  may want just "."

public/images/favicon/browserconfig.xml
```diff
@@ -2,9 +2,11 @@
 <browserconfig>
     <msapplication>
         <tile>
-            <square150x150logo src="/images/favicon/mstile-150x150.png?v=fi5jKKtbwv"/>
-            <square310x310logo src="/images/favicon/mstile-310x310.png?v=fi5jKKtbwv"/>
-            <TileColor>#2973ff</TileColor>
+            <square70x70logo src="/images/favicon/mstile-70x70.png?v=yHh3NA"/>
+            <square150x150logo src="/images/favicon/mstile-150x150.png?v=yHh3NA"/>
+            <square310x310logo src="/images/favicon/mstile-310x310.png?v=yHh3NA"/>
+            <wide310x150logo src="/images/favicon/mstile-310x150.png?v=yHh3NA"/>
```
dajohi: this is missing.

public/images/favicon/manifest.json
chappjc: I'm not too familiar with the manifest.json, but from what I read about `start_url`, it seems like it should be a relative URL. If it is absent, it is the new desktop icon supposedly uses the location from which it was created.  Another common recommendation (MDN and developers.google.com) seems to be  `"start_url": "."`

public/images/favicon/browserconfig.xml
jholdstock: agh, I need to stop using `commit -a`. Fixed

public/images/favicon/manifest.json
jholdstock: good spot. Fixed

jholdstock:
Fixed both issues. I need to go through some other repos and fix the `start_url` as well.

public/images/favicon/manifest.json
chappjc: I will try to test this out on a phone, but it's still a little unclear what `"."` means.  From https://developer.mozilla.org/en-US/docs/Web/Manifest:

> The URL that loads when a user launches the application (e.g. when added to home screen), typically the index. Note that this has to be a relative URL, *relative to the manifest url.*

Emphasis added by me.

chappjc:
I think android icons might need to be refined

image

With Android's newer circle shapes, it doesn't look great.  I just used Android Asset Studio to create a set of icons for different shapes and resolutions, but not sure how packaging works:
[ic_launcher.zip](https://github.com/decred/dcrdata/files/2997250/ic_launcher.zip)

chappjc:
Here's another fun take on material design for our icons: 
[icons(1).zip](https://github.com/decred/dcrdata/files/2997291/icons.1.zip)
image

chappjc:
Android Studio's "Image Asset Studio" makes a more complete icon set like:

image

Although I don't know how this applies to progressive web apps as opposed to native android apps.

[ic_launcher-adaptable.zip](https://github.com/decred/dcrdata/files/2997385/ic_launcher-adaptable.zip)

chappjc:
Am I nitpicking or are we OK with this?  I'd rather see

image

or 

image

instead of

image

But it's not a show-stopper.

jholdstock:
A few changes pushed:
- Seperated the HTML elements in `extras.tmpl` and added some comments. Dealing with those was driving me crazy
- Deleted some unused icons
- Changed `manifest.json` values `start_url` and `display` as discussed on Matrix
- Use correct Decred dark blue
- Changed icons in `manifest.json` to rounded icons provided above, to improve Android support

PR #1165, opened 2019-03-20
update icon set

dajohi:


dajohi:
#1161 is already.

public/images/favicon/manifest.json
```diff
@@ -1,18 +1,20 @@
...
         {
-            "src": "/images/favicon/android-chrome-512x512.png?v=fi5jKKtbwv",
+            "src": "/images/favicon/android-chrome-512x512.png?v=vMddPLeYWy",
             "sizes": "512x512",
             "type": "image/png"
         }
     ],
     "theme_color": "#ffffff",
     "background_color": "#ffffff",
+    "start_url": ".",
```
chappjc: I'm not sure this is right either.  From what I can tell from online docs, `start_url` is a path relative _to the manifest.json` location_.  To have it open to the root of the site, I think we'll need `"/"`.

PR #1170, opened 2019-03-21
agenda db(storm): Add auto update

dmigwi:
Adds the agenda db (storm db version) auto update at intervals of 5 blocks.

### DIFF
--- a/explorer/explorer.go
+++ b/explorer/explorer.go
@@ -268,6 +268,7 @@ type explorerUI struct {
 
 	invsMtx sync.RWMutex
 	invs    *types.MempoolInfo
+	client  agendas.DeploymentSource
 }
 
 // AreDBsSyncing is a thread-safe way to fetch the boolean in dbsSyncing.
@@ -329,6 +330,7 @@ type ExplorerConfig struct {
 	PoliteiaURL       string
 	MainnetLink       string
 	TestnetLink       string
+	RPCClient         agendas.DeploymentSource
 }
 
 // New returns an initialized instance of explorerUI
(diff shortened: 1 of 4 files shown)


explorer/explorer.go
```diff
@@ -618,26 +621,41 @@ func (exp *explorerUI) Store(blockData *blockdata.BlockData, msgBlock *wire.MsgB
...
-		// execution.
-		if newBlockData.Height%20 == 0 {
+		// Update in every 5 blocks implys that in approximately every 25mins
+		// an update will be queried.
+		if newBlockData.Height%5 == 0 {
+			// Update the Agendas DB. Run this asynchronously to avoid
+			// blocking other processes.
 			go func() {
-				err := exp.proposalsSource.CheckProposalsUpdates()
+				err := exp.agendasSource.CheckAgendasUpdates(exp.client,
```
chappjc: The need to make a `client` field in `explorerUI` just to support this operation is not appealing.  Note that a way around this would be to have the explorerUI constructor take as an input argument a `trigger func()`.  But this just makes it clearer that it's really not the place for this. (BTW, seems like agendaSouce should/could have it's own internal client.)  Anyway,  I think the other option of making and registering another a BlockDataSaver (or two) that only calls the agenda ad proposal DB updates, and does not use the actual block data. Each type, the agenda instance and proposal instance can implement Store for this reason.

dmigwi: What if I passed the rpc client as an argument like
 `func NewAgendasDB(dbPath string, client DeploymentSource) (*AgendaDB, error) {`
?
BlockDataSaver interface shall try to pass arguments that may not be that useful in both agendasDB and ProposalDB update.

PR #1174, opened 2019-03-24
api: create endpoint for ticket-specific info

### LINKED ISSUE #1024: api: missed tickets should name the block
Follow-up on the explorer change in https://github.com/decred/dcrdata/pull/1023

buck54321:
Adds the /tx/{txid}/tinfo endpoint, which responds with ticket status, purchase and lottery blocks, and spending txid.

Some endpoints for testing

- immature: /api/tx/10659139fc3a4ef2e00d8539180a761b1c2d1b823645ecff5c8ffba0dc909403/tinfo?indent=1
- live: /api/tx/07f0daf8be13be8e995931afb98b725d8ef23529c26749100ef26821e8eb3025/tinfo?indent=1
- missed, not revoked: /api/tx/08b98e68e18ebb82440d07f3dc70cc8c017df1dcb639493e9b7cd356d66174ae/tinfo?indent=1
- missed, revoked: /api/tx/7405d619e673ca4bb4681951828dcf1893e4004973133ff012e30e7c182bf168/tinfo?indent=1
- voted: /api/tx/5ba836dec32969110487f857be9a59ca98898860c84f66d5497e211f525e66e4/tinfo?indent=1
- expired, not revoked: /api/tx/0974112f3f7a41b2bafbbd4c0c7bd61b9db6f9838721d82e9bc59c42e40f6da6/tinfo?indent=1

Resolves #1024

### DIFF
--- a/db/dbtypes/types.go
+++ b/db/dbtypes/types.go
@@ -484,24 +484,6 @@ func (v VoteChoice) String() string {
 	}
 }
 
-// TicketInfo combines spend and pool statuses and relevant block and spending
-// transaction IDs.
-type TicketInfo struct {
-	Status           string     `json:"status"`
-	PurchaseBlock    *TinyBlock `json:"purchase_block"`
-	MaturityHeight   uint32     `json:"maturity_height"`
-	ExpirationHeight uint32     `json:"expiration_height"`
-	LotteryBlock     *TinyBlock `json:"lottery_block"`
-	Vote             *string    `json:"vote"`
-	Revocation       *string    `json:"revocation"`
-}
-
-// TinyBlock is the hash and height of a block.
-type TinyBlock struct {
-	Hash   string `json:"hash"`
-	Height uint32 `json:"height"`
-}
-
 // ChoiceIndexFromStr converts the vote choice string to a vote choice index.
 func ChoiceIndexFromStr(choice string) (VoteChoice, error) {
 	switch choice {
--- a/db/dbtypes/types.go
+++ b/db/dbtypes/types.go
@@ -484,6 +484,24 @@ func (v VoteChoice) String() string {
 	}
 }
 
+// TicketInfo combines spend and pool statuses and relevant block and spending
+// transaction IDs.
+type TicketInfo struct {
+	Status           string     `json:"status"`
+	PurchaseBlock    *TinyBlock `json:"purchase_block"`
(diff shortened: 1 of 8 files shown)


db/dbtypes/types.go
```diff
@@ -484,6 +484,24 @@ func (v VoteChoice) String() string {
...
+// TicketInfo combines spend and pool statuses and relevant block and spending
+// transaction IDs.
+type TicketInfo struct {
+	Status           string     `json:"status"`
+	PurchaseBlock    *TinyBlock `json:"purchase_block"`
+	MaturityHeight   uint32     `json:"maturity_height"`
+	ExpirationHeight uint32     `json:"expiration_height"`
+	LotteryBlock     *TinyBlock `json:"lottery_block"`
+	Vote             *string    `json:"vote"`
+	Revocation       *string    `json:"revocation"`
```
buck54321: I prefer to have null when these are not valid, rather than omitting or empty strings. But I don't see this done (`*string`) anywhere else, so I'm not sure if this is undesirable or not idiomatic.

db/dcrpg/pgblockchain.go
```diff
@@ -1292,6 +1292,53 @@ func (pgb *ChainDB) ticketPoolVisualization(interval dbtypes.TimeBasedGrouping)
 	return
 }
 
+// GetTicketInfo retrieves information about the pool and spend statuses, the
```
chappjc: Approved, but pgblockchain.go seems to need a go fmt.

db/dcrpg/pgblockchain.go
```diff
@@ -1292,6 +1292,53 @@ func (pgb *ChainDB) ticketPoolVisualization(interval dbtypes.TimeBasedGrouping)
...
+// purchase block, the lottery block, and the spending transaction.
```
chappjc:
```suggestion
// purchase block, the lottery block, and the spending transaction.
```

db/dbtypes/types.go
```diff
@@ -484,6 +484,24 @@ func (v VoteChoice) String() string {
 	}
 }
 
+// TicketInfo combines spend and pool statuses and relevant block and spending
+// transaction IDs.
+type TicketInfo struct {
```
chappjc: I also just noticed this in in dbtypes, but it should really be in api/types.

PR #1177, opened 2019-03-25
Update politeia dep after politeia updated api/types dep.

chappjc:
After this is merged:
1. tag dcrdata `v4.0.0-rc2`
2. Make go.mod for api/types, requiring dcrdata `v4.0.0-rc2`
3. Tag `api/types/v1.0.1`


chappjc:
Unfortunately, this will make us use an older storm dep until we make the api/types module and have Pi use it.

PR #1179, opened 2019-03-25
use replace statements

chappjc:
This adds replace statements to the main go.mod and api/types/go.mod to the local code can be used.

This also changes api/types to require v4.0.0-rc4, and dcrdata to require api/types/v1.0.2

### DIFF
--- a/api/types/go.mod
+++ b/api/types/go.mod
@@ -2,5 +2,6 @@ module github.com/decred/dcrdata/api/types
 
 go 1.11
 
-require github.com/decred/dcrdata v4.0.0-rc3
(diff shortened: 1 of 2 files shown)


api/types/go.mod
```diff
@@ -2,5 +2,6 @@ module github.com/decred/dcrdata/api/types
 
 go 1.11
 
-require github.com/decred/dcrdata v4.0.0-rc3
+require github.com/decred/dcrdata/v4 v4.0.0-rc4
```
chappjc: The missing /v4 here was an error.

PR #1182, opened 2019-03-25
txhelpers/v1.0.0 module

chappjc:
This creates the txhelpers module at v1.0.0.

dcrdata's go.mod is also updated to use the txhelpers module, with a `replace` until the txhelpers mod is tagged and dcrdata's go.sum is updated.


chappjc:
txhelpers has no dependencies in the dcrdata module, so it builds without any replaces, and no go.sum updates will be needed for it.  A `txhelpers/v1.0.0` tag may be created upon merge.

PR #1200, opened 2019-03-26
explorer: fix layout of last block data for small screens

buck54321:
image


chappjc:
Works great on desktop and all virtual mobile devices I tried.  Sneaking this in to final RC of 4.0.

PR #1201, opened 2019-03-26
explorer: turn off auto-focus on the search bar

buck54321:


chappjc:
Working as expected. FYI, everyone: the `\` key gives the search bar focus.

PR #1215, opened 2019-03-27
Add a module graph

chappjc:
modules svg


chappjc:
@buck54321 Given the last several large commits, has anything changed (other than that indirect dcrpg require that I'm ignoring because it's for a test)?

chappjc:
`dot -Tsvg modules.gv > modules.svg`

BTW I'm noticing I've mixed tabs and spaces in modules.gv.  Fixing...

PR #1217, opened 2019-03-28
socket.io test: update test html with address test

chappjc:
Move api/insight/test_socket.html to public/js/test_insight_socket.io.html,
and modify it:

1. subscribe to an address room on connect (Dcur2mcGjmENx4DhNqDctW5wJCVyT3Qeqkx)
2. event handler for the 'Dcur2mcGjmENx4DhNqDctW5wJCVyT3Qeqkx'
event that will be received from the server when subscribed.
3. auto-reconnect handling to the test html.
4. window.location.origin input to `io(...)` allows it to run remotely or locally (e.g. https://explorer.dcrdata.org or http://127.0.0.1:777)

The move to public/js is to allow testing on a web server rather than the
local file system.

Put v2.2.0 of the official socket.io javascript client in public/js/vendor so the
test can be run via the web server instead of the file system.


public/js/test_insight_socket.io.html
```diff
@@ -0,0 +1,54 @@
...
+            var pr = document.createElement('p')
+            pr.textContent = str
+            updates.appendChild(pr)
+        }
+
+        var sio = io(window.location.origin, { path: '/insight/socket.io', transports: ['websocket'], upgrade: false });
+
+        sio.on('connect', function () {
+            update('Connected. Updates will be shown below.')
+            sio.emit('subscribe', 'Dcur2mcGjmENx4DhNqDctW5wJCVyT3Qeqkx');
```
chappjc: Joining a "room" on the server is accomplished by emitting a "subscribe" event. In this case the name of the room/subscription is the address.

public/js/test_insight_socket.io.html
```diff
@@ -0,0 +1,54 @@
...
+        sio.on('disconnect', (reason) => {
+            if (reason === 'io server disconnect') {
+                // Disconnection was initiated by the server. Reconnect manually.
+                sio.connect();
+            }
+            // else the socket will automatically try to reconnect
+        });
+
+        sio.on('Dcur2mcGjmENx4DhNqDctW5wJCVyT3Qeqkx', function (data) {
+            update("New Dcur2mcGjmENx4DhNqDctW5wJCVyT3Qeqkx address event received: " +
```
chappjc: When subscribed to address events, an event handler must be defined. This defines the handler for the 'Dcur2mcGjmENx4DhNqDctW5wJCVyT3Qeqkx' room/subscription event.

public/js/test_insight_socket.io.html
```diff
@@ -0,0 +1,54 @@
...
+    <script type="application/javascript">
+        var updates = document.getElementById('updates')
+
+        var update = function (str) {
+            var pr = document.createElement('p')
+            pr.textContent = str
+            updates.appendChild(pr)
+        }
+
+        var sio = io(window.location.origin, { path: '/insight/socket.io', transports: ['websocket'], upgrade: false });
```
chappjc: `window.location.origin` allows it to run remotely or locally (e.g. https://explorer.dcrdata.org or http://127.0.0.1:777)

chappjc:
@buck54321 I meant to ask you about this.  I've added socket.io.slim.js (and map) to the public/js folder so that the socket.io html client can be tested on the web server.  It doesn't seem to affect any of the packaging, but there may be unforseen consequences.  Is this look OK to you?

buck54321:
Should be fine. Webpack only knows things imported in `index.js` or the controllers.

PR #1219, opened 2019-03-28
CICD: Run tests for all dcrdata modules

dmigwi:
`go test ./...` seems to be working only for the package where `run_tests.sh` scripts is located.
`go test all` runs way too many unnecessary tests. https://go-review.googlesource.com/c/go/+/134095/1/src/cmd/go/internal/test/test.go

Therefore dcrdata submodules have to be passed as arguments as `go tests ./... {dcrdata_submodules}` for all the necessary tests to be run.

### DIFF
--- a/run_tests.sh
+++ b/run_tests.sh
@@ -22,6 +22,25 @@ set -ex
 # Default GOVERSION
 [[ ! "$GOVERSION" ]] && GOVERSION=1.11
 REPO=dcrdata
+ROOTMODULE="github.com/decred/dcrdata"
+ALLMODULES="$ROOTMODULE/api/types
+  $ROOTMODULE/blockdata
+  $ROOTMODULE/db/cache
+  $ROOTMODULE/db/dbtypes
+  $ROOTMODULE/db/dcrpg
+  $ROOTMODULE/db/dcrsqlite
+  $ROOTMODULE/exchanges
+  $ROOTMODULE/explorer/types
+  $ROOTMODULE/gov/agendas
+  $ROOTMODULE/gov/politeia
+  $ROOTMODULE/mempool
+  $ROOTMODULE/middleware
+  $ROOTMODULE/pubsub
+  $ROOTMODULE/pubsub/types
+  $ROOTMODULE/rpcutils
+  $ROOTMODULE/semver
+  $ROOTMODULE/stakedb
+  $ROOTMODULE/txhelpers"
 
 testrepo () {
   TMPDIR=$(mktemp -d)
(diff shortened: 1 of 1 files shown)


run_tests.sh
```diff
@@ -22,6 +22,25 @@ set -ex
...
+  $ROOTMODULE/gov/agendas
+  $ROOTMODULE/gov/politeia
+  $ROOTMODULE/mempool
+  $ROOTMODULE/middleware
+  $ROOTMODULE/pubsub
+  $ROOTMODULE/pubsub/types
+  $ROOTMODULE/rpcutils
+  $ROOTMODULE/semver
+  $ROOTMODULE/stakedb
+  $ROOTMODULE/txhelpers"
```
chappjc: seems like there should be a way to get this list with `go list -m`, I can't a syntax that gets it I suppose this list is ok.  @jrick Know a way to handle this testing other than manually listing the modules?

run_tests.sh
```diff
@@ -40,7 +59,7 @@ testrepo () {
   git clone https://github.com/dcrlabs/bug-free-happiness $TMPDIR/test-data-repo
   tar xvf $TMPDIR/test-data-repo/stakedb/test_ticket_pool.bdgr.tar.xz -C ./stakedb
 
-  env GORACE='halt_on_error=1' go test -v -race ./...
+  env GORACE='halt_on_error=1' go test -v -race ./... $ALLMODULES
```
chappjc: So I understand, the `./...` tests the packages belonging to the root module, and then we have to list the other modules?

chappjc: nvm my comment, doing that would not test packages like `explorer` that are not modules.

run_tests.sh
dmigwi: since the concept of multiple modules in a single project is relatively new, I suspect that such a solution doesn't exist yet. Anyway, I am doing some wide research on the same to see what progress in being made by the go dev team.

jrick: This was already merged, and I'm just seeing it now, but you could do like [dcrd](https://github.com/decred/dcrd/blob/6dbd893/run_tests.sh#L43-L52) or [dcrwallet](https://github.com/decred/dcrwallet/blob/master/run_tests.sh) does.

dmigwi: Sure. Noted, I will add the change.

chappjc: I had tried `'{{.Path}}'` instead of `'{{.Dir}}'` and ended up with the old `v3` module getting listed too.  `'{{.Dir}}'` solves it.  Thanks!

PR #1232, opened 2019-04-02
explorer: check dark mode cookie server-side

buck54321:
Sets the `darkBG` CSS class on the body ahead of time to prevent flashing of light mode before script loads.

I'm also sneaking this in so that user's fiat index preference can be stored and `ExchangeBot` can serve personalized results.

### DIFF
--- a/explorer/explorerroutes.go
+++ b/explorer/explorerroutes.go
@@ -30,7 +30,7 @@ import (
 	humanize "github.com/dustin/go-humanize"
 )
 
-// Information from the request cookies.
+// Cookies contains information from the request cookies.
 type Cookies struct {
 	DarkMode bool
 }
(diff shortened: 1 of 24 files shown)


explorer/explorerroutes.go
```diff
@@ -30,6 +30,11 @@ import (
 	humanize "github.com/dustin/go-humanize"
 )
 
+// Information from the request cookies.
+type Cookies struct {
```
chappjc: for godoc, please begin with `// Cookies`

explorer/explorerroutes.go
```diff
@@ -2040,11 +2048,15 @@ func (exp *explorerUI) StatsPage(w http.ResponseWriter, r *http.Request) {
...
 // This is particularly useful for extras.tmpl, parts of which
 // are used on every page
-func (exp *explorerUI) commonData() *CommonPageData {
+func (exp *explorerUI) commonData(r *http.Request) *CommonPageData {
 	tip, err := exp.blockData.GetTip()
 	if err != nil {
 		log.Errorf("Failed to get the chain tip from the database.: %v", err)
 	}
+	darkMode, ckErr := r.Cookie("dcrdataDarkBG")
+	if err != nil && err != http.ErrNoCookie {
```
chappjc: did you mean to check `ckErr`?  Also it seems `commonData` should return when `GitTip` fails (template execution would probably just fail).

PR #1239, opened 2019-04-05
dcrpg: check and set synchronous_commit

chappjc:
Running PostgreSQL with `synchronous_commit = on` is an absolute performance killer,
and a quite common configuration mistake.  However, it is safe to set it to `"off"`.

These changes check the current setting, and set it to `"off"` if it is not already set that way.


chappjc:
Added verification after changing the setting.

The log message is:

```
[WRN] PSQL: PERFORMANCE ISSUE! The synchronous_commit setting is = "on". Changing it to "off".
```

PR #1243, opened 2019-04-10
exchanges: add websocket support to exchanges package

buck54321:
A websocket layer is implement on PoloniexExchange, since they offer more data by websocket than by HTTP.


exchanges/exchanges.go
```diff
@@ -2153,23 +2153,11 @@ func (poloniex *PoloniexExchange) processTrade(tradeParams []interface{}) (*polo
...
-		existingOrder, found = poloniex.buys[binKey]
-	}
-	if !found {
-		return nil, -1, fmt.Errorf("no order found to match reported poloniex trade")
-	}
-
-	order := &poloniexOrder{
+	trade := &poloniexOrder{
 		price:  price,
-		volume: existingOrder.volume - volume,
```
buck54321: This was wrong. I'm fairly certain that this signal is for trades that take place off of the order books.

exchanges/exchanges.go
```diff
@@ -1821,9 +1956,367 @@ func (r PoloniexCandlestickResponse) translate(bin candlestickKey) Candlesticks
...
+		}
+		poloniex.setWsFail(fmt.Errorf("unknown code in single-element poloniex response: %d", intCode))
+		return
+	}
+	if len(msg) < 3 {
+		poloniex.setWsFail(fmt.Errorf("poloniex websocket message had unexpected length %d", len(msg)))
+		return
+	}
+
+	responseList, ok := msg[2].([]interface{})
```
chappjc: Should it be treated as OK if `len(msg)>3`?  If not, maybe consider a switch:

```go
switch len(msg) {
	case 1:
		// heartbeat -- handle it and return here
		...
	case 3:
		// good -- keep going
	default:
		// error
		...
		return
}
```

exchanges/exchanges.go
```diff
@@ -1821,9 +1956,367 @@ func (r PoloniexCandlestickResponse) translate(bin candlestickKey) Candlesticks
...
+			// trade, direction, err = poloniex.processTrade(updateParams)
+		}
+
+		if direction == poloniexAskDirection {
+			newAsks[int64(order.price*1e8)] = order
+		} else if direction == poloniexBuyDirection {
+			newBids[int64(order.price*1e8)] = order
+		} else {
+			poloniex.setWsFail(fmt.Errorf("Unknown poloniex update direction indicator: %d", direction))
+			return
```
chappjc: Would def go for a switch on `direction` for these.  It's generally considered good practice in go to prefer a switch over if-else if-else chains when appropriate.

exchanges/exchanges.go
```diff
@@ -1821,9 +1956,367 @@ func (r PoloniexCandlestickResponse) translate(bin candlestickKey) Candlesticks
...
+		}
+		return
+	}
+
+	if code != poloniexOrderUpdateKey && code != poloniexTradeUpdateKey {
+		poloniex.setWsFail(fmt.Errorf("Unexpected code in first element of poloniex websocket response list: %s", code))
+		return
+	}
+
+	poloniex.orderMtx.RLock()
```
chappjc: What is being protected against concurrent writes in this section?

exchanges/exchanges_test.go
```diff
@@ -5,6 +5,8 @@ package exchanges
 
 import (
 	"context"
+	"fmt"
+	// "encoding/json"
```
chappjc: Leftover from test data generation?

exchanges/exchanges_test.go
```diff
@@ -118,3 +120,241 @@ func TestSlaveBot(t *testing.T) {
...
+
+var poloniexOrderbookUpdate = []byte(`[
+    14,
+    85537933,
+    [
+        [
+            "o",
+            0,
+            "0.00011358",
+						"0.00000000"
```
chappjc: Peculiar formatting here, but I suppose it doesn't matter.

exchanges/websocket.go
```diff
@@ -0,0 +1,93 @@
...
+	done     chan struct{}
+}
+
+// Read is a wrapper for gorilla's ReadMessage that satisfies websocketFeed.Read.
+func (client *socketClient) Read() (msg []byte, err error) {
+	_, msg, err = client.conn.ReadMessage()
+	return
+}
+
+// Write is a wrapper for gorilla WriteMessage. Satisfies websocketFeed.Write.
```
chappjc: Please mention that the wrapper's function is also to marshal to JSON the data in `msg`.

exchanges/exchanges.go
buck54321: Thanks. When I was processing trades as updates to the order book (incorrectly), the lock was needed, but I corrected that. Dropping the lock now.

exchanges/exchanges.go
```diff
@@ -497,6 +523,98 @@ func (xc *CommonExchange) state() *ExchangeState {
 	return xc.currentState
 }
 
+// WebsocketProcessor is a callback for new websocket messages from the server.
+type WebsocketProcessor func([]byte)
+
+// Only the fields are protected for these, so okay to grab them like this.
```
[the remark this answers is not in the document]
buck54321: Gorilla's read and write methods need concurrency control. The `websocketFeed` itself is safe for concurrent use, but the `commonExchange` field will be updated during a reconnection. 

For `websocketFeed`, writing is protected by the `(socketClient).writeMtx`, and the read is good because `commonExchange` has a single read loop assigned to each connection. 

I was trying to say that once you have the `websocketFeed`, which as you said is a pointer underneath, it is safe to use concurrently. The `commonExchange.ws*` fields need protection, though. 

No need to match request with response for the websocket feeds, but there is a sequence ID I should probably check to make sure orderbook updates are processed in order.

exchanges/exchanges_test.go
buck54321: I just use it a lot for inspection. No need to keep it though.

exchanges/exchanges_test.go
buck54321: Weird. It's not like that for me. I remember seeing some discrepancy in formatting of the raw strings in database queries a while back. Not sure what causes this.

exchanges/exchanges.go
```diff
@@ -1821,9 +1956,367 @@ func (r PoloniexCandlestickResponse) translate(bin candlestickKey) Candlesticks
...
+	poloniexTradeUpdateKey      = "t"
+	poloniexAskDirection        = 0
+	poloniexBuyDirection        = 1
+	poloniexMaxErrors           = 5
+)
+
+// Poloniex has a string code in the result array indicating what type of
+// message it is.
+func firstCode(responseList []interface{}) string {
+	firstElement, ok := responseList[0].([]interface{})
```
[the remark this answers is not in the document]
buck54321: Now checking for zero-length response list in `(*PoloniexExchange).processWsMessage`

chappjc:
rates.dcrdata.org now running with the polo websocket connection.

Issue #1020, opened 2019-02-16
dcrsqlite: possible unclosed connection


chappjc:
After a full sync:

```
2019-02-16 03:41:17.341 [DBG] PSQL: SyncChainDB completed at height 319052.
2019-02-16 03:41:17.341 [INF] DATD: PostgreSQL sync ended at height 319052
2019-02-16 03:41:17.366 [DBG] SQLT: Setting block gate height to 319053
2019-02-16 03:41:17.390 [INF] SQLT: Current best block (chain server):     319053
2019-02-16 03:41:17.390 [INF] SQLT: Current best block (sqlite block DB):  319052
2019-02-16 03:41:17.390 [INF] SQLT: Current best block (stakedb):          319052
2019-02-16 03:41:17.401 [INF] PSQL: Processing blocks 319053 to 319053...
2019-02-16 03:41:17.601 [INF] SQLT: Scanning blocks 319053 to 319053 (40944 live)...
2019-02-16 03:41:47.638 [ERR] SQLT: resyncDBWithPoolValue failed: database is locked
2019-02-16 03:41:47.639 [INF] DATD: SQLite sync ended at height 319052
2019-02-16 03:41:47.639 [ERR] DATD: dcrsqlite.SyncDBAsync failed at height 319052: Unable to store block summary in database: database is locked.
```

chappjc:
Partially resolved by https://github.com/decred/dcrdata/pull/1151.  Changing milestone to 4.1 as we will continue to investigate this.

chappjc:
A CLI switch for dcrdata like `--sqlite-one-connection` should be added to use `SetMaxOpenConns(1)`, hurting performance but hopefully eliminating the possibility for "database is locked" errors.

https://github.com/decred/dcrdata/blob/cb6d76dd69f9adb761330c3592e69d662e8b1bed/db/dcrsqlite/sqlite.go#L345-L355

chappjc:
dcrsqlite is gone, so this issue is irrelevant.

PR #1244, opened 2019-04-11
config: add --sqlite-max-conns=[int] switch

chappjc:
This is a workaround for machines hit by https://github.com/decred/dcrdata/issues/1020


chappjc:
Been working well on beta site.

PR #1247, opened 2019-04-15
Coin Supply chart: Add Inflation chart axis

dmigwi:
The Predicted coin supply from the start till till the latest time a coin was mined has its data plotted on an axis.
An extra legend entry to show percentage change between the Predicted coin supply and the Actual coin supply was added. To calculate the percentage change, the actual supply is used as the base.

Formula used to generate the inflation axis data is derived from this data https://dcrstats.com/subsidy (`Maximum total DCR subsidy`)

*h* = **Given Block Height**
*TCM<sub>h</sub>* = **Total Coins Mined till height `h`**
*R<sub>0</sub>* = **Block reward at genesis block** = `31.19582664` for mainnet

*rInterval* = **Interval of block subsidy reduction interval** = `math.Floor(h / SubsidyReductionInterval)`
*rSubsidyRate* = **Subsidy reduction rate for every `rInterval`** = `MulSubsidy / DivSubsidy` 

equation

**Coin Supply chart has two Axis:**
- Predicted Coin Supply - Should be the upper limit of the max coins that can be mined at any given time.
- Actual Coin Supply - Show the actual number of coins mined at a given time.
 
new

### DIFF
--- a/db/dcrpg/queries.go
+++ b/db/dcrpg/queries.go
@@ -2983,33 +2983,40 @@ func retrieveTicketsPriceByHeight(ctx context.Context, db *sql.DB, interval int6
 func retrieveCoinSupply(ctx context.Context, db *sql.DB, timeArr []dbtypes.TimeDef,
 	sumArr, estimateArr []float64, params *chaincfg.Params) ([]dbtypes.TimeDef,
 	[]float64, []float64, error) {
-	var since, initialTime time.Time
-	var sum, estimate, preminedCoins float64
+	var sum, estimate float64
+	var since, premineTime time.Time
 
-	var reward = dcrutil.Amount(params.BaseSubsidy).ToCoin()
-	var targerBlockTime = float64(params.TargetTimePerBlock)
-	var inflationRate = float64(params.MulSubsidy) / float64(params.DivSubsidy)
-	var subsidyReduction = float64(params.SubsidyReductionInterval)
+	premineBlockInd := 1
+	reward := dcrutil.Amount(params.BaseSubsidy).ToCoin()
+	targetBlockTime := params.TargetTimePerBlock.Minutes()
+	inflationRate := float64(params.MulSubsidy) / float64(params.DivSubsidy)
+	subsidyReduction := float64(params.SubsidyReductionInterval)
 
-	if c := len(timeArr); c > 0 && len(sumArr) == c && len(estimateArr) == c {
+	// https://github.com/decred/dcrd/blob/5076a00512a521cea3c51b443b50970804dbe712/blockchain/subsidy_test.go#L52-L54
+	maxCoinSupply := 20999999.99800912
+
+	// timeArr and estimateArr may contain extra datasets than sumArr that allows
+	// the inflation axis to be extended to one month more than the actual supply
+	// if maximum supply has not yet been reached.
+	if c := len(sumArr); c > 0 && len(timeArr) >= c && len(estimateArr) >= c {
 		since = timeArr[c-1].T
 		sum = sumArr[c-1]
 		estimate = estimateArr[c-1]
+
+		// Delete extra data added before.
+		estimateArr = estimateArr[:c]
+		timeArr = timeArr[:c]
+
+		// Preset the premine Time.
+		if c > 1 {
+			premineTime = timeArr[premineBlockInd].T
+		}
 	} else {
-		// if the count of elems in all arrays doesn't match drops all entries.
+		// If the arrays do not have the same number of elements, drop all elements.
+		// Retrieve and accumulate all data since genesis.
 		sumArr = sumArr[:0]
 		timeArr = timeArr[:0]
 		estimateArr = estimateArr[:0]
-	}
-
-	// premine coins are usually found of the premines block (at height 1).
-	// Its usually the second block after the genesis block.
-	if len(sumArr) > 1 && len(timeArr) > 0 {
-		preminedCoins = sumArr[1]
-		initialTime = timeArr[1].T
-
-		// subtract the premine coins from the prev estimate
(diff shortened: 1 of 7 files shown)


buck54321:
It's funny because I was just looking over inflation stuff yesterday. I won't be able to review until tommorow (on a mini-vacation), but here's a couple of notes. 

I think your formula is mostly right, but I didn't see where you accounted for the lowered block reward from block 3 to block 4095 inclusive. There is no POS reward for those blocks, so you'll need to lower it by 30%, or `float64(params.StakeRewardProportion) / 10`. I think `4096` is `StakeValidationHeight`, but you'll want to verify that. 

It's unfortunate that we're stuck using a time-based estimation for this, but it's simply a result of the `vins` table having no block height column. I have a solution for that will be in a PR tomorrow-ish, so we'll have a conflict in `retrieveCoinSupply`, but we'll make it work.

I also saw that you're grabbing the block 1 value from the database, which is probably fine, but I wanted to note that block 1 value can be parsed from `(chaincfg.Params).BlockOneLedger`.

dmigwi:
The POS reward between block 1-4096 was supposed to be removed but I forgot about it. Thanks for reminding me.

Parsing `(chaincfg.Params).BlockOneLedger` wasn't a choice I was looking forward to consider because on mainnet there are over 3000 entries in that slice to loop through. IMO that is a lot of time that would have been spent parsing data quite often instead of just getting the same data from block 1.

db/dcrpg/queries.go
```diff
@@ -3001,19 +3023,37 @@ func retrieveCoinSupply(ctx context.Context, db *sql.DB, timeArr []dbtypes.TimeD
...
+		}
+
+		var timeDiff = float64(timestamp.Sub(initialTime))
+		var estimatedBlockHeight = int(timeDiff / targerBlockTime)
+		if estimatedBlockHeight <= len(timeArr)-1 {
+			estimatedBlockHeight = len(timeArr) - 1
+		}
+
+		subsidyInterval := math.Floor(float64(estimatedBlockHeight) / subsidyReduction)
+		if value > 0 {
```
chappjc: `value` from the query is the cumulative coin supply at given timestamp.  So, if `value == 0`, that means it is the genesis block since every other block will add to the supply.  If that is correct, I suggest commenting on this logic here.

Also if `value == 0` on genesis, and the `estimate` is not updated, then `preminedCoins` should still be 0. I'm just talking through the logic here because I feel like this loop can be tweaked to make the code more expressive of the intent.

db/dcrpg/queries.go
```diff
@@ -2981,18 +2981,40 @@ func retrieveTicketsPriceByHeight(ctx context.Context, db *sql.DB, interval int6
...
+	}
+
+	// premine coins are usually found of the premines block (at height 1).
+	// Its usually the second block after the genesis block.
+	if len(sumArr) > 1 && len(timeArr) > 0 {
+		preminedCoins = sumArr[1]
+		initialTime = timeArr[1].T
+
+		// subtract the premine coins from the prev estimate
+		estimate -= preminedCoins
```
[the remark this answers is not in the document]
dmigwi: The total coins mined since the premine block are calculated independently without the premine coins for each height/estimated height.

db/dcrpg/queries.go
```diff
@@ -2981,18 +2981,40 @@ func retrieveTicketsPriceByHeight(ctx context.Context, db *sql.DB, interval int6
...
+	var targerBlockTime = float64(params.TargetTimePerBlock)
+	var inflationRate = float64(params.MulSubsidy) / float64(params.DivSubsidy)
+	var subsidyReduction = float64(params.SubsidyReductionInterval)
+
+	if c := len(timeArr); c > 0 && len(sumArr) == c && len(estimateArr) == c {
 		since = timeArr[c-1].T
 		sum = sumArr[c-1]
+		estimate = estimateArr[c-1]
+	} else {
+		// if the count of elems in all arrays doesn't match drops all entries.
```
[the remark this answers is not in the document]
dmigwi: I am sure if a warning log is necessary here since we are logging when the .gob file could not be successfully read.

dmigwi:
Changes included:

- POS reward reduction from the block reward for all the blocks before the stake validation height.
- The inflation axis has been extended by one month.

db/dcrpg/queries.go
```diff
@@ -2983,76 +2983,123 @@ func retrieveTicketsPriceByHeight(ctx context.Context, db *sql.DB, interval int6
...
+	var since, premineTime time.Time
+
+	premineBlockInd := 1
+	reward := dcrutil.Amount(params.BaseSubsidy).ToCoin()
+	targetBlockTime := params.TargetTimePerBlock.Minutes()
+	inflationRate := float64(params.MulSubsidy) / float64(params.DivSubsidy)
+	subsidyReduction := float64(params.SubsidyReductionInterval)
+
+	// https://github.com/decred/dcrd/blob/5076a00512a521cea3c51b443b50970804dbe712/blockchain/subsidy_test.go#L52-L54
+	maxCoinSupply := 20999999.99800912
```
chappjc: Only true for mainnet.  But a quick call to `txhelpers.UltimateSubsidy(params)` will give you the number too (in atoms).

run_tests_with_chartests.sh
```diff
@@ -62,6 +62,9 @@ testrepo () {
     env GORACE='halt_on_error=1' go test -v -race -tags chartests $(cd $MODPATH && go list -m)
   done
 
+  # check linters
+  ./lint.sh
```
chappjc: good catch.  I forgot this script.  Although you'll want to remove the call to `golangci-lint` a few lines down.

PR #1249, opened 2019-04-15
[4.1] Correct the usage of the stake version

chappjc:


chappjc:
Working well on beta site (4.1).

PR #1265, opened 2019-04-16
dcrpg: update to v1.1.0 of github.com/lib/pg

chappjc:
This updates the dcrpg module to use the latest tagged version (v1.1.0) of github.com/lib/pq.

Changes to lib/pq:
-  > pq: support returning multiple result sets from a single query. Tested with Cockroach v2.1.6 and PostgreSQL 11.  https://github.com/lib/pq/commit/ddb99717ca7f59f1678a241f0ab22e253868bd79
- >  add SCRAM-SHA-256 authentication to the pq library, by adding a slightly modified version of the mgo scram library and implementing it's functionality in the auth section of conn.go.  https://github.com/lib/pq/commit/51e2106eed1cea199c802d2a49e91e2491b02056
- > pq: pipe the context through to the dialer, make use of the Connector
- > implement driver.Pinger interface


chappjc:
A follow-up for the v5 dcrdata module is required after this merge.

chappjc:
PG backend is working as before.  We'll have to keep an eye on Postgresql after merging.

PR #1273, opened 2019-04-17
[4.0] Fix proposal update

dmigwi:
- [x]  Bug fix that corrects the wrong marshaling of the single proposal payload fetched.
It also adds logs to warn when empty or invalid payloads are detected.

- [x] If update of previously added proposal tokens fails for some, the error returned is logged and update proceeds for those that new data is available.
It will help in updating past such an error if new data exists.
```
2019-04-15 19:17:13.120 [ERR] DATD: updating proposals db failed: RetrieveProposalByToken failed: retrieving 5d9cfb07aefb338ba1b74f97de16ee651beabc851c7f2b5f790bd88aea23b3cb proposal vote status failed: request (https://proposals.decred.org/api/v1/proposals/5d9cfb07aefb338ba1b74f97de16ee651beabc851c7f2b5f790bd88aea23b3cb/votestatus) failed with status code: 400 Bad Request 
```
- [x] Added this to proposal details pages that do not show proposal votes charts when there data is missing.
Screenshot from 2019-04-16 09-57-16

### DIFF
--- a/gov/politeia/piclient/piclient.go
+++ b/gov/politeia/piclient/piclient.go
@@ -5,7 +5,6 @@
 package piclient
 
 import (
-	"bytes"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
@@ -14,24 +13,6 @@ import (
 
 	pitypes "github.com/decred/dcrdata/gov/politeia/types"
 	piapi "github.com/decred/politeia/politeiawww/api/v1"
-)
-
-var (
-	// nullProposalData defines a case scenario where the proposal could have had
-	// a null pointer attached to it.
-	nullProposalData = []byte(`{"proposal": null}`)
-
-	// nullVotesData defines a case scenario where the votes could have had a null
-	// pointer attached to it.
-	nullVotesData = []byte(`{"votes": null}`)
-
-	// nullProposalsData defines a case scenario where the proposals array could
-	// have had a null pointer attached to it.
-	nullProposalsData = []byte(`{"proposals": null}`)
-
-	// nullVotesStatusData defines a case scenario where the votesstatus array
-	// could have had a null pointer attached to it.
-	nullVotesStatusData = []byte(`{"votesstatus": null}`)
 )
 
 // HandleGetRequests accepts a http client and API URL path as arguments. If the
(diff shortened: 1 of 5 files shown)


chappjc:
Thanks for doing 4.0 also.  Please remove the package lock change though

gov/politeia/piclient/piclient.go
```diff
@@ -95,15 +95,15 @@ func RetrieveAllProposals(client *http.Client, APIRootPath, URLParams string) (
...
 	data, err := HandleGetRequests(client, proposalRoute)
 	if err != nil {
 		return nil, fmt.Errorf("retrieving %s proposal details failed: %v", token, err)
 	}
 
-	var proposal pitypes.ProposalInfo
+	var proposal pitypes.Proposal
 	err = json.Unmarshal(data, &proposal)
 	if err != nil {
 		return nil, err
```
chappjc: Also  need to check if `proposal.Data` is `nil`. `json.Unmarshal` will not error, but `proposal.Data` will be `nil` if Pi returns:

```json
{
    "proposal": null
}
```

https://play.golang.org/p/0nLHmiJaWny

chappjc: Similarly should check any other pointer types in `ProposalInfo` that could cause a panic later.

chappjc: Since it unmarshals without error, I just meant to check `if proposal.Data == nil { return nil, [error] }`

gov/politeia/piclient/piclient.go
```diff
@@ -15,6 +16,24 @@ import (
 	piapi "github.com/decred/politeia/politeiawww/api/v1"
 )
 
+var (
+	// nullProposalData defines a case scenario where the proposal could have had
+	// a null pointer attached to it.
+	nullProposalData = []byte(`{"proposal": null}`)
```
chappjc: A little fragile if pi starts returning indented or otherwise formatted responses.  I think the safe thing to do is check the pointer on the unmarshalled response.

dmigwi: /o\, I totally missed the point earlier. Let me add the fix.

gov/politeia/piclient/piclient.go
```diff
@@ -134,29 +103,24 @@ func RetrieveProposalByToken(client *http.Client, APIRootPath, token string) (*p
...
-
 	var proposal pitypes.Proposal
 	err = json.Unmarshal(data, &proposal)
 	if err != nil {
 		return nil, err
 	}
 
+	// Check if null proposal data was returned as part of the proposal details.
+	if proposal.Data == nil {
+		return nil, fmt.Errorf("invalid proposal with null data found for %s", token)
```
dmigwi: This feels like the only place a null pointer would wreck havoc if were to be returned by the API. 
Please validate this before I can proceed to made changes to the rest of PRs.

chappjc: Yeah, looks like it to me.  Thanks.

PR #1274, opened 2019-04-17
cache: implement encoded chart caching and default day-binned views

buck54321:
Moves chart management to db/cache. A pre-encoded cache system reduces API times by ~95%, once the cache is primed. A single `ChartData` pointer is shared between dcrsqlite and dcrpg, with explorer handling synchronization. Aligns all block-binned data sets on block height. Consequently, partially validates DB data, throwing errors if a corrupted database is detected.

Datasets are binned on blocks or difficulty window, but a secondary day-binned dataset is generated from the block set as well. The day view is set as the default view on /charts. "Rolling average" control
on /charts is replaced with a simpler day/block bin size selector. The x-axis for difficulty-window-based data is now time instead of height.

@dmigwi : This PR ran head-first into #1247. Sorry about that. One thing to note about this PR is that the query for coin supply is modified to return data based on height. This allows us to skip querying and storing the dataset on the server-side at all, since the max theoretical supply is a deterministic function of block height. It's now like a 5 or 10 line javascript function instead of a couple thousand data points. It's also cool if #1247 is merged first while this PR is still open and I can handle the conflicts.


db/dcrpg/pgblockchain.go
```diff
@@ -2833,93 +2782,50 @@ func (pgb *ChainDB) getChartData(data map[string]*dbtypes.ChartsData,
...
-	data[dbtypes.DurationBTW] = durPerB
-	data[dbtypes.HashRate] = chainData[1]
-	data[dbtypes.POWDifficulty] = diff
-	data[dbtypes.TicketByWindows] = ticketsByWin
-	data[dbtypes.TicketPrice] = tickets
-	data[dbtypes.TicketsByBlocks] = tByAllBlocks
-	data[dbtypes.TicketSpendT] = ticketSpendT
-	data[dbtypes.TxPerBlock] = txPerB
-	data[dbtypes.TxPerDay] = txRate
+	// tByAllBlocks := pgb.getChartData(data, dbtypes.TicketsByBlocks)
```
buck54321: There are a couple of data sets that I haven't re-implemented. The solo/pooled tickets and unspent/revoked tickets by purchase block. I question their value, but I am happy to grab that data if requested. The solo/pooled accounting in particular has come up as innacurate due to ticket splitting.

dmigwi:
Here is my 2 cents; 
Moving the charts cache managements to db/cache is quite welcome since it provides that much performance improvements that you have mentioned.

>One thing to note about this PR is that the query for coin supply is modified to return data based on height.

There was a task on the way that I was supported to have worked by now where all charts wherever possible would provide the support plotting for x-axis as either height or time. The person viewing the charts was supposed to have a button to toggle between height and time for the x-axis. What you've done is a step in the right direction but I just hope you will not do away with the previous implementation.

>unspent/revoked tickets by purchase block

I am not aware of any requests or need to do away with this implementation, thus I will consult with @raedah and @chappjc first before sharing my thoughts on the same.

> The solo/pooled tickets 

I was also supposed to add a change to the description listed there but keep the implementation. In the next couple of days I was supposed to drop all references to solo and pooled tickets and instead just replace it with a label that says `# of outputs`. On this one too, I am not aware of plans to do away with the implementation.

dmigwi:
In fact if your hand aren't full now, you could add the support for charts x-axis data to be toggled between height and time wherever possible.

buck54321:
Thanks @dmigwi . I have a handful of changes I wanted to make to /charts too, but I'd like to do them separately. Let's coordinate a bit. 

A height axis option will be easy now. Since the data is validated, the height is simply the index of the data in the block-binned data set. No need to send a separate array of heights. 

To be clear, I'm not doubting the value of the aforementioned data itself, just it's value as a chart. The solo/pooled chart is pretty much just noise, with nothing of interest. Is there a particular metric of interest 
in there and is there a better way to display it?

buck54321:
I just wanted to touch again on the charts I'm questioning. 

The solo/pooled chart looks like 

image

The one interesting feature is the change when the algorithm was fixed, but that particular feature is better demonstrated in the ticket price chart already. Other than that, it's just noise, and unless something drastic changes in people's staking habits, I wouldn't expect anything to change. 

The missed/revoked chart looks like 

image

Which is basically just a probabilistic curve controlled by network parameters. Here again, unless people's staking habits change, particularly their revoking of missed tickets, I don't expect this chart to ever look qualitatively different than it does now. 

I'm hoping somebody will tell me exactly what metric they are hoping to visualize here so that I can find a better way to display it.

buck54321:
Reduced duplication has reduced the gob file size from ~65 MB to ~12 MB.

chappjc:
This PR has some awesome work.  Nice to reduce the gob size too.

Regarding the pool/solo ticket guesses, and their visual representation in any form, I'm still in favor of removing this.  (Even if it were more than just noise, and it does not seem to be.)  This includes the circle charts at the bottom of the ticket pool page.

The blue part data in the Ticket Spend Types chart is also represented on the histogram on /ticketpool, so that is redundant anyway.  The green data (revoked over time) could only be of interest if a whale went offline for a while, came back online and revoked a large number of tickets.  But in normal situations, the chart is mostly uninformative.

go.sum
```diff
@@ -131,6 +131,7 @@ github.com/decred/dcrd/wire v1.1.0 h1:G+3CugtxNbToUN8RKWqm74yLfzJJ2BKMOr2RgWc4Ty
 github.com/decred/dcrd/wire v1.1.0/go.mod h1:/JKOsLInOJu6InN+/zH5AyCq3YDIOW/EqcffvU8fJHM=
 github.com/decred/dcrd/wire v1.2.0 h1:HqJVB7vcklIguzFWgRXw/WYCQ9cD3bUC5TKj53i1Hng=
 github.com/decred/dcrd/wire v1.2.0/go.mod h1:/JKOsLInOJu6InN+/zH5AyCq3YDIOW/EqcffvU8fJHM=
+github.com/decred/dcrdata v2.1.3+incompatible h1:DLzrqgOOgHwrUSbuOVDAUzfQePhCGl9UsjEgZI1D0Kg=
```
chappjc: I'm guessing these v2.1.3+incompatible lines got in here by accident, but best to remove them.

buck54321:
So my solution to the chart question right now is this. 

1. Deprecated the solo/pooled chart. Leaving the code for now, but updated the docs to explain the orphan. 

2. For unspent/revoked, I've deprecated the chart, but created a new endpoint at `/api/stake/powerless` which returns missed and expired tickets sorted by revocation status. The data is pulled directly with no caching. Load time for that endpoint is ~200 ms on my system. 

I'm using a new term, ***powerless***, to describe either missed or expired. `powerless = missed | expired`.

I think its ready for review. Removing WIP. Live tests should be run with https://github.com/dcrlabs/bug-free-happiness/pull/3/commits/6c35a59c7cfaadf07ab0f58901dae5fc8a514678 until that PR is updated and accepted.

db/cache/go.mod
```diff
@@ -3,8 +3,12 @@ module github.com/decred/dcrdata/db/cache
 go 1.12
 
 require (
+	github.com/decred/dcrd/chaincfg v1.4.0
 	github.com/decred/dcrd/chaincfg/chainhash v1.0.1
+	github.com/decred/dcrd/wire v1.2.0
 	github.com/decred/dcrdata/api/types v1.0.7-0.20190416202529-23d1eb95ca1b
+	github.com/decred/dcrdata/blockdata v1.0.1
 	github.com/decred/dcrdata/db/dbtypes v1.0.2-0.20190416202529-23d1eb95ca1b
+	github.com/decred/dcrdata/txhelpers v1.0.2-0.20190416155607-25afe5d5e790
```
chappjc: Would you update the module dependency graph to reflect the added blockdata and txhelpers dependency?

db/cache/charts.go
```diff
@@ -0,0 +1,1015 @@
...
+	err := charts.writeCacheFile(dumpPath)
+	if err != nil {
+		log.Errorf("ChartData.writeCacheFile failed: %v", err)
+	} else {
+		log.Debug("Dumping the charts cache data was successful")
+	}
+}
+
+// Store triggers an Update. Satisfies blockdata.BlockDataSaver.
+func (charts *ChartData) Store(_ *blockdata.BlockData, _ *wire.MsgBlock) error {
```
chappjc: The `blockdata` dependency is somewhat unfortunate, esp. since it does not actually need the data.  But it seems that other `BlockDataSavers` are already running `charts.Update`, so can this Store method be removed?

buck54321: I'm actually hoping to go the other way, removing charts from the explorer package altogether, since explorer doesn't use them. 

I also hate the unneeded dependency. I've got an idea for a solution.

db/cache/go.mod
buck54321: Will do.

db/cache/charts.go
```diff
@@ -0,0 +1,1015 @@
...
+}
+
+// Translate the uints to a slice of the differences between each. The provided
+// data is assumed to be monotinically increasing. The first element is always
+// 0 to keep the data length unchanged.
+func btw(data ChartUints) ChartUints {
+	d := make(ChartUints, 0, len(data))
+	dataLen := len(data)
+	if dataLen == 0 {
+		return d
```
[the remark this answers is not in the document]
buck54321: Right now, the only paths to this function will never have a nil argument. If something got messedup and that did  happen, the result would be a javascipt error on `/charts`, which would be the least of our concerns at that point.

explorer/explorer.go
```diff
@@ -318,6 +319,13 @@ func New(cfg *ExplorerConfig) *explorerUI {
 	exp.NetName = netName(exp.ChainParams)
 	exp.MeanVotingBlocks = txhelpers.CalcMeanVotingBlocks(params)
 
+	var premine int64
+	ledger := params.BlockOneLedger
+	for i := range ledger {
+		premine += ledger[i].Amount
```
chappjc: While this is spot on, there's actually func `(p *Params) BlockOneSubsidy() int64`, which we might as well use:

https://github.com/decred/dcrd/blob/e69a789183f39c2991a0da78c0b056887abe1d36/chaincfg/params.go#L664-L677

public/js/controllers/charts_controller.js
```diff
@@ -25,30 +29,49 @@ function intComma (amount) {
   return amount.toLocaleString(undefined, { maximumFractionDigits: 0 })
 }
 
+function blockReward (height) {
+  if (height >= stakeValHeight) return baseSubsidy * Math.pow(100 / 101.0, Math.floor(height / subsidyInterval))
```
chappjc: I'd feel better if there was a comment referring to MulSubsidy and DivSubsidy.  Being that same on all networks presently is convenient, but it's certainly possible that will change.

explorer/explorer.go
buck54321: Oh nice.

public/js/controllers/charts_controller.js
buck54321: I'll just use the actual net params then.

Issue #1221, opened 2019-03-29
preload high priority web resources


chappjc:
`<link rel=preload>` can potentially improve improve page load time if applied to certain resources that should be downloaded as soon as possible.

Notably:

/fonts/icomoon.ttf
/fonts/source-sans-pro-v9-latin-regular.woff
/fonts/inconsolata-v15-latin-regular.woff

njirap:
hallo.

This issue was not auto-closed on PR merge.

It still shows up as open and I bumped into it while searching for my first issue to work on. Maybe stale issues should be closed. Thanks. I will proceed with my search

PR #1289, opened 2019-04-19
views: preload webfonts and svg

chappjc:
This enables browsers that support `rel="preload"` to preload the fonts and svg symbols instead of waiting to process style.css.

Working: Chrome, Epiphany (probably Safari too), Firefox ([network.preload setting](https://user-images.githubusercontent.com/9373513/56443923-3db29c00-62bc-11e9-9980-e0c63a12eab1.png))
https://caniuse.com/#search=preload

**Without preload (first 200ms of page load):**

image

The webfonts, logo.svg, and connecting.svg are delayed for about 160 ms while the stylesheet is processed.  Also, Chrome reports the initiator is Turbolinks.

Also note that connected.svg is not loaded in this time period because it is loaded only after the websocket connects.

**With preload (first 200ms of page load):**

image

The webfonts, the logo.svg, and connecting.svg are loaded immediately.  Also, Chrome reports the initiator is the index (html head).

Both connected.svg and disconnected.svg are loaded in this time period rather than after the websocket connects or disconnects. Note that if the server is down or not reachable, the browser will not be able to retrieve disconnected.svg, so it makes sense to do it while the connection is up.

Resolving https://github.com/decred/dcrdata/issues/1221.

### DIFF
--- a/views/extras.tmpl
+++ b/views/extras.tmpl
@@ -39,7 +39,15 @@
 
     <meta name="turbolinks-cache-control" content="no-cache">
     <title>{{.}}</title>
+    <link rel="preload" href="fonts/icomoon.ttf?g003m6" as="font" type="font/ttf" crossorigin="anonymous" />
+    <link rel="preload" href="fonts/inconsolata-v15-latin-regular.woff" as="font" type="font/woff" crossorigin="anonymous" />
+    <link rel="preload" href="fonts/source-sans-pro-v9-latin-regular.woff" as="font" type="font/woff" crossorigin="anonymous" />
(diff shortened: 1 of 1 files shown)


chappjc:
This also eliminates a flash of the screen when reloading.  Try on alpha.dcrdata.org

chappjc:
Before

image

After

image

There is no longer a flash/flicker of the content on the page during page load.

Even with an extra 30ms network latency to alpha vs. explorer:
First meaningful paint (FMP): 647 ms -> 396 ms
Onload: 669 ms -> 459 ms

chappjc:
It sure was an easy win.  Certain browsers like Firefox seem to be better at loading resources early even without this, but Chrome really seems take a cue from preload.

PR #1290, opened 2019-04-22
[4.1] Compare all proposal data to check for updates

dmigwi:
Instead of comparing `votestatus` and `timestamp` only to check for updates, compare more `proposalInfo` fields to capture lots more changes when they happen.

```Golang
if proposal.Data.VoteStatus == statuses[0] || proposal.Data.Timestamp == val.Timestamp {
```
In the conditions above, the `Timestamp` was expected to change every time a minor change was effected on the proposal but was only found to change when the proposal author changed the proposal version.

Also Lots of other properties like `Censorship`, `Name`, `State`, `NumComments`, `StatusChangeMsg`, `CensoredDate`, `AbandonedDate`, `PublishedDate`, `Token`, `VoteStatus`, `TotalVotes` and `VoteResults` have a high likelihood of changing as the proposals moves through the various life cycle stages but were not initially considered.

### DIFF
--- a/gov/politeia/proposals.go
+++ b/gov/politeia/proposals.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"reflect"
 	"sync"
 	"time"
 
@@ -103,6 +104,11 @@ func (db *ProposalDB) saveProposals(URLParams string) (int, error) {
 		if data == nil || data.Data == nil {
 			// Should help detect when API changes are effected on Politeia's end.
 			log.Warnf("invalid or empty data entries were returned")
+			break
+		}
+
+		if len(data.Data) == 0 {
+			// No updates found.
 			break
 		}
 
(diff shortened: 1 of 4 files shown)


gov/politeia/proposals.go
```diff
@@ -279,21 +285,21 @@ func (db *ProposalDB) updateInProgressProposals() (int, error) {
...
 		}
 
-		// 2. The new proposals status is NotAuthorized(has not changed).
-		// 3. The last update timestamp has not changed.
-		if proposal.Data.VoteStatus == statuses[0] || proposal.Data.Timestamp == val.Timestamp {
+		// Add the generated field.
+		proposal.Data.ID = val.ID
+
+		// 2. The new proposal data has not changed.
+		if reflect.DeepEqual(*proposal.Data, *val) {
```
chappjc: I'd like to keep `reflect.DeepEqual` in tests only.  Please make an `IsEqual` method instead.

package-lock.json
```diff
@@ -6107,9 +6107,9 @@
       "dev": true
     },
     "js-yaml": {
-      "version": "3.13.0",
-      "resolved": "https://registry.npmjs.org/js-yaml/-/js-yaml-3.13.0.tgz",
-      "integrity": "sha512-pZZoSxcCYco+DIKBTimr67J6Hy+EYGZDY/HCWC+iAEA9h1ByhMXAIVUXMcMFpOCxQ/xjXmPI2MkDL5HRm5eFrQ==",
+      "version": "3.13.1",
+      "resolved": "https://registry.npmjs.org/js-yaml/-/js-yaml-3.13.1.tgz",
+      "integrity": "sha512-YfbcO7jXDdyj0DGxYVSlSeQNHbD7XPWvrVWeVUujrQEoZzWJIRrCPoyk6kL6IAjAG2IolMK4T0hNUe0HOUs5Jw==",
```
chappjc: I'll cherry pick this diff for 4.1 in a different commit.

PR #1293, opened 2019-04-24
explorer: enable dark mode without js

buck54321:


explorer/explorermiddleware.go
```diff
@@ -200,3 +206,34 @@ func ProposalPathCtx(next http.Handler) http.Handler {
...
+				log.Errorf("Cookie dcrdataDarkBG retrieval error: %v", err)
+			} else {
+				if err == http.ErrNoCookie {
+					cookie = &http.Cookie{
+						Name:   darkModeCoookie,
+						Value:  "1",
+						MaxAge: 0,
+					}
+				} else {
+					cookie.Value = "0"
```
chappjc: At first I was thinking this should be a toggle, but I see that no cookie is set in `commonData` if the dark mode cookie value is not "1".  Seems good, but I'm just curious if the idea was to limit cookie use to only when dark mode is enabled, because it seems like the cookie could store both on or off values.

buck54321: Either way is fine. My thoughts are that we will always have to check for the existence of the cookie first anyway, with a missing cookie yielding some default behavior. Since we have to check anyway, it might save a step to delete the cookie. Do you prefer the other way?

chappjc: Nope, I like your way.  Was just wondering if I was following the intended logic.

PR #1295, opened 2019-04-26
Update piparser dependency version

dmigwi:
https://github.com/dmigwi/go-piparser/commit/8412e0f44f5560605f0af8993b0fa0ef78e79a87 `time.NewTimer(` has been replaced with `time.NewTicker(`


db/dcrpg/go.mod
```diff
@@ -23,7 +23,6 @@ require (
 	github.com/decred/dcrdata/testutil/dbconfig v1.0.0
 	github.com/decred/dcrdata/txhelpers v1.0.2-0.20190416204615-70a58657e02f
 	github.com/decred/slog v1.0.0
-	github.com/dmigwi/go-piparser/proposals v0.0.0-20190415205832-3992790812af
```
chappjc: I think these should remain in the dcrpg go.mod and go.sum

PR #1296, opened 2019-04-29
exchanges: add bittrex websocket support

buck54321:
Bittrex uses a SignalR wrapper for its websocket, so a library is used for the SignalR connection. Much of the websocket implementation from Poloniex exchange is re-used, including some types and methods being elevated to a common type or CommonExchange method.

Bittrex docs claim that the "full order book" is sent on the websocket, but it looks like they limit it to 500 each of buy or sell orders. Nevertheless, the websocket data range is still a vast improvement over the REST API.


raedah:
Is the 500 limit just for the initial fetch? Will subsequent streaming order updates complete it gradually over time?

buck54321:
I think that it can grow beyond 500 with updates, but the outliers are probably long-term limit orders that don't get updated often, so it would take a long time and you'd never really know if there are some big ones hiding. 

I can run some longer tests today and see what happens to the counts.

buck54321:
So this was a little more complicated than I thought. Not only do they only send 500 max, they actively maintain it at 500. Right now there are 500 orders on the ask side. If they send a new order through, it will always come paired with a remove order, and the remove order will remove it from the high end of the current 500. Likewise, when they send a remove order, it comes paired with an add order at the high end. In short, they are keeping the length at 500 at all times.

raedah:
Thats strange of bittrex, but it explains how the current chart looks. Thats all they show for the depth on their own site, so I guess it was just a design decision they made.

buck54321:
I've got a couple fixes to push here.

exchanges/exchanges.go
```diff
@@ -232,6 +238,23 @@ func (sticks Candlesticks) needsUpdate(bin candlestickKey) bool {
 	return time.Now().After(lastStick.Start.Add(bin.duration() * 2))
 }
 
+// UTILITY FUNCTIONS
+
+// Quickly print the thing to a JSON-encoded string.
+func jsonify(thing interface{}) string {
+	s, _ := json.MarshalIndent(thing, "", "    ")
+	return fmt.Sprint(string(s))
```
chappjc: What's the Sprintf for?

buck54321: I heard you like strings with your strings, so I made your string into a string. :-P

exchanges/exchanges.go
```diff
@@ -1154,6 +1318,266 @@ type BittrexCandlestickResponse struct {
...
+		if err != nil {
+			log.Errorf("bittrex orderbook re-encode error: %v", err)
+			return
+		}
+		err = json.Unmarshal(bookBytes, book)
+		if err != nil {
+			bittrex.setWsFail(fmt.Errorf("Bittrex order book Unmarshal error: %v", err))
+			return
+		}
+		if book == nil {
```
chappjc: Travis pointed out that this can never be `nil`.  Perhaps checking the nonce or the order slice lengths would indicate if Bittrex returned crap data?

PR #1298, opened 2019-04-29
Insight root page and api redirect

chappjc:
This makes a basic HTML page for the `/insight` to provide information about the Insight API on dcrdata.

image

The above shows "https://alpha.dcrdata.org/insight".  To support reverse proxy configurations that map requests on the root path to /insight on the backend, a middleware (`OriginalRequestURI`) is added to embed the original request URI.  This middleware allows the new Insight page handler to generate links relative to *the original client request URI*, not the host root or the request URI as known by the backend.  Again, this will happen with a reverse proxy that modifies the request URI.  For example, here is http://insight.alpha.dcrdata.org:

image

Note that the "API sub-path" now links to /api/status instead of /insight/api/status since the proxy prefixes all requests with /insight behind the scenes.

Also, the `/insight/api` endpoint now redirects to `/insight/api/status`. This endpoint's handler also uses the `OriginalRequestURI` since it appends to the request URI and it must be the original request URI, not the URI passed to the backend by the reverse proxy.

### DIFF
--- a/api/insight/apirouter.go
+++ b/api/insight/apirouter.go
@@ -1,13 +1,13 @@
 // Copyright (c) 2018-2019, The Decred developers
 // Copyright (c) 2017, The dcrdata developers
 // See LICENSE for details.
-//
-// This is tested against Insight API Implementation v5.3.04beta
 
-// Package insight handles the insight api
+// Package insight implements the Insight API.
 package insight
 
 import (
+	"net/http"
+
 	m "github.com/decred/dcrdata/middleware"
 	"github.com/didip/tollbooth"
 	"github.com/didip/tollbooth_chi"
@@ -49,6 +49,10 @@ func NewInsightApiRouter(app *InsightApi, useRealIP, compression bool) ApiMux {
 	if compression {
 		mux.Use(middleware.NewCompressor(3).Handler())
 	}
+
+	mux.With(m.OriginalRequestURI).Get("/", func(w http.ResponseWriter, r *http.Request) {
+		http.Redirect(w, r, r.URL.Path+"/status", http.StatusSeeOther)
+	})
 
 	// Block endpoints
 	mux.With(BlockDateLimitQueryCtx).Get("/blocks", app.getBlockSummaryByTime)
--- a/middleware/apimiddleware.go
+++ b/middleware/apimiddleware.go
@@ -13,6 +13,7 @@ import (
 	"io"
 	"io/ioutil"
 	"net/http"
+	"net/url"
 	"strconv"
 	"strings"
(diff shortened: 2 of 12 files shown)


api/insight/apirouter.go
```diff
@@ -50,6 +49,10 @@ func NewInsightApiRouter(app *InsightApi, useRealIP, compression bool) ApiMux {
 		mux.Use(middleware.NewCompressor(3).Handler())
 	}
 
+	mux.Get("/", func(w http.ResponseWriter, r *http.Request) {
+		http.Redirect(w, r, r.URL.Path+"/status", http.StatusSeeOther)
```
chappjc: Not redirecting to `"/insight/api/status"` because a server may have `"/insight"` mounted at the root.

views/insight_fallback.tmpl
```diff
@@ -0,0 +1,23 @@
...
+<!DOCTYPE html>
+<html lang="en">
+
+{{template "html-head" "Insight Fallback Page"}}
+    {{template "navbar" . }}
+    <div class="container main">
+        <div>
+            <h2 style="text-align: center; margin-top: 0px">Looking for the Insight API?</h2>
+            <p style="text-align: center; margin-bottom: 5px">
+                The Insight API is available under the API sub-path. (e.g. <a href="{{.RequestURI}}/api/status">{{.RequestURI}}/api/status</a>).
```
chappjc: Using RequestURI for the same reason.  TODO: see how this interacts with a reverse proxy, which makes it's own requests, and attempt to get the request path as it hits the proxy from the client if necessar.

middleware/apimiddleware.go
```diff
@@ -177,6 +178,33 @@ func GetRawHexTx(r *http.Request) (string, error) {
...
 }
 
+// OriginalRequestURI checks the X-Original-Request-URI HTTP request header for
+// a valid URI, and patches the request URL's Path and RawPath. This may be
+// useful in the event that a reverse proxy maps requests on path A to path B,
+// but the request handler requires the original path A (e.g. generating links
+// relative to path A).
+func OriginalRequestURI(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		origRequestURI := r.Header.Get("X-Original-Request-URI")
```
chappjc: Note that without `X-Original-Request-URI`, the page is still served, but if the downstream handlers use `r.URL.Path` (or `r.URL.RequestURI()`), the path may be different from the client's original request URI if a proxy is changing the request headers on the way to the dcrdata backend.

PR #1300, opened 2019-04-30
exchanges: add an aggregated orderbook

buck54321:
Combines order book data into a single aggregated orderbook available at (ExchangeBot).QuickDepths, with special token 'aggregated'. Data is available through the API at `/api/chart/market/aggregated/depth`.

A stacked depth chart is added to /charts. The depth chart gap width is also now displayed.

Screenshot is without #1296, so will get more interesting when the larger range of Bittrex data is available.

image


raedah:
Aggregate (stacking) can be a bit deceiving. Polo looks thinner and steeper here, but its just because its stacked on top. Would like a way to toggle on and off exchanges and choose to aggregate or overlay.

chappjc:
Agreed, but perhaps there is a simpler way to provide a visual cue to make it more clear.  I'm thinking the only solid line is the aggregate.  The fact that the top (aggregate) line is the Binance color immediately suggests that is the depth chart for Binance.  But if the top line were black (or another color), and the other lines were not shown or shown in some other color or format, it might suggest more clearly that the depth charts are summed/stacked.

buck54321:
How about both?

image

Note the new stacking selector. I dropped the fill colors in the unstacked case, because it didn't look very good, IMO.

raedah:
These look good. Only other thought is if it would be worth displaying the total aggregated black line on the non stacking view as well, but not sure if that is a good idea. May be outside of the scale.

raedah:
A zoom selector based on percentage would be useful. Display 10% 20% 30% range outside of current best bid/ask.

buck54321:
Got a zoom selector. Default is 20%, but your zoom will not be changed when switching between exchanges.

image

PR #1303, opened 2019-05-02
ticketpool: Edit the solo and pooled description

dmigwi:
- Remove the solo/pooled description from the ticketpool page.
Screenshot from 2019-05-02 19-53-06

- Replace `All Time` button with `Blocks`
Screenshot from 2019-05-02 19-53-00

### DIFF
--- a/db/dcrpg/internal/txstmts.go
+++ b/db/dcrpg/internal/txstmts.go
@@ -190,13 +190,13 @@ const (
 		) b
 		WHERE block_hash = b.hash;`
 
-	SelectTicketsByType = `SELECT block_time, num_vout
+	SelectTicketsByType = `SELECT DISTINCT num_vout, COUNT(*)
 		FROM transactions
 		JOIN tickets
 		ON transactions.id=purchase_tx_db_id
 		WHERE pool_status=0
 		AND tickets.is_mainchain = TRUE
-		ORDER BY block_time;`
+		GROUP BY num_vout;`
 
 	SelectTxnByDbID = `SELECT block_hash, block_height, tx_hash FROM transactions WHERE id = $1;`
 
--- a/db/dcrpg/internal/txstmts.go
+++ b/db/dcrpg/internal/txstmts.go
@@ -190,12 +190,13 @@ const (
 		) b
 		WHERE block_hash = b.hash;`
 
-	SelectTicketsByType = `SELECT
-		width_bucket(num_vout, array[3, 5, 6]) as ticket_bucket,
-		count(*)
-		FROM transactions JOIN tickets
-		ON transactions.id=purchase_tx_db_id WHERE pool_status=0
-		AND tickets.is_mainchain = TRUE GROUP BY ticket_bucket;`
+	SelectTicketsByType = `SELECT block_time, num_vout
+		FROM transactions
+		JOIN tickets
+		ON transactions.id=purchase_tx_db_id
+		WHERE pool_status=0
+		AND tickets.is_mainchain = TRUE
+		ORDER BY block_time;`
 
 	SelectTxnByDbID = `SELECT block_hash, block_height, tx_hash FROM transactions WHERE id = $1;`
 
--- a/public/js/controllers/ticketpool_controller.js
+++ b/public/js/controllers/ticketpool_controller.js
@@ -3,6 +3,7 @@ import ws from '../services/messagesocket_service'
 import { barChartPlotter } from '../helpers/chart_helper'
 import { getDefault } from '../helpers/module_helper'
 import axios from 'axios'
+import dompurify from 'dompurify'
 
 let Dygraph // lazy loaded on connect
 
@@ -14,6 +15,7 @@ function legendFormatter (data) {
     var labeledData = ' <span style="color: ' + series.color + ';">' + series.labelHTML + ': ' + series.yHTML
     html += '<br>' + series.dashHTML + labeledData + '</span>'
   })
+  dompurify.sanitize(html)
   return html
 }
 
@@ -74,8 +76,17 @@ function priceGraphData (items, memP) {
   return p
 }
 
-function outputsGraphData (items) {
-  return items.time.map((n, i) => { return [new Date(n), items.outputs[i]] })
+function populateOutputs (data) {
+  var totalCount = data.count.reduce((a, n) => { return a + n }, 0)
+  var tableData = '<thead><td># of Outputs</td><td>Count</td><td>% Occurence</td></thead>'
+  data.outputs.map((n, i) => {
+    var count = data.count[i]
+    tableData += '<tr><td class="pr-2 lh1rem vam nowrap xs-w117 font-weight-bold">' + n + '</td>' +
+    '<td><span class="hash lh1rem"> ' + count + ' </span></td>' +
+    '<td><span class="hash lh1rem"> ' + ((count * 100) / totalCount).toFixed(4) + '% </span></td></tr>'
+  })
+  dompurify.sanitize({ IN_PLACE: true }, tableData)
+  return tableData
 }
 
 function getWindow (val) {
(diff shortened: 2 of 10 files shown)


raedah:
I dont think the 3, 5, more then 5 distinction is useful here. Should just be straight list by # of outputs. I dont think anyone is a fan of the donut chart either.

dmigwi:
Ok. I will do away with the donut chart then

dmigwi:
I've replaced the donut chart with a bar graph
Screenshot from 2019-05-03 14-59-48

raedah:
The time view of that data is not useful. I will msg you to discuss how it can be shown in a better context.

chappjc:
Please coordinate with @buck54321 regarding the #outputs charts and graphs.  This was discussed in the charts cache PR here: https://github.com/decred/dcrdata/pull/1274#issuecomment-486314557

buck54321:
I think the only number that means anything anymore is the **solo staking ratio**. *How many of the tickets appear to be solo stakers?* A single figure with some appropriate averaging, e.g. daily average, or rolling N-block average. If a history chart was desired, it would be a line chart, not by block.

dmigwi:
In place of the new chart that had been added this new table has replaced it.
Screenshot from 2019-05-08 04-00-47

raedah:
@buck54321 there was a prior discussion that happened in the matrix room where several people voiced that they did not think it was accurate to infer 'solo' or otherwise based on output count. Thats why the labels have been dropped. I do find the chart migwi just posted useful, and a historic chart would be useful too.

chappjc:
I think the table is interesting.  A good replacement for previous charts.

db/dcrpg/internal/txstmts.go
```diff
@@ -190,12 +190,13 @@ const (
...
-		FROM transactions JOIN tickets
-		ON transactions.id=purchase_tx_db_id WHERE pool_status=0
-		AND tickets.is_mainchain = TRUE GROUP BY ticket_bucket;`
+	SelectTicketsByType = `SELECT DISTINCT num_vout, COUNT(*)
+		FROM transactions
+		JOIN tickets
+		ON transactions.id=purchase_tx_db_id
+		WHERE pool_status=0
+		AND tickets.is_mainchain = TRUE
+		GROUP BY num_vout;`
```
chappjc: ~60 ms.  Works for me.

public/js/controllers/ticketpool_controller.js
```diff
@@ -75,14 +74,17 @@ function priceGraphData (items, memP) {
...
-
-function outputsGraphData (items) {
-  return [
-    getVal(items.solo),
-    getVal(items.pooled),
-    getVal(items.txsplit)
-  ]
+function populateOutputs (data) {
+  var totalCount = 0
+  data.count.map((n) => { totalCount += n })
```
buck54321: `var totalCount = data.count.reduce((a, n) => { return a + n }, 0)`

public/js/controllers/ticketpool_controller.js
```diff
@@ -75,14 +74,17 @@ function priceGraphData (items, memP) {
...
+  var tableData = '<thead><td># of Outputs</td><td>Count</td><td>% Occurence</td></thead>'
+  data.outputs.map((n, i) => {
+    var count = data.count[i]
+    tableData += '<tr><td class="pr-2 lh1rem vam nowrap xs-w117 font-weight-bold">' + n + '</td>' +
+    '<td><span class="hash lh1rem"> ' + count + ' </span></td>' +
+    '<td><span class="hash lh1rem"> ' + ((count * 100) / totalCount).toFixed(4) + '% </span></td></tr>'
+  })
```
buck54321: We are trying to use template literals for this kind of thing.

Also, anytime we are mixing HTML with network data we should sanitize the HTML before inserting. In this case, since a floating `thead` is not valid HTML, you'll want to create a node and sanitize `{ IN_PLACE: true }`, otherwise `dompurify` will sometimes behave weirdly. 

Alternatively, you could just pass your values through a `parseInt` or `parseFloat`, since nothing should be a string.

public/js/controllers/ticketpool_controller.js
```diff
@@ -75,14 +74,17 @@ function priceGraphData (items, memP) {
...
+  document.getElementById('outputs-data').innerHTML = tableData
```
buck54321: For consistency, let's switch up this implementation to use a stimulus `data-target` instead of an `id` please? Either make `populateOutputs` a controller method, or return the HTML from `populateOutputs`, and set it in `processData`.

public/scss/charts.scss
```diff
@@ -36,24 +36,13 @@
   color: black;
 }
 
-.tickets {
-  width: 100%;
-  height: 300px;
+.tp-charts {
+  width: 100% !important;
+  height: 25vh !important;
```
buck54321: Why the `!important`s? They don't seem necessary.

public/js/controllers/ticketpool_controller.js
```diff
@@ -75,14 +76,17 @@ function priceGraphData (items, memP) {
...
+function populateOutputs (data) {
+  var totalCount = data.count.reduce((a, n) => { return a + n }, 0)
+  var tableData = '<thead><td># of Outputs</td><td>Count</td><td>% Occurence</td></thead>'
+  data.outputs.map((n, i) => {
+    var count = data.count[i]
+    tableData += '<tr><td class="pr-2 lh1rem vam nowrap xs-w117 font-weight-bold">' + n + '</td>' +
+    '<td><span class="hash lh1rem"> ' + count + ' </span></td>' +
+    '<td><span class="hash lh1rem"> ' + ((count * 100) / totalCount).toFixed(4) + '% </span></td></tr>'
+  })
+  dompurify.sanitize({ IN_PLACE: true }, tableData)
```
buck54321: Your arguments are switched here. And when sanitizing `IN_PLACE`, you must provide a node, not raw HTML. This is why I suggested you might use `parseFloat` and `parseInt` instead of dompurify. Use a template literal, and do something like

```
tableData += `<tr><td class="pr-2 lh1rem vam nowrap xs-w117 font-weight-bold">${parseInt(n)}</td>
    <td><span class="hash lh1rem">${parseInt(count)}</span></td>
    <td><span class="hash lh1rem">${((parseInt(count) * 100) / parseFloat(totalCount)).toFixed(4)}% </span></td></tr>`
```
 
and you won't need `dompurify` at all. You can only use this approach when all of your variables are expected to be numbers. 

If you do want to use dompurify, you'll need to create a node, and then dump the child. i.e.

```
tbody = document.createElement('tbody')
tbody.innerHTML = tableData
dompurify.sanitize(tbody, { IN_PLACE: true })
return tbody.firstChild

...
// And then below
while(this.outputsTarget.firstChild) this.outputsTarget.removeChild(this.outputsTarget.firstChild)
this.outputsTarget.appendChild(populateOutputs(data['outputs_chart']))
```

IMO, using `parseInt` and `parseFloat` to sanitize is easier in this case.

views/ticketpool.tmpl
```diff
@@ -17,35 +17,28 @@
...
-          </canvas>
-
-          <div class="ticket_by_types">
-            <b style="color:#2971FF;">3 Outputs:</b> - typically solo.<br><br>
-            <b style="color:#FF8C00;">5 Outputs:</b> - typically Voting Service Provider (VSP).<br><br>
-            <b style="color:#41BF53;">More than 5 Outputs:</b> - likely multi-party (split) tickets.<br>
+        <div class="justify-content-between">
+          <div class="dygraph-label dygraph-title mb-1">Ticket Outputs Distribution</div>
+          <div class="row col-lg-12 col-sm-8 d-flex text-center m-auto">
+            <table class="tp-outputs-align" data-target="ticketpool.outputs"></tr></table>
```
buck54321: ~~`</tr>`~~. Just noticed this. 

Then you're good.

dmigwi: I think all is well. Nothing looks out of place according to me. Right?

buck54321: ~~`</tr>`~~

dmigwi: Seen.

PR #1310, opened 2019-05-07
cache: validate snip length and test ReorgHandler

buck54321:
I'm still not certain what was the cause of the panic on master, but I think I've corrected a couple of weak spots, and added a test. 

From appearances, it appears the reorg was being attempted on a zero-length day-binned data set. This would have been possible if there was an error in `(ChartData).Lengthen`, in which case the data set would have been emptied to trigger a full refresh. I don't see the associated `Warnf` anywhere though, so it's not clear that's what happened. 

Regardless, the `Snip` length in `ReorgHandler` needed to be checked anyway.

### DIFF
--- a/db/cache/charts.go
+++ b/db/cache/charts.go
@@ -211,6 +211,9 @@ type zoomSet struct {
 
 // Snip truncates the zoomSet to a provided length.
 func (set *zoomSet) Snip(length int) {
+	if length < 0 {
+		length = 0
+	}
 	set.Height = set.Height.snip(length)
 	set.Time = set.Time.snip(length)
 	set.PoolSize = set.PoolSize.snip(length)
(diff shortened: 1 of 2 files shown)


db/cache/charts.go
```diff
@@ -468,14 +474,18 @@ func (charts *ChartData) ReorgHandler(wg *sync.WaitGroup, c chan *txhelpers.Reor
 			commonAncestorHeight := uint64(data.NewChainHeight) - uint64(len(data.NewChain))
 			charts.mtx.Lock()
 			defer charts.mtx.Unlock()
+			newHeight := int(commonAncestorHeight) + 1
+			log.Debug("ChartData.ReorgHandler snipping blocks height to %d", newHeight)
 			charts.Blocks.Snip(int(commonAncestorHeight) + 1)
```
chappjc: Did you want to feed `newHeight` to `Snip`?

buck54321: I need some coffee.

PR #1313, opened 2019-05-08
 charts: fix reorg deadlock

buck54321:
Also a few followups to address @dmigwi's orphaned review from #1274.

### DIFF
--- a/db/cache/charts.go
+++ b/db/cache/charts.go
@@ -870,23 +870,55 @@ func accumulate(data ChartUints) ChartUints {
 	return d
 }
 
-// Translate the uints to a slice of the differences between each. The provided
-// data is assumed to be monotonically increasing. The original dataset minus
-// the first element is returned for convenience.
-func btw(data ChartUints) (ChartUints, ChartUints) {
-	d := make(ChartUints, 0, len(data))
-	dataLen := len(data)
+// Translate the times slice to a slice of differences. The original dataset
+// minus the first element is returned for convenience.
+func blockTimes(blocks ChartUints) (ChartUints, ChartUints) {
+	times := make(ChartUints, 0, len(blocks))
+	dataLen := len(blocks)
 	if dataLen < 2 {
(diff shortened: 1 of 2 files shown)


db/cache/charts.go
```diff
@@ -931,9 +932,11 @@ func coinSupplyChart(charts *ChartData, zoom ZoomLevel) ([]byte, error) {
 func durationBTWChart(charts *ChartData, zoom ZoomLevel) ([]byte, error) {
 	switch zoom {
 	case BlockZoom:
-		return charts.encode(charts.Blocks.Time, btw(charts.Blocks.Time))
+		t, d := btw(charts.Blocks.Time)
+		return charts.encode(t, d)
```
chappjc: No problem, but note that go will bind the return values of `btw` to the parameters of `charts.encode` if you do:

```go
charts.encode(btw(charts.Blocks.Time))
```

(not just because `charts.encode` is variadic, but because the spec allows this kind of thing.  https://play.golang.org/p/8mPe9bDAT-x

buck54321: that sweet sugar! I'm gonna update it.

buck54321:
Also fixed a deficiency in the way daily block time averages were being calculated.

db/cache/charts.go
```diff
@@ -870,22 +870,55 @@ func accumulate(data ChartUints) ChartUints {
...
+	}
+	times := make(ChartUints, 0, len(ticks)-1)
+	avgs := make(ChartUints, 0, len(ticks)-1)
+	workingOn := ticks[0]
+	nextIdx := 1
+	next := ticks[nextIdx]
+	lastIdx := 0
+	for i, t := range blocks {
+		if t > next {
+			_, pts := blockTimes(blocks[lastIdx:i])
```
chappjc: A check that len(ticks) == len(blocks) would give me the warm fuzzies.

Issue #1314, opened 2019-05-08
charts data cache should be deleted just before re-saving


chappjc:
If dcrdata dies ungracefully, the charts data cache gob file will not be saved.  However, since the charts data cache file gets deleted on startup (just after loading it), there will be no charts data cache when dcrdata restarts after a crash.

```
PSQL: Cache dump data loading failed: open /mnt/lssd/dcrdata_datadir/testnet3/chartscache.gob: no such file or directory
```

If the charts data file is not deleted after loading, and dcrdata crashes at some point, the next start will at least find the charts cache gob file.  However, since the gob was saved at block A, but the DBs can be at any future block B, some care may be needed to ensure the charts data is at DB height after loading it.

chappjc:
Resolved by https://github.com/decred/dcrdata/pull/1316

PR #1316, opened 2019-05-08
charts: Update derived data on db charts data query

dmigwi:
Once `func (charts *ChartData) Update() {` was invoked to fetch the latest charts data update, the derived data `charts.Days` dataset wasn't updated by invoking `func (charts *ChartData) Lengthen() error {`.

Failure to invoke `func (charts *ChartData) Lengthen() error {` on data update.
- Allowed stale data to be persisted till the next system start up.
- When a fresh data query (cache .gob file missing) was run the derived data was bound to return some empty datasets thus affecting functionality of some charts.

### DIFF
--- a/db/cache/charts.go
+++ b/db/cache/charts.go
@@ -712,9 +712,15 @@ func (charts *ChartData) Update() {
 		}
 		cancel()
 		if err != nil {
-			log.Warnf("%v", err)
+			log.Errorf("%v", err)
 			return
 		}
+	}
+
+	// Since the charts db data query is complete. Update chart.Days derived dataset.
+	if err := charts.Lengthen(); err != nil {
+		log.Errorf("(*ChartData).Lengthen failed %v", err)
+		return
 	}
 }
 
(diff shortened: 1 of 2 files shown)


dmigwi:
Also the old cached dump from now will be persisted till newer data is available. #1314

db/cache/charts.go
```diff
@@ -838,9 +842,16 @@ func (charts *ChartData) encode(sets ...lengther) ([]byte, error) {
...
 		return nil, fmt.Errorf("encode called without arguments")
 	}
+
 	smaller := sets[0].Length()
 	for _, x := range sets {
 		l := x.Length()
+
+		// Detect empty datasets and return an error if found.
+		if l == 0 {
+			return nil, fmt.Errorf("empty dataset found")
```
buck54321: I'm torn on whether this should be an error. This would not necessarily be an error for a `Days` data set on the day of a chain's genesis block, but that case is extremely limited. On the other hand, there may also be larger bins (week, month) implemented someday. 

Now that you've added the missing `Lengthen` call, all validation should already be done there, so this error is probably not needed anyway. Maybe `encode` should stay permissive of empty data sets and validation can stay in `Lengthen`.

dmigwi: You are right on this that its occurrence would be rarer now but in case it ever occurred for whatever reason in the future, its effects would be disastrous. No chart that can load with one or more of the axis data set empty.
Also I don't imagine an ideal case scenario where empty data sets for the historical charts will be returned. It is not possible (that straight forward) to run the dcrdata server with only the genesis block loaded.
Still think that an error isn't suitable?

dmigwi: On second though let me convert it to a warning.

buck54321: Let's get rid of it, with no warning. In the case of actual invalid data, there will already be a message from `Lengthen`. It's not necessary for `encode` to care about the content of the data.

buck54321: Just to keep things modular. Separation of concerns.

PR #1318, opened 2019-05-09
pgblockchain: recover from piparser panic

buck54321:
~~Tested by forcing a panic in `path/to/go/pkg/mod/github.com/dmigwi/go-piparser/proposals@v0.0.0-20190426030541-8412e0f44f55/parser.go` in `UpdateSignal`.~~

Never mind. Don't do that. Use a replace in the `go.mod` and point it to a repo somewhere.

### DIFF
--- a/db/dcrpg/pgblockchain.go
+++ b/db/dcrpg/pgblockchain.go
@@ -1113,6 +1113,13 @@ func (pgb *ChainDB) VotesInBlock(hash string) (int16, error) {
 	return voters, nil
 }
 
+// piparser panics should not cause dcrdata panics.
+func recoverPiparser() {
+	if r := recover(); r != nil {
+		log.Errorf("recovered from piparser panic: %v", r)
+	}
+}
+
 // proposalsUpdateHandler runs in the background asynchronous to retrieve the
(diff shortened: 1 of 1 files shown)


db/dcrpg/pgblockchain.go
```diff
@@ -1113,6 +1113,13 @@ func (pgb *ChainDB) VotesInBlock(hash string) (int16, error) {
 	return voters, nil
 }
 
+// piparser panics should not cause dcrdata panics.
+func recoverPiparser() {
+	if r := recover(); r != nil {
+		log.Errorf("recovered from piparser panic: %v", r)
+	}
+}
```
chappjc: I'm tempted to say relaunch `proposalsUpdateHandler`, but it's possible or even likely to panic again depending on the root cause.  Perhaps relaunch after a short delay.  Thoughts?

buck54321: That sounds good. Should we print the `debug.Stack()` ?

chappjc: Ah, yeah, totally.

PR #1321, opened 2019-05-13
API: /api/status/happy and periodic node peer connection check

chappjc:
This adds a new API endpoint, `/api/status/happy`, for bottom line health checks.

The happy endpoint responds with a 503 (service unavailable) if dcrdata is not "happy".
A JSON response marshaled from the new `api/types.Happy` struct is returned
regardless of 200 (OK) or 503 (service unavailable).

An example response with HTTP response code 200:

```json
{
  "happy": true,
  "api_ready": true,
  "tip_age": 1530,
  "node_connections": 1
}
```

This may seem redundant with the `/api/status` data, but the configuration of health checks
is often very limited.  The 503 response code is returned when `"happy"` is `false`.
The condition for `"happy"` is `api_ready && node_connections > 0 && tip_age > 90min`.

To keep `node_connections` current, a 5 second ticker triggers an RPC and update of `node_connections`.

`(*Status).SetHeight` and `(*Status).SetConnections` account for `Status.nodeConnections`.

Add `writeJSONWithStatus` to write a JSON response to the http.ResponseWriter
with a custom http status code.

### DIFF
--- a/api/apiroutes.go
+++ b/api/apiroutes.go
@@ -175,8 +175,8 @@ func (c *appContext) updateNodeConnections() error {
 	nodeConnections, err := c.nodeClient.GetConnectionCount()
 	if err != nil {
 		// Assume there arr no connections if RPC had an error.
-		nodeConnections = 0
-		err = fmt.Errorf("failed to get connection count: %v", err)
+		c.Status.SetConnections(0)
+		return fmt.Errorf("failed to get connection count: %v", err)
 	}
 
 	// Before updating connections, get the previous connection count.
@@ -184,7 +184,7 @@ func (c *appContext) updateNodeConnections() error {
 
 	c.Status.SetConnections(nodeConnections)
 	if nodeConnections == 0 {
-		return err
+		return nil
 	}
 
 	// Detect if the node's peer connections were just restored.
--- a/api/apiroutes.go
+++ b/api/apiroutes.go
@@ -171,29 +171,66 @@ func NewContext(cfg *AppContextConfig) *appContext {
 	}
 }
 
+func (c *appContext) updateNodeConnections() error {
+	nodeConnections, err := c.nodeClient.GetConnectionCount()
+	if err != nil {
+		// Assume there arr no connections if RPC had an error.
+		nodeConnections = 0
+		err = fmt.Errorf("failed to get connection count: %v", err)
+	}
+
+	// Before updating connections, get the previous connection count.
+	prevConnections := c.Status.NodeConnections()
+
+	c.Status.SetConnections(nodeConnections)
+	if nodeConnections == 0 {
+		return err
+	}
+
+	// Detect if the node's peer connections were just restored.
+	if prevConnections != 0 {
+		// Status.ready may be false, but since connections were not lost and
+		// then recovered, it is not our job to check other readiness factors.
+		return nil
+	}
+
+	// Check the reconnected node's best block, and update Status.height.
+	_, nodeHeight, err := c.nodeClient.GetBestBlock()
+	if err != nil {
+		c.Status.SetReady(false)
+		return fmt.Errorf("node: getbestblock failed: %v", err)
+	}
+
+	// Update Status.height with current node height. This also sets
+	// Status.ready according to the previously-set Status.dbHeight.
+	c.Status.SetHeight(uint32(nodeHeight))
+
+	return nil
+}
+
 // StatusNtfnHandler keeps the appContext's Status up-to-date with changes in
 // node and DB status.
 func (c *appContext) StatusNtfnHandler(ctx context.Context, wg *sync.WaitGroup) {
 	defer wg.Done()
+	// Check the node connection count periodically.
+	rpcCheckTicker := time.NewTicker(5 * time.Second)
 out:
 	for {
 	keepon:
 		select {
+		case <-rpcCheckTicker.C:
+			if err := c.updateNodeConnections(); err != nil {
+				log.Warn("updateNodeConnections: ", err)
+				break keepon
+			}
+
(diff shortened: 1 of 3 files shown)


api/types/apitypes.go
```diff
@@ -460,6 +460,29 @@ func (s *Status) API() APIStatus {
...
+	s.RLock()
+	blockAge := time.Since(time.Unix(s.dbLastBlockTime, 0))
+	h := Happy{
+		APIReady:        s.ready,
+		TipAge:          int64(blockAge.Seconds()),
+		NodeConnections: s.nodeConnections,
+	}
+	s.RUnlock()
+
+	h.Happy = h.APIReady && blockAge < 90*time.Minute && h.NodeConnections > 0
```
chappjc: I'm unsure if 90 minutes is a safe happy threshold.  Will have to check the distribution to pick a duration that is statistically extremely unlikely.

api/types/apitypes.go
```diff
@@ -471,37 +494,56 @@ func (s *Status) Height() uint32 {
...
+}
+
+// SetConnections sets the node connection count.
+func (s *Status) SetConnections(conns int64) {
 	s.Lock()
-	defer s.Unlock()
 	s.nodeConnections = conns
-	s.ready = height == s.dbHeight
-	s.height = height
+	s.ready = s.ready && s.nodeConnections > 0
```
chappjc: Note that going back from 0 to N>0 connections does not automatically make it ready.  Restoring connections requires checking and setting node height, as done in `updateNodeConnections`.

api/types/apitypes.go
```diff
@@ -471,37 +494,56 @@ func (s *Status) Height() uint32 {
...
-// is set to true if Status.height is the same as Status.dbHeight.
+// is set to true if Status.height is the same as Status.dbHeight and the node
+// has connections.
 func (s *Status) DBUpdate(height uint32, blockTime int64) {
 	s.Lock()
-	defer s.Unlock()
 	s.dbHeight = height
 	s.dbLastBlockTime = blockTime
-	s.ready = s.dbHeight == height
+	s.ready = s.dbHeight == s.height
```
chappjc: I'm pretty sure this was a bug.

api/types/apitypes.go
chappjc: For an exponential distribution with a mean of 300 seconds, the probability of a block taking 5400 seconds to mine is about 0.9999999847700203 (close to 1 - 5e-8).  Network incidents not considered, it seems that 90 minutes is not too long.

chappjc: BTW, that's computed from `1 - exp(-5400/300) = 1 - 1/exp(18)`.

buck54321:
I noticed during testing that the client appears to be capable of automatically reconnecting if a node connection is lost, but we are disabling it [here](https://github.com/decred/dcrdata/blob/aab33cbbf837e649fd4654c4285f20e4dc89f0ca/main.go#L1322). Could or should we make that a config option?

api/types/apitypes.go
buck54321: I like those numbers.

chappjc:
Regarding the the auto-reconnect setting here:

https://github.com/decred/dcrdata/blob/aab33cbbf837e649fd4654c4285f20e4dc89f0ca/rpcutils/rpcclient.go#L44

We are disabling reconnect (of the RPC client to the node's RPC server) because we haven't addressed what happens if when it does reconnect a block or even a reorg has happened without dcrdata being notified.  It needs to be solved better than just having the next RPC call result in an error an dcrdata shutting down, but I think rpcclient needs connect/disconnect notifications.  Otherwise we'll have to do a lot of client abstraction to turn off RPC requests when a connection is not available.

But for testing the peer connections dropping to zero, I start dcrd with a single peer via `--connect` and then use the `node` RPC to disconnect (e.g. `dcrctl --notls node remove 192.168.1.2:9108`).  That way the RPC connection is still alive, but the `"node_connections"` field would be zero.  This simulates the issue I've hit in production where dcrd suddenly loses its peers, then it doesn't get new ones for one reason or another.

api/apiroutes.go
```diff
@@ -171,29 +171,66 @@ func NewContext(cfg *AppContextConfig) *appContext {
...
+		nodeConnections = 0
+		err = fmt.Errorf("failed to get connection count: %v", err)
+	}
+
+	// Before updating connections, get the previous connection count.
+	prevConnections := c.Status.NodeConnections()
+
+	c.Status.SetConnections(nodeConnections)
+	if nodeConnections == 0 {
+		return err
```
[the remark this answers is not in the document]
chappjc: I actually confused myself while working on this, but it can be non-nil because the `if err != nil {` check above does not return.  Rather it sets `nodeConnections=0` (and supplements `err`).  So this `if nodeConnections == 0` can happen if (1) `GetConnectionCount` just returns zero node connections, or (2) if the RPC call has an error.  We don't return above in the non-nil error case because we still want to `c.Status.SetConnections(nodeConnections)`.  Since this is somewhat tortuous logic, I'll rewrite this part as follows for clarity:

```go
	nodeConnections, err := c.nodeClient.GetConnectionCount()
	if err != nil {
		// Assume there arr no connections if RPC had an error.
		c.Status.SetConnections(0)
		return fmt.Errorf("failed to get connection count: %v", err)
	}

	// Before updating connections, get the previous connection count.
	prevConnections := c.Status.NodeConnections()

	c.Status.SetConnections(nodeConnections)
	if nodeConnections == 0 {
		return nil
	}
```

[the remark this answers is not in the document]
buck54321: Oh, right. Thanks.

PR #1328, opened 2019-05-15
Hashrate chart: labeling and legend units update

dmigwi:
The following changes were made and UI changes are shown on the screenshot below:
1. The chart was renamed from `Network Hashrate` to `Hashrate`.
2. The previous y-Axis and legend used `K` instead of hash rate specific labeling like `TeraHash/s`, `PentaHash/s` for the legend and `Th`, `Ph` for the axis.
3. The charts cache versioning implementation was added back - It should help in managing the cache especially when its data has new structural changes whose only solution is to trigger a fresh update.

**UI Changes**

Screenshot from 2019-05-15 07-38-02-updated

### DIFF
--- a/public/js/controllers/charts_controller.js
+++ b/public/js/controllers/charts_controller.js
@@ -29,13 +29,18 @@ function intComma (amount) {
   return amount.toLocaleString(undefined, { maximumFractionDigits: 0 })
 }
 
-function formatHashRate (s) {
-  var rates = ['Hash/s', 'KHash/s', 'MHash/s', 'GHash/s', 'THash/s', 'PHash/s', 'ExHash/s']
-  for (var i = 0; i < rates.length; i++) {
-    var r = Math.pow(1000, i + 1)
-    var val = Math.floor(s / r)
-    if (val <= 1000 || i + 1 === rates.length) {
-      return intComma(val) + ' ' + rates[i]
+function formatHashRate (value, displayType) {
+  value = parseInt(value)
+  if (value <= 0) return value
+  var shortUnits = ['Th', 'Ph', 'Eh']
+  var labelUnits = ['TeraHash/s', 'PentaHash/s', 'ExaHash/s']
+  for (var i = 0; i < labelUnits.length; i++) {
+    var quo = Math.pow(1000, i)
+    var max = Math.pow(1000, i + 1)
+    if ((value > quo && value <= max) || i + 1 === labelUnits.length) {
+      var data = intComma(Math.floor(value / quo))
+      if (displayType === 'axis') return data + ' ' + shortUnits[i]
+      return data + ' ' + labelUnits[i]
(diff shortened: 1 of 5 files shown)


public/js/controllers/charts_controller.js
```diff
@@ -29,6 +30,22 @@ function intComma (amount) {
   return amount.toLocaleString(undefined, { maximumFractionDigits: 0 })
 }
 
+function formatHashRate (value, displayType) {
+  value = parseInt(value)
+  if (value <= 0) return value
+  var shortUnits = ['Th', 'Ph', 'Eh']
+  var labelUnits = ['TeraHash/s', 'PentaHash/s', 'ExaHash/s']
```
chappjc: penta -> peta

chappjc: The Hash part does not need caps.  Actually, I'm not sure should be any caps (terahash/s)

public/js/controllers/charts_controller.js
```diff
@@ -319,14 +342,14 @@ export default class extends Controller {
...
-          false, 'Cumulative Chainwork (exahash)', 'Date', undefined, true, false))
+        assign(gOptions, mapDygraphOptions(d, ['Date', 'Cumulative Chainwork (exahash/s)'],
+          false, 'Cumulative Chainwork (exahash/s)', 'Date', undefined, true, false))
         break
 
       case 'hashrate': // Total chainwork over time
         d = zipYvDate(data)
-        assign(gOptions, mapDygraphOptions(d, ['Date', 'Network Hashrate (terahash/s)'],
-          false, 'Network Hashrate (terahash/s)', 'Date', undefined, true, false))
+        assign(gOptions, mapDygraphOptions(d, ['Date', 'Hashrate'], false, 'Hashrate (per seconds)', 'Date', undefined, false, false))
```
chappjc: `Hashrate (per seconds)` doesn't make too much sense since it's already described as a rate.  I suggest having this be just `Hashrate` or `Hashrate (hashes per second)` if you feel that adds anything, but I don't think it does.

public/js/controllers/charts_controller.js
```diff
@@ -319,14 +338,14 @@ export default class extends Controller {
 
       case 'chainwork': // Total chainwork over time
         d = zipYvDate(data)
-        assign(gOptions, mapDygraphOptions(d, ['Date', 'Cumulative Chainwork (exahash)'],
-          false, 'Cumulative Chainwork (exahash)', 'Date', undefined, true, false))
+        assign(gOptions, mapDygraphOptions(d, ['Date', 'Cumulative Chainwork (exahash/s)'],
+          false, 'Cumulative Chainwork (exahash/s)', 'Date', undefined, true, false))
```
buck54321: Old unit was correct.

PR #1333, opened 2019-05-16
explorer: give StatusPage the http.Request

chappjc:
StatusPage panics because `dummyRequest` has a nil `*URL`.  This give `StatusPage` and `timeoutErrorPage` the `*http.Request`.


buck54321:
Hmmm. I ran into this a while back and snuck in a fix with #1274.  Or so I thought.

https://github.com/decred/dcrdata/blob/4b8bea7c72c66baa793c9d80b7573dd66b41d435/explorer/explorerroutes.go#L38

chappjc:
Ah, that's why it's only panicing on 4.1.  I'll close these PRs down and back port that `init` for explorerroutes.  We presently don't have a need to pass the http.Request to StatusPage, other than for the URL, so we won't do it.

PR #1335, opened 2019-05-16
[4.1] resolve lite mode panics

chappjc:
lite mode is regrettably still available in dcrdata 4.1.  This resolves panics that prevented it from working.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,5 +1,7 @@
 module github.com/decred/dcrdata/v4
 
+replace github.com/decred/dcrdata/gov/agendas => ./gov/agendas
+
 require (
 	github.com/caarlos0/env v3.5.0+incompatible
 	github.com/chappjc/logrus-prefix v0.0.0-20180227015900-3a1d64819adb
--- a/gov/agendas/tracker.go
+++ b/gov/agendas/tracker.go
@@ -467,6 +467,10 @@ func (tracker *VoteTracker) newVoteSummary() *VoteSummary {
 // never be modified by VoteTracker, so can be used read-only by any number
 // of threads.
 func (tracker *VoteTracker) Summary() *VoteSummary {
+	// lite mode hack
+	if tracker == nil {
+		return nil
+	}
 	tracker.mtx.RLock()
 	defer tracker.mtx.RUnlock()
 	return tracker.summary
--- a/main.go
+++ b/main.go
@@ -678,16 +678,16 @@ func _main(ctx context.Context) error {
 
 	// This dumps the cache charts data into a file for future use on system
 	// exit.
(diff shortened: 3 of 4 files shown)


go.mod
```diff
@@ -1,5 +1,7 @@
 module github.com/decred/dcrdata/v4
 
+replace github.com/decred/dcrdata/gov/agendas => ./gov/agendas
```
chappjc: The change to agendas.VoteTracker is a hack just for lite mode.  Further, 4.1 is about to become legacy, so we aren't going to bother with a new gov/agendas module tag.

gov/agendas/tracker.go
```diff
@@ -467,6 +467,10 @@ func (tracker *VoteTracker) newVoteSummary() *VoteSummary {
 // never be modified by VoteTracker, so can be used read-only by any number
 // of threads.
 func (tracker *VoteTracker) Summary() *VoteSummary {
+	// lite mode hack
+	if tracker == nil {
+		return nil
```
chappjc: Resolving it this way since the proper fix is not needed at all on `master`/5.0 and lite mode is essentially never used.

main.go
```diff
@@ -564,10 +564,13 @@ func _main(ctx context.Context) error {
...
 
 	// A vote tracker tracks current block and stake versions and votes.
-	tracker, err := agendas.NewVoteTracker(activeChain, dcrdClient,
-		auxDB.AgendaVoteCounts, activeChain.Deployments)
-	if err != nil {
-		return fmt.Errorf("Unable to initialize vote tracker: %v", err)
+	var tracker *agendas.VoteTracker
+	if usePG {
+		tracker, err = agendas.NewVoteTracker(activeChain, dcrdClient,
+			auxDB.AgendaVoteCounts, activeChain.Deployments)
```
chappjc: was a panic on `auxDB.AgendaVoteCounts`

main.go
```diff
@@ -674,11 +679,13 @@ func _main(ctx context.Context) error {
 	// This dumps the cache charts data into a file for future use on system
 	// exit.
 	defer func() {
-		er := explorer.WriteCacheFile(dumpPath)
-		if er != nil {
-			log.Errorf("WriteCacheFile failed: %v", er)
-		} else {
-			log.Debug("Dumping the charts cache data was successful")
+		if usePG {
+			er := explorer.WriteCacheFile(dumpPath)
```
chappjc: Not useful in lite mode.

chappjc: pretty sure, but do check this assertion please, @dmigwi

dmigwi: You right. You could probably include the whole defer function in the if statement.

PR #1339, opened 2019-05-18
Charts: Dual x-axis support for all historical charts

dmigwi:
All historical charts have a button that enables switching between the two x-axis types namely: Block Height and Block Time. A group button at the chart bottom was added to enable easy swapping of either of the two x-axis. Since dygraph doesn't support display of multiple x-axis the two distinct x-axes cannot not be displayed together.

No changes to the data stored in charts cache was made to enable an extra x-axis to be swapped on a chart with the same y-axis data set.
The query below allows the index of each record retrieved to be used like a `Block Height` since the query is ordered by height. Therefore no extra data is required when the bin value is set to blocks.
```sql
SelectBlockStats = `SELECT height, size, time, chainwork, numtx
	FROM blocks
	WHERE is_mainchain
        AND height > $1
	ORDER BY height;`
``` 

When the bin value is set to day a pre-populated height array entry existed.

**Charts UI**
Screenshot from 2019-05-18 08-52-48

### DIFF
--- a/api/apiroutes.go
+++ b/api/apiroutes.go
@@ -1602,11 +1602,11 @@ func (c *appContext) getAddressTxAmountFlowData(w http.ResponseWriter, r *http.R
 
 func (c *appContext) ChartTypeData(w http.ResponseWriter, r *http.Request) {
 	chartType := m.GetChartTypeCtx(r)
-	zoom := r.URL.Query().Get("zoom")
-	chartData, err := c.charts.Chart(chartType, zoom)
+	bin := r.URL.Query().Get("bin")
+	chartData, err := c.charts.Chart(chartType, bin)
 	if err != nil {
 		http.NotFound(w, r)
-		log.Warnf(`Error fetching chart %s at zoom level '%s': %v`, chartType, zoom, err)
+		log.Warnf(`Error fetching chart %s at bin level '%s': %v`, chartType, bin, err)
 		return
 	}
 	writeJSONBytes(w, chartData)
--- a/api/apiroutes.go
+++ b/api/apiroutes.go
@@ -1603,6 +1603,10 @@ func (c *appContext) getAddressTxAmountFlowData(w http.ResponseWriter, r *http.R
 func (c *appContext) ChartTypeData(w http.ResponseWriter, r *http.Request) {
 	chartType := m.GetChartTypeCtx(r)
 	bin := r.URL.Query().Get("bin")
+	// Support the deprecated URL parameter "zoom".
+	if bin == "" {
+		bin = r.URL.Query().Get("zoom")
+	}
 	axis := r.URL.Query().Get("axis")
 	chartData, err := c.charts.Chart(chartType, bin, axis)
 	if err != nil {
--- a/api/apiroutes.go
+++ b/api/apiroutes.go
@@ -1603,7 +1603,8 @@ func (c *appContext) getAddressTxAmountFlowData(w http.ResponseWriter, r *http.R
 func (c *appContext) ChartTypeData(w http.ResponseWriter, r *http.Request) {
 	chartType := m.GetChartTypeCtx(r)
 	bin := r.URL.Query().Get("bin")
-	chartData, err := c.charts.Chart(chartType, bin)
+	axis := r.URL.Query().Get("axis")
+	chartData, err := c.charts.Chart(chartType, bin, axis)
 	if err != nil {
 		http.NotFound(w, r)
 		log.Warnf(`Error fetching chart %s at bin level '%s': %v`, chartType, bin, err)
--- a/db/cache/charts.go
+++ b/db/cache/charts.go
@@ -66,6 +66,16 @@ func ParseZoom(zoom string) binLevel {
 	return DefaultZoomLevel
 }
 
+// ParseAxis returns the matching axis type, else the default of time axis.
+func ParseAxis(aType string) axisType {
+	switch axisType(aType) {
+	case HeightAxis:
+		return HeightAxis
+	default:
+		return TimeAxis
+	}
+}
+
 const (
 	// aDay defines the number of seconds in a day.
 	aDay = 86400
@@ -770,8 +780,8 @@ func NewChartData(height uint32, genesis time.Time, chainParams *chaincfg.Params
 }
 
 // A cacheKey is used to specify cached data of a given type and ZoomLevel.
-func cacheKey(chartID string, zoom binLevel) string {
-	return chartID + "-" + string(zoom)
+func cacheKey(chartID string, zoom binLevel, axis axisType) string {
+	return chartID + "-" + string(zoom) + "-" + string(axis)
 }
 
 // Grabs the cacheID associated with the provided ZoomLevel. Should be called
@@ -789,9 +799,9 @@ func (charts *ChartData) cacheID(zoom binLevel) uint64 {
 }
 
 // Grab the cached data, if it exists. The cacheID is returned as a convenience.
-func (charts *ChartData) getCache(chartID string, zoom binLevel) (data *cachedChart, found bool, cacheID uint64) {
+func (charts *ChartData) getCache(chartID string, zoom binLevel, axis axisType) (data *cachedChart, found bool, cacheID uint64) {
 	// Ignore zero length since bestHeight would just be set to zero anyway.
-	ck := cacheKey(chartID, zoom)
+	ck := cacheKey(chartID, zoom, axis)
 	charts.cacheMtx.RLock()
 	defer charts.cacheMtx.RUnlock()
 	cacheID = charts.cacheID(zoom)
@@ -800,8 +810,8 @@ func (charts *ChartData) getCache(chartID string, zoom binLevel) (data *cachedCh
 }
 
 // Store the chart associated with the provided type and ZoomLevel.
-func (charts *ChartData) cacheChart(chartID string, zoom binLevel, data []byte) {
-	ck := cacheKey(chartID, zoom)
+func (charts *ChartData) cacheChart(chartID string, zoom binLevel, axis axisType, data []byte) {
+	ck := cacheKey(chartID, zoom, axis)
 	charts.cacheMtx.Lock()
 	defer charts.cacheMtx.Unlock()
 	// Using the current best cacheID. This leaves open the small possibility that
@@ -815,7 +825,7 @@ func (charts *ChartData) cacheChart(chartID string, zoom binLevel, data []byte) 
 
 // ChartMaker is a function that accepts a chart type and ZoomLevel, and returns
 // a JSON-encoded chartResponse.
-type ChartMaker func(charts *ChartData, zoom binLevel) ([]byte, error)
+type ChartMaker func(charts *ChartData, zoom binLevel, axis axisType) ([]byte, error)
 
 var chartMakers = map[string]ChartMaker{
 	BlockSize:       blockSizeChart,
@@ -834,9 +844,10 @@ var chartMakers = map[string]ChartMaker{
 
 // Chart will return a JSON-encoded chartResponse of the provided type
 // and ZoomLevel.
-func (charts *ChartData) Chart(chartID, zoomString string) ([]byte, error) {
+func (charts *ChartData) Chart(chartID, zoomString, axisString string) ([]byte, error) {
 	zoom := ParseZoom(zoomString)
-	cache, found, cacheID := charts.getCache(chartID, zoom)
+	axis := ParseAxis(axisString)
+	cache, found, cacheID := charts.getCache(chartID, zoom, axis)
 	if found && cache.cacheID == cacheID {
 		return cache.data, nil
 	}
@@ -847,12 +858,12 @@ func (charts *ChartData) Chart(chartID, zoomString string) ([]byte, error) {
 	// Do the locking here, rather than in encodeXY, so that the helper functions
 	// (accumulate, btw) are run under lock.
 	charts.mtx.RLock()
-	data, err := maker(charts, zoom)
+	data, err := maker(charts, zoom, axis)
 	charts.mtx.RUnlock()
 	if err != nil {
 		return nil, err
 	}
-	charts.cacheChart(chartID, zoom, data)
+	charts.cacheChart(chartID, zoom, axis, data)
 	return data, nil
 }
 
@@ -948,7 +959,10 @@ func avgBlockTimes(ticks, blocks ChartUints) (ChartUints, ChartUints) {
 	return times, avgs
 }
 
-func blockSizeChart(charts *ChartData, zoom binLevel) ([]byte, error) {
+func blockSizeChart(charts *ChartData, zoom binLevel, axis axisType) ([]byte, error) {
+	if zoom == DayZoom && axis == HeightAxis {
+		return charts.encode(charts.Days.Height, charts.Days.BlockSize)
+	}
 	switch zoom {
 	case BlockZoom:
 		return charts.encode(charts.Blocks.Time, charts.Blocks.BlockSize)
@@ -958,7 +972,10 @@ func blockSizeChart(charts *ChartData, zoom binLevel) ([]byte, error) {
 	return nil, InvalidZoomErr
 }
 
-func blockchainSizeChart(charts *ChartData, zoom binLevel) ([]byte, error) {
+func blockchainSizeChart(charts *ChartData, zoom binLevel, axis axisType) ([]byte, error) {
+	if zoom == DayZoom && axis == HeightAxis {
+		return charts.encode(charts.Days.Height, accumulate(charts.Days.BlockSize))
+	}
 	switch zoom {
 	case BlockZoom:
 		return charts.encode(charts.Blocks.Time, accumulate(charts.Blocks.BlockSize))
@@ -968,7 +985,10 @@ func blockchainSizeChart(charts *ChartData, zoom binLevel) ([]byte, error) {
 	return nil, InvalidZoomErr
 }
 
-func chainWorkChart(charts *ChartData, zoom binLevel) ([]byte, error) {
+func chainWorkChart(charts *ChartData, zoom binLevel, axis axisType) ([]byte, error) {
+	if zoom == DayZoom && axis == HeightAxis {
+		return charts.encode(charts.Days.Height, charts.Days.Chainwork)
+	}
 	switch zoom {
 	case BlockZoom:
 		return charts.encode(charts.Blocks.Time, charts.Blocks.Chainwork)
@@ -978,7 +998,10 @@ func chainWorkChart(charts *ChartData, zoom binLevel) ([]byte, error) {
 	return nil, InvalidZoomErr
 }
 
-func coinSupplyChart(charts *ChartData, zoom binLevel) ([]byte, error) {
+func coinSupplyChart(charts *ChartData, zoom binLevel, axis axisType) ([]byte, error) {
+	if zoom == DayZoom && axis == HeightAxis {
+		return charts.encode(charts.Days.Height, accumulate(charts.Days.NewAtoms), charts.Days.Height)
+	}
 	switch zoom {
 	case BlockZoom:
 		return charts.encode(charts.Blocks.Time, accumulate(charts.Blocks.NewAtoms))
@@ -988,7 +1011,10 @@ func coinSupplyChart(charts *ChartData, zoom binLevel) ([]byte, error) {
 	return nil, InvalidZoomErr
 }
 
-func durationBTWChart(charts *ChartData, zoom binLevel) ([]byte, error) {
+func durationBTWChart(charts *ChartData, zoom binLevel, axis axisType) ([]byte, error) {
+	if zoom == DayZoom && axis == HeightAxis {
+		return charts.encode(avgBlockTimes(charts.Days.Time, charts.Blocks.Time))
+	}
 	switch zoom {
 	case BlockZoom:
 		return charts.encode(blockTimes(charts.Blocks.Time))
@@ -1026,7 +1052,10 @@ func hashrate(time, chainwork ChartUints) (ChartUints, ChartUints) {
 	return t, y
 }
 
-func hashRateChart(charts *ChartData, zoom binLevel) ([]byte, error) {
+func hashRateChart(charts *ChartData, zoom binLevel, axis axisType) ([]byte, error) {
+	if zoom == DayZoom && axis == HeightAxis {
+		return charts.encode(charts.Days.Height, charts.Days.Chainwork)
+	}
 	switch zoom {
 	case BlockZoom:
 		t, y := hashrate(charts.Blocks.Time, charts.Blocks.Chainwork)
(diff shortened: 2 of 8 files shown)


api/apiroutes.go
```diff
@@ -1560,11 +1560,12 @@ func (c *appContext) getAddressTxAmountFlowData(w http.ResponseWriter, r *http.R
 
 func (c *appContext) ChartTypeData(w http.ResponseWriter, r *http.Request) {
 	chartType := m.GetChartTypeCtx(r)
-	zoom := r.URL.Query().Get("zoom")
-	chartData, err := c.charts.Chart(chartType, zoom)
+	bin := r.URL.Query().Get("bin")
```
dmigwi: I renamed most of all usages of the word and variable `zoom` to `bin` to reduce confusion from the zoom front-end charts feature.

chappjc: I'm disinclined to make such changes since it breaks existing links.  How about supporting "zoom" of no "bin" is provided, and then redirecting to the link with "zoom"?

buck54321: I've used those words a little too liberally, for sure.  I think my thinking was that `zoom`  would correspond to a `zoom level`, and some day in the future, the default bin sizes returned might depend on how far the user is zoomed in, à la google maps. But it's not that yet, so it's probably just a little confusing for now. 

@chappjc's suggestion sounds good for now, but can we mark the date in the docs and give a deprecation warning in the js console so that we can eventually simplify things?

dmigwi: I have added support for the deprecated "zoom" URL parameter. I think the change might be to small to warrant a deprecation warning on the front-end. Also the previous usage of "zoom" URL parameter is easily supported without making too many adjustments.

public/js/controllers/charts_controller.js
```diff
@@ -434,66 +453,69 @@ export default class extends Controller {
     this.lastZoom = Zoom.object(start, end)
     this.settings.zoom = Zoom.encode(this.lastZoom)
     this.query.replace(this.settings)
-    this.setSelectedZoom(Zoom.mapKey(this.lastZoom, this.limits))
   }
 
-  setSelectedZoom (zoomKey) {
```
buck54321: The two different functions `onZoom` and `setSelectedZoom` were for two different purposes. One just set the selected button, one was a button callback which set `this.settings.zoom` and redrew the chart. They can't be combined in the way that you've done here. The zoom selector is no longer working.

image

public/js/controllers/charts_controller.js
```diff
@@ -434,66 +453,69 @@ export default class extends Controller {
...
+    if (!option) return
+    this.setActiveOptionBtn(option, this.binSizeTargets)
+    if (option === e) return // Exit if running for the first time.
     selectedChart = null // Force fetch
     this.selectChart()
   }
 
-  setBinButton (bin) {
-    this.binSizeTargets.forEach(li => {
-      if (li.dataset.bin === bin) {
```
buck54321: Here too, `setBinButton` sets the selected button, while `setBin` changes `this.settings.bin` and causes a chart refresh, which is not desirable during initialization.

public/js/controllers/charts_controller.js
```diff
@@ -197,20 +222,11 @@ export default class extends Controller {
     baseSubsidy = parseInt(this.data.get('bs'))
     subsidyInterval = parseInt(this.data.get('sri'))
     subsidyExponent = parseFloat(this.data.get('mulSubsidy')) / parseFloat(this.data.get('divSubsidy'))
+    windowSize = parseInt(this.data.get('windowSize'))
 
-    this.settings = TurboQuery.nullTemplate(['chart', 'zoom', 'scale', 'bin'])
-    this.query.update(this.settings)
-    if (this.settings.zoom) {
-      this.setSelectedZoom(this.settings.zoom)
```
buck54321: See notes below about why you can't replace `setSelectedZoom` with `setZoom` quite so easily.

db/cache/charts.go
```diff
@@ -913,82 +931,93 @@ func blockTimes(blocks ChartUints) (ChartUints, ChartUints) {
...
-	return nil, InvalidZoomErr
+	return nil, InvalidBinErr
 }
 
-func coinSupplyChart(charts *ChartData, zoom ZoomLevel) ([]byte, error) {
-	switch zoom {
-	case BlockZoom:
+func coinSupplyChart(charts *ChartData, bin binLevel, axis axisType) ([]byte, error) {
+	if bin == DayBin && axis == HeightAxis {
+		return charts.encode(charts.Days.Height, accumulate(charts.Days.NewAtoms), charts.Days.Height)
```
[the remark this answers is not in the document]
dmigwi: I am not sure why it was there in the previous implementation.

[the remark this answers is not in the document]
buck54321: It wasn't. The x-axis was time before.

db/cache/charts.go
```diff
@@ -913,82 +931,93 @@ func blockTimes(blocks ChartUints) (ChartUints, ChartUints) {
...
-	return nil, InvalidZoomErr
+	return nil, InvalidBinErr
 }
 
-func durationBTWChart(charts *ChartData, zoom ZoomLevel) ([]byte, error) {
-	switch zoom {
-	case BlockZoom:
+func durationBTWChart(charts *ChartData, bin binLevel, axis axisType) ([]byte, error) {
+	if bin == DayBin && axis == HeightAxis {
+		return charts.encode(charts.Days.Height, avgBlockTimes(charts.Days.Time, charts.Blocks.Time))
```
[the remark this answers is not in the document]
dmigwi: But `avgBlockTimes` function does that validation.

db/cache/charts.go
```diff
@@ -826,10 +843,11 @@ var chartMakers = map[string]ChartMaker{
...
 // Chart will return a JSON-encoded chartResponse of the provided type
-// and ZoomLevel.
-func (charts *ChartData) Chart(chartID, zoomString string) ([]byte, error) {
-	zoom := ParseZoom(zoomString)
-	cache, found, cacheID := charts.getCache(chartID, zoom)
+// and BinLevel.
+func (charts *ChartData) Chart(chartID, binString, axisString string) ([]byte, error) {
+	bin := ParseBin(binString)
+	axis := ParseAxis(axisString)
+	cache, found, cacheID := charts.getCache(chartID, bin, axis)
```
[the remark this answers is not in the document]
dmigwi: `TimeAxis` value has major effect only when `DayBin` is set otherwise it doesn't get used.
I also cant see where `WindowBin` is used in the code unless one manually edits the url params to return the "window" value.

db/cache/charts.go
```diff
@@ -913,82 +931,93 @@ func blockTimes(blocks ChartUints) (ChartUints, ChartUints) {
...
+func coinSupplyChart(charts *ChartData, bin binLevel, axis axisType) ([]byte, error) {
+	if bin == DayBin && axis == HeightAxis {
+		return charts.encode(charts.Days.Height, accumulate(charts.Days.NewAtoms), charts.Days.Height)
+	}
+	switch bin {
+	case BlockBin:
 		return charts.encode(charts.Blocks.Time, accumulate(charts.Blocks.NewAtoms))
-	case DayZoom:
+	case DayBin:
 		return charts.encode(charts.Days.Time, accumulate(charts.Days.NewAtoms), charts.Days.Height)
```
dmigwi: I meant about this when I was referencing the previous implementation.

dmigwi: Does it have any significance?

buck54321: Yes. The first argument to `encode` is `charts.Days.Time` here. You have changed it to `charts.Days.Height`. Before, it was a time axis, a coin supply axis, and a height axis. Now it is two identical axes of height and one axis of coin supply.

db/cache/charts.go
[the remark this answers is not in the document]
buck54321: `avgBlockTimes` returns a slice of length `len(charts.Days.Time) - 1`, but `charts.Days.Height` is length `len(charts.Days.Height)` = `len(charts.Days.Time)`, so it has one more element. 

Now that I look at it again, even though you are encoding slices of different lengths, since `Height` is the longer one, it will be truncated during encoding by removing an element from the end. So by luck, you will end up with the correct axis alignment. But it's much better IMO to just create the data sets correctly to begin with instead of relying on a fallback feature of `encode`. 

So actually, I had it backwards, the correct way to size the x axis is `charts.Days.Height[:len(charts.Days.Height)-1]`, since we associate a day with it's midnight.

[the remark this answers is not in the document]
dmigwi: I thought returning to your previous implementation was better as I would not have to check for the array length.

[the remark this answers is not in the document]
buck54321: The previous implementation did not deal with `Height` arrays. You're just adding those now, so you have to adjust the height array. It just needs the last element removed.

dmigwi:
@buck54321 I have addressed all the concerns you raised earlier, Kindly take a look at it once again.

db/cache/charts.go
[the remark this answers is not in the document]
buck54321: While `WindowBin` is not used explicitly in `ticketPriceChart` and others window-binned `ChartMaker`s, it is still part of the `cacheKey` for those charts. 

The `axis` is not used for any bin size except `DayBin`. So what happens when you get two requests, one for `/api/chart/hashrate?bin=block&axis=time`, and then another request to `/api/chart/hashrate?bin=block&axis=height`? The same exact data is returned for both, but since the `cacheKey` is different, you are encoding the data twice, unneccesarily. 

Instead of doing that, I'd recommend one of two things

1. When the bin is not `DayBin`, set the axis to some common value, maybe some new value like `NoAxis = ""`. 
2. Find a way to only include the axis in the `cacheKey` when the bin is `DayBin`.

public/js/helpers/zoom_helper.js
```diff
@@ -112,19 +126,18 @@ export default class Zoom {
 
   // mapKey returns the corresponding map key, if the zoom meets the correct
   // range and position within the limits, else null.
-  static mapKey (zoom, limits, scale) {
+  static mapKey (zoom, limits, scale, isTimeAxis, ratio) {
```
buck54321: This doesn't need to be changed. The scale parameter already handled the transformation, it just needs to be set correctly.  Use `scale = 300000` (300 seconds per block) when it is the blocks axis.

public/js/helpers/zoom_helper.js
```diff
@@ -81,15 +94,16 @@ export default class Zoom {
 
   // validate will shift and clamp the proposed zoom window to accommodate the
   // range limits and minimum size.
-  static validate (proposal, limits, minSize, scale) {
+  static validate (proposal, limits, minSize, scale, isTimeAxis) {
```
buck54321: See the other note about using the `scale` parameter.

public/js/controllers/charts_controller.js
```diff
@@ -379,24 +415,15 @@ export default class extends Controller {
...
     await animationFrame()
     this.chartWrapperTarget.classList.add('loading')
     await animationFrame()
     let oldLimits = this.limits || this.chartsView.xAxisExtremes()
     this.limits = this.chartsView.xAxisExtremes()
-    if (this.selectedZoom) {
-      this.lastZoom = Zoom.validate(this.selectedZoom, this.limits, blockTime)
+    var selected = this.selectedZoom()
+    if (selected) {
+      this.lastZoom = Zoom.validate(selected, this.limits, avgBlockTime, this.isTimeAxis())
```
buck54321: Leave `Zoom.validate` the way it was, and use the `minSize` and `scale` parameters like this

```
this.lastZoom = Zoom.validate(selected, this.limits, this.isTimeAxis() ? avgBlockTime : 1, this.isTimeAxis() ? 1 : avgBlockTime)
```

public/js/controllers/charts_controller.js
```diff
@@ -431,69 +457,76 @@ export default class extends Controller {
...
-      } else {
-        button.classList.remove('active')
-      }
-    })
+  setZoom (e) {
+    var target = e.srcElement || e.target
+    var option
+    if (!target) {
+      let ex = this.chartsView.xAxisExtremes()
+      option = Zoom.mapKey(e, ex, 1, this.isTimeAxis(), avgBlockTime)
```
buck54321: Leave `Zoom.mapKey` the way it was, and use the `scale` parameter like 

```
option = Zoom.mapKey(e, ex, this.isTimeAxis() ? 1 : avgBlockTime)
```

public/scss/_variables.scss
```diff
@@ -35,9 +35,9 @@ $progress-bar-bg: #74e06b;
 $progress-bar-color: #333;
 
 // nav
-$nav-pills-border-radius: 0;
-$nav-link-padding-y: .55rem;
-$nav-link-padding-x: .55rem;
+$nav-pills-border-radius: .2rem;
+$nav-link-padding-y: .3rem;
+$nav-link-padding-x: .45rem;
```
buck54321: Good call. This looks much better.

public/js/helpers/zoom_helper.js
```diff
@@ -113,18 +113,15 @@ export default class Zoom {
...
   // range and position within the limits, else null.
   static mapKey (zoom, limits, scale) {
+    scale = scale || 1
     let lims = tryDecode(limits)
     let decoded = this.decode(zoom, lims)
-    if (decoded.end !== lims.end) return null
-    if (decoded.start === lims.start) return 'all'
-    let keys = Object.keys(zoomMap)
-    for (let idx in keys) {
-      let k = keys[idx]
```
buck54321: These changes break `mapKey`. Please undo the changes made to this function.

dmigwi: This the only part whose change breaks my implementation. I am not aware how `null` value returned falls into place because the zoom button doesn't expect it. IMO `all` should be the default and returning `null` complicates things.

Secondly, In the previous implementation the `decoded.start` and `decoded.end` seems to have more significance than just getting the range. I think comparing the range was a better way to implement this `mapKey`.

Lets work on a compromise that works on all cases.

buck54321: No problem. Let's figure it out. What is the behavior that you are expecting?

The zoom buttons correspond to very specific ranges. If none of those ranges are currently displayed, then no zoom button should be selected. There is no default, only specific ranges for which a button should be visually distinguished. 

As an example, compare this PR to alpha by opening a chart, clicking the `year` button, and then grabbing the left zoom handle and dragging the window back to full range. On this PR, the `year` button will remain highlighted, even though it is not a year, and is actually the correct range for the `all` button. On master, when you begin dragging, the `year` button is no longer selected, because it is no longer a window of a year. Also, when you drag back to full range, the `all` button will become selected.

buck54321: Oh yeah. As far as the ranges go, I think it would be fine if the button is selected based on just the range, dropping the *most recent* requirement. It seems like you were trying to do that. The only reason I hesitate at all is because for the url parameter `?zoom=day`, it should display the most recent, but that doesn't necessarily have to match the behavior of the buttons.

dmigwi: >There is no default, only specific ranges for which a button should be visually distinguished.

I don't fully agree with the above statement because if it was true, when no zoom button is active there should never be a chart that is visible. Instead when no zoom button is active, the chart whose zoom corresponds to zoomMap key `all` is visible. IMO this introduces inconsistencies as the other button adjacent to zoom nav do not behave in the same way (i.e. their default values are always active). It also becomes hard to describe a chart if some button sets have an active one while others don't.

Concerning the slider drag feature, well I wasn't aware of its existence and thanks for letting me know about it. And because of the huge volume of data set that historical charts typically deal with, I would rather just use the zoom buttons than try to move the slider in order to select a `year` range or any other range. This ranges deals with exact values (i.e start and end values) and honestly I cannot imagine how someone will use the range slide to select a zoom type.

Don't get me wrong, matching the range selected to its matching zoom type is a cool feature which should remain in place but I doubt users who focus on the range slider at the page bottom will ever notice that the active zoom button at the page top has changed.

dmigwi: My use case of `mapKey` focuses on obtaining the correct zoom type from an old URL while your use case focuses on the slider drag feature.
I suggest we add variable to the `mapKey` function that differentiate those two use cases so as to resolve this slider feature break.

dmigwi: Based on the implementation when `?zoom=day`, It should work seamlessly as when any other correct case scenario is set. Kindly share more on the exact case scenario(s) where `?zoom=day` would be unpredictable.

buck54321: I'm going to use `month` because it's easier to work with. Below are two views of the same chart. 

image

Which view should the user get when they navigate to `/charts?chart=ticket-price&zoom=month`? 

I think most would agree that it's the bottom chart. Then the question is whether the zoom keyword `month` is only associated with the most recent month? 

But again, we can separate the button behavior from the definition of the URL parameter behavior if it improves the experience. 

I'm still not certain what behavior you are expecting that you aren't seeing. I'm not quite sure what you mean when you say 

> obtaining the correct zoom type from an old URL

dmigwi: With old URLs, I meant this: https://github.com/decred/dcrdata/pull/1339#discussion_r286082512

As mentioned here let me add a variable to differentiate the two use cases: https://github.com/decred/dcrdata/pull/1339#discussion_r287894900

buck54321: This is a little confusing. The comment you linked to is about the URL parameter of the `/api/charts` API endpoint , not the `/charts` page. The use of `mapKey` is only used to set the `zoom` range parameter the `/charts` page URL, and is not used for the API endpoint. 

I'd also like to swing back to something above

> when no zoom button is active there should never be a chart that is visible. Instead when no zoom button is active, the chart whose zoom corresponds to zoomMap key all is visible. IMO this introduces inconsistencies as the other button adjacent to zoom nav do not behave in the same way (i.e. their default values are always active). It also becomes hard to describe a chart if some button sets have an active one while others don't.

It seems like you are suggesting that we should always have a button selected, even when it does not accurately describe the displayed range, which I vehemently disagree with. But maybe I'm not understanding your position correctly.

dmigwi: >This is a little confusing. The comment you linked to is about the URL parameter of the /api/charts API endpoint , not the /charts page. The use of mapKey is only used to set the zoom range parameter the /charts page URL, and is not used for the API endpoint.

But on the user facing end, parameters  used to query `/api/charts`  API endpoint are stored on the browser's  `/charts` page URL. Therefore IMO if a specific chart has been referenced, the `/charts` page URL will be used instead of `/api/charts` API URL. I made an effort to ensure that the referenced chart page URL will show a chart similar to the originally referenced when accessed.

With the slider feature, it makes sense as to why a `null` value might be returned and this is why I will be adding an extra parameter to differentiate the two use cases rather than create a separate implementation of `mapKey`.

public/js/helpers/zoom_helper.js
```diff
@@ -112,16 +112,31 @@ export default class Zoom {
...
-      let v = zoomMap[k]
-      if (v / scale === range) return k
+
+    if (isLegacyURL) { // maps the legacy url zoom lookup
+      let range = decoded.end - decoded.start
+      for (let k in zoomMap) {
+        let v = zoomMap[k]
+        if (v / scale === range) return k
+      }
+      return 'all'
```
buck54321: image

I am still completely uncertain about what you are trying to accomplish by changing this function. The output of this function is only used to set what button is blue. It does not affect any URL query parameters. It only sets the button color. Was there an issue with the way the buttons were behaving? What did you see that made you want to change this? 

To make this work correctly with `setActiveOptionBtn`, you should return this function to its original form but make sure you pass a numerical `scale` (or add your `scale = scale || 1` in `mapKey`). Remove this line https://github.com/dmigwi/dcrdata/blob/0a5b19c78d91cf325dd3691e39b423c0bbfeaeaa/public/js/controllers/charts_controller.js#L473

and add this line

```
this.setActiveOptionBtn(Zoom.mapKey(this.settings.zoom, this.limits), this.zoomOptionTargets)
```

at the end of `_zoomCallback` and `_drawCallback`.

dmigwi: The zoom on this url http://localhost:7777/charts?chart=ticket-pool-value&zoom=juy57sw0-jw5luyo0&bin=day&axis=time corresponds to `month` zoom interval. 
Without my `mapKey` implementation the zoom in that url doesn't show the month zoomed chart.

buck54321: I'm willing to bet it's because you were passing a bad value for the `scale` argument at some point. I've tested with the changes I recommended above, and it is working correctly.

dmigwi: Removing this line https://github.com/dmigwi/dcrdata/blob/0a5b19c78d91cf325dd3691e39b423c0bbfeaeaa/public/js/controllers/charts_controller.js#L473 will mean that  when a chart corresponding to `all` zoom type is visible then the `all` button may or may not active.

I think the effect you need functioning should be fixing by just adding:
```
this.setActiveOptionBtn(Zoom.mapKey(this.settings.zoom, this.limits), this.zoomOptionTargets)
```
to the 
`_zoomCallBack`.

dmigwi: All should be well now.

buck54321: That will not fix the problem in the image I pasted above. The function `mapKey` function was working perfectly before. There was no need to change it.

buck54321: Removing that line will not make it work incorrectly.

dmigwi: The same issue I was trying to fix exists on alpha: https://alpha.dcrdata.org/charts?chart=ticket-price&zoom=juzfs5nk-jw6wfbfk. This zoom is set to the month, but the month button isn't activated or page hard reloading.

Try it on your end using alpha:
1. Select any zoom button of your choice and copy the chart URL.
2. Open a brand new tab and load the chart URL you just copied. Let me know if the specific zoom button you selected earlier is activated on the new tab.

This is what my `mapKey` implementation is supposed to fix.

buck54321: I think I might see what you're saying. So even though the URL ` https://alpha.dcrdata.org/charts?chart=ticket-price&zoom=month` will set the range and selected button correctly, an encoded zoom range like `zoom=juzfs5nk-jw6wfbfk`, which should correspond to a month, only sets the range correctly but not the selected button. 

Does that sound right?

dmigwi: Thats it.

buck54321: So there are a couple of subtleties here. 

When you give somebody a URL for the charts page, it should always bring them to exactly the view that's on the chart when you copied the URL (which is why the keyword is immediately replaced with an encoded range). As mentioned before, the keywords correspond to only the most recent window. So if the *range* is say a month, but the end of the range is not at the rightmost point of the available data, the button is not shown as selected. That means that an encoded-range URL that corresponds to a valid button window will only set the button selection if used before the next block is found, at which point it's rightmost point is no longer at the extent of the available data.

It is true that even if you immediately re-paste the URL with encoded range it does not set the selected button correctly. It's certainly reasonable to want to fix that, though again, it only matters til the next block is found anyway. It's also reasonable to want show the button as selected based only on range and not on whether it is the most recent possible window of that duration. Neither of those solutions requires any new parameters or special case handling in the `mapKey` function.

How about you use your range matching logic, but with a couple tweaks as below. 

1. Default value needs to be `null`. 
1. Special check for `all`
```
  static mapKey (zoom, limits, scale) {
    scale = scale || 1
    let lims = tryDecode(limits)
    let decoded = this.decode(zoom, lims)
    let range = decoded.end - decoded.start
    if (range === lims.end - lims.start) return 'all'
    for (let k in zoomMap) {
      let v = zoomMap[k]
      if (v / scale === range) return k
    }
    return null
  }
```

You will need to remove the `if (!option) return` line for it to work correctly since `null` is a valid value. 

If after those changes, it still doesn't select the button correctly, even if you paste the URL before the next block, then it will probably just be a math error somewhere and we can chase that down.

dmigwi: This should work, I will be testing that mapKey implementation in a few. Its looks just like the compromise we set out to achieve.

buck54321: I do like this change. It'll be a nice indicator that the window hasn't changed size.

PR #1340, opened 2019-05-18
debug exchanges

buck54321:
This PR is for testing. The exchanges package seems to have some issues lately that may or may not be related to recent work to implement websockets. This PR effectively just adds some trace output, though the foundations are being laid for some more sophisticated error monitoring. 

If you are testing, run this PR with `--debuglevel=XBOT=trace` on dcrdata, and `--loglevel=trace` on rateserver. Feel free to give me your logs.

### DIFF
--- a/exchanges/exchanges.go
+++ b/exchanges/exchanges.go
@@ -574,6 +574,7 @@ func (xc *CommonExchange) connectWebsocket(processor WebsocketProcessor, cfg *so
 	}
 
 	xc.wsMtx.Lock()
+	// Ensure that any previous websocket is closed.
 	if xc.ws != nil {
 		select {
 		case <-xc.ws.Done():
@@ -634,6 +635,10 @@ func (xc *CommonExchange) setWsFail(err error) {
 	sr := xc.signalr()
 	if sr != nil && !xc.wsFailed() {
 		sr.Close()
+	}
+	ws, _ := xc.websocket()
+	if ws != nil {
+		ws.Close()
 	}
 	xc.wsMtx.Lock()
 	defer xc.wsMtx.Unlock()
@@ -761,7 +766,7 @@ func (xc *CommonExchange) wsDepths() *DepthData {
 func (xc *CommonExchange) wsDepthStatus(connector func()) (tryHttp, initializing bool, depth *DepthData) {
 	if !xc.wsListening() {
 		if xc.wsFailed() {
-			log.Tracef("using http fallback for bittrex orderbook data")
+			log.Tracef("using http fallback for %s orderbook data", xc.token)
 			tryHttp = true
 			errCount := xc.wsErrorCount()
 			if errCount < wsMaxErrors {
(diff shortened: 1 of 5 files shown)


exchanges/exchanges_test.go
```diff
@@ -283,7 +298,11 @@ func (p *fakePoloniexWebsocket) Write(interface{}) error {
 }
 
 func (p *fakePoloniexWebsocket) Close() {
-	close(poloniexDoneChannel)
+	select {
+	case <-poloniexDoneChannel:
+	default:
+		close(poloniexDoneChannel)
```
chappjc: Is Close() possibly called in different places?  I'm trying to understand why this select is needed.  But it is a race since the channel could be closed by another goroutine if Close() is called elsewhere (they could both end up at the `close` statement and one would panic)

exchanges/websocket.go
```diff
@@ -77,7 +77,14 @@ func (client *socketClient) Done() chan struct{} {
 
 // Close is wrapper for gorilla's Close that satisfies websocketFeed.Close.
 func (client *socketClient) Close() {
+	client.mtx.Lock()
+	defer client.mtx.Unlock()
 	client.conn.Close()
+	select {
+	case <-client.done:
+	default:
+		close(client.done)
```
chappjc: Same comment as with the select in the test file.

chappjc: BTW, `IsOpen` does not race per se, but the caller is likely racing.

exchanges/exchanges_test.go
buck54321: I think it's unlikely during the test, but easy to add some sequencing.

exchanges/exchanges.go
```diff
@@ -614,29 +610,21 @@ func (xc *CommonExchange) wsSend(msg interface{}) error {
...
-		sr := xc.signalr()
-		if sr == nil {
-			return false
-		}
-		return sr.On()
-	}
-	return ws.On()
+	xc.wsMtx.RLock()
+	defer xc.wsMtx.RUnlock()
+	return xc.wsSync.init.After(xc.wsSync.fail)
```
buck54321: Should have been this way already.

exchanges/exchanges_test.go
```diff
@@ -27,157 +23,6 @@ func enableTestLog() {
...
-	killSwitch := make(chan os.Signal, 1)
-	signal.Notify(killSwitch, os.Interrupt)
-	return killSwitch
-}
-
-func testExchanges(asSlave, quickTest bool, t *testing.T) {
-	enableTestLog()
-
-	// Skip this test during automated testing.
-	if os.Getenv("GORACE") != "" {
```
buck54321: This was a sloppy way to skip this test during CI testing, but it's not compatible with the changes in #1347.

chappjc: Is it because the test takes too much time that you wanted to skip it?  We can leverage the `-short` switch.  https://golang.org/pkg/testing/#hdr-Skipping e.g.

```go
if !testing.Short() {
    // do long tests
...
}
```

chappjc: Or use `t.Skip` above I suppose.

buck54321: Yeah. All the live tests take a long time. I put a tag at the top of the _live file, so you have to enable it e.g. `go test -tags-livexc`.

chappjc: Oh, perfect.

buck54321:
@chappjc I think I've addressed the concerns you've raised. Before I remove the draft status, could you check out those areas and give me a thumbs up or down.

chappjc:
Great, thanks! I'll go though it this evening.

Issue #1287, opened 2019-04-18
Proposals page: add voting outcome data


RichardRed0x:
I have been using the [proposals page ](https://alpha.dcrdata.org/proposals) to find proposal charts, if it had information about proposal voting outcomes (numbers of votes, approval % etc,) it would be a very useful page.

buck54321:
I'd like to see the /proposals and /agendas page come together into a single governance portal different than but closer to the design of voting.decred.org. Likewise, the /agenda and /proposal pages could either be reduced to a single page, or at least adopt some common styling. 

If I don't hear any objections to that plan, I'm going to start working on it as early as this weekend.

chappjc:
Thank would be great for sure.  Go for it!

buck54321:
Your understanding in #1 is correct, but not the solution we need right now. Let's just get the approval % and vote count in the table on  `/proposals` for now. Don't remove it from `/proposal/{token}`. The information is arguably the most important figure associated with the proposal, so it deserves some space in the `/proposals` table too. 

I think that data is already available in the `ProposalInfo` (embedded from `ProposalVotes`) from `gov/politeia/types` being used [in `proposals.tmpl` here.] (https://github.com/decred/dcrdata/blob/71bb36a72193bb61f8bd4882d14d5d06a0aaa1ff/views/proposals.tmpl#L116) Just need to add the columns.

njirap:
ok got it..Thanks

njirap:
@buck54321 is the next iteration of this issue to have `/agendas` and `/proposals` as one page?

buck54321:
That is the plan. I imagine that being a pretty large undertaking that will also involve some consolidation of types from the `gov` packages. That's not to say that you can't work on it. Please do. I just want to make sure that we're attacking it from the best angle. 

For a new page, it might be good to work together on some kind of wireframe or prototype before we get too deep. Can anybody recommend a good open-source mockup/prototyping tool? 

@njirap, have you worked with bootstrap CSS? If not, it might be an informative project for you to make the table on `/proposals` mobile-friendly, which will require you to use the Bootstrap and make some hard decisions about how to best display things in the table. The work would carry over to the combined page as well. We could collab on a mockup while making incremental improvements to page elements before finally combining things.

[...]

PR #1341, opened 2019-05-20
ticketpool: Presents the ticket pool PercentTarget as a diff from 100% target

### LINKED ISSUE #1225: home: ticket pool size diff from target percentage
Instead of

`Ticket Pool Size: 41,285 (100.79% of target 40,960)`

show

`Ticket Pool Size: 41,285 (+0.79% of target 40,960)`

it takes less space, faster to read, and in case of negative number -1% is more useful than 99% (imo)

edit: if there are other places than home page where this is shown, update them too.

njirap:
Resolves #1225 

Instead of

`Ticket Pool Size: 41,285 (100.79% of target 40,960)`

shows

`Ticket Pool Size: 41,285 (+0.79% of target 40,960)`

It takes less space, faster to read, and in case of negative number -1% is more useful than 99%

Screen Shot 2019-05-20 at 12 18 28 PM

### DIFF
--- a/public/js/controllers/homepage_controller.js
+++ b/public/js/controllers/homepage_controller.js
@@ -173,7 +173,7 @@ export default class extends Controller {
     this.windowIndexTarget.textContent = ex.window_idx
     this.posBarTarget.style.width = `${(ex.window_idx / ex.params.window_size) * 100}%`
     this.poolSizeTarget.innerHTML = humanize.decimalParts(ex.pool_info.size, true, 0)
-    this.targetPctTarget.textContent = parseFloat(ex.pool_info.percent_target).toFixed(2)
+    this.targetPctTarget.textContent = parseFloat(ex.pool_info.percent_target - 100).toFixed(2)
     this.rewardIdxTarget.textContent = ex.reward_idx
     this.powBarTarget.style.width = `${(ex.reward_idx / ex.params.reward_window_size) * 100}%`
     this.poolValueTarget.innerHTML = humanize.decimalParts(ex.pool_info.value, true, 0)
--- a/views/home.tmpl
+++ b/views/home.tmpl
@@ -245,7 +245,9 @@
                                 </span>
                             </div>
                             <div class="fs12 lh1rem text-black-50">
-                                <span data-target="homepage.targetPct">{{printf "%.2f" .PoolInfo.PercentTarget}}</span> % of target&nbsp;<span>{{intComma .PoolInfo.Target}}</span>
+                                <span data-target="homepage.targetPct">{{printf "%.2f" (toAbsValue $.PercentChange)}}</span>%
+                                {{if lt $.PercentChange 0.0}} under {{else}} over {{end}}
+                                target
                             </div>
                         </div>
                         <div class="col-12 mb-3 mb-sm-2 mb-md-3 mb-lg-3">
--- a/explorer/explorerroutes.go
+++ b/explorer/explorerroutes.go
@@ -180,13 +180,14 @@ func (exp *explorerUI) Home(w http.ResponseWriter, r *http.Request) {
 
 	str, err := exp.templates.execTemplateToString("home", struct {
 		*CommonPageData
-		Info        *types.HomeInfo
-		Mempool     *types.MempoolInfo
-		BestBlock   *types.BlockBasic
-		BlockTally  []int
-		Consensus   int
-		Blocks      []*types.BlockBasic
-		Conversions *homeConversions
+		Info          *types.HomeInfo
+		Mempool       *types.MempoolInfo
+		BestBlock     *types.BlockBasic
+		BlockTally    []int
+		Consensus     int
+		Blocks        []*types.BlockBasic
+		Conversions   *homeConversions
+		PercentChange float64
 	}{
 		CommonPageData: exp.commonData(r),
 		Info:           homeInfo,
@@ -196,6 +197,7 @@ func (exp *explorerUI) Home(w http.ResponseWriter, r *http.Request) {
 		Consensus:      consensus,
 		Blocks:         blocks,
 		Conversions:    conversions,
+		PercentChange:  homeInfo.PoolInfo.PercentTarget - 100,
 	})
 
 	inv.RUnlock()
--- a/explorer/templates.go
+++ b/explorer/templates.go
@@ -569,5 +569,6 @@ func makeTemplateFuncMap(params *chaincfg.Params) template.FuncMap {
 			}
 			return addrPKH.EncodeAddress()
 		},
+		"toAbsValue": math.Abs,
 	}
 }


explorer/explorer.go
```diff
@@ -505,7 +505,7 @@ func (exp *explorerUI) Store(blockData *blockdata.BlockData, msgBlock *wire.MsgB
 			Value:         blockData.PoolInfo.Value,
 			ValAvg:        blockData.PoolInfo.ValAvg,
 			Percentage:    stakePerc * 100,
-			PercentTarget: 100 * float64(blockData.PoolInfo.Size) / float64(tpTarget),
+			PercentTarget: 100 * (float64(blockData.PoolInfo.Size)/float64(tpTarget) - 1),
```
buck54321: See note below about changing the implementation. While it seems less important here because there is no websocket signal that uses this value, it is still semantically questionable. 

I see two ways to handle this. 

1. Add a `PercentChange` field to the `HomeInfo` struct, and populate accordingly. 
2. Just do the arithmetic for a new field in the `explorer.Home` template struct around [here](https://github.com/decred/dcrdata/blob/5ceb56b5c8e1ed53b6b92b5deee6e2a05903f61f/explorer/explorerroutes.go#L189)

I prefer the second way.

views/home.tmpl
```diff
@@ -245,7 +245,7 @@
                                 </span>
                             </div>
                             <div class="fs12 lh1rem text-black-50">
-                                <span data-target="homepage.targetPct">{{printf "%.2f" .PoolInfo.PercentTarget}}</span> % of target&nbsp;<span>{{intComma .PoolInfo.Target}}</span>
+                                <span data-target="homepage.targetPct">{{printf "%+.2f" .PoolInfo.PercentTarget}}</span> % of target&nbsp;<span>{{intComma .PoolInfo.Target}}</span>
```
buck54321: I think this is fine as is, so don't feel the need to change anything, but as another option, instead of `+/-x% of target 40960`, it could instead say

`x% over target` or `x% under target`

which I think reads better and is more compact.

njirap:
Thanks working on the changes now

njirap:
@buck54321 changes made this is how it looks now

Screen Shot 2019-05-20 at 12 18 28 PM

views/home.tmpl
```diff
@@ -1,5 +1,6 @@
 {{define "home"}}
 {{$conv := .Conversions}}
+{{$percentchange := .PercentChange}}
```
chappjc: You can eliminate this line and use `$.PercentChange` anywhere in the document.

views/home.tmpl
```diff
@@ -245,7 +246,9 @@
                                 </span>
                             </div>
                             <div class="fs12 lh1rem text-black-50">
-                                <span data-target="homepage.targetPct">{{printf "%.2f" .PoolInfo.PercentTarget}}</span> % of target&nbsp;<span>{{intComma .PoolInfo.Target}}</span>
+                                <span data-target="homepage.targetPct">{{printf "%.2f" $percentchange}}</span>
+                                % {{if lt $percentchange 0.0}} under {{else}} over {{end}} a target of&nbsp;
```
chappjc: The space between the percent sign and the number is unusual.  Instead of `0.5 %`, I believe this should read `0.5%`.

chappjc: `over a target of 40960` might be better as `over target (40,960)` (or something without "a").

njirap:
@chappjc done. this is the new look

Screen Shot 2019-05-20 at 8 10 17 PM

views/home.tmpl
buck54321: I would personally prefer to drop the value altogether. Just `x% over/under target`. The value of the target can be found in the appropriate place somebody is interested.

views/home.tmpl
```diff
@@ -245,7 +245,9 @@
                                 </span>
                             </div>
                             <div class="fs12 lh1rem text-black-50">
-                                <span data-target="homepage.targetPct">{{printf "%.2f" .PoolInfo.PercentTarget}}</span> % of target&nbsp;<span>{{intComma .PoolInfo.Target}}</span>
+                                <span data-target="homepage.targetPct">{{printf "%.2f" $.PercentChange}}%</span>
+                                {{if lt $.PercentChange 0.0}} under {{else}} over {{end}}target&nbsp;
```
dmigwi: Since this `x% over/under target` is the case going forward then on the UI the value `%x` should always be positive.
I think part of this implementation could be moved to the golang side so it can be easy to use `math.Abs(` function. 
 Alternatively the `math.Abs(` implementation could be added to the template.go file and invoked here.

dmigwi: Actually in `func makeTemplateFuncMap(params *chaincfg.Params) template.FuncMap {` in template.go adding 
```go
"ToAbsValue": math.Abs,
```
then doing this:
should fix the issue,

` <span data-target="homepage.targetPct">{{printf "%.2f" ToAbsValue($.PercentChange)}}%</span>`

buck54321: @dmigwi Just to be sure I understand, you're saying to just generate the entire string in `explorer.Home`, rather than trying to parse things in the template?

dmigwi: That was option 1, the option 2 was to add a `math.Abs` function in the template.go file.

buck54321: Oh. Sorry. Didn't see your second comment before. Yeah, I think either of those would be fine.

njirap: @buck54321 I understand this changing the current

`-0.3% under target` to `0.3% under target` hence the comment by @dmigwi to make the `%x`

an absolute

buck54321: That's sounds right.

njirap:
@dmigwi the change is now being absoluted. Not sure how to demonstrate a negative change though since I'm working with live data

Also @buck54321 the target number is now gone

Screen Shot 2019-05-20 at 11 33 46 PM

buck54321:
Looks like you'll need a `-100` at https://github.com/decred/dcrdata/blob/5ceb56b5c8e1ed53b6b92b5deee6e2a05903f61f/public/js/controllers/homepage_controller.js#L176
too

njirap:
@buck54321 done

views/home.tmpl
```diff
@@ -245,7 +245,9 @@
                                 </span>
                             </div>
                             <div class="fs12 lh1rem text-black-50">
-                                <span data-target="homepage.targetPct">{{printf "%.2f" .PoolInfo.PercentTarget}}</span> % of target&nbsp;<span>{{intComma .PoolInfo.Target}}</span>
+                                <span data-target="homepage.targetPct">{{printf "%.2f" (toAbsValue $.PercentChange)}}%</span>
```
buck54321: The percent sign needs to be outside of the `data-target="homepage.targetPct"` span, otherwise it is lost when the homepage controller overwrites the value when new data is received by websocket. Here is what it looks like after a block comes in

image


The percent sign is lost.

njirap: on it

njirap: this is done

public/js/controllers/homepage_controller.js
```diff
@@ -173,7 +173,7 @@ export default class extends Controller {
     this.windowIndexTarget.textContent = ex.window_idx
     this.posBarTarget.style.width = `${(ex.window_idx / ex.params.window_size) * 100}%`
     this.poolSizeTarget.innerHTML = humanize.decimalParts(ex.pool_info.size, true, 0)
-    this.targetPctTarget.textContent = parseFloat(ex.pool_info.percent_target).toFixed(2)
+    this.targetPctTarget.textContent = parseFloat(ex.pool_info.percent_target - 100).toFixed(2)
```
dmigwi: @njirap You might need to add the `%` at the end of this line.

dmigwi: This is probably what buck meant.

buck54321: She fixed it in the template, which is better

njirap:
@buck54321 Ok I'll look into #1287

PR #1343, opened 2019-05-20
ticketpool: Count only sstxcommitments outputs

dmigwi:
Since counting the `sstxcommitments` outputs is simpler than counting all outputs of a given ticket, the `sstxcommitment` tx count are used to represent the output count for a given ticket in the ticketpool.

**New Tickets Output Distribution Table**
Screenshot from 2019-05-21 02-06-06

### DIFF
--- a/db/dcrpg/queries.go
+++ b/db/dcrpg/queries.go
@@ -1094,7 +1094,10 @@ func retrieveTicketsGroupedByType(ctx context.Context, db *sql.DB) (*dbtypes.Poo
 		}
 
 		tickets.Count = append(tickets.Count, count)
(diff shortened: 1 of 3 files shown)


chappjc:
That's pretty  much what I had in mind, except to be clearer, the first column would need to say "# of `sstxcommitment` outputs" (to be technical).  But perhaps something in plain language like "reward addresses"

dmigwi:
Table update:
Screenshot from 2019-05-21 19-46-37

chappjc:
Works for me.  If "# of sstxcommitment outputs" is too mystifying for the average user, it can be reworded in the future to something like reward addresses. @buck54321 @raedah Opinions?

PR #1348, opened 2019-05-22
proposals: add votes count and approval percentage columns

### LINKED ISSUE #1287: Proposals page: add voting outcome data

njirap:
Partiallyresolves(editing this so that the main issue is not closed) #1287 

add votes count and approval percentage columns in the proposals view

Screen Shot 2019-05-22 at 9 43 35 PM

### DIFF
--- a/explorer/templates.go
+++ b/explorer/templates.go
@@ -570,5 +570,15 @@ func makeTemplateFuncMap(params *chaincfg.Params) template.FuncMap {
 			return addrPKH.EncodeAddress()
 		},
 		"toAbsValue": math.Abs,
+		"toFloat64": func(x uint32) float64 {
+			return float64(x)
+		},
+		"toInt": func(str string) int {
+			intStr, err := strconv.Atoi(str)
+			if err != nil {
+				return 0
+			}
+			return intStr
+		},
 	}
 }
--- a/views/proposals.tmpl
+++ b/views/proposals.tmpl
@@ -106,8 +106,8 @@
                     <tr>
                         <th class="text-left">Title</th>
                         <th class="text-left">Author</th>
-                        <th class="text-left">Proposal Status</th>
-                        <th class="text-left">Vote Status</th>
+                        <th class="text-left">Proposal Vote Status</th>
+                        <th class="text-left">Vote Count</th>
                         <th class="text-right">Age</th>
                         <th class="text-right">Updated</th>
                     </tr>
@@ -118,12 +118,43 @@
                     <tr>
                         <td class="text-left"><a href="/proposal/{{.RefID}}">{{.Name}}</a></td>
                         <td class="text-left">{{.Username}}</td>
-                        {{with .Status}}
-                            <td class="text-left">{{toTitleCase .String}}</td>
-                        {{end}}
                         <td class="text-left">
-                            {{if .VoteStatus}}{{with .VoteStatus}}{{.ShortDesc}}{{end}}{{else}}N/A{{end}}
+                            {{if .VoteStatus}}
+                                {{if eq .VoteStatus.ShortDesc "Not Authorized"}}
+                                    <span class="text-abandoned">
+                                        In Discussion
+                                    </span>
+                                {{else}}
+                                    {{range $i, $vr := .VoteResults}}
+                                        {{if eq $vr.Option.OptionID "yes"}}
+                                            {{$votesPercent :=  percentage $vr.VotesReceived $v.TotalVotes}}
+                                            {{if lt $.Tip.Height (toInt $v.Endheight)}}
+                                                <span class="text-progress">
+                                                    In Progress ({{printf "%.0f" ($votesPercent)}}%)
+                                                </span>
+                                            {{else if lt (percentage $v.TotalVotes $v.NumOfEligibleVotes) (toFloat64 $v.QuorumPercentage)}}
+                                                <span class="text-no-quorum">No quorum</span>
+                                            {{else}}
+                                                {{if lt $votesPercent (toFloat64 $v.PassPercentage)}}
+                                                    <span class="text-failed">
+                                                        Failed ({{printf "%.0f" ($votesPercent)}}%)
+                                                    </span>
+                                                {{else}}
+                                                    <span class="text-green">
+                                                        Passed ({{printf "%.0f" ($votesPercent)}}%)
+                                                    </span>
+                                                {{end}}
+                                            {{end}}
+                                        {{end}}
+                                    {{end}}
+                                {{end}}
+                            {{else}}
+                                <span class="text-abandoned">
+                                    Abandoned
+                                </span>
+                            {{end}}
                         </td>
+                        <td class="text-left">{{.TotalVotes}}</td>
                         <td class="text-right" data-target="time.age" data-age="{{.Timestamp}}"></td>
                         <td class="text-right">{{timeWithoutDateAndTimeZone .Timestamp}}</td>
                     </tr>
--- a/public/scss/themes.scss
+++ b/public/scss/themes.scss
@@ -183,4 +183,8 @@ body.darkBG {
   .text-green {
     color: #2eefa1;
   }
+
+  .text-abandoned {
+    color: #c4cbd2;
+  }
 }
--- a/public/scss/typography.scss
+++ b/public/scss/typography.scss
@@ -159,6 +159,22 @@ a.grayed {
   color: #00a100;
 }
 
+.text-progress {
+  color: #2970ff;
+}
+
+.text-no-quorum {
+  color: #feb8a5;
+}
+
+.text-failed {
+  color: #ed6d47;
+}
+
+.text-abandoned {
+  color: #3d5873;
+}
+
 // numbers
 .decimal-parts {
   .int {


views/proposals.tmpl
```diff
@@ -108,13 +108,15 @@
...
+                        <th class="text-left">Approval Percentage</th>
                         <th class="text-right">Age</th>
                         <th class="text-right">Updated</th>
                     </tr>
                 </thead>
                 <tbody data-controller="time">
-                {{range $i, $v := .Proposals}}
-                {{with $v}}
+                {{range $i, $p := .Proposals}}
+                {{with $p}}
```
njirap: hmm just seen this is not necessary

chappjc:
This is a step in the right direction, but the specific outcome needs to be shown (instead of just Finished, it needs to say if it was approved/passed or disapproved/failed).  Although the approval percentage is now a column, the outcome depends on whether quorum was met and what the threshold for passing is.

njirap:
@chappjc do you mean we should populate the `Vote Status` column with the specific outcome or add another column altogether?

chappjc:
I would prefer not to add another column, so unless changing the Vote Status column this way creates ambiguity, let's put the specific outcome in Vote Status.

njirap:
ok..

views/proposals.tmpl
```diff
@@ -108,6 +108,8 @@
                         <th class="text-left">Author</th>
                         <th class="text-left">Proposal Status</th>
                         <th class="text-left">Vote Status</th>
+                        <th class="text-left">Votes Count</th>
+                        <th class="text-left">Approval Percentage</th>
```
chappjc: This column can just be "Approval" since there's a "%" with each value.  In general, it's best to keep the table width as narrow as possible.

njirap:
@chappjc Can we get the `approved/disapproved` status by checking if the approval is >/< 60%? Or do we have to pull the `approved/disapproved` from politeia?

chappjc:
Have a look at `explorer.(*explorerUI).ProposalPage` and and the `VoteResults` loop in proposal.tmpl.

dmigwi:
This was the suggested format:
https://matrix.to/#/!LmKzrmxJIXNHNiEmIh:decred.org/$155482664511711QpoJf:decred.org?via=decred.org&via=matrix.org&via=zettaport.com
Screenshot from 2019-05-22 23-52-52

Merging the two status columns probably shouldn't be done now.

buck54321:
> Merging the two status columns probably shouldn't be done now.

Why can't we do this? Seems like a great idea to me.

njirap:
This is what I currently have
Screen Shot 2019-05-23 at 9 56 15 PM

views/proposals.tmpl
```diff
@@ -118,14 +120,28 @@
...
-                        {{with .Status}}
-                            <td class="text-left">{{toTitleCase .String}}</td>
-                        {{end}}
                         <td class="text-left">
-                            {{if .VoteStatus}}{{with .VoteStatus}}{{.ShortDesc}}{{end}}{{else}}N/A{{end}}
+                            {{if .VoteStatus}}
+                                {{range $i, $vr := .VoteResults}}
+                                    {{if eq $vr.Option.OptionID "yes"}}
+                                        {{$percent :=  percentage $vr.VotesReceived $v.TotalVotes}}
+                                        {{if lt $percent 60.0}}
```
njirap: This 60.0 should probably be the `PassPercentage` field in `ProposalInfo`

views/proposals.tmpl
```diff
@@ -118,14 +120,28 @@
...
+                                            <span style="color:#ED6D47;">Failed
+                                        {{else}}
+                                            <span style="color:#2DD8A3;">Passed
+                                        {{end}}
+                                        ({{printf "%.0f" ($percent)}}%)</span>
```
njirap: the way the close of the `span` intertwines with the `if` statement feels funny but it has produced the desired effect

views/proposals.tmpl
njirap: there's an error about non matching data types with using the `PassPercentage`

views/proposals.tmpl
buck54321: I agree that its a little funny looking, but it works. If you can improve the readability, feel free.

views/proposals.tmpl
buck54321: Yeah. It is a little strange that `PassPercentage` is a `uint32` instead of a float. I guess the way you have it is best.

buck54321:
I'll get you a proper review later today, but in the meantime there is still one issue I'm not certain about. Perhaps @dmigwi or @chappjc can answer. 

On Politeia, a 20% quorum must be achieved for the results to be accepted, i.e. 20% of all eligible on-chain tickets must submit a vote of yes or no through Politeia. For on-chain voting, if quorum is not achieved, I believe the vote is restarted immediately. I'm not sure what happens in Politeia.

views/proposals.tmpl
dmigwi: You could convert the value obtained from `PassPercentage` and pass into the template in whatever type you want. IMO changing `PassPercentage` value type now might not be worth the effort.

dmigwi: Should you decide to pursue this:
> IMO changing PassPercentage value type now might not be worth the effort.

Just change the `PassPercentage` value type all through the code and increase the [semver version](https://github.com/decred/dcrdata/blob/master/gov/politeia/proposals.go#L32) to maybe v1.0.1.
That should fix it just as you need it to be.

views/proposals.tmpl
dmigwi: You need to add the `InProgress(% approval)` option whereby `proposalInfo.ProposalVotes.Endheight` will help know the last block when voting should be complete after which `Failed(% approval)` or `Passed(% approval)` status can be displayed.

views/proposals.tmpl
njirap: @dmigwi I actually prefer changing the data type to hard coding the `60.0`, that is assuming there's no external party who may be consuming it as a `uint32`

dmigwi: I am not aware of any external party that could be affected as now.

chappjc: Changing the data type of an exported field would bump the major version since it's a breaking change.

I'll ask in the Politeia matrix channel why this is uint32 (https://github.com/decred/politeia/blob/master/politeiawww/api/www/v1/api.md#proposal-vote-status), but only out of curiosity.

Solving this is doable without changing the data type of `gov/politeia/types.ProposalVotes.PassPercentage` or hard coding the value in the template.  I suggest following the example in templates.go of `"int64": toInt64`.  That is, make a `"float64": toFloat64,` entry and then define `toFloat64`.  Except don't make a variable that's a function like was done for `toInt64`; just define a regular function and refer to it by name.

buck54321:
If a proposal does not achieve quorum, it is considered failed, regardless of the approval rate. You can maybe mark this condition, `$v.QuorumPercentage < 20`, with `No quorum`, without an approval rate. 

Add the quorum logic, and dmigwi's *In Progress` status when `$.Tip.Height > $v.EndHeight`, and I think we're in business.

views/proposals.tmpl
```diff
@@ -118,12 +118,35 @@
...
-                        {{with .Status}}
-                            <td class="text-left">{{toTitleCase .String}}</td>
-                        {{end}}
                         <td class="text-left">
-                            {{if .VoteStatus}}{{with .VoteStatus}}{{.ShortDesc}}{{end}}{{else}}N/A{{end}}
+                            {{if .VoteStatus}}
+                                {{range $i, $vr := .VoteResults}}
+                                    {{if eq $vr.Option.OptionID "yes"}}
+                                        {{$votesPercent :=  percentage $vr.VotesReceived $v.TotalVotes}}
+                                        {{if lt $.Tip.Height $v.Endheight}}
```
njirap: the` Endheight` should be cast from string. I will be doing that later today.

buck54321: That is annoying. Ugh. Should def be a number already.

njirap:
`No quorum` works.

I request for the colors hex codes to be used for `In Progress`, `No quorum` and `Abandoned`..for now they are the red `#ED6D47`

Screen Shot 2019-05-26 at 3 55 31 PM

buck54321:
I swear I'm not trying to make things complicated, but I just realized that there needs to be a differentiation between votes that failed due to `No quorum`, and votes that are live but haven't yet achieved quorum. For live votes, `.VoteStatus.ShortDesc() == "Started"`, the approval rate needs to be shown regardless of quorum status. 

re: colors. You can and should use colors from the Decred palette whenever you can make them work, so your red is good. There is also a `text-green` class that has good contrast in both light and dark mode.

image

njirap:
@buck54321 votes that are live are captured by `{{if lt $.Tip.Height (toInt $v.Endheight)}}` regardless of the quorum status...by the time the `else` that checks for quorum checks in we are already dealing with votes that have ended. So I think the No-quorum-lstill-live scenario is already captured by the check for votes that are live

buck54321:
I'll take a look, but that first proposal, `A Journey to the Future ...` is live right now, but showing as `No quorum` in your screenshot.

njirap:
Interesting..

njirap:
On logging the two values `.Tip.Height` and `v.Endheight` I confirm that `.Tip.Height` is indeed lesser than the end height yet it are not noted as `In Progress`..I'm investigating this but I suspect it's the casting function I have that's casting the `Endheight` to int

njirap:
I've seen it's the `if` statement that doesn't work

explorer/templates.go
```diff
@@ -570,5 +570,15 @@ func makeTemplateFuncMap(params *chaincfg.Params) template.FuncMap {
 			return addrPKH.EncodeAddress()
 		},
 		"toAbsValue": math.Abs,
+		"toFloat64": func(x uint32) float64 {
+			return float64(x)
+		},
+		"toInt": func(str string) int {
+			intStr, err := strconv.Atoi(str)
+			if err != nil {
```
njirap: this is where the bug is.. resolving it now

njirap:
@buck54321 bug terminated. and below is how the colors look in both dark and light

njirap:
@buck54321 bug terminated. and below is how the colors look in both dark and light

Screen Shot 2019-05-26 at 6 01 47 PM
Screen Shot 2019-05-26 at 6 03 20 PM

chappjc:
One last thing from me.  A space is needed between the text and the parenthesized percentage ("Passed (63%)")

njirap:
@chappjc Done.

Screen Shot 2019-05-26 at 6 13 06 PM

views/proposals.tmpl
```diff
@@ -106,8 +106,8 @@
                     <tr>
                         <th class="text-left">Title</th>
                         <th class="text-left">Author</th>
-                        <th class="text-left">Proposal Status</th>
-                        <th class="text-left">Vote Status</th>
+                        <th class="text-left">Proposal Vote Status</th>
+                        <th class="text-left">Votes Count</th>
```
buck54321: Probably `Vote Count`

views/proposals.tmpl
```diff
@@ -118,12 +118,38 @@
...
-                            <td class="text-left">{{toTitleCase .String}}</td>
-                        {{end}}
                         <td class="text-left">
-                            {{if .VoteStatus}}{{with .VoteStatus}}{{.ShortDesc}}{{end}}{{else}}N/A{{end}}
+                            {{if .VoteStatus}}
+                                {{range $i, $vr := .VoteResults}}
+                                    {{if eq $vr.Option.OptionID "yes"}}
+                                        {{$votesPercent :=  percentage $vr.VotesReceived $v.TotalVotes}}
+                                        {{if lt $.Tip.Height (toInt $v.Endheight)}}
+                                            <span style="color:#2970ff;">
```
buck54321: As a general rule, no inline styles in the templates. Some of these text colors may already exists as a class defintion, otherwise they should be created, New text coloring classes can be defined in `public/scss/typograghy.scss`. You can also add a `body.darkBG` descendant selector in `themes.scss` if you need. See the class `text-green` as an example. 

I would recommend having a dark mode color for the `Abandoned` proposals, as the look isn't great with `#091440` in both modes.

njirap: I'm working on the styling now

njirap:
2 bugs:
1. Only `text-green` css class works. I have been scratching my head on what I could be missing to make the new classes work
2. 1 proposal is currently `In discussion` on politeia but it's not marked by any status. What condition do we check for `In discussion`?

Screen Shot 2019-05-29 at 4 20 34 AM
Screen Shot 2019-05-29 at 4 20 57 AM

chappjc:
Sorry if this is already clear, but have you repacked the static assets (`npm run build` as in dev/build.sh`)?

njirap:
Actually the problem was saved cache..It works in incognito and when cache is disabled, sorry for the alarm

njirap:
`Abandoned` now has dark and light modes

Screen Shot 2019-05-29 at 9 59 46 AM
Screen Shot 2019-05-29 at 10 00 04 AM

njirap:
I'm getting the `In discussion` status in now

njirap:
`In discussion` now works

Screen Shot 2019-05-29 at 11 13 14 AM

explorer/templates.go
njirap: this is resolved

views/proposals.tmpl
```diff
@@ -118,12 +118,43 @@
                     <tr>
                         <td class="text-left"><a href="/proposal/{{.RefID}}">{{.Name}}</a></td>
                         <td class="text-left">{{.Username}}</td>
-                        {{with .Status}}
-                            <td class="text-left">{{toTitleCase .String}}</td>
-                        {{end}}
                         <td class="text-left">
-                            {{if .VoteStatus}}{{with .VoteStatus}}{{.ShortDesc}}{{end}}{{else}}N/A{{end}}
+                            {{if .VoteStatus}}
+                                {{if eq .VoteStatus.ShortDesc "Not Authorized"}}
```
njirap: I'm open to any other better way of achieving this

buck54321: That's perfect

PR #1349, opened 2019-05-23
db/{cache,dcrpg}: improved memory management

chappjc:
### In-flight address UTXOs

The Insight API's `/addr/{address}/utxo` endpoint can consume excessive memory
for addresses with many UTXOs.  This imposes a limit on the number of "in-flight"
UTXOs to restrict how much memory can be consumed at the request of a client.

The `.../utxo` endpoint is handed by `(*InsightApi).getAddressesTxnOutput`, to which
these edits add in-flight UTXO counting and limiting.  To do this, a new field is added to
`InsightApi` called `inflightUTXOs` that tracks how many UTXOs are currently loaded
from DB queries by the `getAddressesTxnOutput` method.

Limiting is implemented by:
1. Restricting the number of address UTXO queries to one (1) concurrent request with a `Mutex`.
2. Atomically updating the `inflightUTXOs` counter while the lock is still held.
3. Checking if the counter is over a global limit.
   - If the counter is over the limit, hold the lock until the http handler (including JSON
   marshaling) completes. This handler call has gained "priority" status.
   - If the counter is not over the limit, release the lock for other concurrent requests.
4. When the http handler completes, decrement the total in-flight UTXOs by the
  number processed by the returning handler.

### Cache memory management

* Create a new `dbtypes.AddressTxnOutput` type for caching instead of using the `api/types.AddressTxnOutput` type that is very inefficient in several respects. Add the `RetrieveAddressDbUTXOs` function for returning an address' UTXOs in this type.
* Change the cached address rows and UTXOs to slices of pointers to structs instead of slices of structs (e.g. `[]*dbtypes.AddressRowCompact` instead of `[]dbtypes.AddressRowCompact`).  This allows fewer copies in places, particularly in `debitAddressRowsMerged` and `creditAddressRowsMerged`.
* Avoid a second `range map` loop in `MergeRows`, `MergeRowsCompact`, and `MergeRowsCompactRange`.
* In `(*ChainDB).AddressUTXO`, make cached data eligible for garbage collection before performing the update queries that generate data that will just replace the cached data. See the added `(*AddressCache).ClearRows` and `(*AddressCache).ClearUTXOs` functions.
* Also in `AddressUTXO`, set the local copy of the UTXOs slice to `nil` before trying to obtain the update lock (or waiting for another updater goroutine).
* Add a GNU Octave script for live plotting heap-in-use (`runtime.MemStats.HeapInuse`). Uses [`dcrps`](https://github.com/dcrlabs/dcrps) and requires the binary to be called "dcrdata".

### DIFF
--- a/api/insight/apiroutes.go
+++ b/api/insight/apiroutes.go
@@ -13,6 +13,8 @@ import (
 	"net/http"
 	"sort"
 	"strconv"
+	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/decred/dcrd/chaincfg"
@@ -42,19 +44,27 @@ const (
 	// "from" URL query parameters. Note that each transaction requires a
 	// getrawtransaction RPC call to dcrd.
 	maxInsightAddrsTxns = 250
+
+	// inflightUTXOLimit is a soft limit on the number of in-flight UTXOs that
+	// may be handled by the /addr/{address}/utxo endpoint. It is set slightly
+	// higher than maxInsightAddrsUTXOs to allow smaller requests to be
+	// processed concurrently with a very large request.
+	inflightUTXOLimit = maxInsightAddrsUTXOs * 5 / 4
 )
 
 // InsightApi contains the resources for the Insight HTTP API. InsightApi's
 // methods include the http.Handlers for the URL path routes.
 type InsightApi struct {
-	nodeClient     *rpcclient.Client
-	BlockData      *dcrpg.ChainDBRPC
-	params         *chaincfg.Params
-	mp             rpcutils.MempoolAddressChecker
-	status         *apitypes.Status
-	JSONIndent     string
-	ReqPerSecLimit float64
-	maxCSVAddrs    int
+	nodeClient      *rpcclient.Client
+	BlockData       *dcrpg.ChainDBRPC
+	params          *chaincfg.Params
+	mp              rpcutils.MempoolAddressChecker
+	status          *apitypes.Status
+	JSONIndent      string
+	ReqPerSecLimit  float64
+	maxCSVAddrs     int
+	inflightUTXOs   int64
+	inflightLimiter sync.Mutex
 }
 
 // NewInsightApi is the constructor for InsightApi.
@@ -93,7 +103,7 @@ func writeJSON(w http.ResponseWriter, thing interface{}, indent string) {
 	encoder := json.NewEncoder(w)
 	encoder.SetIndent("", indent)
 	if err := encoder.Encode(thing); err != nil {
-		apiLog.Infof("JSON encode error: %v", err)
+		apiLog.Warnf("JSON encode error: %v", err)
 	}
 }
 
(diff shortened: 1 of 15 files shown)


db/dbtypes/types.go
```diff
@@ -939,27 +939,25 @@ func CountMergedRowsCompact(rows []AddressRowCompact, txnView AddrTxnViewType) (
 // positive or not, although the Value field is an absolute value (always
 // positive). MergedRows will return a non-nil error of a merged row is detected
 // in the input since only non-merged rows are expected.
-func MergeRows(rows []*AddressRow) ([]AddressRowMerged, error) {
-	hashes := make([]chainhash.Hash, 0, len(rows))
-	merged := make(map[chainhash.Hash]*AddressRowMerged)
+func MergeRows(rows []*AddressRow) ([]*AddressRowMerged, error) {
+	numUniqueHashesGuess := len(rows)*2/3 + 1
```
chappjc: The number of unique transaction hashes is not known from the input address rows since an address may have multiple appearances in a given transaction.  Still pre-allocate, since we have an idea of the ballpark size of the result, but try not to overshoot.

db/dbtypes/types.go
```diff
@@ -975,11 +973,12 @@ func MergeRows(rows []*AddressRow) ([]AddressRowMerged, error) {
...
 			}
 
-			merged[*Hash] = &mr
+			hashMap[*Hash] = &mr
+			mergedRows = append(mergedRows, &mr)
 			continue
 		}
 
-		// Update existing transaction.
+		// Update existing transaction in the mergedRows slice.
```
chappjc: `row` comes out of the `hashMap`, not the slice, but the struct it points to is the same.

db/dbtypes/types.go
```diff
@@ -988,32 +987,25 @@ func MergeRows(rows []*AddressRow) ([]AddressRowMerged, error) {
 		}
 	}
 
-	// Build the slice.
-	mergedRows := make([]AddressRowMerged, 0, len(merged))
-	for i := range hashes {
-		mergedRows = append(mergedRows, *merged[hashes[i]])
-	}
+	fmt.Printf("MergeRows: guess = %d, actual = %d\n", numUniqueHashesGuess, len(mergedRows))
```
chappjc: will remove

db/dbtypes/types.go
```diff
@@ -1042,16 +1035,12 @@ func MergeRowsCompact(rows []AddressRowCompact) []AddressRowMerged {
 		}
 	}
 
-	// Build the slice.
-	mergedRows := make([]AddressRowMerged, 0, len(merged))
-	for i := range hashes {
-		mergedRows = append(mergedRows, *merged[hashes[i]])
-	}
+	fmt.Printf("MergeRowsCompact: guess = %d, actual = %d\n", numUniqueHashesGuess, len(mergedRows))
```
chappjc: will remove

db/dbtypes/types.go
```diff
@@ -988,32 +987,25 @@ func MergeRows(rows []*AddressRow) ([]AddressRowMerged, error) {
 		}
 	}
 
-	// Build the slice.
-	mergedRows := make([]AddressRowMerged, 0, len(merged))
-	for i := range hashes {
-		mergedRows = append(mergedRows, *merged[hashes[i]])
```
chappjc: We eliminate this final loop over the map, but as a consequence, we cannot accurately pre-allocate the `mergedRows` slice.

db/dcrpg/pgblockchain.go
```diff
@@ -4139,6 +4164,6 @@ func (pgb *ChainDBRPC) GetChainWork(hash *chainhash.Hash) (string, error) {
 func (pgb *ChainDB) GenesisStamp() int64 {
 	tDef := dbtypes.NewTimeDefFromUNIX(0)
 	// Ignoring error and returning zero time.
-	pgb.db.QueryRowContext(pgb.ctx, internal.SelectGenesisTime).Scan(&tDef)
+	_ = pgb.db.QueryRowContext(pgb.ctx, internal.SelectGenesisTime).Scan(&tDef)
```
chappjc: silencing error not checked

dev/heap_in_use.m
```diff
@@ -0,0 +1,9 @@
+heapuse=[];
+h=figure;
+ax=axes('parent',h);
+while true,
+  [s,out] = system("dcrps memstats dcrdata | grep heap-in-use | awk '{print $3}' | sed 's/(//'");
+  heapuse(end+1)=str2double(out);
+  plot(ax, heapuse); drawnow
+  pause(0.05)
+end
```
chappjc: GNU Octave script for live plotting `heap-in-use` (`runtime.MemStats.HeapInuse`). Uses [`dcrps`](https://github.com/dcrlabs/dcrps) and requires the binary to be called "dcrdata".

db/dbtypes/types.go
buck54321: Any particular reasoning behind this guess?

api/insight/apiroutes.go
```diff
@@ -324,11 +334,70 @@ func (iapi *InsightApi) getAddressesTxnOutput(w http.ResponseWriter, r *http.Req
...
+
+		// While locked, check in-flight UTXO count. If over the limit, become
+		// the prioritized goroutine and hold the lock until this http handler
+		// completes (and decrements the inflight UTXO count).
+		if !priority {
+			if totalInflight >= inflightUTXOLimit {
+				// Over the limit, but it's our turn to wrap it up.
+				priority = true
+				apiLog.Info("Becoming prioritized getAddressesTxnOutput goroutine"+
+					"with %d total in-flight UTXOs.", totalInflight)
```
buck54321: Needs an `Infof` and a space.

db/dbtypes/types.go
chappjc: The actual number could be smaller or even larger than in input number of address rows.  Since we don't know what it is, I decided not to over allocate space, which is wasted.  On the other hand, when go needs to append to a slice with insufficient capacity, it automatically expands the capacity by re-allocating (and copying the elements to the new larger buffer), which not free.  As for the exact fraction, it's more or less pulled out of thin air.  But when go expands capacity to append, it doubles capacity until it hits 1024, after which further appends will cause a 25% capacity increase.  Considering a large slice, of say 409,600 pointers, where it's more costly to reallocate, it is best not to start with a zero-capacity slice as it would involve 36 reallocations and 14,632,472 bytes copied in the process.  https://play.golang.org/p/wZTdS7TqjX3
So `2/3` is just balancing the desire to avoid reallocation and copying, with the desire to not waste space when the actual output length is much less than the input length (which does happen, even in our `insight-addr` attack profile.
I definitely do not think I picked the optimal number here, so I'm open to other ideas.  One thought I had was to overallocate and when the loop finishes, make a new slice with the correct capacity.

PR #1351, opened 2019-05-23
misc cleanup and func argument interfacing

chappjc:


stakedb/stakedb.go
```diff
@@ -331,22 +331,20 @@ func (db *StakeDatabase) PopulateLiveTicketCache() error {
...
 	promisesGetRawTransaction := make([]promiseGetRawTransaction, 0, len(liveTickets))
-
-	// Send all the live ticket requests
-	for _, hash := range liveTickets {
+	for i := range liveTickets {
 		promisesGetRawTransaction = append(promisesGetRawTransaction, promiseGetRawTransaction{
-			result: db.NodeClient.GetRawTransactionAsync(&hash),
-			ticket: hash,
+			result: db.NodeClient.GetRawTransactionAsync(&liveTickets[i]),
+			ticket: liveTickets[i],
```
chappjc: With the previous loop, `hash` was a copy of a `chainhash.Hash` (not a pointer to one), and when creating the promise with `&hash`, the copy was likely persisting until the promises slice was cleaned up.  Further, another copy was being made for the `tickets` field in the promise container struct, so there were too many copies.

db/dcrpg/queries.go
```diff
@@ -2462,7 +2462,7 @@ func SetSpendingForVinDbID(db *sql.DB, vinDbID uint64) (int64, error) {
 
 	// Set the spending tx info (addresses table) for the funding transaction
 	// rows indicated by the vin DB ID.
-	N, err := setSpendingForFundingOP(dbtx, prevOutHash, prevOutVoutInd,
+	N, err := SetSpendingForFundingOP(dbtx, prevOutHash, prevOutVoutInd,
```
dmigwi: Do Methods and functions that are not used outside the dcrpg package need to be exported?
IMO this makes to much than necessary information to be on display in the package's godoc page.

db/dcrpg/pgblockchain.go
```diff
@@ -44,6 +44,8 @@ var (
 	zeroHashStringBytes = []byte(chainhash.Hash{}.String())
 )
 
+const errMsgRetry = "retry"
```
dmigwi: This error message looks too short to be informative and clear.

chappjc: It's not for logging, just programmatically recognizing an outcome.  This could be `errMsgRetry = errors.New("retry")` to avoid doing kind of check like `err.Error() == "some string"`, but the message itself is not for humans.

db/dcrpg/queries.go
chappjc: No, it could be removed since it's only used in dcrpg, but since there has been an exported `SetSpendingForFundingOP` for a very long time (and most other functions in queries.go are exported), I'm hesitant to unexport this one now.  We can consider unexporting the bulk of them in a followup PR that investigates this possibility however.

db/dcrpg/pgblockchain.go
dmigwi: ok. Noted. 

You could create an new error type that guarantees `const errMsgRetry = ...` will still be used ensuring that the error remains immutable.

chappjc: Can you clarify that?  Do you mean a new actual `type` or just like `const errMsgRetry = errors.New("retry")`?

dmigwi: I don't think its possible to have `const errMsgRetry = errors.New("retry")` https://play.golang.org/p/-TzVytclrMy

I meant something like this.
```golang
type shortError string

func (s shortError) Error() string {
 	return string(s)
}

const errMsgRetry = shortError("retry")
```

chappjc: Gotcha... implement the `error` interface.  That's cleaner.

chappjc: I went with a type assertion.

db/dcrpg/go.mod
```diff
@@ -31,4 +31,5 @@ require (
 replace (
 	github.com/decred/dcrdata/api/types => ../../api/types
 	github.com/decred/dcrdata/db/cache => ../cache
+	github.com/decred/dcrdata/db/dbtypes => ../dbtypes
```
chappjc: We'll clean out these replaces pretty soon.

PR #1354, opened 2019-05-25
[4.1] Change the reverse chronologically sorting url to return new data

chappjc:
This is a backport of https://github.com/decred/dcrdata/pull/1352 to 4.1-stable.

It is presently deployed to https://explorer.dcrdata.org/proposals.

### DIFF
--- a/gov/politeia/proposals.go
+++ b/gov/politeia/proposals.go
@@ -119,8 +119,7 @@ func (db *ProposalDB) saveProposals(URLParams string) (int, error) {
 			break
 		}
 
-		copyURLParams = fmt.Sprintf("%s?after=%v", URLParams,
-			data.Data[pageSize-1].Censorship.Token)
+		copyURLParams = fmt.Sprintf("?before=%v", data.Data[pageSize-1].Censorship.Token)
 	}
 
 	// Save all the proposals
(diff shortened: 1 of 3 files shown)


gov/politeia/proposals.go
```diff
@@ -237,7 +236,7 @@ func (db *ProposalDB) CheckProposalsUpdates() error {
 }
 
 func (db *ProposalDB) lastSavedProposal() (lastP []*pitypes.ProposalInfo, err error) {
-	err = db.dbP.All(&lastP, storm.Limit(1), storm.Reverse())
+	err = db.dbP.Select().Limit(1).OrderBy("Timestamp").Reverse().Find(&lastP)
```
chappjc: @dmigwi At least in a sql statement, the order by has to come before limit.  Does this orm has a similar requirement?

I'm not seeing newer proposals on explorer.dcrdata.org/proposals, so something appears to be wrong despite it looking OK on alpha.

dmigwi: The ORM has this example: 
```Golang
// Calls can also be chained
query = query.Limit(10).Skip(20).OrderBy("Age").Reverse()
```
https://github.com/asdine/storm

This is the newest proposal https://alpha.dcrdata.org/proposal/a-journey-to-the-future-of-work-telling-the-dcr-story-through-politeia

dmigwi: Let me test the 4.1 PR on my machine.

chappjc: Thanks.  I'll scour the logs later.

dmigwi: This test on my machine will not bear any fruits because my proposals db is upto date. From the https://explorer.dcrdata.org/proposals seems like upto six proposal could be missing.

Seems any relevant error or warning message on the logs?

dmigwi: The main problem here is that dependencies used are linked to a specific version.
Since we've updated the Politeia package, you would need to update its dependency as linked in the `go.mod` file or just use a replace statement.

chappjc: Good catch.  I forget about this since we're using replace for everything on master.

Issue #996, opened 2019-02-10
create chart for % of coin supply staked


raedah:

xaur:
That would help me a lot for Decred Journal. dcred.eu/powStats has incorrect chart of locked %. Someone recommended a replacement at [charts.dcr.farm](https://charts.dcr.farm/) -- need to check it out but I already see it wants too much perms. In any case, having this on dcrdata would be great.

xaur:
charts.dcr.farm doesn't open anymore for me. dcred.eu has wrong % numbers.

chappjc:
Resolved by https://github.com/decred/dcrdata/pull/1355

PR #1355, opened 2019-05-27
chart: Add a percent staked coins chart

dmigwi:
New percent staked coin chart and fixes some charts legends.

Screenshot from 2019-05-27 10-39-34

Resolving https://github.com/decred/dcrdata/issues/996

### DIFF
--- a/db/cache/charts.go
+++ b/db/cache/charts.go
@@ -34,6 +34,7 @@ const (
 	TicketPoolSize  = "ticket-pool-size"
 	TicketPoolValue = "ticket-pool-value"
 	WindMissedVotes = "missed-votes"
+	PercentStaked   = "stake-participation"
 )
 
 // binLevel specifies the granularity of data.
@@ -86,8 +87,10 @@ const (
 	HashrateAvgLength = 120
 )
 
-// cacheVersion helps detect when cache data stored has its structure changed.
-var cacheVersion = semver.NewSemver(5, 0, 0)
+// cacheVersion helps detect when the cache data stored has changed its
+// structure or content. A change on the cache version results to recomputing
+// all the charts data a fresh thereby making the cache to hold the latest changes.
+var cacheVersion = semver.NewSemver(6, 0, 0)
 
 // versionedCacheData defines the cache data contents to be written into a .gob file.
 type versionedCacheData struct {
@@ -886,6 +889,7 @@ var chartMakers = map[string]ChartMaker{
 	TicketPoolSize:  ticketPoolSizeChart,
 	TicketPoolValue: poolValueChart,
 	WindMissedVotes: missedVotesChart,
+	PercentStaked:   stakedCoinsChart,
 }
 
 // Chart will return a JSON-encoded chartResponse of the provided type
@@ -1209,4 +1213,22 @@ func missedVotesChart(charts *ChartData, _ binLevel, _ axisType) ([]byte, error)
 	}
 	return charts.encode(charts.Windows.Time[stakeValidWindows:],
 		charts.Windows.MissedVotes[stakeValidWindows:])
+}
+
+func stakedCoinsChart(charts *ChartData, bin binLevel, axis axisType) ([]byte, error) {
+	switch bin {
+	case BlockBin:
+		return charts.encode(charts.Blocks.Time, accumulate(charts.Blocks.NewAtoms),
+			charts.Blocks.PoolValue)
+	case DayBin:
+		switch axis {
+		case HeightAxis:
+			return charts.encode(charts.Days.Height, accumulate(charts.Days.NewAtoms),
+				charts.Days.PoolValue)
+		default:
+			return charts.encode(charts.Days.Time, accumulate(charts.Days.NewAtoms),
+				charts.Days.PoolValue)
+		}
+	}
+	return nil, InvalidBinErr
 }
--- a/public/js/controllers/charts_controller.js
+++ b/public/js/controllers/charts_controller.js
@@ -22,6 +22,7 @@ const lineScales = ['ticket-price']
 const yValueRanges = { 'ticket-price': [1] }
 var ticketPoolSizeTarget, premine, stakeValHeight, stakeShare
 var baseSubsidy, subsidyInterval, subsidyExponent, windowSize, avgBlockTime
+var rawCoinSupply, rawPoolValue
 
 function usesWindowUnits (chart) {
   return windowScales.indexOf(chart) > -1
@@ -125,7 +126,16 @@ function legendFormatter (data) {
           yVal = series.y.toFixed(4) + '%'
           break
       }
-      return `${nodes} <div class="pr-2">${series.dashHTML} ${series.labelHTML}: ${yVal}</div>`
+      let result = `${nodes} <div class="pr-2">${series.dashHTML} ${series.labelHTML}: ${yVal}</div>`
+
+      var i = data.dygraph.getOption('legendIndex')
+      if (series.label.toLowerCase() === 'stake participation' && rawCoinSupply.length === rawPoolValue.length &&
+          rawPoolValue.length !== 0 && i !== null) {
+        result += `<div class="pr-2"><div class="dygraph-legend-line"></div> Ticket Pool Value: ${intComma(rawPoolValue[i])} DCR</div>
+          <div class="pr-2"><div class="dygraph-legend-line"></div> Coin Supply: ${intComma(rawCoinSupply[i])} DCR</div>`
+      }
+
+      return result
     }, '')
 
     html = `<div class="d-flex flex-wrap justify-content-center align-items-center">
@@ -221,7 +231,9 @@ function percentStakedFunc (gData, isHeightAxis, isDayBinned) {
     } else {
       xAxisVal = new Date(n * 1000)
     }
-    return [ xAxisVal, percentStaked, poolValue, coinSupply ]
+    rawPoolValue.push(poolValue)
+    rawCoinSupply.push(coinSupply)
+    return [ xAxisVal, percentStaked ]
   })
   return [data, extremeStaked]
 }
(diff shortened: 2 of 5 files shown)


views/charts.tmpl
```diff
@@ -31,6 +31,7 @@
                             <option value="ticket-price">Ticket Price</option>
                             <option value="ticket-pool-size">Ticket Pool Size</option>
                             <option value="ticket-pool-value">Ticket Pool Value</option>
+                            <option value="percent-staked-coins">Percent Staked Coins</option>
```
chappjc: For the option text, I think "Stake Participation" would be good.  It's a commonly used phrase, and the chart y-axis will clarify how stake participation is quantified.

db/cache/charts.go
```diff
@@ -68,7 +69,7 @@ const (
 )
 
 // cacheVersion helps detect when cache data stored has its structure changed.
-var cacheVersion = semver.NewSemver(1, 0, 0)
+var cacheVersion = semver.NewSemver(1, 1, 0)
```
chappjc: Does any difference in the semantic version result in a complete recompute of all charts data?  Ideally a minor version bump (1.0.0 -> 1.1.0) would allow loading a cache dump with the previous minor version (1.0.0) and then completely building just the added chart data and updating everything else.  If that is not current behavior, it seems like we should simply bump the major version.  If that behavior is not planned, then we should just be using an integer version.  Can you clarify, @dmigwi?  Also, the docs for `cacheVersion` should explain how the version components are used and what actions are taken when the version changes.

dmigwi: Yes. Recomputing of all charts data was the intended effect effect in case of a version change.

I will add the needed documentation.

chappjc: In that case, what is the meaning of each major/minor/patch component of the semver if any change requires recomputing all charts data?  If there are no plans to ever have a partial recompute with minor updates, or patching/checks with a patch update, then I suggest we just bump the major version, and change all logging to report the version as a single integer (e.g. cache dump file v2).

dmigwi:
Moved the legend to the space between the chart and the control buttons.

Screenshot from 2019-05-30 11-18-16

db/cache/charts.go
```diff
@@ -1158,3 +1162,21 @@ func poolValueChart(charts *ChartData, bin binLevel, axis axisType) ([]byte, err
 	}
 	return nil, InvalidBinErr
 }
+
+func stakedCoinsChart(charts *ChartData, bin binLevel, axis axisType) ([]byte, error) {
+	switch bin {
+	case BlockBin:
+		return charts.encode(charts.Blocks.Time, accumulate(charts.Blocks.NewAtoms),
+			charts.Blocks.PoolValue)
```
buck54321: We can reduce the API response size by more that 1/3 by performing the conversion to staked percent here. That's at least 5 MB on the blocks view. You can then reconstruct the data client-side using a similar procedure to `circulationFunc` in `charts_controller`.

public/js/helpers/zoom_helper.js
```diff
@@ -121,7 +121,8 @@ export default class Zoom {
     for (let k in zoomMap) {
       let v = zoomMap[k]
       if (v === 0) continue
-      if (v / scale === range) return k
+      // support an error of ±1 due to precision loss while dividing.
+      if (Math.abs(v / scale - range) < 1) return k
```
buck54321: Good call. Note that this only works when the range is large, >> 1, which it is for time and block height. This will break if ever used for a chart with say range 0 < x < 1. 

A more flexible solution would be to check if the node is within say +/-1 **%** of the visible window, so this could be 

```
let scaled = v / scale
if (Math.abs((scaled - range) / scaled) < 0.01) return k 
```
*untested

public/js/controllers/charts_controller.js
```diff
@@ -320,11 +354,21 @@ export default class extends Controller {
           }
         }
         break
+      case 'stake-participation':
+        d = percentStakedFunc(data, isHeightAxis, isDayBinned)
+        assign(gOptions, mapDygraphOptions(d[0], [xlabel, 'Stake Participation', 'Ticket Pool Value',
+          'Coin Supply'], true, 'Stake Participation (%)', undefined, true, false))
+        gOptions.valueRange = [null, d[1] + 5]
```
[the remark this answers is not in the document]
dmigwi: This is why I mentioned the bug here: https://github.com/decred/dcrdata/pull/1359#discussion_r288960156 The fix in #1359 will work effectively on this chart too thus correcting that chart you were able to create by messing with the rangeselector.

I thought it wise not to repeat the same code in this PR so as not to have many merge conflicts while merging both PRs.

public/js/helpers/zoom_helper.js
dmigwi: I think the ±1% check is way too huge and could distort information more than ±1.

Consider this:

       A day has a range of 288 blocks 
       1% of 288 is `2.88` 
       This means  `± 2.88` blocks is the range where `Month` zoom button would be active.
       I think `±1` is more better than `± 2.88` when factoring precision loss.
       Also because integer division just affects the decimal places, `±1` should be okay.

public/js/controllers/charts_controller.js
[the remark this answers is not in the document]
dmigwi: >The hiccup here is that the data supplied to the legendFormatter doesn't include the data point's index in the series, so d[2] would need to be associative, i.e. d[2] = { 'x1': y1, 'x2': y2, ... } which adds a little time to the lookup.

I don't think we should take this route since the value range fix works.

public/js/helpers/zoom_helper.js
buck54321: > `± 2.88` blocks is the range where `Day` zoom button would be active

That actually sounds great, but you can set it tighter if you want. The way you have it now, you are setting a 0.000000064% margin on time-based data, so there a lot of wiggle room to add. For use of `Zoom` on charts that with 0 < x < 1, this will simply fail, which is not acceptable. `Zoom` must handle data in a scale independent way.

public/js/controllers/charts_controller.js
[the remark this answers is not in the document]
buck54321: You are still tripling the rendering time. Loading blocks-based data on this PR's chart takes about 40 seconds for me. Compare that to ~ 7 seconds for other blocks-based charts where there isn't extraneous tripling of data. 

`valueRange` should simply be avoided, as it's implementation is incomplete.

public/js/helpers/zoom_helper.js
dmigwi: >For use of Zoom on charts that with 0 < x < 1, this will simply fail, which is not acceptable.

What does "x" here mean? A chart with only one block to plot?

buck54321: `x` is whatever the x-axis is. `Zoom` doesn't know or need to know what a block. It should handle fractional x values identically to integral x values.

dmigwi: An error of `±0.99` should be better than ±1 since we don't have a 0.99 block (no such a counting number). On the time-based data this error has no much effect since its in microseconds and an error of ±0.99 microseconds is so small to be even noticed.

db/cache/charts.go
dmigwi: I have had a looks at doing the % calculations on the backend but that implementation results to a increase in data needed to plot this chart. IMO we should still do the % calculations on the frontend.

public/js/controllers/charts_controller.js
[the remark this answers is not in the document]
dmigwi: I still don't know how 40 sec or 7 sec were obtained but on my end I get much lower time with the same data.

**Here is a screenshot of my terminal.**
Screenshot from 2019-06-04 12-45-44

**Screenshot of the browser metrics.**
Screenshot from 2019-06-04 12-34-57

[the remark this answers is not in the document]
dmigwi: >Possible solution: Pass the poolValue, coinSupply values back from percentStakedFunc as a separate structure. Set it as a custom option on `gOptions.myCustomOptionName = d[2]`, and then use `Dygraph.getOption` to retrieve the information during legend formatting.

I am also curious to know how you got `Dygraph.getOption` working because its returning errors.
Screenshot from 2019-06-04 13-05-35

Also from the API, legendformatter doesn't pass the current graph instance: https://github.com/danvk/dygraphs/pull/683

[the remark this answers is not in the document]
buck54321: Yeah you may need to make it a controller method. 

The timings are for rendering, not downloading. For blocks data, the download is a pretty small fraction of the time to display. Most of the time is locked in Dygraphs data processing and drawing. That's why adding a lot of `file` data that you aren't actually displaying is a bad idea. Dygraphs is trying to draw ~700,000 extra line segments that you are just hiding with `valueRange`.

public/js/helpers/zoom_helper.js
buck54321: It's fine for now.

db/cache/charts.go
buck54321: If we drop the pool size and circulation from the legend, and just send percents, the API response goes from 1 array of timestamps and two arrays of atoms to 1 timestamps and 1 floats. It should be about ~1/3 smaller. For blocks data that is about 5 MB. 

I'm only pointing out that the cost for the extra legend data is pretty steep. It's fine though.

chappjc:
Got a couple conflicts now that https://github.com/decred/dcrdata/pull/1359 is merged.

db/cache/charts.go
```diff
@@ -84,8 +85,10 @@ const (
 	HashrateAvgLength = 120
 )
 
-// cacheVersion helps detect when cache data stored has its structure changed.
-var cacheVersion = semver.NewSemver(3, 0, 0)
+// cacheVersion helps detect when the cache data stored has changed its
+// structure or content. A change on the cache version results to recomputing
+// all the charts data a fresh thereby making the cache to hold the latest changes.
+var cacheVersion = semver.NewSemver(2, 0, 0)
```
chappjc: Ooops, the version is at 3 already.  Better change this to 4.

buck54321:
Hey @dmigwi, check out #1455

That will allow you to get the data index in the `legendFormatter`. 

Store any undisplayed data as a Dygraph option (not part of `file`) and retrieve it in the `legendFormatter` using `data.dygraph.getOption` and get the corresponding value with the index, allowing you to format the legend.

public/js/controllers/charts_controller.js
```diff
@@ -208,6 +218,8 @@ function zipXYZData (gData, isHeightAxis, isDayBinned, yCoefficient, zCoefficien
 
 function percentStakedFunc (gData, isHeightAxis, isDayBinned) {
   var extremeStaked = 0
+  rawPoolValue = []
+  rawCoinSupply = []
```
buck54321: Could you reset these before every new chart is loaded? Somewhere around https://github.com/dmigwi/dcrdata/blob/17fdf489f8c49fb899772da364eb88bc9fbe4bd1/public/js/controllers/charts_controller.js#L412

PR #1357, opened 2019-05-27
explorer: websocket oopsies

chappjc:
For a very long time we've been ignoring some error messages when `explorer`'s websocket controller shuts down on dcrdata shutdown.

```
[INF] EXPR: Stopping websocket hub.
[ERR] EXPR: close channel already closed. This should not happen.
[ERR] EXPR: Do not send on stopPing channel, only close it.
```

These are bugs that were fixed in the `pubsub` package's version of the websocket controller, and these changes make the fixes in the `explorer` package.

### DIFF
--- a/explorer/websocket.go
+++ b/explorer/websocket.go
@@ -149,8 +149,8 @@ func (wsh *WebsocketHub) unregisterClient(c *hubSpoke) {
 	delete(wsh.clients, c)
 	wsh.setNumClients(len(wsh.clients))
 
-	// Close the channel, but make sure the client didn't do it
-	safeClose(*c)
+	// Close the channel.
+	close(*c)
 }
 
 // unregisterAllClients should only be called from the loop in run() or when no
@@ -180,10 +180,7 @@ func (wsh *WebsocketHub) pingClients() chan<- struct{} {
 			select {
 			case <-ticker.C:
 				wsh.HubRelay <- pstypes.HubMessage{Signal: sigPingAndUserCount}
-			case _, ok := <-stopPing:
-				if !ok {
-					log.Errorf("Do not send on stopPing channel, only close it.")
-				}
+			case <-stopPing:
 				return
 			}
 		}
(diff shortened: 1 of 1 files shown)


explorer/websocket.go
```diff
@@ -180,10 +180,7 @@ func (wsh *WebsocketHub) pingClients() chan<- struct{} {
 			select {
 			case <-ticker.C:
 				wsh.HubRelay <- pstypes.HubMessage{Signal: sigPingAndUserCount}
-			case _, ok := <-stopPing:
-				if !ok {
```
chappjc: This should have been `if ok` to check if it was a send rather than a close, but this entire check was just for debugging.  No one is sending on `stopPing`

explorer/websocket.go
```diff
@@ -192,22 +189,13 @@ func (wsh *WebsocketHub) pingClients() chan<- struct{} {
 	return stopPing
 }
 
-func safeClose(cc hubSpoke) {
```
chappjc: This is totally not safe.  It's inherently racy.

explorer/websocket.go
```diff
@@ -192,22 +189,13 @@ func (wsh *WebsocketHub) pingClients() chan<- struct{} {
...
-
-// Stop kills the run() loop and unregisteres all clients (connections).
+// Stop kills the run() loop and unregisters all clients (connections).
 func (wsh *WebsocketHub) Stop() {
-	// end the run() loop, allowing in progress operations to complete
+	// End the run() loop, allowing in-progress operations to complete.
 	wsh.quitWSHandler <- struct{}{}
+	// Lastly close the hub relay channel sine the quitWSHandler signal is
+	// handled in the Run loop.
+	close(wsh.HubRelay)
```
chappjc: As in pubsub/websocket.go, close the `HubRelay` channel so that `(*WebsocketHub).run` will return.

PR #1358, opened 2019-05-28
Piparser: Complete its decoupling the in case of its failure

dmigwi:
Shielding the dcrdata from errors obtained from piparser wasn't fixed conclusively before. When I had some internet outage earlier today, dcrdata crashed because of the nil parser instance value passed though the `ProposalsFetcher` interface.
Instead of using `reflect` package to check if the interface has a nil value, a parser validity flag is set when the parser instance created is nil.

Proposals charts were set to display on the correct UTC time instead of the browser's timezone.

### DIFF
--- a/main.go
+++ b/main.go
@@ -185,6 +185,10 @@ func _main(ctx context.Context) error {
 		log.Error(err)
 	}
 
+	// SetPiParserValidity sets the invalid parser flag if a nil parser instance
+	// was found.
+	dcrpg.SetPiParserValidity(parser == nil)
+
 	// Auxiliary DB (PostgreSQL)
 	var newPGIndexes, updateAllAddresses, updateAllVotes bool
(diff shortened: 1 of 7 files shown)


main.go
```diff
@@ -185,6 +185,10 @@ func _main(ctx context.Context) error {
 		log.Error(err)
 	}
 
+	// SetPiParserValidity sets the invalid parser flag if a nil parser instance
+	// was found.
+	dcrpg.SetPiParserValidity(parser == nil)
```
chappjc: I think the solution we discussed for this with jrick would be a bit simpler:

```go
var piParser dcrpg.ProposalsFetcher
if parser != nil {
	piParser = parser
}
```

Then use `piParser` as input to `NewChainDBWithCancel`.

This avoids the need for `reflect`, and it minimizes the solution's complexity.

PR #1359, opened 2019-05-30
chart: Add purchased tickets axis to the ticket price chart

dmigwi:
Screenshot from 2019-05-30 14-38-59

### DIFF
--- a/public/js/controllers/charts_controller.js
+++ b/public/js/controllers/charts_controller.js
@@ -236,6 +236,9 @@ export default class extends Controller {
       'axisOption',
       'binSelector',
       'scaleSelector',
+      'ticketsPurchase',
+      'ticketsPrice',
+      'vSelector',
       'binSize'
     ]
   }
@@ -295,7 +298,6 @@ export default class extends Controller {
       labelsDiv: this.labelsTarget,
       legendFormatter: legendFormatter,
       highlightCircleSize: 4,
-      xlabel: 'Date',
       ylabel: 'Ticket Price',
       y2label: 'Tickets Bought',
       labelsUTC: true
@@ -321,6 +323,7 @@ export default class extends Controller {
       zoomCallback: null,
       drawCallback: null,
       logscale: this.settings.scale === 'log',
+      axes: {},
       y2label: null,
       stepPlot: false
     }
@@ -330,10 +333,10 @@ export default class extends Controller {
 
     switch (chartName) {
       case 'ticket-price': // price graph
-        d = zipXYZData(data, isHeightAxis, isDayBinned, atomsToDCR, 1, windowSize)
+        d = zipXYZData(data, isHeightAxis, false, atomsToDCR, 1, windowSize)
         gOptions.stepPlot = true
-        assign(gOptions, mapDygraphOptions(d, ['Date', 'Ticket Price', 'Tickets Bought'], true,
-          'Price (DCR)', 'Date', undefined, false, false))
+        assign(gOptions, mapDygraphOptions(d, [xlabel, 'Ticket Price', 'Tickets Bought'], true,
+          'Price (DCR)', xlabel, undefined, false, false))
         gOptions.y2label = 'Tickets Bought'
         gOptions.series = { 'Tickets Bought': { axis: 'y2' } }
         gOptions.axes.y2 = {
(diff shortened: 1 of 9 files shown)


public/js/controllers/charts_controller.js
```diff
@@ -16,17 +17,45 @@ const aDay = 86400 * 1000 // in milliseconds
...
 
+function isScaleDisabled (chart) {
+  return lineScales.indexOf(chart) > -1
+}
+
 function intComma (amount) {
   return amount.toLocaleString(undefined, { maximumFractionDigits: 0 })
 }
 
+function axesToRestoreYRange (chartName, origYRange, newYRange) {
```
dmigwi: When a dygraph chart has a fixed y-axis `valueRange`, zooming in and later out returning to the full chart view alters the set `valueRange`.  `axesToRestoreYRange` helps to check if a specific chart has had its value range altered and recreates the required value range parameters needed to revert to the original value range.

The issue has been mentioned here https://groups.google.com/forum/#!topic/dygraphs-users/WbrphwBcRgQ but no viable solution was found. This problem is only triggered by sliding the rangeselector only.

views/charts.tmpl
```diff
@@ -156,6 +156,38 @@
...
                 </div>
 
+                <div class="chart-control-wrapper mr-2 mb-1" data-target="charts.vSelector">
+                    <div class="chart-control-label">VISIBILITY</div>
+                    <div class="chart-control">
+                        <ul
+                            class="nav nav-pills"
+                        >
+                            <li class="nav-item">
+                                <label class="customcheck ml-2 d-inline-flex">Tickets Price
```
chappjc: Just "Price" for the text here is good I'd say.

buck54321:
Was there a request for this data as a chart? As discussed at https://github.com/decred/dcrdata/pull/1274#issuecomment-485773848, we should avoid adding charts that are likely to always look exactly the same. This type of data is better served through an API endpoint.

dmigwi:
@buck54321 An effort has been made to avoid too much noise as in the charts you referenced:
- by ensuring that the second axis occupies only a 1/3 of the chart display so as to reduce overlapping.
- checkboxes were added to control which axes are visible at any given time.
Screenshot from 2019-05-30 20-09-58

buck54321:
Even though the two data sets are mathematically related through the stake diff algo, the plots show little correlation other than that they get noisy together. When plotted together, almost half of the range is muddled. Double-axis charts are very difficult to pull off in a compelling way and they have accessibility issues. I'd also like to register here again my recommendation to make this an API endpoint instead. The data is just not interesting enough to justify a chart. 

I'd love to hear other opinions on this.

chappjc:
I frequently see the following chart from another site posted and in articles:

image

Perhaps the motivation of this PR is to show the same data?  On that page the number of tickets purchased bars are scaled down much more so that they are not too distracting.  But even if we do the same, I too don't see a lot of value in the number of tickets bars.

If the tickets purchased data is unchecked in the visibility controls, does the unused y-axis also get removed?  If so, I'd say this new chart is acceptable if,
- by default only the ticket price is visible
- the y-axis limits for the tickets purchased bars are fixed at [0, 20000] (or maybe even higher to compress the data further)
- There are no regressions in the ticket price chart, on both mobile and desktop

If it's not reasonable to achieve these goals, I also support making this an API-only data set.

dmigwi:
>If the tickets purchased data is unchecked in the visibility controls, does the unused y-axis also get removed? If so, I'd say this new chart is acceptable if,
>- by default only the ticket price is visible
>- the y-axis limits for the tickets purchased bars are fixed at [0, 20000] (or maybe even higher to compress the data further)
>- There are no regressions in the ticket price chart, on both mobile and desktop

This requirements are possible to achieve. I will be making the changes in a few.

The visibility of the unused axis is handled by dygraph that ensures its data is no longer visible even on the legend until its visibility control is activated again. I think the unused data is just hidden from the user but not totally removed because to show the axis back no fresh data load is required.

buck54321:
Even if you can make it look nice, there's is still a simple question that isn't answered here. What does putting these two data sets together add? They don't appear to correlate in any way, even if there is a relationship somehow. Two-axis charts are useful for showing data correlation. This doesn't do that, so why have a two-axis char?

chappjc:
TBH, I agree with @buck54321.  If there is an appropriate data set for plotting on the same chart as ticket price, it would be ticket pool size (as per the sdiff algo).  Tickets bought is of course related to pool size (almost a derivative except for tickets leaving the pool by votes), but it's hard to make out any correlation from a chart like this one.

chappjc:
But ticket pool size is essentially a low-pass filtered version of ticket price, so that's arguably not of much value either. Ticket pool size modulated by percent staked might be of interest though.

dmigwi:
**Default ticket price view:**

Screenshot from 2019-06-01 01-14-40

**On activating Tickets Bought Axis:**
- Tickets Bought axis is a line graph now instead of a bar graph.
Screenshot from 2019-06-01 01-14-56

PR #1362, opened 2019-05-31
chart: Add missed votes chart

dmigwi:
Screenshot from 2019-06-01 01-02-08

### DIFF
--- a/db/cache/charts.go
+++ b/db/cache/charts.go
@@ -33,6 +33,7 @@ const (
 	Fees            = "fees"
 	TicketPoolSize  = "ticket-pool-size"
 	TicketPoolValue = "ticket-pool-value"
+	WindMissedVotes = "missed-votes"
 )
 
 // binLevel specifies the granularity of data.
@@ -85,7 +86,7 @@ const (
 )
 
 // cacheVersion helps detect when cache data stored has its structure changed.
-var cacheVersion = semver.NewSemver(3, 0, 0)
+var cacheVersion = semver.NewSemver(4, 0, 0)
 
 // versionedCacheData defines the cache data contents to be written into a .gob file.
 type versionedCacheData struct {
@@ -305,6 +306,35 @@ func newWindowSet(size int) *windowSet {
 	}
 }
 
+// missedSet is for missed votes data that is grouped at an interval of 144
+// blocks on mainnet. missed votes txs don't appear till block 4096 on mainnet
+// when the first POS votes txs were made.
+type missedSet struct {
+	cacheID uint64
+	Count   ChartUints
+	Time    ChartUints
+	Height  ChartUints
+}
+
+// Snip truncates the missedSet to a provided length.
+func (set *missedSet) Snip(length int) {
+	if length < 0 {
+		length = 0
+	}
+	set.Count = set.Count.snip(length)
+	set.Time = set.Time.snip(length)
+	set.Height = set.Height.snip(length)
+}
+
+// Constructor for a sized missedSet.
+func newMissedSet(size int) *missedSet {
+	return &missedSet{
+		Count:  newChartUints(size),
+		Time:   newChartUints(size),
+		Height: newChartUints(size),
+	}
+}
+
 // ChartGobject is the storage object for saving to a gob file. ChartData itself
 // has a lot of extraneous fields, and also embeds sync.RWMutex, so is not
 // suitable for gobbing.
@@ -356,9 +386,11 @@ type ChartData struct {
 	ctx          context.Context
 	genesis      uint64
 	DiffInterval int32
+	StartPOS     int32
 	Blocks       *zoomSet
 	Windows      *windowSet
 	Days         *zoomSet
+	MissedPos    *missedSet
 	cacheMtx     sync.RWMutex
 	cache        map[string]*cachedChart
 	updaters     []ChartUpdater
@@ -412,6 +444,18 @@ func (charts *ChartData) Lengthen() error {
 	}
 	if shortest == 0 {
 		// No blocks yet. Not an error.
+		return nil
+	}
+
+	// Make sure the database has set an equal number of missed votes window in each data set.
+	missedVotes := charts.MissedPos
+	shortest, err = ValidateLengths(missedVotes.Time, missedVotes.Height, missedVotes.Count)
+	if err != nil {
+		log.Warnf("ChartData.Lengthen: block data length mismatch detected. truncating missed votes windows length to %d", shortest)
+		missedVotes.Snip(shortest)
+	}
+	if shortest == 0 {
+		// No missed votes windows yet. Not an error.
 		return nil
 	}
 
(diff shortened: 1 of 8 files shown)


db/cache/charts.go
```diff
@@ -302,22 +303,54 @@ func newWindowSet(size int) *windowSet {
...
+
+// Constructor for a sized missedSet.
+func newMissedSet(size int) *missedSet {
+	return &missedSet{
+		Count:  newChartUints(size),
+		Time:   newChartUints(size),
+		Height: newChartUints(size),
+	}
+}
+
```
buck54321: Why not use the `windowSet`?

dmigwi: Because missed votes do not appear till after the stake validation height.
I had made a try to match the `windowSet` data with missed votes but I experienced some slice length inconsistencies. I will make another try.

db/dcrpg/queries.go
```diff
@@ -3111,6 +3111,41 @@ func appendCoinSupply(charts *cache.ChartData, rows *sql.Rows) error {
 	return nil
 }
 
+// retrieveMissedVotes fetches the missed votes data from the misses and tarnsactions tables.
```
buck54321: typo `tarnsactions`.

db/dcrpg/queries.go
```diff
@@ -3111,6 +3111,41 @@ func appendCoinSupply(charts *cache.ChartData, rows *sql.Rows) error {
...
+	return rows, nil
+}
+
+// Append the results from retrieveMissedVotes to the provided ChartData.
+// This is the Appender half of a pair that make up a cache.ChartUpdater.
+func appendMissedVotes(charts *cache.ChartData, rows *sql.Rows) error {
+	defer closeRows(rows)
+	windows := charts.Windows
+	windowSize := uint64(charts.DiffInterval)
+	startHeight := windowSize*uint64(len(windows.MissedVotes)) + uint64(charts.StartPOS)
```
buck54321: Instead of messing with lengths and validations in `charts.go`, it would be easier just to prepend some number of zeroes right here if the `MissedVotes` is empty. 

```
if len(windows.MissedVotes)  == 0 {
    windows.MissedVotes = append(windows.MissedVotes, make([]int64, charts.StartPOS / charts.DiffInterval)...)
}
```
*untested

It's even technically correct that there were zero missed tickets during those windows.

db/cache/charts.go
```diff
@@ -275,33 +276,49 @@ func newDaySet(size int) *zoomSet {
...
 // Snip truncates the windowSet to a provided length.
 func (set *windowSet) Snip(length int) {
 	if length < 0 {
 		length = 0
 	}
+	missedSize := length - set.stakeValid
+	if missedSize < 0 {
+		missedSize = 0
+	}
+
```
buck54321: See a possible simpler solution in the my comment on the appender.

db/cache/charts.go
```diff
@@ -416,7 +435,11 @@ func (charts *ChartData) Lengthen() error {
 	}
 
 	windows := charts.Windows
-	shortest, err = ValidateLengths(windows.Time, windows.PowDiff, windows.TicketPrice, windows.StakeCount)
+	// expectedMissedSlice shows the full slice size for the missed votes slice if
+	// the stake validation height was zero.
+	expectedMissedSlice := append(make(ChartUints, windows.stakeValid), windows.MissedVotes...)
```
buck54321: Generating slice just to spoof validation is not great. The zero-prepending solution solves this too.

db/dcrpg/internal/stakestmts.go
```diff
@@ -342,6 +342,16 @@ const (
...
 			AND blocks.is_mainchain = TRUE;`
 
+	SelectMissesVotesChartData = `SELECT blocks.height, blocks.time, count(*)
+		FROM blocks
+		JOIN misses
+		ON misses.block_hash = blocks.hash
+		WHERE blocks.height > $1
+		AND blocks.is_valid
+		AND blocks.is_mainchain
+		GROUP BY blocks.height, blocks.time
```
buck54321: Do you need the `blocks.time` here?

public/js/controllers/charts_controller.js
```diff
@@ -141,15 +141,16 @@ function nightModeOptions (nightModeOn) {
   }
 }
 
-function zipXYData (gData, isHeightAxis, isDayBinned, coefficient, windowS) {
-  coefficient = coefficient || 1
+function zipXYData (gData, isHeightAxis, isDayBinned, coefficient, windowS, initValue) {
```
buck54321: No need for this if you prepend the zeros.

dmigwi: On this I think `initValue` makes sense as plotting zero missed votes before the stake validation height doesn't sound so good.

buck54321: If you really want to truncate the zeros, do it in the `ChartMaker`. That's the place for pre-processing of datasets.

db/cache/charts.go
```diff
@@ -1164,3 +1184,12 @@ func poolValueChart(charts *ChartData, bin binLevel, axis axisType) ([]byte, err
 	}
 	return nil, InvalidBinErr
 }
+
+func missedVotesChart(charts *ChartData, _ binLevel, _ axisType) ([]byte, error) {
+	stakeValidWindows := int(charts.StartPOS / charts.DiffInterval)
+	if stakeValidWindows >= len(charts.Windows.MissedVotes) {
+		stakeValidWindows = 0
+	}
+	return charts.encode(charts.Windows.Time[stakeValidWindows:],
```
chappjc: Please add a check that `stakeValidWindows` is in range.

db/cache/charts_test.go
```diff
@@ -45,7 +45,7 @@ func TestMain(m *testing.M) {
 func TestChartsCache(t *testing.T) {
 	gobPath := filepath.Join(tempDir, "log.gob")
 	ctx, shutdown := context.WithCancel(context.Background())
-	charts := NewChartData(0, time.Unix(0, 0), &chaincfg.MainNetParams, ctx)
+	charts := NewChartData(ctx, 0, time.Unix(0, 0), &chaincfg.MainNetParams)
```
chappjc: @buck54321 Having context as the first arg is good practice.  Might as well make the same change in https://github.com/decred/dcrdata/pull/1384 too if you decide to modify the signature of `NewChartData`

db/dcrpg/internal/stakestmts.go
```diff
@@ -342,6 +342,16 @@ const (
...
 
+	SelectMissesVotesChartData = `SELECT blocks.height, count(*)
+		FROM blocks
+		JOIN misses
+		ON misses.block_hash = blocks.hash
+		WHERE blocks.height > $1
+		AND blocks.is_valid
+		AND blocks.is_mainchain
+		GROUP BY blocks.height
+		ORDER BY blocks.height;`
```
chappjc: ~70ms with warm cache.  Works for me.

db/cache/charts.go
dmigwi: I thought this should work:
```golang
        if stakeValidWindows >= len(charts.Windows.MissedVotes) {
		stakeValidWindows = 0
	}
```

dmigwi: Adding `charts.Windows.Time` too.

dmigwi: This one should work:
```golang
	if stakeValidWindows >= len(charts.Windows.MissedVotes) ||
		stakeValidWindows >= len(charts.Windows.Time) {
		stakeValidWindows = 0
	}
```

PR #1367, opened 2019-06-04
proposal listing: Add tooltip to the vote status

dmigwi:
Added a tooltip to the proposal vote status column for the entries showing the approval percentage.

Screenshot from 2019-06-04 14-50-54

Added the `In Discussion` vote status to include the period when the proposal has been approved but no votes exist yet.


chappjc:
I'm OK with this, but I would think this "authorized but lacking any votes" state would be "Authorized" or "In Progress" (without the percent approval).

PR #1368, opened 2019-06-04
explorer/charts: async line plotter

buck54321:
This has some conflicts with recent commits and there will be more once @dmigwi's PRs are merged. I'll resolve them all after that. Experimental and WIP. 

Very large datasets can cause the browser to become unresponsive for up to 10s of seconds while Dygraphs processes and plots. This PR introduces an asynchronous plotter and some synchronization utilities. Plotting and data crunching from the event loop enables continued interaction with page elements, and even selection of different data sets or zoom levels before the current set is finished plotting. The Job class adds synchronization of functions and loops and context cancellation.


buck54321:
I may swing back to this at a later date. Closing for now.

PR #1372, opened 2019-06-05
sync: Disable multiple home page redirects

dmigwi:
The "redirecting in 20 seconds" function keeps redirecting to the homepage every couple of seconds (until you close the tab and open a new one) after the db sync is completed.

### DIFF
--- a/public/js/controllers/status_controller.js
+++ b/public/js/controllers/status_controller.js
@@ -63,6 +63,8 @@ function doNotification () {
   newBlockNtfn.show()
 }
 
+let hasRedirected = false
+
 export default class extends Controller {
   static get targets () {
     return ['statusSyncing', 'futureBlock', 'init', 'address', 'message']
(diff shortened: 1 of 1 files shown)


public/js/controllers/status_controller.js
```diff
@@ -63,6 +63,8 @@ function doNotification () {
   newBlockNtfn.show()
 }
 
+let isRedicting = true
```
chappjc: `isRedirecting`? (spelling)

But since it starts as true, the variable is actually "not redirecting".  Perhaps you wanted to start with false, negate the test, and then set to true the first time it recognizes `'sync complete'`?

public/js/controllers/status_controller.js
```diff
@@ -84,13 +86,14 @@ export default class extends Controller {
         while (bar.firstChild) bar.removeChild(bar.firstChild)
         bar.innerHTML = buildProgressBar(v)
 
-        if (v.subtitle === 'sync complete') {
+        if (v.subtitle === 'sync complete' && isRedicting) {
+          isRedicting = false // block consecutive calls.
```
chappjc: Can this event handler be run multiple times concurrently?  I don't know the limits of JS, but this would be a data race in go.

dmigwi: I have tested this possibility several time but `hasRedirected` variable set outside the controller implementation ensures that redirecting function only runs once.

PR #1378, opened 2019-06-06
gov: VoteTracker handle no vote info from node

buck54321:
When dcrd has no deployments, as in simnet, `VoteTracker` registered that as an error, which was wrong.


chappjc:
Reworking things to do without Deployments from the chainparams.Params struct can be in a follow-up PR.

PR #1381, opened 2019-06-07
piparser: Add piparser disable flag

### LINKED ISSUE #1374: need option to disable politeia code
On simnet, the politeia code makes no sense.  The following should be skipped along with all other Pi operations:

```
[INF] DATD: Setting up the Politeia's proposals clone repository. Please wait...
```

dmigwi:
A new flag to disable Piparser and Politea API query is added.

 `--disable-piparser` ensures that `/proposals` and `/proposal` page returns:

Screenshot from 2019-06-07 15-04-50

The API returns:
```bash
$ curl http://localhost:7777/api/proposal/a4f2a91c8589b2e5a955798d6c0f4f77f2eec13b62063c5f4102c21913dcaf32 
piparser is disabled.
```

Fixes #1374

### DIFF
--- a/db/dbtypes/go.sum
+++ b/db/dbtypes/go.sum
@@ -97,6 +97,7 @@ github.com/decred/dcrdata/db/cache v1.0.1/go.mod h1:WPkmV72Cal2OOjt4bG35XrjzSqsp
 github.com/decred/dcrdata/db/dbtypes v1.0.1/go.mod h1:d9qVcMl+l5rOEY7zJB5nqQ0WDEHof9vWlW/9hpK0xpc=
 github.com/decred/dcrdata/db/dbtypes v1.0.2-0.20190416155607-25afe5d5e790/go.mod h1:yPNF/Tsj3pLBxofu2Pr70hqejMYpCZHvcudkexvq0hQ=
 github.com/decred/dcrdata/db/dbtypes v1.0.2-0.20190416204615-70a58657e02f/go.mod h1:yPNF/Tsj3pLBxofu2Pr70hqejMYpCZHvcudkexvq0hQ=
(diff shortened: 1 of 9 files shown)


db/dbtypes/go.sum
```diff
@@ -149,6 +150,7 @@ github.com/kr/pretty v0.1.0/go.mod h1:dAy3ld7l9f0ibDNOQOHHMYYIIbhfbHSm3C4ZsoJORN
 github.com/kr/pty v1.1.1/go.mod h1:pFQYn66WHrOpPYNljwOMqo10TkYh1fy3cYio2l3bCsQ=
 github.com/kr/text v0.1.0 h1:45sCR5RtlFHMR4UwH9sdQ5TC8v0qDQCHnXt+kaKSTVE=
 github.com/kr/text v0.1.0/go.mod h1:4Jbv+DJW3UT/LiOwJeYQe1efqtUx/iVham/4vfdArNI=
+github.com/lib/pq v1.0.0/go.mod h1:5WUZQaWbwv1U+lTReE5YruASi9Al49XbQIvNi/34Woo=
```
chappjc: We'll be updating all the module file shortly, so let's revert these changes for now.

chappjc:
Sorry for the conflict.  Just need to have Politeia capitalized in config.go on those two lines.

dmigwi:
I could add the automatic disabling while on simnet functionality.

db/dbtypes/go.sum
dmigwi: Seems like `github.com/decred/dcrdata/db/dcrpg` has an older version of that package that won't disappear unless I update package `dcrpg` too.

chappjc: It's because of the one "blackbox" test file.  No worries.

PR #1384, opened 2019-06-07
cache/charts: check for genesis time 0

buck54321:


chappjc:
Genesis time was one of the things I ran realized was funny. Although I think the `NewChartData` call sites are part of the problem too.  e.g. In `main`, instead of:

```go
charts := cache.NewChartData(uint32(heightDB), time.Unix(pgDB.GenesisStamp(), 0),
	activeChain, ctx)
```

I changed it to:

```go
gt := activeChain.GenesisBlock.Header.Timestamp
charts := cache.NewChartData(uint32(heightDB), gt, activeChain, ctx)
```

But given that the chainparams GenesisBlock header has the timestamp, I think we can actually that constructor input arg and use the Params that are already passed to it.

I worked back from the panic in `Avg`, discovering that the panic was from on of the `intervals` being `[2]int{-1, 0}`.  It takes some mental gymnastics to work backwards and figure out how first iteration of the following loop would hit `t>next` while `startIdx` still being -1.

```go
next := start + aDay
startIdx := -1
for i, t := range blocks.Time {
	fmt.Println(i, t, next, startIdx)
	if t < start {
		continue
	} else if t >= next {
		intervals = append(intervals, [2]int{startIdx, i})
		days.Time = append(days.Time, start)
		start = next
		next += aDay
		startIdx = i
		if t > end {
			break
		}
	} else if startIdx == -1 {
		startIdx = i
	}
}
```

Can this loop be made any more intuitive, and maybe check for pathological conditions like the one we hit?

buck54321:
Oh. I didn't realize the timestamp was in the params. Nice. 

Yeah. That loop is unintuitive, but it's pretty efficient. I'll look into clarifying it.

buck54321:
Nevermind. That loop sucked. It's better now. 

This still needs simnet and dbload testing.

buck54321:
Looks good on simnet/reorg testing. Dbload testing is probably not crtitical, and will need to wait anyway as that stuff needs to be updated to the new db schema changes (meta table).

PR #1390, opened 2019-06-12
views/explorer: improve address page memory utilization

chappjc:
Three main chainges:

* views: Reduce byte size of `addressTable` template html.

Also avoid the unneeded template if and range variables.

* explorer: reduce max address rows

* explorer: use `strings.Builder` instead of `bytes.Buffer`

In `(*templates).execTemplateToString`, use `strings.Builder` instead of
`bytes.Buffer` to avoid an unnecessary copy when calling `String()`.
TODO: look into preallocating its buffer.

### DIFF
--- a/views/address.tmpl
+++ b/views/address.tmpl
@@ -4,9 +4,9 @@
 
 {{template "html-head" printf "Decred Address %s" .Data.Address}}
     {{template "navbar" . }}
-    {{with .Data}}
-    {{$TxnCount := add .TxnCount .NumUnconfirmed}}
-    {{$txType := .TxnType}}
+    {{- with .Data}}
+    {{- $TxnCount := add .TxnCount .NumUnconfirmed}}
+    {{- $txType := .TxnType -}}
     <div class="container main"
       data-controller="address newblock"
       data-address-offset="{{.Offset}}"
@@ -16,11 +16,11 @@
     >
     <div class="row pb-4 px-2">
       <div class="col-24 col-xl-10 bg-white px-3 py-3 position-relative">
-          {{if eq .Address $.DevAddress}}
+          {{- if eq .Address $.DevAddress}}
               <div class="fs22 pb-3">Decred Treasury</div>
-          {{else}}
+          {{- else}}
               <div class="fs22 pb-3">Address</div>
-          {{end}}
+          {{- end}}
           <div class="text-left d-flex align-items-start flex-wrap">
             <div class="fs15 font-weight-bold break-word d-inline-block hash-box mb-3" data-target="address.addr">{{.Address}}</div>
             <a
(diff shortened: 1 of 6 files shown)


views/extras.tmpl
```diff
@@ -1,368 +1,358 @@
...
-                    <td class="text-right">unspent</td>
-                {{end}}
+			{{else if or (eq $txType "credit") .IsFunding}}  <!-- .IsFunding = true && txType = "all" is a credit -->
+				<td class="text-right fs15">
+					{{template "decimalParts" (float64AsDecimalParts .ReceivedTotal 8 false)}}
+				</td>
+				{{if ne .MatchedTx ""}}
+				<td class="text-right"><a href="/tx/{{.MatchedTx}}/in/{{.MatchedTxIndex}}"
+					data-txid="{{.MatchedTx}}"
+					data-action="mouseover->address#hashOver mouseout->address#hashOut"
```
chappjc: This may seem like madness, but there were a ton of spaces and the html is ~2 kB smaller just for the `addressTable` template.

views/extras.tmpl
```diff
@@ -1,368 +1,358 @@
...
-          {{end}}
-          <th class="d-none d-sm-table-cell text-right">Time</th>
-          <th class="text-right">Age</th>
-          <th class="text-right"><span class="d-sm-none position-relative" data-tooltip="Confirmations">Cons</span><span class="d-none d-sm-inline">Confirms</span></th>
-          <th class="d-none d-sm-table-cell text-right">Size</th>
-      </thead>
-      <tbody>
-          {{range $i, $v := .Transactions}}
-          <tr {{if eq .Confirmations 0}} data-target="address.pending" data-txid="{{.TxID}}" {{end}}>
-              {{with $v}}
```
chappjc: Doing `{{range .Transactions}}` makes the current value (was `$v`) become `.`.

views/address.tmpl
```diff
@@ -260,11 +260,12 @@
...
             <option {{if eq $Txlen 20}}selected{{end}} value="20"{{if lt $TxnCount 20}} disabled{{end}}>20</option>
-            <option {{if eq $Txlen 100}}selected{{end}} value="100"{{if lt $TxnCount 100}} disabled{{end}}>100</option>
-              {{if lt $TxnCount 1000}}
-                  <option{{if eq $Txlen $TxnCount}} selected{{end}} value="{{$TxnCount}}"{{if le $TxnCount 100}} disabled{{end}}>{{$TxnCount}}</option>
+            <option {{if eq $Txlen 40}}selected{{end}} value="40"{{if lt $TxnCount 40}} disabled{{end}}>40</option>
+            <option {{if eq $Txlen 80}}selected{{end}} value="80"{{if lt $TxnCount 80}} disabled{{end}}>80</option>
+              {{if lt $TxnCount 160}}
+                  <option{{if eq $Txlen $TxnCount}} selected{{end}} value="{{$TxnCount}}"{{if le $TxnCount 160}} disabled{{end}}>{{$TxnCount}}</option>
               {{else}}
-                  <option{{if eq $Txlen 1000}} selected{{end}} value="1000">1000</option>
```
chappjc: Could get OOM in certain situations.

views/extras.tmpl
buck54321: It's only madness because github treats a tab as 8 spaces, which is... excessive, IMO. Looks great in my IDE.

chappjc:
Good call about putting comments in the template nodes.  And great suggestion about the white space scrubbers.  I had not noticed that function was added in go 1.6.  Really useful, especially for the micro-templates like `decimalParts` and `hashelide`.  I'm gonna give that a go.

chappjc:
I neglected to update the address controller, which has the hard limit of 1000 set there too.

Also, I added two commits that use white space scrubbing.

views/extras.tmpl
```diff
@@ -1,369 +1,341 @@
...
+	<script
+		src="/dist/js/app.bundle.js"
+		data-turbolinks-eval="false"
+		data-turbolinks-suppress-warning
+	></script>
 </footer>
 {{end}}
 
-{{define "decimalParts"}}
+{{define "decimalParts" -}}
```
chappjc: `decimalParts`, `fmtPercentage`, and `hashElide` are "microtemplates" that get executed perhaps thousands of times, plus the indentation is very unlikely to match the surrounding html.  Thus, I think it's best to have these execute into html on a single line.

chappjc:
Before:

```
[DBG] EXPR: "address" template HTML size: 55.40 kiB (DsfBwMjVusgQu93t45TAmhsxbmjp3MAPo63, all, 20)
Next
[DBG] EXPR: "addresstable" template HTML size: 35.28 kiB (DsfBwMjVusgQu93t45TAmhsxbmjp3MAPo63, all, 20)
100
[DBG] EXPR: "addresstable" template HTML size: 181.04 kiB (DsfBwMjVusgQu93t45TAmhsxbmjp3MAPo63, all, 100)
1000 (877)
[DBG] EXPR: "addresstable" template HTML size: 1673.86 kiB (DsfBwMjVusgQu93t45TAmhsxbmjp3MAPo63, all, 877)
160
[DBG] EXPR: "address" template HTML size: 314.95 kiB (DsfBwMjVusgQu93t45TAmhsxbmjp3MAPo63, all, 160)
```

After:

```
[DBG] EXPR: "address" template HTML size: 41.55 kiB (DsfBwMjVusgQu93t45TAmhsxbmjp3MAPo63, all, 20)
Next
[DBG] EXPR: "addresstable" template HTML size: 23.32 kiB (DsfBwMjVusgQu93t45TAmhsxbmjp3MAPo63, all, 20)
100
[DBG] EXPR: "address" template HTML size: 142.14 kiB (DsfBwMjVusgQu93t45TAmhsxbmjp3MAPo63, all, 100)
160
[DBG] EXPR: "addresstable" template HTML size: 197.63 kiB (DsfBwMjVusgQu93t45TAmhsxbmjp3MAPo63, all, 160)
```

PR #1395, opened 2019-06-13
[4.1] dcrpg/dbtypes: fix first uncached load of AddressHistory

chappjc:
Backport of https://github.com/decred/dcrdata/pull/1394 to 4.1

The output of `(*ChainDB).updateAddressRows` was not being sliced in
`(*ChainDB).AddressHistory`.  This change select the correct type and
range of address rows, merging if needed.  This only happens on a cache
miss.

This adds `SliceAddressRows` and variants, and `MergeRowsRange` to dbtypes.


chappjc:
Fix confirmed working on 4.1.

PR #1424, opened 2019-06-15
chart: Update duration between blocks chart

dmigwi:
The old version of the chart plotted **duration btw blocks vs height/block time**. This format was found to be inefficient especially when data plotted was grouped by any zoom type other than `blocks`. Further data processing was required to make the data more useful.

It was decided that a graph plotting **Count of blocks sharing similar duration vs exact duration** would be more effective. The data to be plotted was supposed to be similar to [this reddit table](https://www.reddit.com/r/decred/comments/8dszhf/expected_versus_actual_block_production_times/) only that the new chart would plot the actual block count at the given duration interval instead of the percentage exponential growths shown.

Screenshot from 2019-06-16 00-15-45

- The green bars plot the data from a P.M.F of the Poisson Distribution. (Expected Blocks Count)
Screenshot from 2019-06-13 18-14-22
- The blue line plots the actual count block vs duration (Actual Blocks Count)
- X-axis data is grouped in buckets of size (interval) 30 Secs. This plots the maximum bars for the Poisson distribution.
- The mean of the distribution is a approx. 300 i.e `μ = 300`
- lambda = `μ / bucketSize`
- event x = `x-axis value / bucketSize`  *(was divided by the `bucketSize` to help reduce the value to be used in the P.M.F of the Poission distribution)*. 

It was expected that both the Poisson distribution and the actual data, would have almost similar bell(n)-shaped graphs but that isn't the case.

I could use some help to understand why the two charts plotted are too different!


dmigwi:
New charts control buttons were added to help limit the data plotted. When `Year` is selected, data to be plotted is of the blocks made in the latest year ending at the current best block.

Screenshot from 2019-06-12 01-39-18

raedah:
The data for the blue line, listed as Actual Block Count, cant be correct. Its showing most blocks occurring in less then 100 seconds. We know this is not the case because mainnet has a block target (average) time of 5 min, 300 seconds.

davecgh:
I'm really not sure what the intent to show here is, but the idea of there being a bell curve is not at all the case in a Poisson distribution with one event per interval, which is what block production is.  As I mentioned in Matrix, block production is the special case of a Poisson distribution with `λ = 1` and `k = 0` which results in an exponential distribution.  That green chart is plotting multiple events in an interval which is not at all modelling how block production works.

The formula you want to use for the expected percentage of blocks produced is the CDF for an exponential distribution with a target mean of 300 seconds (5 minutes) due to the aforementioned.  In particular, because you're trying to produce a chart that shows what percentage of blocks are expected to be produced within a given time frame you'll want to use `λ = 1/300` in the exponential CDF which results in the equation:

image

For example, if you plug in 10 for x, you'll get ~3.28% which is the expected number of blocks to be produced within 10 seconds.  Similarly, if you plug in 300 for x, you'll get ~63.21% which is the expected number of blocks to be produced within 5 minutes (300 seconds).

Therefore, since there are currently around 354000 total blocks, you would expect around 223771 of them to have been produced within 5 minutes.

dmigwi:
I had earlier drawn a chart using the exponential C.D.F.  for the expected blocks count percent (in green bars), the blue line shows the actual cummulative data. @davecgh It this the chart you are referring to? Should it be the appropriate chart?

Screenshot from 2019-06-12 22-49-45
.

raedah:
Using the same data from the reddit post, it is possible to calculate each rows proportion of the overall blocks as I have done below. Showing the proportion of the total would be a useful way to represent the data visually. Note that the distribution is entirely dependent on the different amounts of time chosen for the elapsed amount per row, ranging from 10 seconds to 10 minutes. Its not immediately clear what would be the best way to choose the value for the elapsed time per row. Since its an exponential distribution, having the elapsed time increase exponentially per row makes sense.

Elapsed	|Expected  |Proportion of Total
--------------|---------------|---------------------------
10 sec	|3.28%	  |9.52-3.28 = 6.24%
30 sec	|9.52%	  |18.13-9.52 = 8.61%
1 min	|18.13%	  |32.97-18.13 = 14.84%
2 min	|32.97%	  |45.12-32.97 = 12.15%
3 min	|45.12%	  |55.07-45.12 = 9.95%
4 min	|55.07%	  |63.21-55.07 = 8.14%
5 min	|63.21%	  |69.88-63.21 = 6.67%
6 min	|69.88%	  |76.34-69.88 = 6.46%
7 min	|75.34%	  |75.34-69.88 = 5.46%
10 min	|86.47%	  |86.47-75.34 = 11.13%
20 min	|98.17%	  |98.17-86.47 = 11.7%
30 min	|99.75%	  |99.75-98.17 = 1.58%
greater than 30 min | 100% |0.25%

davecgh:
That looks accurate, although I agree it isn't really super readable because the actual is so close to the expected (which is a good thing of course).

Since I believe the overall idea should be to highlight any deviations from the expected, it might make sense to show it as a variance chart with a shaded exponential curve.  Similar to my ticket vote chance by number of days chart, except it would have time on the x axis and percentage on the y axis.  Then you could so the expected curve with the actual curve overlaid in a different color.

raedah:
I like @davecgh's idea better since it doesn't require picking different elapsed time values per row. Here is the chart being referenced. 
ticket_vote_chance_by_number_of_days

chappjc:
@dmigwi The example I had linked to was in the `happy` API endpoint PR where we're trying to guess with high certainty if the node or dcrdata are stuck at an old block.  https://github.com/decred/dcrdata/pull/1321#discussion_r283856216

davecgh:
> 
> 
> Using the same data from the reddit post, it is possible to calculate each rows proportion of the overall
> ....
> Elapsed 	Expected 	Proportion of Total
> ....
> 30 min 	99.75% 	99.75-98.17 = 1.58%
> greater than 30 min 	100% 	0.25%

A correction.  Greater than 30 minutes isn't 100%.  It technically never reaches 100%, but if you want to round to 2 decimal places, as in the remaining cases, it would be at `2972` seconds which results in `0.9999501...` that rounds to `99.995%` -> `100%`.

That said, I would suggest an upper bound that results in at least `99.9999%` (without rounding) of all blocks having been produced so that it will only be one in a million that is not.  That happens at `4155` seconds, but for a more round number for human consumption purposes, something like `1 hour and 10 minutes = 70 minutes = 4200 seconds` would be good.  `4200` seconds gives `99.999917%`.

dmigwi:
Below is the finalized duration btw blocks chart that shows the actual (in blue) and expected (in green) count of blocks to be mined on a given time interval.
Screenshot from 2019-06-29 08-16-06

The expected blocks count was calculated using the formula below:

render

chappjc:
Thanks for the update, and apologies for the delay in review on my end.  We'll get to it asap!

db/cache/charts.go
```diff
@@ -959,57 +981,76 @@ func accumulate(data ChartUints) ChartUints {
...
+
+		xValue = append(xValue, timeVal)
+		blockCount = append(blockCount, tracker[timeVal])
+		expectedDistr = append(expectedDistr, math.Floor(expectedCount*1e2)/1e2)
+
+		lastDistr = distr
+		lastTimestamp = timeInterval
 	}
-	return times, avgDiffs
+	return xValue, blockCount, expectedDistr
```
chappjc: Let's add some unit tests for `blockTimes`, covering all the edge cases we can think of.

e.g.

```go
func Test_blockTimes(t *testing.T) {
	type args struct {
		blocks ChartUints
		limit  int
	}
	blockSeconds := ChartUints{1, 20, 33, 43, 56, 60, 79}
	tests := []struct {
		name  string
		args  args
		want0 ChartUints
		want1 ChartUints
		want2 ChartFloats
	}{
		{
			name: "empty",
			args: args{
				blocks: nil,
				limit:  0,
			},
			want0: ChartUints{},
			want1: ChartUints{},
			want2: ChartFloats{},
		},
		{
			name: "basicOK6",
			args: args{
				blocks: blockSeconds,
				limit:  6,
			},
			want0: ChartUints{4, 10, 13, 19},
			want1: ChartUints{1, 1, 2, 1},
			want2: ChartFloats{0.31, 0.18, 0.14, 0.08},
		},
		{
			name: "basicOK7",
			args: args{
				blocks: blockSeconds,
				limit:  7,
			},
			want0: ChartUints{4, 10, 13, 19},
			want1: ChartUints{1, 1, 2, 2},
			want2: ChartFloats{0.35, 0.22, 0.17, 0.11},
		},
		{
			name: "allBig",
			args: args{
				blocks: blockSeconds,
				limit:  999999,
			},
			want0: ChartUints{4, 10, 13, 19},
			want1: ChartUints{1, 1, 2, 2},
			want2: ChartFloats{0.35, 0.22, 0.17, 0.11},
		},
		{
			name: "all0",
			args: args{
				blocks: blockSeconds,
				limit:  0,
			},
			want0: ChartUints{4, 10, 13, 19},
			want1: ChartUints{1, 1, 2, 2},
			want2: ChartFloats{0.35, 0.22, 0.17, 0.11},
		},
		{
			name: "1",
			args: args{
				blocks: blockSeconds,
				limit:  1,
			},
			want0: ChartUints{},
			want1: ChartUints{},
			want2: ChartFloats{},
		},
		{
			name: "2",
			args: args{
				blocks: blockSeconds,
				limit:  2,
			},
			want0: ChartUints{19},
			want1: ChartUints{1},
			want2: ChartFloats{0.01}, // Is this right???
		},
	}
	for _, tt := range tests {
		t.Run(tt.name, func(t *testing.T) {
			got, got1, got2 := blockTimes(tt.args.blocks, tt.args.limit)
			if !reflect.DeepEqual(got, tt.want0) {
				t.Errorf("blockTimes() got = %v, want %v", got, tt.want0)
			}
			if !reflect.DeepEqual(got1, tt.want1) {
				t.Errorf("blockTimes() got1 = %v, want %v", got1, tt.want1)
			}
			if !reflect.DeepEqual(got2, tt.want2) {
				t.Errorf("blockTimes() got2 = %v, want %v", got2, tt.want2)
			}
		})
	}
}
```

public/js/controllers/charts_controller.js
```diff
@@ -597,7 +635,7 @@ export default class extends Controller {
   }
 
   isTimeAxis () {
-    return this.selectedAxis() === 'time'
+    return this.selectedAxis() === 'time' && customXLabel(this.chartSelectTarget.value) !== ''
```
chappjc: This test `customXLabel(this.chartSelectTarget.value) !== ''` is used in several functions to dictate different behaviors.  If this test were made into a function, what would it be called?  `specialXAxis` perhaps, or something more descriptive (I don't know)?  My concern is that it's just `'duration-btw-blocks'` right now in `altXLabel`, but it's set up to hold multiple chart types.  Would this test always be applicable to everything with any custom x axis label?

views/charts.tmpl
```diff
@@ -222,6 +225,17 @@
                         </li>
                     </ul>
                 </div>
+                <div class="chart-control chart-control-wrapper d-hide" data-target="charts.customAxisSelector">
+                    <ul class="nav nav-pills">
+                        <li class="nav-item active">
+                            <a
+                                class="nav-link active"
+                                href="javascript:void(0);"
```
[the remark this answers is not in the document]
chappjc: Maybe look into `preventDefault`.

db/cache/charts.go
```diff
@@ -959,57 +981,76 @@ func accumulate(data ChartUints) ChartUints {
...
 	}
-	last := blocks[0]
-	for _, v := range blocks[1:] {
+
+	last := blocks[k]
+	k++
+	tracker := make(map[uint64]uint64)
+	for _, v := range blocks[k:] {
 		dif := v - last
 		if int64(dif) < 0 {
```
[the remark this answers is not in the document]
dmigwi: diff is of type `uint64` which doesn't have negative values. conversion to `int64` enables us to detect a negative value.

db/cache/charts.go
```diff
@@ -959,57 +981,76 @@ func accumulate(data ChartUints) ChartUints {
...
+// minus the first element is returned for convenience. limit defines the number
+// of blocks to consider for the current limitType from the best block.
+func blockTimes(blocks ChartUints, limit int) (ChartUints, ChartUints, ChartFloats) {
 	dataLen := len(blocks)
+	keys := make(ChartUints, 0, len(blocks))
 	if dataLen < 2 {
 		// Fewer than two data points is invalid for btw. Return empty data sets so
 		// that the JSON encoding will have the correct type.
-		return times, times
+		return keys, keys, newChartFloats(len(blocks))
```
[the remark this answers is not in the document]
dmigwi: I will reduce the capacity to be equal to `dataLen` which will be a max of 1 then.

views/charts.tmpl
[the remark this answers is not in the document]
dmigwi: `event.preventDefault()` requires an event but `javascript:void(0);` seems to be accurately used here to prevent the browser from loading a new page. 

The following seem to achieve the same goal. @buck54321 can confirm this.
```html
<a href="/" onclick="return false">Click here</a>
or
<a href="/" onclick="event.preventDefault()">Click here</a>
or
<a href="javascript:void(0);" >Click here</a>
```

db/cache/charts.go
```diff
@@ -959,57 +981,76 @@ func accumulate(data ChartUints) ChartUints {
...
+	}
+
+	var k int
+	if limit > 0 && limit < len(blocks) {
+		k = len(blocks) - limit
 	}
-	last := blocks[0]
-	for _, v := range blocks[1:] {
+
+	last := blocks[k]
```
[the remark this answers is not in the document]
dmigwi:
```Golang
dataLen := len(blocks)
	if dataLen < 2 {
            ...
```

The above code in lines before that highlighted code snippet ensures that the case scenario described above will never happen.

db/cache/charts.go
```diff
@@ -959,57 +981,76 @@ func accumulate(data ChartUints) ChartUints {
...
+	k++
+	tracker := make(map[uint64]uint64)
+	for _, v := range blocks[k:] {
 		dif := v - last
 		if int64(dif) < 0 {
 			dif = 0
```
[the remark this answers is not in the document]
dmigwi: I don't think this logging is necessary as the negative value showing up is normal due to the up and down of the actual blocks count line graph plot.

db/cache/charts.go
```diff
@@ -958,58 +985,94 @@ func accumulate(data ChartUints) ChartUints {
...
-	times := make(ChartUints, 0, len(blocks))
+// blockTimes enables two line graph charts showing the expected and actual
+// blocks count mined by a given time duration to be generated.
+// It translates the times slice to a slice of differences. The original dataset
+// minus the first element is returned for convenience. limitBlocks defines the
+// number of blocks from the best block (most recent block mined) back towards
+// the genesis block that were supposed to be mined in the limit string passed as
+// the API limit parameter. When the limit string API parameter selected is 'all'
+// the limitBlocks value passed is 0 indicating that all blocks will be used.
+func blockTimes(blocks ChartUints, limitBlocks int) (ChartUints, ChartUints, ChartFloats) {
```
buck54321: Everything for the theoretical distribution should be done client side. No changes needed here.  The only values you need to plot the distribution are the target block time, which the charts controller already has, and the number of blocks, which is already encoded in your data. 

As a general rule, if you can avoid sending data by pushing the processing to JS, you should.

db/cache/charts.go
```diff
@@ -958,58 +985,94 @@ func accumulate(data ChartUints) ChartUints {
...
+		if _, ok := tracker[dif]; !ok {
+			keys = append(keys, dif)
+		}
+		tracker[dif]++
 	}
-	return blocks[1:], times
-}
-
-// Take the average block times on the intervals defined by the ticks argument.
-func avgBlockTimes(ticks, blocks ChartUints) (ChartUints, ChartUints) {
```
buck54321: Here again, no changes necessary. Do the theoretical distribution client-side.

db/cache/charts.go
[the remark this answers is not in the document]
chappjc: I don't like relying on integer underflow behavior.  Please convert v and last and compute dif directly in int64.

dmigwi:
@buck54321 Won't moving the theoretical calculation to the js side add undue pressure to the frontend? Remember the current implementation with the cache allows us to generate the data **once and reuse** it as many times as possible. With that suggested implementation, every page refresh would require that data to be **calculated afresh**.

I think having as many calculations as possible run on the multi-threaded backend will result to much more responsiveness for all the of devices that will be accessing that chart's page.

If you've got a reason why your suggested implementation is better, kindly share it!

buck54321:
> Won't moving the theoretical calculation to the js side add undue pressure to the frontend?

Not at all. I think if you run some tests, you'll see that this type of calculation is not very resource intensive. 

> Remember the current implementation with the cache allows us to generate the data once and reuse it as many times as possible.

You're doubling the size of every request. Downloading the extra data will add much more time to completion compared to calculating on the front-end.  

> I think having as many calculations as possible run on the multi-threaded backend will result to much more responsiveness for all the of devices that will be accessing that chart's page.

This may be true if you weren't increasing the size of the response. You are doubling the size of the response, which more than offsets any gains from pre-calculating.  

Calculating on the front end is a win-win. It's both faster for the client and easier on the server.

dmigwi:
@buck54321 Here is the latest update:

I have moved the expected block count calculations to the front-end as recommend to reduce the payload size sent from the back-end.

The payload *(Time Interval and Actual Block Count)* sent from the back-end has been compressed to only the necessary entries which will always range from 2k to 3k rows instead of the previous 300k and plus rows for all blocks. With this compressed data, the `"limit"` parameter passed to the back-end will still be processed in the back-end since the data available on the front-end cannot support its implementation on the front-end.

public/js/controllers/charts_controller.js
```diff
@@ -605,9 +669,20 @@ export default class extends Controller {
   }
 
   _zoomCallback (start, end) {
+    let currentSettings = {}
```
buck54321: This is not the place to be changing arbitrary information in the `settings` object. If you need to update the settings for a new chart, it should be done in `plotGraph`.

public/js/controllers/charts_controller.js
```diff
@@ -16,25 +16,26 @@ let Dygraph // lazy loaded on connect
...
-  return lineScales.indexOf(chart) > -1
-}
-
-function intComma (amount) {
-  return amount.toLocaleString(undefined, { maximumFractionDigits: 0 })
-}
+var isBinDisabled = (chart) => chartsToHideBin.indexOf(chart) > -1
+var isScaleDisabled = (chart) => chartsToHideScale.indexOf(chart) > -1
+var isAllControlsDisabled = (chart) => chartsToHideAll.indexOf(chart) > -1
+var customXLabel = (chart) => altXLabel[chart] || ''
```
buck54321: This comment applies to the general approach you've taken to handling the new axis here and below. By my count, you've added 10s of new functions to handle 1 new button/axis unit. I see that there is some special handling to be done for the new chart because of the new type of axis, but this approach is a bit tedious. Can't you just set the button visibility and null settings in the switch/case in `plotGraph`?

I'm happy to work with you towards a better solution here. Let me know if I can help.

public/js/controllers/charts_controller.js
```diff
@@ -605,9 +669,20 @@ export default class extends Controller {
   }
 
   _zoomCallback (start, end) {
+    let currentSettings = {}
+    if (this.isSpecialXAxis()) {
+      this.settings.limit = this.selectedZoom()
+      currentSettings = this.settings
+      this.query.nullEntry(currentSettings, ['zoom', 'axis', 'bin', 'scale'])
```
buck54321: previous two line are the same as `this.query.nullEntry(this.settings, ['zoom', 'axis', 'bin', 'scale'])`

In javascript, all objects and arrays are assigned by reference.

public/js/controllers/charts_controller.js
```diff
@@ -605,9 +669,20 @@ export default class extends Controller {
...
+      this.query.replace(currentSettings)
```
buck54321: Also identical to `this.query.replace(this.settings)`

public/js/controllers/charts_controller.js
```diff
@@ -605,9 +669,20 @@ export default class extends Controller {
...
+      return
+    }
+
     this.lastZoom = Zoom.object(start, end)
     this.settings.zoom = Zoom.encode(this.lastZoom)
-    this.query.replace(this.settings)
+    currentSettings = this.settings
+    this.query.nullEntry(currentSettings, ['limit'])
+    this.query.replace(currentSettings)
```
buck54321: See notes above about assignment by reference in javascript.

views/charts.tmpl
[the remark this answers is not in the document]
buck54321: No inline scripting. In fact, the use of the anchor element here is perplexing to begin with. Very un-semantic. Why aren't we just styling the list items directly. Or using button elements?

public/js/controllers/charts_controller.js
```diff
@@ -629,6 +708,42 @@ export default class extends Controller {
...
+    isBinControls ? binControl.add('d-hide') : binControl.remove('d-hide')
+  }
+
+  toggleAxisControls (isAxesControls) {
+    var dualAxes = this.dualAxesSelectorTarget.classList
+    var customAxis = this.customAxisSelectorTarget.classList
+    isAxesControls ? dualAxes.add('d-hide') : dualAxes.remove('d-hide')
+    isAxesControls ? customAxis.remove('d-hide') : customAxis.add('d-hide')
+  }
+
```
buck54321: Can we maybe reduce all of these functions to just one function that takes a list of elements to show and hides all the rest?

public/js/controllers/charts_controller.js
```diff
@@ -16,25 +16,26 @@ let Dygraph // lazy loaded on connect
...
-
-function intComma (amount) {
-  return amount.toLocaleString(undefined, { maximumFractionDigits: 0 })
-}
+var isBinDisabled = (chart) => chartsToHideBin.indexOf(chart) > -1
+var isScaleDisabled = (chart) => chartsToHideScale.indexOf(chart) > -1
+var isAllControlsDisabled = (chart) => chartsToHideAll.indexOf(chart) > -1
+var customXLabel = (chart) => altXLabel[chart] || ''
+var intComma = (amount) => amount.toLocaleString(undefined, { maximumFractionDigits: 0 })
+var VisibilityLabels = (chart) => chartsToShowVisibility[chart] || null
```
buck54321: This comment applies to the general implementation of the new axis. By my count, you've added 10+ new functions to handle a single new axis unit. I see that there is some special handling needed, but this approach is a bit tedious. Can't you just set the button visibility and `this.settings` values in `plotGraph`?

I'm happy to work through a better solution with you. Let me know?

public/js/controllers/charts_controller.js
```diff
@@ -537,35 +596,40 @@ export default class extends Controller {
...
     if (selectedChart !== selection || this.settings.bin !== this.selectedBin() ||
-      this.settings.axis !== this.selectedAxis()) {
-      let url = '/api/chart/' + selection
-      if (usesWindowUnits(selection)) {
-        this.binSelectorTarget.classList.add('d-hide')
-      } else {
-        this.binSelectorTarget.classList.remove('d-hide')
+      (this.settings.axis !== this.selectedAxis() && this.selectedAxis()) ||
+      isAllControlsDisabled(selection)) {
+      let searchParams = new URLSearchParams()
```
buck54321: I don't think that `URLSearchParams` works in IE. You could use the `Url` class from  `url-parse`.  We are already using the module in other places. https://www.npmjs.com/package/url-parse

buck54321:
I put together some suggestion for this PR. 

PR on your branch: https://github.com/dmigwi/dcrdata/pull/7

Comparison with master: https://github.com/decred/dcrdata/compare/master...buck54321:dmigwi?expand=1

dmigwi:
@buck54321 Kindly help me out here.

I have just experienced a massive conflict that was a little bit hard to fix, Kindly help me check if all the required functionality/code exists in perfect condition.

buck54321:
Sorry again for the delay here. Can you rebase again. I think you'll see another conflict from the database upgrade, which switched units for fees from DCR to atoms.

public/js/controllers/charts_controller.js
```diff
@@ -322,6 +327,19 @@ function missedVotesFunc (data) {
   return zipWindowHvY(data.missed, data.window, 1, data.offset * data.window)
 }
 
+function durationBTwBlocksFunc (rawData) {
+  var times = rawData.x
+  var counts = rawData.y
+  var blockCount = counts.reduce((total, n) => total + n)
```
chappjc: `rawData` has `t` and `duration` arrays, but no `x` or `y`.

chappjc: The API hit for this is `http://127.0.0.1:7777/api/chart/duration-btw-blocks?limit=all?bin=day`

dmigwi: Looking into this.

chappjc: Still broken in this function.  Are you able to address this, or should we put this PR on hold?

chappjc:
Labeled as "on hold", closing for now.

Issue #1405, opened 2019-06-14
Green line VISIBILITY - Charts 


Michae2xl:
Alpha -> Charts -> VISIBILITY (both selected) -> Week and day filter, the green line maybe disappear 

<img width="1244" alt="charts" src="https://user-images.githubusercontent.com/44928059/59520257-8ab78680-8e97-11e9-8a9d-5a269a5433a5.png">

chappjc:
Thanks for the reports.  This appears to be resolved by a fix in https://github.com/decred/dcrdata/pull/1426

Hopefully we'll have it merged shortly.

![image](https://user-images.githubusercontent.com/9373513/60110487-ae15e780-975b-11e9-85e5-7c5cafe95a62.png)

buck54321:
Resolved by #1426

Michae2xl:
Thanks, Masters!

PR #1426, opened 2019-06-18
dcrpg: combine tests and use a single TestMain

chappjc:
The `dcrpg` package tests no longer worked for various reasons.  This attempts to fix that.

The `chartests` build tag is removed in favor of `chartdata` (`go test . -v -tags chartdata`), and `mainnettest` is replaced by `pgonline` for all the tests. Note that `chartdata` is a subset of `chartdata`.  Also, a new build tag `fullpgdb` is added to use a different database rather than the test database from the bug-free-happiness repo.

This requires having only one `TestMain` (and no other redefined objects).  Further, the db/dbtypes test file is merged into the `dcrpg` tests with the `pgonline` tag.

The slog backends are enabled using stdout for both the `db/dcrpg` and `db/cache` packages.

The `testutil/dbconfig` module is modified so that the package in the `testutil/dbconfig` folder is now called `dbconfig` instead of `testsconfig`.  The various postgresql db config variables are renamed with the `PGTests` prefix.


chappjc:
The console output:

```
2019-06-18 12:34:55.546 [INF] dcrpg: Switching PostgreSQL time zone to UTC for this session.
2019-06-18 12:34:55.546 [INF] dcrpg: PostgreSQL 11.3 on x86_64-pc-linux-gnu, compiled by gcc (GCC) 8.3.0, 64-bit
2019-06-18 12:34:55.548 [INF] dcrpg: DB schema version 1.1.0
2019-06-18 12:34:56.030 [DBG] dcrpg: meta height 353294 / blocks height 353294
2019-06-18 12:34:56.030 [INF] dcrpg: Pre-loading unspent ticket info for InsertVote optimization.
2019-06-18 12:34:56.150 [INF] dcrpg: Storing data for 43724 unspent tickets in cache.
2019-06-18 12:34:56.156 [INF] dcrpg: Setting PostgreSQL DB statement timeout to 1h0m0s.
2019-06-18 12:34:56.171 [INF] dcrpg: Creating the "testing" table.
=== RUN   TestPgCharts
353295
0
2019-06-18 12:35:00.652 [WRN] db/cache: charts.ValidateLengths: dataset at index 1 has mismatched length 0 != 353295
0
2019-06-18 12:35:00.652 [WRN] db/cache: charts.ValidateLengths: dataset at index 2 has mismatched length 0 != 353295
353295
353295
353295
353295
2019-06-18 12:35:00.652 [WRN] db/cache: ChartData.Lengthen: block data length mismatch detected. truncating blocks length to 0
0
0
0
0
35330
35330
35330
--- FAIL: TestPgCharts (4.48s)
    pgblockchain_charts_test.go:43: &{0 [] [] [] [] [] [] [] [] []}
    pgblockchain_charts_test.go:44: unexpected empty blocks data
FAIL
FAIL	github.com/decred/dcrdata/db/dcrpg/v3	5.130s
```

buck54321:
Took me a while, but I think I know what's happening. If I'm remembering correctly, `(ChartData).Lengthen` did not used to be called at the end of `(ChartData).Update`, but was invoked separately from somewhere else. At one point, I think it was even accidentally removed, but then found it's place in `Update`. The tests were not calling `Lengthen` either, but were checking the subset of data from the individual DB packages, with custom validators defined at the top of `TestPgCharts` and `TestSqliteCharts`. 

Quick and dirty way would be to add a boolean argument to `(ChartData).Update` that would skip the call to `Lengthen`. There's probably something more elegant though.

Also, the test PG DB didn't have a row in the transactions table for the genesis block, causing the coin supply datasets to be short an element. It was a typo. I opened https://github.com/dcrlabs/bug-free-happiness/pull/5 to fix it.

chappjc:
How about in the test, `stubDb.RegisterCharts(charts)`, where `stubDb` has updaters for `"fee info"` and `"pool info"` that do no more than set the expected length of the `Fees`, `PoolSize`, and `PoolSize` slices?

chappjc:
BTW, it seems that the `Fees` slice is not included in any of the `ValidateLengths` calls in `(*ChartData).Lengthen`.  Is that a bug or is the `Fees` slice expected to be a different length than the others?

chappjc:
FYI, I have commits to push that fix several bugs with charts data, including the misses data set.

db/dcrpg/pgblockchain_charts_test.go
```diff
@@ -72,26 +39,45 @@ func TestMain(m *testing.M) {
...
-		_, err := cache.ValidateLengths(blocks.Time, blocks.Chainwork, blocks.TxCount, blocks.BlockSize)
+		_, err := cache.ValidateLengths(blocks.Time, blocks.Chainwork,
+			blocks.TxCount, blocks.BlockSize, blocks.Fees, blocks.PoolSize,
+			blocks.PoolValue)
 		if err != nil {
 			t.Fatalf("%s blocks length validation error: %v", tag, err)
 		}
-		_, err = cache.ValidateLengths(windows.TicketPrice, windows.PowDiff, windows.Time)
+		_, err = cache.ValidateLengths(windows.TicketPrice, windows.PowDiff,
+			windows.Time, windows.StakeCount, windows.MissedVotes)
```
chappjc: The test now validates all zoom and window datasets.

db/dcrpg/pgblockchain_charts_test.go
```diff
@@ -72,26 +39,45 @@ func TestMain(m *testing.M) {
 // It also checks the order and duplicates in the x-axis dataset.
 func TestPgCharts(t *testing.T) {
 	charts := cache.NewChartData(context.Background(), 0, &chaincfg.MainNetParams)
-	charts.DiffInterval = 10
```
chappjc: Setting this is no longer OK since `DiffInterval` and other `chaincfg.Params` fields are used in the Windows data sets.

db/cache/charts.go
```diff
@@ -410,11 +414,12 @@ func (charts *ChartData) Lengthen() error {
 
 	// Make sure the database has set an equal number of blocks in each data set.
 	blocks := charts.Blocks
-	shortest, err := ValidateLengths(blocks.Time, blocks.PoolSize,
-		blocks.PoolValue, blocks.BlockSize, blocks.TxCount,
-		blocks.NewAtoms, blocks.Chainwork)
+	shortest, err := ValidateLengths(blocks.Height, blocks.Time,
```
dmigwi: `blocks.Height` is unnecessary since the data stored is the same as the indexes of any array in the `blocks` binType.
This is how the frontend shows the height axis even without a `blocks.Height` in the backend.

db/dcrpg/pgblockchain_test.go
```diff
@@ -50,29 +54,63 @@ func (m MemStats) String() string {
...
 	return cleanUp, err
 }
 
 func TestMain(m *testing.M) {
+	// Setup dcrpg logger.
+	UseLogger(slog.NewBackend(os.Stdout).Logger("dcrpg"))
+	log.SetLevel(slog.LevelTrace)
+
+	// Setup db/cache logger.
+	cache.UseLogger(slog.NewBackend(os.Stdout).Logger("db/cache"))
```
chappjc: There was no logging from either dcrpg or cache, so we were missing the warnings about failed validation, which did not result in an error.  @buck54321 I'm wondering if we should have error values back from `Update`, etc.

db/dcrpg/queries.go
```diff
@@ -3278,6 +3278,7 @@ func appendChartBlocks(charts *cache.ChartData, rows *sql.Rows) error {
 			// datasets sized properly.
 			bigwork = big.NewInt(int64(blocks.Chainwork[len(blocks.Chainwork)-1]))
 		}
+		blocks.Height = append(blocks.Height, height)
```
chappjc: Bug fix 7: `Blocks.Height` was never updated, so length validation with it was not possible.

db/dcrpg/pgblockchain_charts_test.go
```diff
@@ -72,26 +39,45 @@ func TestMain(m *testing.M) {
...
 	blocks := charts.Blocks
 	windows := charts.Windows
 
 	validate := func(tag string) {
 		// Not checking NewAtoms right now, as the test database does not appear to
 		// contain stakebase and coinbase vins.
-		_, err := cache.ValidateLengths(blocks.Time, blocks.Chainwork, blocks.TxCount, blocks.BlockSize)
+		_, err := cache.ValidateLengths(blocks.Time, blocks.Chainwork,
+			blocks.TxCount, blocks.BlockSize, blocks.Fees, blocks.PoolSize,
+			blocks.PoolValue)
```
dmigwi: `blocks.NewAtoms` is missing.

db/dcrpg/queries.go
```diff
@@ -3315,28 +3316,63 @@ func retrieveWindowStats(ctx context.Context, db *sql.DB, charts *cache.ChartDat
...
+		} // else height < nextWindowHeight-1
 	}
 
+	// Store the partial window miss count.
+	if !fullWindow {
+		windows.TicketPrice = append(windows.TicketPrice, price)
+		windows.PowDiff = append(windows.PowDiff, difficulty)
+		windows.Time = append(windows.Time, uint64(timestamp.Unix()))
+		windows.StakeCount = append(windows.StakeCount, ticketsCount)
+	}
```
chappjc: A partial window (the current window) is included now.

db/dcrpg/queries.go
dmigwi:
```golang
	SelectBlockStats = `SELECT height, size, time, chainwork, numtx
		FROM blocks
		WHERE is_mainchain
		AND height > $1
		ORDER BY height;`
```
This `ORDER BY height` condition makes it unnecessary to have `blocks.height` cached data.

db/dcrpg/queries.go
```diff
@@ -3315,28 +3316,63 @@ func retrieveWindowStats(ctx context.Context, db *sql.DB, charts *cache.ChartDat
...
 		if err := rows.Scan(&price, &timestamp, &difficulty, &height, &count); err != nil {
 			return err
 		}
-
 		ticketsCount += count
-		if (height % windowSize) == 0 {
+
+		// If that was the last block in the current sdiff window, append the
+		// data, and reset for the next window.
+		fullWindow = height == nextWindowHeight-1 // e.g. mainnet block 143, 287, etc.
```
dmigwi: Nice call.

db/dcrpg/queries.go
```diff
@@ -3372,38 +3408,65 @@ func appendCoinSupply(charts *cache.ChartData, rows *sql.Rows) error {
 
 // retrieveMissedVotes fetches the missed votes data from the misses and transactions tables.
 func retrieveMissedVotes(ctx context.Context, db *sql.DB, charts *cache.ChartData) (*sql.Rows, error) {
-	rows, err := db.QueryContext(ctx, internal.SelectMissesVotesChartData, charts.MissedVotesTip())
+	fmt.Println("MissedVotesTip:", charts.MissedVotesTip())
```
dmigwi: You mind need to do away with this.

db/dcrpg/queries.go
```diff
@@ -3372,38 +3408,65 @@ func appendCoinSupply(charts *cache.ChartData, rows *sql.Rows) error {
...
+				height, nextWindowHeight)
+		} // else height < nextWindowHeight-1
+	}
+
+	// Store the partial window miss count.
+	if !fullWindow {
+		windows.MissedVotes = append(windows.MissedVotes, uint64(windowMisses))
 	}
+
+	fmt.Println("appendMissedVotesPerWindow:", len(windows.MissedVotes))
```
dmigwi: This too might be unnecessary

db/dcrpg/internal/stakestmts.go
```diff
@@ -342,16 +342,25 @@ const (
...
 
+	SelectMissCountPerBlock = `SELECT blocks.height,
+			CASE WHEN misses.block_hash IS NULL THEN 0 ELSE count(*) END
+		FROM blocks
+		LEFT JOIN misses
+		ON misses.block_hash = blocks.hash
+		WHERE blocks.height > $1
+			AND blocks.is_mainchain
+			GROUP BY blocks.hash, blocks.height, misses.block_hash
+			ORDER BY blocks.height;`
```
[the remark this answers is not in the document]
chappjc: This was likely the cause of https://github.com/decred/dcrdata/issues/1405, which would probably appear similarly for the other window data sets like misses.

db/dcrpg/types_online_test.go
```diff
@@ -208,14 +154,14 @@ func TestTimeRoundTripIncorrectValuerTimeDefLocal(t *testing.T) {
 	// Using dbtypes.TimeDefLocal, where Value ensures Local, stores the
 	// incorrect time.
 	var id uint64
-	err := db.QueryRow(`INSERT INTO testing (timestamp) VALUES ($1) RETURNING id;`,
+	err := sqlDb.QueryRow(`INSERT INTO testing (timestamp) VALUES ($1) RETURNING id;`,
```
dmigwi: `INSERT INTO testing (timestamp) VALUES ($1) RETURNING id;` and `SELECT timestamp FROM testing;` sql statements seem to be repeated quite often. Maybe it would be cleaner if the statements were referenced through variables e.g.
```golang
const(
   insertStatement = "INSERT INTO testing (timestamp) VALUES ($1) RETURNING id;"
   fetchStatement = "SELECT timestamp FROM testing;"
)
```

db/dcrpg/pgblockchain_test.go
```diff
@@ -50,29 +54,63 @@ func (m MemStats) String() string {
...
 	}
 	var err error
 	db, err = NewChainDB(&dbi, &chaincfg.MainNetParams, nil, true, true,
-		addrCacheCap, nil, nil, nil)
+		addrCacheCap, nil, new(dummyParser), nil)
 	cleanUp := func() error { return nil }
 	if db != nil {
 		cleanUp = db.Close
+	} else {
+		return cleanUp, err
```
dmigwi: Why not have this instead:
```golang
return nil, err
```

db/dcrpg/queries.go
dmigwi: ~mind~ => might

chappjc:
Version bumps:

- cache - v2.0.1 (patch)
- dcrpg - v3.0.2 (patch)
- testutil/dbconfig - v2.0.0 (major)
- dcrsqlite - v3.0.1 (patch)

db/dcrpg/pgblockchain_test.go
chappjc: Just because `TestMain` has `defer cleanUp()` regardless of the error value.  It could be changed not to do that, but this doesn't make a different really.

db/dcrpg/types_online_test.go
chappjc: Good call.  Done.

db/dcrpg/queries.go
chappjc: It's helpful for the sanity check here: https://github.com/decred/dcrdata/pull/1426/files/2eb2d445c52be6fc94fcb7f0b79d17606aaf28a8#diff-87f7a577b6af45d0284e4c3fbb98ddf5R3292
But I get what you mean about implicit ordering and not using a slice of heights when plotting.

db/dcrpg/pgblockchain_charts_test.go
chappjc: added

db/cache/charts.go
chappjc: I'm gonna leave it for now, with a comment, to avoid a major version bump of the cache module.

db/dcrpg/pgblockchain_charts_test.go
chappjc: @buck54321 You had a comment a couple lines up about omitting NewAtoms because it's not in the test DB.  Is that OK now given the data you regenerated a few days ago?  It works for me with a full DB, but I've not tried the test DBs we have in https://github.com/dcrlabs/bug-free-happiness/tree/master/pgdb

db/dcrpg/pgblockchain_test.go
dmigwi: ok. Makes sense.

db/dcrpg/queries.go
```diff
@@ -3315,28 +3316,63 @@ func retrieveWindowStats(ctx context.Context, db *sql.DB, charts *cache.ChartDat
...
+		// Trim off the last window, as done by charts.MissedVotesTip(), since
+		// it is not known to be complete.
+		windows.TicketPrice = windows.TicketPrice[0 : fullWindows-1]
+		windows.PowDiff = windows.PowDiff[0 : fullWindows-1]
+		windows.Time = windows.Time[0 : fullWindows-1]
+		windows.StakeCount = windows.StakeCount[0 : fullWindows-1]
+		fullWindows = len(windows.TicketPrice)
+	}
+	nextWindowHeight := windowSize * (fullWindows + 1)
+	fmt.Println(fullWindows, nextWindowHeight)
```
buck54321: oops

db/dcrpg/queries.go
buck54321: It's a little confusing that the last data point will not be indicative of a full window. For instance, the ticket count of the last point will only be the count of a single block's tickets at one point. 

Not from this PR, but the comment above this block seems like copypasta.

db/dcrpg/pgblockchain_charts_test.go
```diff
@@ -72,26 +39,45 @@ func TestMain(m *testing.M) {
...
+	t.Log("Validating initial state.")
+	validate("pre-update")
+
+	// Perform the DB queries and append charts data.
+	t.Log("Performing initial DB query and dataset updates.")
 	charts.Update()
+
+	// Ensure all the slice lengths generated by ChainDB's updaters are consistent.
+	t.Log("Validating initial update.")
 	validate("initial update")
```
buck54321: Since `(ChartData).Update` calls `Lengthen`, won't these always pass validation now? You could potentially return an `error` from `(ChartData).Update` to catch validations errors.

db/dcrpg/queries.go
chappjc: Yeah, that's true.  This change wasn't necessary.  Will revert.

db/dcrpg/pgblockchain_charts_test.go
chappjc: Ah, I recall this now.   I though this was causing trouble in some of the charts updates, but perhaps that's not an issue now with the other fixes.  `chainParams.StakeDiffWindowSize` was being used inconsistent with the fudged DiffInterval somewhere.  I'll try to sort that out.

db/dcrpg/pgblockchain_charts_test.go
chappjc: That's true, although validating after `Snip` seems like a worth while check too.

db/dcrpg/queries.go
chappjc: Migwi mentioned that the heights were implicit.  Still, I like filling in the heights slice, since we have it.  It might be useful for debugging at some point, such as if the slice lengths do not agree with the height, as might happen if there's a missing datapoint in the DB somewhere.

chappjc:
@buck54321 I'm now thinking about splitting `charttests` back out from `mainnettest` since most of the other "online" tests that come with `mainnettest` don't pass on the test DB (things like missing indexes, etc.).  Gotta review this tomorrow.

db/dcrpg/queries.go
chappjc: I've addressed the review comments.  It works with the test db and a full db.

blockdata/datasaver.go
```diff
@@ -252,15 +252,13 @@ func JSONFormatBlockData(data *BlockData) (*bytes.Buffer, error) {
 	return &jsonAllIndented, err
 }
 
-// BlockTrigger wraps a simple function of builtin-typed hash and height that
-// reports no error.
+// BlockTrigger wraps a simple function of builtin-typed hash and height.
 type BlockTrigger struct {
-	Saver func(string, uint32)
+	Saver func(string, uint32) error
```
chappjc: Unfortunately, another breaking change, which bumps the major revision again, to v3.0.0

db/cache/charts.go
```diff
@@ -479,24 +485,25 @@ func (charts *ChartData) Lengthen() error {
...
-	daysLen, err := ValidateLengths(days.PoolSize, days.PoolValue, days.BlockSize,
-		days.TxCount, days.NewAtoms, days.Chainwork, days.Fees, days.Height, days.Time)
+	daysLen, err := ValidateLengths(days.Height, days.Time, days.PoolSize,
+		days.PoolValue, days.BlockSize, days.TxCount, days.NewAtoms,
+		days.Chainwork, days.Fees)
 	if err != nil {
 		return fmt.Errorf("day bin: %v", err)
 	} else if daysLen == 0 {
-		return fmt.Errorf("unexpected zero-length day-binned data")
+		log.Warnf("(*ChartData).Lengthen: Zero-length day-binned data!")
```
chappjc: @buck54321 With 199 blocks in the test DB, we don't get 1 day.  Is this OK?

db/cache/charts.go
```diff
@@ -622,17 +629,18 @@ func (charts *ChartData) readCacheFile(filePath string) error {
 
 // Load loads chart data from the gob file at the specified path and performs an
 // update.
-func (charts *ChartData) Load(cacheDumpPath string) {
+func (charts *ChartData) Load(cacheDumpPath string) error {
```
chappjc: For the new `BlockTrigger.Saver`.

db/cache/charts.go
```diff
@@ -646,8 +654,12 @@ func (charts *ChartData) Dump(dumpPath string) {
 }
 
 // TriggerUpdate triggers (*ChartData).Update.
-func (charts *ChartData) TriggerUpdate(_ string, _ uint32) {
-	charts.Update()
+func (charts *ChartData) TriggerUpdate(_ string, _ uint32) error {
```
chappjc: For the new `BlockTrigger.Saver`.

db/cache/charts.go
buck54321: With a zero-based mainnet test db, we need 334 blocks to get a single day point, and 691 to get 2. 

I'm okay with a bigger test db if you are. Or we could explore artificially adding time to the block times to keep things small.

chappjc: Moving forward on master (after 5.0), I'd probably go with a bigger db, 2 days sounds good.

For getting the tests working for 5.0 release, do you see an issue with having this day-binned data error become a warning log?  It could be a legit error of course, when there really is 1 or more full days, but it seems like it should not be an error in this case where there just aren't that many blocks yet.

buck54321: I think that's the right move.

db/dcrpg/internal/stakestmts.go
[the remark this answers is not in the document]
chappjc: Confirmed that https://github.com/decred/dcrdata/issues/1405 is fixed.

db/dcrpg/pgblockchain_fullpgdb_test.go
```diff
@@ -0,0 +1,145 @@
+// +build pgonline,fullpgdb
```
chappjc: comma separated tags are an AND operation

db/dcrpg/pgblockchain_common_test.go
```diff
@@ -0,0 +1,115 @@
+// +build pgonline chartdata
```
chappjc: space separated tags are an OR operation

db/dcrpg/queries.go
chappjc: I've bumped the cache version from 4 -> 5 since v4 cache gobs will have an empty `Heights` slice and fail validation

run_tests.sh
```diff
@@ -4,25 +4,22 @@
 # ./run_tests.sh                         # local, go 1.12
 # ./run_tests.sh docker                  # docker, go 1.12
 # ./run_tests.sh podman                  # podman, go 1.12
+#
+# To use build tags:
+#  TESTTAGS="chartdata" ./run_tests.sh
+#  TESTTAGS="pgonline" ./run_tests.sh
+#  TESTTAGS="pgonline fullpgdb" ./run_tests.sh
```
chappjc: Note this tag convention is used with `go test`. e.g. `go test . -v -tags "pgonline fullpgdb" -count 1`

testutil/dbconfig/config_fullpgdb.go
```diff
@@ -0,0 +1,12 @@
...
+
+package dbconfig
+
+// Test DB server and database config.
+const (
+	PGTestsHost   = "localhost" // "/run/postgresql" for UNIX socket
+	PGTestsPort   = "5432"      // "" for UNIX socket
+	PGTestsUser   = "postgres"  // "dcrdata" for full database rather than test data repo
+	PGTestsPass   = ""
+	PGTestsDBName = "dcrdata_mainnet_full"
```
chappjc: The DB config you get with `-tags fullpgdb` will use a different database.  But because of the `dcrpg` replace statement for dbconfig, any of these settings can be modified.

db/dcrpg/go.mod
```diff
@@ -2,6 +2,10 @@ module github.com/decred/dcrdata/db/dcrpg/v3
 
 go 1.11
 
+replace github.com/decred/dcrdata/testutil/dbconfig => ../../testutil/dbconfig
```
chappjc: So the db config can be changed and modified tests run without banging your head against the wall.

db/dcrpg/queries.go
```diff
@@ -3370,40 +3383,49 @@ func appendCoinSupply(charts *cache.ChartData, rows *sql.Rows) error {
...
+		if fullWindow {
+			windows.MissedVotes = append(windows.MissedVotes, uint64(windowMisses))
 
-			count = 0
-			startHeight += windowSize
-		}
+			// Next sdiff window
+			windowMisses = 0
+			nextWindowHeight += windowSize
+		} else if height == nextWindowHeight {
```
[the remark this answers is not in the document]
chappjc: agreed.

PR #1427, opened 2019-06-19
set default market view to aggregated depth

### LINKED ISSUE #1425: make Aggregate default on /markets
Currently it is defaulting to showing binance orderbooks. It also not obvious that the exchanges on the left hand side are buttons since there is no highlighting on the currently selected row. Aggregate is a more interesting default view and also does not preference one exchange over another.

buck54321:
Resolves #1425

### DIFF
--- a/public/js/controllers/market_controller.js
+++ b/public/js/controllers/market_controller.js
@@ -582,7 +582,7 @@ export default class extends Controller {
       settings.chart = depth
     }
     if (settings.xc == null) {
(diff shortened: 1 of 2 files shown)


chappjc:
Woops, I merge this too hastily. npm build says:

```
/home/jon/github/decred/dcrdata/public/js/controllers/market_controller.js
  18:7  error  'binance' is assigned a value but never used  no-unused-vars
```

PR #1432, opened 2019-06-25
js/controllers: remove unused const binance

chappjc:


njirap:
Hi @chappjc 

On grepping 'binance' I find it hard coded in the same file maybe it's the variable you are removing that was supposed to be used at https://github.com/decred/dcrdata/blob/master/public/js/controllers/market_controller.js#L585

buck54321:
That is what I intended, @njirap. Looks like I just hard-coded it on accident. Not sure why I didn't see the linter message.

chappjc:
@buck54321 So you'd prefer to keep and use the `const`?  Seemed like extra code weight.

buck54321:
Let's drop the `const` then.

PR #1434, opened 2019-06-25
mod updates

chappjc:
After merge, tag the following:
- gov v1.0.1
- middleware v2.2.0
- exchanges v2.0.1 - NOTE: only updated dcrdata and rateserver deps!
- blockdata - v3.0.0 (major from v2 to v3)
- testutil/dbconfig - v2.0.0 (major from v1 to v2)
- cache - v2.1.0 - before dcrpg

After this commit is merged, update and tag:
- pubsub v2.0.1 (updated for blockdata major bump)
- db/dcrpg v3.0.2
- db/dcrsqlite v3.0.1

### DIFF
--- a/middleware/apimiddleware.go
+++ b/middleware/apimiddleware.go
@@ -772,6 +772,12 @@ func AgendaIdCtx(next http.Handler) http.Handler {
 	})
 }
 
+// AgendIdCtx returns a http.HandlerFunc that embeds the value at the url part
+// {agendaId} into the request context. This is here for backward compatibility.
+func AgendIdCtx(next http.Handler) http.Handler {
+	return AgendaIdCtx(next)
+}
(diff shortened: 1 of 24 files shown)


middleware/apimiddleware.go
```diff
@@ -772,6 +772,12 @@ func AgendaIdCtx(next http.Handler) http.Handler {
 	})
 }
 
+// AgendaIdCtx returns a http.HandlerFunc that embeds the value at the url part
+// {agendaId} into the request context. This is here for backward compatibility.
+func AgendIdCtx(next http.Handler) http.Handler {
+	return AgendaIdCtx(next)
```
chappjc: It's not worth making a middleware/v3 just for this function renaming.

Issue #1435, opened 2019-06-25
scanQueries does not seem to be separating individual queries


chappjc:
scanQueries does not seem to be separating individual queries, although this does not seem to be breaking anything, just making it difficult to debug in case of a query/exec failure.

dmigwi:
>does not seem to be separating individual queries

What does it mean? Like all queries are run whether one of them fails or not?

chappjc:
That wasn't very specific - sorry. What I mean was that in this loop, `text` contains the entire contents of the .sql file, not a single query:
https://github.com/decred/dcrdata/blob/f11bfb37ceba8d7844d11a2d72c03966393ce7b0/testutil/dbconfig/config.go#L41
(just a single iteration of the loop does everything)

chappjc:
Resolved by https://github.com/decred/dcrdata/pull/1437

PR #1437, opened 2019-06-26
Run each testutil query individually

dmigwi:
Since the file is scanned several bytes at a time, my custom token scanner using `LastIndex(` function was checking the last occurrence of `;\n` characters in the chunk already scanned.

`Index(` now returns the first occurrence of `;\n` in the scanned chunk thus you can not debug the tests.

### DIFF
--- a/testutil/dbconfig/config.go
+++ b/testutil/dbconfig/config.go
@@ -61,10 +61,18 @@ func scanQueries(data []byte, atEOF bool) (advance int, token []byte, err error)
 
 // Comments start with (--) and end with a new line.
 func deleteComments(a []byte) []byte {
-	re := regexp.MustCompile(`^--[\s*\w*[[:punct:]]*]*$\n`)
+	// delete sql comment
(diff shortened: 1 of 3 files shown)


dmigwi:
https://github.com/decred/dcrdata/issues/1435

chappjc:
Thanks for the quick turnaround on this!  This certainly seems like the right fix.  Will give a proper review shortly.

testutil/dbconfig/config.go
```diff
@@ -86,7 +85,7 @@ func scanQueries(data []byte, atEOF bool) (advance int, token []byte, err error)
 	if atEOF && len(data) == 0 {
 		return 0, nil, bufio.ErrFinalToken
 	}
-	if i := bytes.LastIndex(data, []byte(";\n")); i >= 0 {
+	if i := bytes.Index(data, []byte(";\n")); i >= 0 {
```
chappjc: Would you add a test for `scanQueries` with `data` that has a few queries?  Just to catch future breakage. Thanks.

Issue #48, opened 2017-06-27
Version the API


chappjc:
I've wrestled with this for a while, and it seems like using the Accept header with version in a MIME type parameter (e.g. `Accept: application/json; version=1`), is perhaps the "right way" to do this.  The thing is, forget that and all the headaches that go with it.  **I want to go with version in the URL.**

Much like described in this blog post: https://cloudplatform.googleblog.com/2017/06/versioning-APIs-at-Google.html

The API (not the app) should have a X.Y version where the URL contains /vX.  Changes in X are thus breaking (removed or renamed fields, etc.), while changes in Y are not breaking (added fields or endpoints, etc.).

ukane-philemon:
@chappjc, this seems to be overdue. I'd like to pick this up.

chappjc:
OK, but there's not much to do for this, except mirror the entire /api paths on /api/v1.  Obviously all /api links have to keep working.
When we make breaking changes, we'll add to /api/v2.

chappjc:
Please use Chi's routing capabilities and sub-routers as needed to avoid logic our API handler implementations where possible.  Review the chi methods and consider a single Router that can be shared between the two parent paths /api and /api/vX

https://pkg.go.dev/github.com/go-chi/chi/v5#Mux.Route
https://pkg.go.dev/github.com/go-chi/chi/v5#Mux.Mount

ukane-philemon:
we can mount the same router on different paths. on `/api/` as default and `/api/v{version}/`.

chappjc:
> we can mount the same router on different paths. on `/api/` as default and `/api/v{version}/`.

Right.  That will probably do it.

For planning for the future, here's an interesting approach: https://github.com/go-chi/chi/tree/v5.0.8/_examples/versions
...subrouters, common handlers, and an `api.version` context value for handlers to modify their behavior.

Regarding the `api/types` package, we could consider making that it's own module (again, since it was previously), but the _module version_ would already be at v6 since we were at v5 before we reabsorbed it into the main dcrdata module.  Alternatively, we could just have _subfolders_ like `api/types/v1`, which seems better for a REST API's data type versioning.

ukane-philemon:
> For planning for the future, here's an interesting approach: https://github.com/go-chi/chi/tree/v5.0.8/_examples/versions
...subrouters, common handlers, and an api.version context value for handlers to modify their behavior.

Yes, this makes sense. This will enable us to use one entry point for requests and handle them differently.

[...]

PR #1439, opened 2019-06-26
insight: set time with MempoolTime from mempool tx store

chappjc:
For unconfirmed transactions, this sets a non-zero time value for the following Insight API endpoints:
- `/insight/api/addrs/{address}/txs`
- `/insight/api/txs?address={address}`

Note that this sets the `"time"` JSON value for unconfirmed transactions only. After confirmation, both `"time"` and `"blocktime"` will be set to the value from the `getrawtransaction` RPC (the mempool time is no longer known).

### DIFF
--- a/api/insight/apiroutes.go
+++ b/api/insight/apiroutes.go
@@ -448,11 +448,11 @@ func (iapi *InsightApi) getAddressesTxnOutput(w http.ResponseWriter, r *http.Req
 					Address:       address,
 					TxnID:         fundingTx.Hash().String(),
 					Vout:          f.Index,
+					BlockTime:     fundingTx.MemPoolTime,
 					ScriptPubKey:  hex.EncodeToString(txOut.PkScript),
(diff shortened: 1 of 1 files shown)


buck54321:
There's a chance, however small, that someone might be using the absence of a time field as an indicator of lack of confirmation. Is this a breaking change, and if so does it require any special versioning?

chappjc:
Hmm, that change of behavior could be breaking, but I'm inclined to ignore that possibility since a client doing that is essentially abusing a bug.  We're also not versioning the Insight API in any way (either with a `/v1` url prefix or other mechanism), so there are essentially no promises, which BTW is also what the official Insight API seems to do.  At least one of the recent changes to our Insight API implementation were just to keep up with some random changes made to bitpay's Insight implementation.

This reminds me of one of the oldest issues, ["Version the API"](https://github.com/decred/dcrdata/issues/48), which refers to our own API.  We're essentially treating the API as having v0 status, but that should change when we feel no more changes are needed.

PR #1442, opened 2019-06-28
explorer: add semi-bold font

buck54321:
Adds a semi-bold font and uses it for table headers throughout. Also applied to address on address page and txid on tx page.

A medium-weight font was called for in a number of places in #451, but I could never get it to work with the fonts we had, and I'm admittedly not very knowledgeable on fonts. 

image

@chappjc You're usually on top of minifying SVGs, but I wasn't sure about fonts since they contain possibly important metadata.


chappjc:
That's much easier on the eyes.

Yeah, minifying font SVGs is likely to mess it up.  I wanna have a look though anyway to see if there's anything obvious bloat the files.

chappjc:
Well, there doesn't seem to be much to "clean up" in source-sans-pro-semibold.svg, but it compresses well. 

Original | compressed
--- | ---
1,193,865 | 132,890

go.sum
```diff
@@ -131,6 +131,7 @@ github.com/decred/dcrd/wire v1.1.0 h1:G+3CugtxNbToUN8RKWqm74yLfzJJ2BKMOr2RgWc4Ty
 github.com/decred/dcrd/wire v1.1.0/go.mod h1:/JKOsLInOJu6InN+/zH5AyCq3YDIOW/EqcffvU8fJHM=
 github.com/decred/dcrd/wire v1.2.0 h1:HqJVB7vcklIguzFWgRXw/WYCQ9cD3bUC5TKj53i1Hng=
 github.com/decred/dcrd/wire v1.2.0/go.mod h1:/JKOsLInOJu6InN+/zH5AyCq3YDIOW/EqcffvU8fJHM=
+github.com/decred/dcrdata v2.1.3+incompatible h1:DLzrqgOOgHwrUSbuOVDAUzfQePhCGl9UsjEgZI1D0Kg=
```
chappjc: Could you revert this line pls?

PR #1443, opened 2019-06-28
explorer: fullscreen address chart

### LINKED ISSUE #1430: Support full screen on address page charts
Current charts on address pages (such as the treasury page) show the chart in the top right hand corner. It would be useful for the user to be able to view the charts using the full screen. This could be done with a icon/button for expanding the chart area so that it covers a large part of the available screen space, and which can be closed/minimized for the user to return to the original address view.

buck54321:
Resolve #1430

Allows expanding the chart on the address page to full screen.

image


chappjc:
Just made a conflict for you.  Looks like it's just taking the file delete.

PR #1448, opened 2019-06-30
explorer: reformat unminted legend value in charts

### LINKED ISSUE #1444: show dcr amount difference on coin supply chart
-Predicted coin supply is formatted incorrectly, but maybe already fixed in master.

-Currently shows  
Change: -1.39%. 
More of a comparison then a change, so Id label it Difference instead.

-Also suggest showing the actual DCR amount difference, ~140,624 dcr
Label could be Not Minted:

buck54321:
Resolves #1444

image

### DIFF
--- a/public/js/controllers/charts_controller.js
+++ b/public/js/controllers/charts_controller.js
@@ -95,13 +95,15 @@ function legendFormatter (data) {
     // The circulation chart has an additional legend entry showing percent
     // difference.
     if (data.series.length === 2 && data.series[0].label.toLowerCase().includes('coin supply') &&
-      data.series[1].label.toLowerCase().includes('coin supply')) {
+      data.series[1].label.toLowerCase().includes('inflation limit')) {
       data.series.sort((a, b) => a.y > b.y ? 1 : -1)
       let actual = data.series[0].y
       let predicted = data.series[1].y
-      let unminted = predicted - actual
-      let change = ((unminted / predicted) * 100).toFixed(2)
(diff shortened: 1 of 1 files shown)


chappjc:
I always felt that "predicted" was a bit misleading because it's only a prediction assuming full block rewards, but once there's a single reduced block reward and missed vote it's a no longer a prediction; it is certain not to hit that amount.  So it's really like a target/ideal coin supply.

With that in mind, I wonder if we should (a) rename "predicted" and (b) not show the predicted line for the past, just the future.

buck54321:
I'm guessing removal of the ideal line will result in complaints. Maybe just a better name? What is a compact but unambiguous way to describe it? Using a dashed line might be good too.

buck54321:
How about "inflation limit"?

image

chappjc:
That's good I suppose.  Although I still think it's confusing to plot a line into the future that cannot be reached because at present there are already coins that will never be mined.  Thus why I feel like the predicted supply or inflation limit line makes the most sense for data points in the future. It's not a big deal to me personally though.

It's just that the two supply lines starting at genesis only say what coins did not get minted because of missed votes.  Showing the inflation limit line for the future is where it gets confusing I think because that suggests that's where DCR supply could end up, but really it's that line minus the unminted value at present.

buck54321:
Let's try it out. I like it this way too. 

image

chappjc:
Exactly what I had in mind.  I like this.  Let's ping the dcrdata chan for feedback.

PR #1449, opened 2019-07-01
gov/politeia: Manage duplicate proposal ref id instances

### LINKED ISSUE #1441: Proposal Title edit resulted to 2 instances of same Proposal.
This two proposals links point to the same proposal on politeia:
- https://alpha.dcrdata.org/proposal/decreed-media-campaign-proposal-crypto-economy-2019-2020
- https://alpha.dcrdata.org/proposal/decred-media-campaign-proposal-crypto-economy-2019-2020

Screenshot from 2019-06-28 01-51-27

Proposal title edits made dcrdata identify them as two different proposals.

I will be fixing this.

dmigwi:
This PR fixes two bugs:
1. Fixes #1441 
2. The [vetted proposals(/api/v1/proposals/vetted)](https://github.com/decred/politeia/blob/master/politeiawww/api/www/v1/api.md#vetted) URL returns a snapshot of most recent proposals to be added in Politeia. This implies that `?after=` should be used when running a fresh sync and `?before=` when fetching new updates.
Param `before=` works like `next=` and `after=` like `previous=` based on politeia usage.

### DIFF
--- a/gov/politeia/proposals.go
+++ b/gov/politeia/proposals.go
@@ -214,29 +214,33 @@ func (db *ProposalDB) saveProposals(publicProposals pitypes.Proposals) (int, err
 		// that generates a RefID similar to one already in the db and appending
 		// integers to it till it becomes unique is the solution.
 		if err == storm.ErrAlreadyExists {
+			var data *pitypes.ProposalInfo
 			// Check if the proposal token already exists in the db.
-			data, newErr := db.proposal("TokenVal", val.TokenVal)
-			if newErr == nil && data != nil {
+			data, err = db.proposal("TokenVal", val.TokenVal)
+			if err == nil && data != nil {
 				// The proposal token already exists thus trigger an update with
 				// the latest details.
 				valCopy := *val
 				valCopy.ID = data.ID
-				prefixStr := ""
+				suffixStr := ""
 
 				for k := 1; k <= maxLoop; k++ {
-					valCopy.RefID += prefixStr
+					valCopy.RefID += suffixStr
 					// Attempt to update the old entry.
-					newErr = db.dbP.Update(&valCopy)
-					if newErr == nil || newErr != storm.ErrAlreadyExists {
-						err = nil
-						break
+					err = db.dbP.Update(&valCopy)
+					if err == storm.ErrAlreadyExists {
+						suffixStr = strconv.Itoa(k)
+						continue
 					}
-					prefixStr = strconv.Itoa(k)
+					if err != nil {
+						log.Error("storm DB update failed: %v", err)
+					}
+					break
 				}
 			}
 
-			// First try wasn't successful if newErr != nil.
-			if newErr != nil {
+			// First try wasn't successful if err != nil.
+			if err != nil {
 
 				for c := 1; c <= maxLoop; c++ {
 					// Drop the previously assigned ID.
@@ -245,10 +249,13 @@ func (db *ProposalDB) saveProposals(publicProposals pitypes.Proposals) (int, err
 					val.RefID += strconv.Itoa(c)
 					// Attempt to save a new entry.
 					err = db.dbP.Save(val)
-					if err == nil || err != storm.ErrAlreadyExists {
-						err = nil
-						break
+					if err == storm.ErrAlreadyExists {
+						continue
 					}
+					if err != nil {
+						log.Error("storm DB save failed: %v", err)
+					}
+					break
 				}
 			}
 		}
--- a/gov/politeia/proposals.go
+++ b/gov/politeia/proposals.go
@@ -141,8 +141,7 @@ func generateCustomID(title string) (string, error) {
 	return reg.ReplaceAllString(strings.ToLower(title), "-"), nil
 }
 
-// saveProposals adds the proposals data to the db.
-func (db *ProposalDB) saveProposals(URLParams string) (int, error) {
+func (db *ProposalDB) fetchAPIData(URLParams string) (pitypes.Proposals, error) {
 	copyURLParams := URLParams
 	pageSize := int(piapi.ProposalListPageSize)
 	var publicProposals pitypes.Proposals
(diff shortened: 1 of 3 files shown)


gov/politeia/proposals.go
```diff
@@ -29,7 +29,7 @@ var (
 	errDef = fmt.Errorf("ProposalDB was not initialized correctly")
 
 	// dbVersion is the current required version of the proposals.db.
-	dbVersion = semver.NewSemver(1, 0, 0)
+	dbVersion = semver.NewSemver(2, 0, 0)
```
dmigwi: This version change guarantees that the db will be rebuilt.

gov/politeia/proposals.go
```diff
@@ -141,19 +141,30 @@ func generateCustomID(title string) (string, error) {
 	return reg.ReplaceAllString(strings.ToLower(title), "-"), nil
 }
 
-// saveProposals adds the proposals data to the db.
-func (db *ProposalDB) saveProposals(URLParams string) (int, error) {
+// fetchAPIData returns the API data fetched from the Politeia API endpoints.
+// NB: "/api/v1/proposals/vetted" path returns the latest snapshot of the API
+// data that currently exits. This implies that parameter "after=" should be used
+// when syncing the data from scratch and "before=" should be used to fetch newer
+// updates.
```
dmigwi: Bug 2:

gov/politeia/proposals.go
```diff
@@ -188,26 +207,61 @@ func (db *ProposalDB) saveProposals(URLParams string) (int, error) {
...
 		err = db.dbP.Save(val)
 
-		// In the rare case scenario that the current proposal has a duplicate refID
-		// append an integer value to it till it becomes unique.
+		// When a duplicate CensorshipRecord struct is detected this "already exists"
+		// error is thrown, this means that some edits were made to an older version
+		// of the proposal and its fixed by updating the new changes. In another
+		// case that is more rare, is that the current proposal could have a Name
+		// that generates a RefID similar to one already in the db and appending
+		// integers to it till it becomes unique is the solution.
```
dmigwi: Bug 1: Fix

gov/politeia/proposals.go
```diff
@@ -188,26 +207,61 @@ func (db *ProposalDB) saveProposals(URLParams string) (int, error) {
...
+					prefixStr = strconv.Itoa(k)
+				}
+			}
+
+			// First try wasn't successful if newErr != nil.
+			if newErr != nil {
+
+				for c := 1; c <= maxLoop; c++ {
+					// Drop the previously assigned ID.
+					val.ID = 0
```
dmigwi: `db.dbP.Save(val)` automatically assigns the `ID` value thus dropping the ID value ensures that a conflict is not likely to arise just in case another proposal was saved since the last time `db.dbP.Save(val)` run.

gov/politeia/proposals.go
chappjc: I recall a fix for the before/after semantics already.  Is this the same issue, just in a different part of the gov code, or?

dmigwi: It is an addition to that fix.

This bug will be noticed when one does a fresh sync on proposals.db (storm db) file otherwise on normal updates where only the latest data is fetched, all works well.

The actual description highlighted above doesn't exist in the [docs](https://github.com/decred/politeia/blob/master/politeiawww/api/www/v1/api.md#vetted), in fact lack of that description is the reason why we had that previous issue.

gov/politeia/proposals.go
```diff
@@ -188,26 +207,61 @@ func (db *ProposalDB) saveProposals(URLParams string) (int, error) {
...
+				// the latest details.
+				valCopy := *val
+				valCopy.ID = data.ID
+				prefixStr := ""
+
+				for k := 1; k <= maxLoop; k++ {
+					valCopy.RefID += prefixStr
+					// Attempt to update the old entry.
+					newErr = db.dbP.Update(&valCopy)
+					if newErr == nil || newErr != storm.ErrAlreadyExists {
```
chappjc: I don't understand this condition. First, `newErr != storm.ErrAlreadyExists` is covered by `newErr == nil`.  Second, what about if `newErr` is some other fatal error?

dmigwi: The loop and if statement are attempting to solve `storm.ErrAlreadyExists` error thus the loop should stop once an error other than `storm.ErrAlreadyExists` and `nil` is returned.
There exists two error variables `err` and `newErr` where `err` is the main error to be considered if the attempted update/save new data fails. 

Its expected that by the time the loop above exits either of the following will have happened;
1. Update wasn't successful if all the suffixes added by the fifth (`maxLoop`) did not make it unique.
2. Another error other than `storm.ErrAlreadyExists` was returned thus update failed.
3. Update was successful and `newErr` is finally equal to `nil`.

dmigwi: Actually `newErr` variable is not necessary.

dmigwi: I have dropped the `newErr` error variable as it wasn't necessary.

gov/politeia/proposals.go
```diff
@@ -214,29 +214,29 @@ func (db *ProposalDB) saveProposals(publicProposals pitypes.Proposals) (int, err
...
 
 				for k := 1; k <= maxLoop; k++ {
-					valCopy.RefID += prefixStr
+					valCopy.RefID += suffixStr
 					// Attempt to update the old entry.
-					newErr = db.dbP.Update(&valCopy)
-					if newErr == nil || newErr != storm.ErrAlreadyExists {
-						err = nil
+					err = db.dbP.Update(&valCopy)
+					if err == nil || err != storm.ErrAlreadyExists {
```
chappjc: The issue I was pointing out still exists.  Consider `err != storm.ErrAlreadyExists`.  this is true if `err` is `nil` (assuming `storm.ErrAlreadyExists` is not equal to `nil`).  Thus, the `err == nil` part of the condition is not necessary.

Also, you still want to catch (and log or whatever) legit errors.

```go
if err == storm.ErrAlreadyExists {
	suffixStr = strconv.Itoa(k)
	continue
}
if err != nil {
	log.Error("storm DB Update failed: %v", err)
	// do something else?  return err?
}
break
```

PR #1456, opened 2019-07-04
proposals page: Tweak the UI

dmigwi:
Screenshot from 2019-07-04 04-57-23

The following UI changes were made;
- The previous tooltip on the proposals vote status was found not to be that informative as anticipated. Adding the word `approval` was found to be effective. The approval percentage description was highlighted in grey for all status with a percentage.
- Column  Title/Proposal Name was combined with Author column.
- The proposal sync times status was shifted from bottom center to top just after the page title.
- More description about the Time remaining was added to the `In progress` vote status cell.
- The vote count for all abandoned proposals shows `__` instead of 0.

### DIFF
--- a/views/proposals.tmpl
+++ b/views/proposals.tmpl
@@ -6,7 +6,17 @@
         <div class="container main" data-controller="pagenavigation">
             <div class="row justify-content-between">
                 <div class="col-sm-lg-14 col-sm-12 d-flex">
-                    <h4 class="mb-2">Politeia Proposals</h4>
+                    <h4 class="mb-2">
+                        Politeia Proposals
+                        <span class="mb-0 fs11 text-black-50">
+                            &nbsp;(<span class="mr-2">Proposals Sync:&nbsp;
+                                <span data-controller="time" data-target="time.age" data-age="{{$.LastPropSync}}"></span>&nbsp;ago
+                            </span>
+                            <span>Votes Sync:&nbsp;
+                                <span data-controller="time" data-target="time.age" data-age="{{$.LastVotesSync}}"></span>&nbsp;ago
+                            </span>)
+                        </span>
+                    </h4>
                 </div>
             </div>
             <div class="mb-1 fs13">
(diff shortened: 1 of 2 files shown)


buck54321:
Let's look at this after #1383 is merged. There's a lot of overlap.

chappjc:
PR #1383 is in.  Please rebase and see what overlap there is to address.

dmigwi:
Latest Changes:

**Desktop View**
Screenshot from 2019-07-05 20-52-12


**Mobile View**
Screenshot from 2019-07-05 20-51-48

views/proposals.tmpl
```diff
@@ -138,29 +148,45 @@
...
-                                    <span class="text-progress">
+                                    <span class="text-progress position-relative" data-tooltip="Voting in progress">
                                         In Progress
                                     </span>
                                 {{else}}
                                     {{range $i, $vr := .VoteResults}}
                                         {{if eq $vr.Option.OptionID "yes"}}
                                             {{$votesPercent :=  percentage $vr.VotesReceived $v.TotalVotes}}
-                                            {{if lt $.Tip.Height (toInt $v.Endheight)}}
+                                            {{$blocksRemaining := subtract (int64 (toInt $v.Endheight)) (int64 $.Tip.Height)}}
```
buck54321: I know it has come up before, but `Endheight` should really already be an integer type. Probably outside the scope of this PR.

dmigwi: This conversion should be a PR on its own as data from back-end returns the field as a string. It would be optimal to change `EndHeight` string value to integer during the API data unmarshaling.

dmigwi:
This was to create more space so that users will smaller devices can have a better user experience while viewing the page.
Also those two data entries can also be combined the way we did and still convey the required information.

buck54321:
Since we're hiding the author column at low widths now (since #1383), there's no need to combine the columns. Title and author are unrelated pieces of information that should have their own columns if possible.

dmigwi:
>Title and author are unrelated pieces of information....

In terms of the Title and Author being unrelated, I hold a different opinion. I think they have a very low probability of confusing anyone when joined.

The Author column is only hidden for screens with width less than 500px. Thus the Ipad view and any other view with width ranging from 500px to 900px looks better with that joined column.

Screenshot from 2019-07-19 22-31-09

PR #1458, opened 2019-07-06
Add step plot for the address balance chart

dmigwi:
**Before Adding the step plot**

Screenshot from 2019-07-06 17-09-38

**After**

Screenshot from 2019-07-06 17-09-19


public/js/controllers/address_controller.js
```diff
@@ -464,6 +464,7 @@ export default class extends Controller {
 
       case 'balance':
         options = balanceGraphOptions
+        options.stepPlot = true
```
buck54321: Probably set `stepPlot` as part of `balanceGraphOptions`.

PR #1460, opened 2019-07-08
Align section heights of the homepage

### LINKED ISSUE #992: home: make left and right blocks of same height
Increase the number of rows in Blocks and Mempool sections so that together they occupy same height as Voting+Mining+Distribution on the right.

Suggested [here](https://www.reddit.com/r/decred/comments/anu91e/new_decred_block_explorer_by_dcrdataorg/efxtg21/).

### LINKED ISSUE #1457: align section heights on home page
From the comment on reddit by Vadimusnews123

"I propose to increase the number of rows in the tables to align the columns in height."

https://www.reddit.com/r/decred/comments/anu91e/new_decred_block_explorer_by_dcrdataorg/efxtg21/?context=8&depth=9

5KRrn6I

njirap:
Resolves #1457
Resolves #992

- Adds an extra row in Blocks section
- Makes the 2 columns flexbox in order to evenly distribute their content
- Removes bottom margin from Distribution section so it aligns with end of the column

Full desktop width
Screen Shot 2019-07-08 at 1 27 25 PM

Width 900
Screen Shot 2019-07-08 at 1 28 45 PM

Width 864
Screen Shot 2019-07-08 at 1 29 25 PM

### DIFF
--- a/views/home.tmpl
+++ b/views/home.tmpl
@@ -7,7 +7,7 @@
     {{ template "navbar" . }}
     <div class="container main" data-controller="time blocklist homepage">
         <div class="row">
-            <div class="col-md-15 p-0">
+            <div class="d-flex flex-column col-md-15 p-0">
 
                 <div class="bg-white mb-1 py-2 px-3 my-0">
                     <div class="d-inline-block position-relative p-2">
@@ -95,7 +95,7 @@
                     </div>
                 </div> <!-- end blocks card -->
 
-                <div class="py-1 px-2 mb-1 bg-white">
+                <div class="py-1 px-2 mb-1 bg-white flex-grow-1">
 
                     <div class="d-flex align-items-baseline my-2">
                         <div class="position-relative col nowrap">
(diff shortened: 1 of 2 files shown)


views/home.tmpl
```diff
@@ -7,7 +7,7 @@
     {{ template "navbar" . }}
     <div class="container main" data-controller="time blocklist homepage">
         <div class="row">
-            <div class="col-md-15 p-0">
+            <div class="d-flex flex-column justify-content-between col-md-15 p-0">
```
buck54321: This doesn't quit work with exchange data. Turn on exchange monitoring in your configuration to see that. 

image

You'll need to be a little more strategic with the flexbox. I'd imagine you'll need to set flex on the children elements (the "cards") too.

Also, you seem to be missing your treasury balance, which may be problematic. Are you seeing any errors?

njirap: I have replicated that display bug..looking into it.

What do you mean by my treasury balance? Is it the 626,369 DCR under the Distribution section?

njirap:
Screen Shot 2019-07-16 at 4 02 50 PM

views/home.tmpl
buck54321: Sorry I missed this. I meant that the `Treasury` was missing from your original screenshots.

njirap:
Not sure how to add mempool so I've added a block and flexed both columns

Screen Shot 2019-07-16 at 9 59 48 PM

Issue #1459, opened 2019-07-07
insight: pageNum URL query param ignored


chappjc:
The `/txs?address=` endpoint is not checking for a `pageNum` query param.

Presently this endpoint only returns up to 10 transactions for the address.  Further, the `pagesTotal` is set to the total number of transactions, rather than the number of pages of 10.

While we fix that, can you get what you need from `/api/addr/{address}`, which has `to=` and `from=` URL query params, or from `/api/addrs/{address}/txs`.

PR #1463, opened 2019-07-12
insight: add pageNum middleware and use in /txs endpoint

chappjc:
The `pageNum` URL query param was being ignored with the `/txs` endpoint.
This checks for the `pageNum` parameter and goes to the correct page of
transactions.

This also fixes the computation of the `pagesTotal` JSON response field,
which previously was set to the total number of transactions rather than
than the number of pages.

Resolves https://github.com/decred/dcrdata/issues/1459

### DIFF
--- a/api/insight/apiroutes.go
+++ b/api/insight/apiroutes.go
@@ -543,6 +543,7 @@ func removeSliceElements(txOuts []*apitypes.AddressTxnOutput, inds []int) []*api
 }
 
 func (iapi *InsightApi) getTransactions(w http.ResponseWriter, r *http.Request) {
+	pageNum := m.GetPageNumCtx(r)
 	hash, blockerr := m.GetBlockHashCtx(r)
 	addresses, addrerr := m.GetAddressCtx(r, iapi.params, 1)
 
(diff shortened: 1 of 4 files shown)


api/insight/apiroutes.go
```diff
@@ -565,31 +568,32 @@ func (iapi *InsightApi) getTransactions(w http.ResponseWriter, r *http.Request)
...
+		// Go to the last page if pageNum is greater than the number of pages.
+		if pageNum > pagesTotal {
+			pageNum = pagesTotal
+		}
+
+		// Grab the transactions for the given page (1-based index).
+		skipTxns := (pageNum - 1) * txPageSize
+		txsOld := []*dcrjson.TxRawResult{}
+		for i := skipTxns; i < txcount && i < txPageSize+skipTxns; i++ {
+			txsOld = append(txsOld, &mergedTxns[i])
```
chappjc: @buck54321 If you have a minute, could you review my page number logic in the 20 or so lines above?

chappjc: Any time in the next few days that is.  No hurry, just prioritizing 5.0.1 patch commits.

Issue #1465, opened 2019-07-12
insight: /addrs/txs omits vin.vout for all inputs


chappjc:
The previous outpoint's tx index (should be called `vout`) is missing

chappjc:
Resolved by https://github.com/decred/dcrdata/pull/1466

PR #1466, opened 2019-07-15
api/types: do not omit zero-value InsightVin.Vout

chappjc:
`Vout` and `Sequence` should not be omitted when zero. Instead, use a nil
pointer to omit them.
`Value` and `ValueSat` never need be omitted.

These changes require a *major* version bump to the api/types module.


chappjc:
Resolves https://github.com/decred/dcrdata/issues/1465

Issue #1464, opened 2019-07-12
insight: /addrs/txs omits vin.addr for unconfirmed txns


chappjc:

chappjc:
Resolved by https://github.com/decred/dcrdata/pull/1468

PR #1468, opened 2019-07-15
insight: DcrToInsightTxns can RPC for prevout addresses

chappjc:
For non-coinbase/stakebase inputs, the previous outpoint itself may be
in mempool, in which case it is necessary to retrieve the inputs' txns
and decode the pkScript.

In addition to the above fix, which uses `txhelpers.OutPointAddressesFromString`,
amount summation is performed in integers rather than floats, and fee computation
is now correct for both coinbase and stakebase (vote) transactions.


chappjc:
Resolves https://github.com/decred/dcrdata/issues/1464

PR #1481, opened 2019-07-26
insight: count client connections and set max to 16384

chappjc:
By default, the socket.io server is limited to 1000 connections. This increases it to 16384.  Note that scaling higher may be achieved by adding more backends.  This limit may be made configurable in future work, but it is hard coded for a patch release to dcrdata 5.0.

These edits also supplement the socket.io log messages to give the current connected user count on each connect and disconnect event.


chappjc:
Ran well in testing and now running well in production.

PR #1487, opened 2019-07-27
explorer: market: add fiat and percent values to the market depth gap

### LINKED ISSUE #1484: Show bid/ask spread percentage
Currently on https://dcrdata.decred.org/market "delta" is displayed in BTC terms exclusively. It's useful to know what that is in percentage terms, the BTC value could be placed in brackets next to that value.

buck54321:
Resolves #1484 

Modifies the depth chart gap display to show values in BTC, fiat, and percent, where percent is taken as (gapWidth/midGap*100).

image

Adds negative value support to `humanize.threeSigFigs`.


jzbz:
Looks perfect.

Issue #1473, opened 2019-07-20
fees display in scientific notation


raedah:
For some reason the fees in this block are displaying in scientific notation. Expected behavior is for it to show DCR is in the regular format with 8 decimal places.

https://explorer.dcrdata.org/block/14207

Fees: | 1e-08

chappjc:
Resolved by https://github.com/decred/dcrdata/issues/1489

PR #1489, opened 2019-07-28
explorer: show 8 decimal places in /tx tables and /block fees

### LINKED ISSUE #1486: Amounts are rounded to 6 decimal places on tx pages
DCR amounts are rounded to 6 decimal places on tx pages, despite the fact that DCR has precision to 8 places, and 8 places are displayed everywhere else AFAIK.

For example here: https://explorer.dcrdata.org/tx/ecadccbd7c19ab2fedd6e9c118759df595fda805eb0a69743eda49812171f39c we see `125.239848` for the first output.

screen

However, on the address page for that output (https://explorer.dcrdata.org/address/DsY1suSrkn8DYnF5BPLvnvSfFbeXTQusUVh), we see the full amount of `125.23984754`.

screen

It is extremely confusing to have block explorer rounding the actual amounts on the blockchain. Not to mention the fact that this inconsistency between the tx and address pages makes it even more confusing. Can we get this fixed and display the full 8 places everywhere?

Thanks for the great explorer otherwise, BTW!

buck54321:
Resolves #1486, #1473

~~In places where "decimalParts" was being used 1. in a table and 2. with 8 decimal places, replace with the simpler `{{printf "%.8f" .Value}}`. The only effect of "decimalParts" seemed to be to lighten the trailing zeros, which to me sort of just looked like a rendering problem when stacked in a table~~. ~~If anyone really liked that feature, let me know.  "decimalParts is still left in all other places besides full precision DCR values in tables~~.

### DIFF
--- a/views/block.tmpl
+++ b/views/block.tmpl
@@ -108,7 +108,7 @@
                       ><span class="d-sm-none">Tkt Price</span>:
                     <td class="text-left text-secondary">{{printf "%.8f" .SBits}}</td>
                     <td class="text-right font-weight-bold text-nowrap pr-2">Fees: </td>
-                    <td class="text-left text-secondary">{{.MiningFee}}</td>
+                    <td class="text-left text-secondary">{{printf "%.8f" .MiningFee}}</td>
                     <td class="d-none d-sm-table-cell text-right font-weight-bold text-nowrap pr-2">Pool Size: </td>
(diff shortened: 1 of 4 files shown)


RyanZim:
I actually slightly prefer the greyed-out trailing zeros, but no strong feelings.

buck54321:
No prob. I put 'em back.

PR #1491, opened 2019-07-30
use dcrutil/v2, rpcclient/v4, txscript/v2, and chaincfg/v2

chappjc:
For packages that use the blockchain or blockchain/stake modules, this also introduces a function `txhelpers.ParamsV2ToV1` to convert from `chaincfg/v2.Params` to `chaincfg.Params`.

This will be tagged as txhelpers/v3.0.1.


chappjc:
There will be no major version bumps since we just did a round of major version bumps for the dcrjson/v2 update and this is continues updating the dcrd deps on master, which we advice never to use.
There will also be changes shortly for database/v2 that will be included as part of our major version update.

davecgh:
Yes, there will ultimately be a `blockchain/v2` updated to use all the latest major module versions.

Issue #1500, opened 2019-07-31
Fix time remaining time on proposal chart


s-ben:
The proposal [Decred Fundamental Metrics Research Proposal - Phase 1](https://alpha.dcrdata.org/proposal/decred-fundamental-metrics-research-proposal-phase-1) just started voting yesterday. However the text below says 'Voting now~1d7h remaining'. I believe this is the time since voting started, not time remaining.

![image](https://user-images.githubusercontent.com/1634777/62182925-306e6a00-b30d-11e9-86f6-2d1c56d10cbb.png)

chappjc:
Resolved by https://github.com/decred/dcrdata/pull/1508

PR #1508, opened 2019-08-08
gov/proposal: Correct proposals remaining time

dmigwi:
Fix the proposal remaining time value on the proposal details page.

### DIFF
--- a/gov/politeia/types/types.go
+++ b/gov/politeia/types/types.go
@@ -244,10 +244,6 @@ func (pinfo *ProposalInfo) Metadata(tip, targetBlockTime int64) *ProposalMetadat
 	meta := new(ProposalMetadata)
 	desc := pinfo.VoteStatus.ShortDesc()
 	switch desc {
-	case "Authorized":
-		blocksTil := int64(windowSize)
-		meta.StartHeight = tip + blocksTil
-		meta.SecondsTil = blocksTil * targetBlockTime
 	case "Started", "Finished":
 		endHeight, _ := strconv.ParseInt(pinfo.ProposalVotes.Endheight, 10, 64)
 		meta.StartHeight = endHeight - windowSize
--- a/gov/politeia/types/types.go
+++ b/gov/politeia/types/types.go
@@ -240,12 +240,12 @@ type ProposalMetadata struct {
 // prepare figures for display. Many of these manipulations require a tip height
 // and a target block time for the network, so those must be provided as
 // arguments.
-func (pinfo *ProposalInfo) Metatdata(tip, targetBlockTime int64) *ProposalMetadata {
+func (pinfo *ProposalInfo) Metadata(tip, targetBlockTime int64) *ProposalMetadata {
 	meta := new(ProposalMetadata)
 	desc := pinfo.VoteStatus.ShortDesc()
 	switch desc {
 	case "Authorized":
-		blocksTil := windowSize - tip%windowSize
+		blocksTil := int64(windowSize)
 		meta.StartHeight = tip + blocksTil
 		meta.SecondsTil = blocksTil * targetBlockTime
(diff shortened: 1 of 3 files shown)


dmigwi:
#1500

gov/politeia/types/types.go
```diff
@@ -240,12 +240,12 @@ type ProposalMetadata struct {
...
 // and a target block time for the network, so those must be provided as
 // arguments.
-func (pinfo *ProposalInfo) Metatdata(tip, targetBlockTime int64) *ProposalMetadata {
+func (pinfo *ProposalInfo) Metadata(tip, targetBlockTime int64) *ProposalMetadata {
 	meta := new(ProposalMetadata)
 	desc := pinfo.VoteStatus.ShortDesc()
 	switch desc {
 	case "Authorized":
-		blocksTil := windowSize - tip%windowSize
+		blocksTil := int64(windowSize)
```
buck54321: You're correct that this was wrong, but I'm not sure this fixes it. Actually, I'm not sure we can know the `StartHeight` on an `"Authorized"` proposal. It starts whenever the admin approves. That was my mistake.

gov/politeia/types/types.go
```diff
@@ -246,7 +246,6 @@ func (pinfo *ProposalInfo) Metadata(tip, targetBlockTime int64) *ProposalMetadat
 	switch desc {
 	case "Authorized":
 		blocksTil := int64(windowSize)
-		meta.StartHeight = tip + blocksTil
 		meta.SecondsTil = blocksTil * targetBlockTime
```
buck54321: Can just use `windowSize` now. Don't need `blocksTil`.

views/proposal.tmpl
```diff
@@ -41,8 +41,8 @@
                 <span class="fs18 font-weight-bold lh1rem">Vote scheduled</span>
                 <div class="py-2">
                   {{.Username}} has authorized voting on this proposal.
-                  Voting begins at block {{$metadata.StartHeight}}
-                  (about <span class="medium-sans">{{secondsToShortDurationString $metadata.SecondsTil}}</span>)
+                  Waiting for the Politeia admin to approve the vote start.
+                  (should last for <span class="medium-sans">{{secondsToShortDurationString $metadata.SecondsTil}}</span>)
```
buck54321: How about "Awaiting admin approval (vote duration: 7 days)" or even just "Awaiting admin approval"? Drop the period as well.

`SecondsTil` is not the right field for this value. It's a constant of 7 days here, so we don't actually need a variable or even to list it.

gov/politeia/types/types.go
```diff
@@ -240,14 +240,12 @@ type ProposalMetadata struct {
...
-func (pinfo *ProposalInfo) Metatdata(tip, targetBlockTime int64) *ProposalMetadata {
+func (pinfo *ProposalInfo) Metadata(tip, targetBlockTime int64) *ProposalMetadata {
 	meta := new(ProposalMetadata)
 	desc := pinfo.VoteStatus.ShortDesc()
 	switch desc {
 	case "Authorized":
-		blocksTil := windowSize - tip%windowSize
-		meta.StartHeight = tip + blocksTil
-		meta.SecondsTil = blocksTil * targetBlockTime
+		meta.SecondsTil = int64(windowSize) * targetBlockTime
```
buck54321: Sorry @dmigwi. I was probably not being very clear here. We have no reason to set this field, and doing so is a little misleading. Seconds until what? This value, as you've set it, is a constant. It's just the size of the window converted to time, and is purposefully set to be 7 days. It is not the time til voting starts or ends or until any other time relevant to the metadata for an authorized proposal.

For the metadata, we expect some fields to be unset during certain stages of the proposal process. You can just drop the "Authorized" case altogether here.

Issue #1191, opened 2019-03-26
moving to the next design iteration


buck54321:
We've put in a lot of work towards achieving the vision of #451. Things don't align perfectly, but at this point, I think I'd like to close that PR and move forward with a new discussion. 

The guidance provided by the design team has been critical into making dcrdata one of the jewels of the decred ecosystem. I'd love to hear from the design team again on what we can do to keep this front-end momentum going in the right direction.

buck54321:
Here are a few things that are in my plans for dcrdata. Feedback, mockups, etc. are welcome and appreciated.

1. Better prioritization of homepage data. Really widdle down to the critical summary information, focusing on simple and intuitive graphical representations to go along with each important datum. 

2. A separate dashboard page will combine some of the current homepage elements, as well as charts and other stats (the [/stats](https://alpha.dcrdata.org/stats) page did not make it to prime time).

3. More color. Some dark backgrounds with light text. I'm not very good at it though, so I'd really like the design team's input.

4. Incorporate visual blocks. Not on the homepage. Maybe a /blocks page option?

5. An API info page with instructions for use, links to code samples, and sample responses on demand (AJAX and \<pre\> elements). 

6. An exchange page which displays the current exchange data and explains dcrdata's methodology on the price average. Expect to have market depth charts as well (#1041). Some limited price history might be possible too.

xaur:
I'm still addicted to dcrstats dashboard. The ticket price chart, ticket pool size, time left till ticket price change, price chart - I get a very comprehensive status of the network health by just scrolling this one dashboard for a few seconds, and immediately know if it's the time to buy more tickets.

If replacing that dashboard is something dcrdata team is looking at, I can give it a closer look and write up what elements exactly make it so great for me, and how dcrdata could outcompete it by adding missing features.

P.S. Sorry if this issue is not the proper place for such feedback.

buck54321:
This is the perfect place. Thank you. Definitely let me know what you like about that experience.

s-ben:
This all sounds good to me. Would just add that pretty, engaging data visualizations (even if not prominently displayed), can and do get used in research papers, medium think pieces, social media, etc. More pretty please.

RichardRed0x:
alpha.dcrdata.org is looking great, heading in a good direction.

Number 5 would be most useful for me. API instructions/examples should be relatively easy to produce but could be quite impactful if they allow additional people to start generating new charts/stats easily.

s-ben:
[...]

PR #1509, opened 2019-08-09
explorer: update header and blocks pages

buck54321:
#1191, https://github.com/decred/dcrdesign/issues/139

Based on mockup by @harlovski: https://matrix.decred.org:8448/_matrix/media/r0/download/decred.org/PejFvrGLlVlpiRuZaHeBwChS

I'd like to use a font closer to the one in the mockup. @harlovski, can you tell me what font you used? Your font falls somewhere in-between my source-sans regular and semi-bold. 

The mockup also used a slightly off-white for the header, where I'm still using white. The off-white looked okay on the blocks page, but not very good on others, Can try again later after more pages move towards this style. 

image


explorer/explorerroutes.go
```diff
@@ -540,7 +550,7 @@ func (exp *explorerUI) Blocks(w http.ResponseWriter, r *http.Request) {
 	}
 
 	rows, err := strconv.Atoi(r.URL.Query().Get("rows"))
-	if err != nil || (rows < minExplorerRows && rows == 0) {
```
buck54321: I couldn't figure out a reason why this was here. Seems that it should be one or the other, but not both.

chappjc: There's not much value in enforcing at least `minExplorerRows` either.  How about:

```go
if err != nil || rows < 1 {`
```

harlovski:
> #1191, [decred/dcrdesign#139](https://github.com/decred/dcrdesign/issues/139)

Hi, in current layouts i use regular weight. For table headers use also regular and #596D81
image

views/extras.tmpl
```diff
@@ -55,82 +55,70 @@
...
-<header class="top-nav"
-		id="navBar"
-		data-blocktime="{{.BlockTimeUnix}}">
-	<div class="container">
-		<div class="d-flex align-items-center flex-wrap">
-			<div class="d-flex align-items-center">
-				<div class="col-sm-auto" style="padding: 0 2px">
-					<a href="/" class="no-underline d-block home-link">
-						<div class="dcrdata-logo"></div>
+<header class="top-nav d-flex align-items-center"	id="navBar"	data-blocktime="{{.BlockTimeUnix}}">
```
chappjc: Tabs snuck in here.  Just about all I noticed in a cursory review.  Still testing though.

chappjc:
I think we should think about a different default page size.  IMO the default should be able to fit on the screen without scrolling, especially since clicking next is easier than bouncing up and down in the window.

10-15 might be good.  Here's 10:

image

chappjc:
Another thing that jumps out at me is the semi-transparent dashed line around the new logo.  It was like that before, but for some reason it looks more prominent.  Perhaps we can make the default item the first block in the table.

buck54321:
Should be good now.

PR #1511, opened 2019-08-10
proposals: show voter turnout as percentage

### LINKED ISSUE #1392: proposals list: show voter turnout percentage
Currently https://alpha.dcrdata.org/proposals shows absolute vote counts. That number is interesting and must be shown somewhere, but in this table I think 30% would be more interesting than 12369 (for the DEX spec proposal).

njirap:
Resolves #1392

Screen Shot 2019-08-10 at 1 42 55 PM

### DIFF
--- a/views/proposals.tmpl
+++ b/views/proposals.tmpl
@@ -137,6 +137,7 @@
                 {{range $i, $v := .Proposals}}
                 {{with $v}}
                     <tr>
+                        {{$votesTotalPercent :=  percentage $v.TotalVotes $v.NumOfEligibleVotes}}
                         <td class="text-left">
                             <a href="/proposal/{{.RefID}}">{{.Name}}</a>
                             <span class="d-none d-sm-inline">&nbsp;({{.Username}})</span>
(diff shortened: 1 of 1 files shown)


njirap:
`0` comes as `NaN% turnout` Looking into it

njirap:
It's `NaN` because `NumOfEligibleVotes` = `0`

njirap:
`NaN` fixed to `0`

Screen Shot 2019-08-10 at 2 16 20 PM

njirap:
Removed whitespace before `%`
Screen Shot 2019-08-10 at 2 24 55 PM

chappjc:
Now there's whitespace after the "(".  I think you'll have to sacrifice nice template code formatting for desired effect.

chappjc:
Actually, you can do stuff like `{{-- if eq $v.NumOfEligibleVotes 0}}` to consume white space prior.

njirap:
Whitespace is not stripped from the `0` because it is not in braces `{{}}`

Screen Shot 2019-08-11 at 5 40 29 PM

chappjc:
`-}}` works too.
https://golang.org/pkg/text/template/#hdr-Text_and_spaces

njirap:
Whitespace stripped from `0`
Screen Shot 2019-08-12 at 1 37 56 PM

PR #1513, opened 2019-08-14
hashes/addresses: add copy text icon plus copy text functionality

### LINKED ISSUE #127: Need copy text icon on all hashes and addresses
Ill have my eye out for a good way to do this that does not clutter the UI too much.

njirap:
Resolves #127
Screen Shot 2019-08-14 at 7 48 30 PM
Screen Shot 2019-08-14 at 7 52 40 PM
Screen Shot 2019-08-14 at 7 52 57 PM


public/fonts/icomoon.svg
```diff
@@ -4,29 +4,30 @@
 <metadata>Generated by IcoMoon</metadata>
 <defs>
 <font id="icomoon" horiz-adv-x="1024">
-<font-face units-per-em="1024" ascent="819.2" descent="-204.8" />
+<font-face units-per-em="1024" ascent="819.8005865102639" descent="-204.19941348973606" />
```
buck54321: Instead of using icomoon's font file directly, I've gotten into the habit of copy-pasting just the glyph element into the existing SVG. Then upload that file to generate the TTF and WOFF files. It's a little roundabout, but it prevents drifting on these font metric attributes. 

So the operation is this.

1. Add your icon to an empty set and generate + download the fonts
2. Open the SVG and copy the glyph element. Paste into this file: *public/fonts/icomoon.svg*
3. Upload this file and select all the icons. Make sure to deselect the single icon set from 1.
4. Generate, download, move just the TTF and WOFF files to *public/fonts/*

views/extras.tmpl
```diff
@@ -353,3 +353,10 @@
   data-rev-total="{{$likely.RevokeTotal}}"
   data-rev-count="{{.NumRevokes}}"
 {{end}}
+
+{{define "copyTextIcon"}}
+	&nbsp;
+	<span class="dcricon-copy clickable"
+	data-controller="extras"
```
buck54321: I know it's a WIP and you may not be this far yet, but I just wanted to make sure you didn't write this controller but forget to `git add` it.

njirap: I wrote it it just wasn't presentable yet just a bunch of console logs and window alerts

public/js/controllers/extras_controller.js
```diff
@@ -0,0 +1,24 @@
...
+    var copySupported = document.queryCommandSupported('copy')
+    var textContent = textElement.srcElement.parentNode.textContent.trim()
+    if (copySupported === true) {
+      var copyTextArea = document.createElement('textarea')
+      copyTextArea.value = textContent
+      document.body.appendChild(copyTextArea)
+      copyTextArea.select()
+      try {
+        document.execCommand('copy')
+        window.alert('Copied text ' + textContent + ' to clipboard')
```
njirap: I am open to suggestions for anything that would be better than this alert?

njirap:
Also the icon does not persist... working on that..

public/fonts/icomoon.svg
njirap: this is done

public/js/controllers/extras_controller.js
chappjc: I tend to like sites that gently flash "Copied text to clipboard" or similar, either on top of the copied text with a somewhat opaque background, or in a little blob of text above the button.  but just pulsing/flashing the button to acknowledge the click would be good too.

njirap:
> You seem to be on the right track. One thing that jumped out immediately was this extra vertical space in the table here.
> 
> image
> 
> This table should stay at a single row of text until very small widths, hence the truncated tx IDs.
> 
> I would suggest trying a pattern like the following
> 
> ```
> <a class="hash"><div class="dcricon-copy clipboard"></div>{{.Hash}}</a>
> ```
> 
> and then give the icon absolute positioning and hide it until hover
> 
> ```
> div > .clipboard {
>   position: absolute;
>   left: 0;
>   display: none;
>   z-index: 100;
> }
> 
> div:hover > .clipboard {
>   display: block;
> }
> ```
> 
> *untested
> 
> Absolute positioning removes the element from the layout so no extra space is taken. It's a little harder to trigger on mobile, but I think on mobile you can usually long click and "select link text", so it's less important there.

I am tweaking this approach because the copy icon being inside the `a` means on click of the copy icon the link will also have been clicked and user moved to the next page

njirap:
copy-text

njirap:
The icon now appears on hover
ctext

njirap:
There's a bug for addresses that have the `toggle OP_RETURN`
Screen Shot 2019-08-19 at 11 17 52 PM
Screen Shot 2019-08-19 at 11 18 00 PM

njirap:
All issues fixed here.

njirap:
All issues resolved here

views/extras.tmpl
```diff
@@ -375,3 +375,12 @@
 	</div>
 </div>
 {{end}}
+
+{{define "copyTextIcon"}}
+  <span class="dcricon-copy clickable"
+  data-controller="clipboard"
+  data-action="click->clipboard#copyTextToClipboard this"
```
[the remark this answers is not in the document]
njirap: turns our it's not really it..I've gotten rid of it

chappjc:
The PR is working great.

My last nit is that a pixel or two of padding is needed after the text and before the icon:

image

chappjc:
The "Copied!" box would be much more aesthetically pleasing if the corners were rounded:

image

Also with a touch more internal padding between the text and the boundary of the box.

njirap:
All changes made

chappjc:
Thank, @njirap!  Just one comment about PR workflow is that during review, it's best not to squash subsequent commits since having them makes it easier to see how change requests were handled.

njirap:
noted

chappjc:
The border radius does not seem to be working as intended:

image

njirap:
looking into it

chappjc:
It seems to be due to `border-color` of bootstrap's `alert-success`.  I think that it needs to be `transparent`.

chappjc:
I was just toying around in FF's debugger and I think the following additional changes to `span.alert-copy` would be nice:

```css
border-radius: 3px;
border-width: 3px 4px 5px 4px;
```

chappjc:
Another thing I noticed, is that in some places it can be very difficult to actually click on the copy icon.

image

In the above case, it's really hard to actually click on the copy icon that appears for prev out 0 because (afaict) when you move your mouse over to it, you hit the padding of the adjacent address span.

buck54321:
The mouse interaction issue can likely be solved by setting the copy icon's (or possibly an ancestors) `z-index`.

njirap:
All changes made

public/scss/utils.scss
```diff
@@ -217,3 +217,22 @@ body.darkBG [data-tooltip]:hover::after {
   border-color: #2dd8a3;
   color: #2dd8a3;
 }
+
+.clipboard span.clickable,
+.clipboard span.alert-success {
```
buck54321: Though it likely won't affect anything because it is nested here, let's avoid modifying the Boostrap `alert-success` class. Add a custom class if necessary.

njirap: this is done

chappjc:
Would be good to follow this up with a PR for copy icons for the txhashes on a block page

njirap:
@chappjc I am working on the new designs...Will add that when I get to the block page

chappjc:
I'd like to assign the issues.  Which one are you on now?  Also I believe we might want an additional dev to work in parallel on these issues too, so we'll need to coordinate well.

njirap:
I'm on the mempool page now.. I basically move in the order in which they appear on the designs https://xd.adobe.com/view/800a9bfa-dcb9-4985-58fb-ae312bfb80d2-f63b/screen/061c3e84-82df-49ce-8a06-72bc360be393/Explore-blockchain/

PR #1518, opened 2019-08-16
middleware: use tollbooth/v5 and drop tollbooth_chi

chappjc:
Write our own middleware so we don't need tollbooth_chi.

Requires a fix to tollbooth: https://github.com/didip/tollbooth/pull/77


chappjc:
Upstream PR https://github.com/didip/tollbooth/pull/77 merged.

chappjc:
Re-tested with dependency on fixed tollbooth (`v5.1.1-0.20190817151620-2c720dff9427`).

PR #1528, opened 2019-08-19
multi: rework homepage

buck54321:
Changes the layout and information of the homepage. Replaces the mempool line gauge with a pie chart. Adds a simple line graph. Both new charts have both SVG and canvas versions, so that no-js viewers can have a chart too. Removes the live transactions table. Adds data for unique address count.

This work was started before harlovski's mockups (https://github.com/decred/dcrdesign/issues/139#issuecomment-522582493) were posted, so this is a draft while I come together with the design team. 

The images show three different widths as well as night mode and the no-js chart versions.

image

image


xaur:
Some quick personal perspective on the images above:

- Stake Returns box:
  - use "Rewards" to avoid entrenching further the already widespread ideas about Decred's PoS "returns", "passive income", "dividends", "interest" and similar
  - add "yearly" or "ann" or something to mark that 8.69% is per year
- Mining Profit box: if possible, add a tiny label to explain what is `+1.21%`
- Ticket Price box: the USD price doesn't need the 2 decimals, while the DCR price could use one
- Exchange Rate box: add link to /markets page
- Mempool box: "vote" -> "votes"; consider "tkt" -> "tix" (minor, I've just seen that used more often)
- Ticket Price chart: add bars of tickets purchased in each interval like in dcrstats
- I really miss the USD/BTC price chart
- I like the version where Ticket Price chart and Staking are side by side
  - this places almost all PoS info I need in one horizontal block
- PoS info is scattered across the Ticket Price box, Stake Returns box and the Staking box - how about combining that
  - same for Mining

buck54321:
This obviously needs some re-thinking. Closing for now, though I might open up a couple PRs to incorporate the new data feeds and some other optimizations from this PR. 

@chappjc, did you happen to see the improved algorithm for `func threeSigFigs` in `templates.go`? At least I think it's an improvement.

PR #1529, opened 2019-08-21
docker: bump node and go versions

dajohi:


Dockerfile
```diff
@@ -1,17 +1,17 @@
-FROM golang:1.12.4 as daemon
+FROM golang:1.12.9 as daemon
 
 COPY . /go/src
 WORKDIR /go/src
 RUN env GO111MODULE=on go build
 
-FROM node:10.15.3 as gui
+FROM node:10.16.3 as gui
```
chappjc: Is 11.15 available?

chappjc: As per matrix chat, current (12.x) is preferable for smoking out issues we'll eventually have to face.

dajohi: travis builds the Dockerfile, which will test LTS.
travis also builds using the decred/dcrdata-golang-builder container which has current (12.x)

chappjc:
So we're going to dump Docker and update to Go 1.13, right?

chappjc:
Can I haz github actions?

PR #1531, opened 2019-08-21
require at least go 1.12 in each go.mod

chappjc:
The reason for keeping these updated with latest-1 is that these version lines in go.mod apparently can also activate certain go features, according to the [Go 1.13 beta release notes](https://tip.golang.org/doc/go1.13#language):

> If your code uses modules and your go.mod files specifies a language version, be sure it is set to at least 1.13 to get access to these language changes. You can do this by editing the go.mod file directly, or you can run go mod edit -go=1.13.

I do not know of any such features that are enabled with `go 1.12`, but updating go.mod in this way should be done on each new release.

### DIFF
--- a/api/types/go.mod
+++ b/api/types/go.mod
@@ -1,6 +1,6 @@
 module github.com/decred/dcrdata/api/types/v4
 
-go 1.11
+go 1.12
 
 require (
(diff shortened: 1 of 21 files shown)


chappjc:
And Go 1.13 has landed.  dcrdata no longer supports Go 1.11.

PR #1534, opened 2019-08-23
proposals: fix percentage completion

### LINKED ISSUE #1527: proposal page time remaining
Voting now ~2d 14h remaining  <-- this appears to be wrong.

proposal list view says  55m 0s left  for the same proposal.

njirap:
also adds whitespace to voting now display
Resolves #1527


chappjc:
> Looks good. I wonder if we should take the floor for all of the percentages in this table.

Might be warranted.  Maybe for a followup PR from @njirap if interested.

PR #1542, opened 2019-09-03
dcrpg: signal heightClient channels on new best block

### LINKED ISSUE #1530: /api/status/happy not becoming happy after startup
Likely related to the sync changes when dcrsqlite was removed.

buck54321:
Resolves #1530

### DIFF
--- a/db/dcrpg/pgblockchain.go
+++ b/db/dcrpg/pgblockchain.go
@@ -3583,6 +3583,9 @@ func (pgb *ChainDB) StoreBlock(msgBlock *wire.MsgBlock, isValid, isMainchain,
 			return
 		}
 
+		// Signal updates to any subscribed heightClients.
+		pgb.SignalHeight(dbBlock.Height)
+
 		// Insert the block stats.
 		if isMainchain && tpi != nil {
 			err = InsertBlockStats(pgb.db, blockDbID, tpi)
--- a/db/dcrpg/pgblockchain.go
+++ b/db/dcrpg/pgblockchain.go
@@ -2899,6 +2899,10 @@ func (pgb *ChainDB) Store(blockData *blockdata.BlockData, msgBlock *wire.MsgBloc
 	_, _, _, err := pgb.StoreBlock(msgBlock, isValid, isMainChain,
 		updateExistingRecords, updateAddressesSpendingInfo,
 		updateTicketsSpendingInfo, blockData.Header.ChainWork)
+
+	// Signal updates to any subscribed heightClients.
+	pgb.SignalHeight(msgBlock.Header.Height)
+
 	return err
 }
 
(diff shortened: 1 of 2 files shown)


chappjc:
When starting up dcrdata (about 4000 blocks behind), a sql error appears for each block:

```
...
[ERR] JAPI: BlockData summary is nil for height 372492.
[ERR] PSQL: Unable to retrieve block 372496 summary: sql: no rows in result set
[ERR] JAPI: BlockData summary is nil for height 372496.
[ERR] PSQL: Unable to retrieve block 372497 summary: sql: no rows in result set
[ERR] JAPI: BlockData summary is nil for height 372497.
[INF] PSQL: Processing blocks 372501 to 373000...
[ERR] PSQL: Unable to retrieve block 372503 summary: sql: no rows in result set
[ERR] JAPI: BlockData summary is nil for height 372503.
[ERR] PSQL: Unable to retrieve block 372506 summary: sql: no rows in result set
[ERR] JAPI: BlockData summary is nil for height 372506.
[ERR] PSQL: Unable to retrieve block 372509 summary: sql: no rows in result set
[ERR] JAPI: BlockData summary is nil for height 372509.
[ERR] PSQL: Unable to retrieve block 372521 summary: sql: no rows in result set
[ERR] JAPI: BlockData summary is nil for height 372521.
[ERR] PSQL: Unable to retrieve block 372525 summary: sql: no rows in result set
[ERR] JAPI: BlockData summary is nil for height 372525.
[ERR] PSQL: Unable to retrieve block 372537 summary: sql: no rows in result set
[ERR] JAPI: BlockData summary is nil for height 372537.
[ERR] PSQL: Unable to retrieve block 372541 summary: sql: no rows in result set
...
```

db/dcrpg/pgblockchain.go
```diff
@@ -3583,6 +3583,9 @@ func (pgb *ChainDB) StoreBlock(msgBlock *wire.MsgBlock, isValid, isMainchain,
 			return
 		}
 
+		// Signal updates to any subscribed heightClients.
+		pgb.SignalHeight(dbBlock.Height)
```
chappjc: One way to avoid the block summary sql error during startup sync might be to put this in the `if !pgb.InBatchSync {` block below.  I'd have to look to see how this was handled in the sqlite days.

chappjc: Actually, looking at the error log I posted, it's not every block.

Looking at `(*appContext).StatusNtfnHandler`, I think what we're seeing is a inherent race with `c.DataSource.GetBestBlockSummary()`.  It first gets the best block height from the meta table, and then attempts to retrieve block summary data for a block at that height (from the blocks and stats table).  But in ChainDB.StoreBlock, `InsertBlockStats` (and `SignalHeight`) are called after the meta table is updated.  So, I think SignalHeight needs to be after those two things, and SignalHeight should be after all DB updates.

buck54321:
I think I've got it like it was in dcrsqlite now. One call at the end of `(*ChainDB).SyncChainDB`, and one at the end of `(*ChainDB).Store`, which isn't called during the sync.

PR #1554, opened 2019-09-05
deal with invalid module pseudo-versions in transitive deps

chappjc:
Go 1.13 is more strict when it comes to validating module pseudo-versions.  There are four transitive dependencies with invalid pseudo-versions that must be replaced as described at https://tip.golang.org/doc/go1.13#version-validation

```
go: github.com/decred/dcrdata/gov/v2@v2.0.2 requires
	github.com/decred/politeia@v0.0.0-20190716200512-2a928f2c2084 requires
	github.com/marcopeereboom/sbox@v0.0.0-20190125180204-32a7c85e429a requires
	github.com/golangci/golangci-lint@v1.13.1 requires
	github.com/go-critic/go-critic@v0.0.0-20181204210945-0af0999fabfb: invalid pseudo-version: does not match version-control timestamp (2019-01-08T19:27:14Z)

github.com/golangci/gofmt@v0.0.0-20181105071733-0b8337e80d98: invalid pseudo-version: does not match version-control timestamp (2018-12-22T12:35:16Z)

github.com/golangci/go-tools@v0.0.0-20180109140146-35a9f45a5db0: invalid pseudo-version: does not match version-control timestamp (2019-01-24T09:00:46Z)

github.com/golangci/errcheck@v0.0.0-20181003203344-ef45e06d44b6: invalid pseudo-version: does not match version-control timestamp (2018-12-23T08:41:20Z)
```

These are all coming in through the politeia import and github.com/marcopeereboom/sbox.

PR #1555, opened 2019-09-05
update politeia and go mod tidy

chappjc:
Related to https://github.com/decred/dcrdata/pull/1554

gov: update politeia and go mod tidy
dcrdata: update go.{mod,sum} for new politeia import


chappjc:
Note that this is necessary with Go 1.13 with a cleared modcache.  You may not notice the invalid pseudo-version errors unless you `go clean -modcache` first, which I would advise doing with a new Go version.

PR #1561, opened 2019-09-28
exchanges: better websocket error handling and reconnect logic

buck54321:
Adds a progressive reconnection delay, so even persistent errors can be recovered. Fixes a bug where the error counter was not being reset.

### DIFF
--- a/exchanges/exchanges.go
+++ b/exchanges/exchanges.go
@@ -761,7 +761,7 @@ func (xc *CommonExchange) wsDepthStatus(connector func()) (tryHttp, initializing
 			var delay time.Duration
 			// wsDepthStatus is only called every DataExpiry, so a delay of zero is ok
 			// until there are a few consecutive errors.
-			switch true {
+			switch {
 			case errCount < 5:
 			case errCount < 20:
 				delay = 10 * time.Minute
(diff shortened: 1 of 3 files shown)


exchanges/exchanges.go
```diff
@@ -752,12 +758,23 @@ func (xc *CommonExchange) wsDepthStatus(connector func()) (tryHttp, initializing
 			log.Tracef("using http fallback for %s orderbook data", xc.token)
 			tryHttp = true
 			errCount := xc.wsErrorCount()
-			if errCount < wsMaxErrors {
+			var delay time.Duration
+			// wsDepthStatus is only called every DataExpiry, so a delay of zero is ok
+			// until there are a few consecutive errors.
+			switch true {
```
chappjc: I'd prefer just `switch {`, but it's the same.

chappjc:
Deployed to the rateserver.  Keeping an eye on https://dcrdata.decred.org/market?chart=depth&xc=aggregated&bin=1h&stack=1

chappjc:
Still showing the three enabled exchanges on dcrdata.decred.org.  Interestingly, alpha is missing Polo.  Whatever is bothering alpha, this PR is clearly a great improvement.

PR #1567, opened 2019-10-02
build: replace travis-ci with ci via github actions

dajohi:


lint.sh
```diff
@@ -1,61 +0,0 @@
-#!/usr/bin/env bash
```
chappjc: I do use lint.sh fairly.  Tests and npm run build take so long that I doubt I'll just want to do run_tests.sh instead.  Any way to keep this separate still?

lint.sh
```diff
@@ -1,61 +0,0 @@
...
-
-export GO111MODULE=on
-
-go version
-
-pushd $ROOT > /dev/null
-
-ROOTPATH=$(go list -m -f {{.Dir}} 2>/dev/null)
-ROOTPATHPATTERN=$(echo $ROOTPATH | sed 's/\\/\\\\/g' | sed 's/\//\\\//g')
-MODPATHS=$(go list -m -f {{.Dir}} all 2>/dev/null | grep "^$ROOTPATHPATTERN")
```
chappjc: Are these three vars computed right?  I though there were some changes to these in dcrd and dcrwallet to fix something, but it never made it to dcrdata.

lint.sh
dajohi: i'll just not delete it :)

PR #1571, opened 2019-10-04
pubsub: fix possible dropped messages

chappjc:
When sending a signal to a pubsub client (i.e. `*spoke <- hubMsg`) and
the channel buffer is full, that does not mean the client is dead.
Instead, block the send unless `client.killed` is closed (when its
`WebSocketHandler` returns after both send and receive loops have
returned). There is no longer a `default` case.

Assign each client a random id, and add more logging around address messages.

Remove `SQLT` log subsystem from sample-dcrdata.conf

Add timeout for sends on the `hubSpoke` message channel send to
catch programmer error that could lead to deadlock.

### DIFF
--- a/pubsub/websocket.go
+++ b/pubsub/websocket.go
@@ -5,6 +5,8 @@
 package pubsub
 
 import (
+	"crypto/rand"
+	"encoding/binary"
 	"fmt"
 	"sync"
 	"sync/atomic"
@@ -106,6 +108,7 @@ func (wsh *WebsocketHub) SetReady(ready bool) {
 
 type client struct {
 	sync.RWMutex
+	id     uint64
 	subs   map[pstypes.HubSignal]struct{}
 	addrs  map[string]struct{}
 	killed chan struct{}
@@ -114,6 +117,7 @@ type client struct {
 
 func newClient() *client {
 	return &client{
+		id:     newClientID(),
 		subs:   make(map[pstypes.HubSignal]struct{}, 16),
 		addrs:  make(map[string]struct{}, 16),
 		killed: make(chan struct{}),
@@ -211,6 +215,15 @@ func NewWebsocketHub() *WebsocketHub {
 type clientHubSpoke struct {
 	cl *client
 	c  *hubSpoke
+}
+
+func newClientID() uint64 {
+	var id [8]byte
+	n, err := rand.Read(id[:])
+	if n != 8 || err != nil {
+		panic("newClientID: rand.Read failed!")
+	}
+	return binary.LittleEndian.Uint64(id[:])
 }
 
 // NewClientHubSpoke registers a connection with the hub, and returns a pointer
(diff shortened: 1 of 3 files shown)


pubsub/websocket.go
```diff
@@ -213,6 +228,15 @@ type clientHubSpoke struct {
 	c  *hubSpoke
 }
 
+func newClientID() uint64 {
+	var id [8]byte
+	n, err := rand.Read(id[:])
```
chappjc: crypto/rand, but perhaps this should be a monotonically increasing integer counter instead.

pubsub/websocket.go
```diff
@@ -173,17 +184,21 @@ func (c *client) unsubscribe(msg pstypes.HubMessage) error {
 			return fmt.Errorf("msg.Msg not an AddressMessage (SigAddressTx): %T", msg.Msg)
 		}
 		delete(c.addrs, am.Address)
+		// Unsubscribe from address signals ONLY if this client has no more
+		// watched addresses.
+		if len(c.addrs) == 0 {
+			delete(c.subs, pstypes.SigAddressTx)
+		}
```
chappjc: This is the primary bug fix here.

pubsub/websocket.go
```diff
@@ -397,6 +445,9 @@ func (wsh *WebsocketHub) Run() {
 				}
 			case sigPingAndUserCount:
 				log.Tracef("Signaling ping/user count to %d websocket clients.", clientsCount)
+				// Ping all clients (not a subscription).
```
chappjc: *no longer* a subscription.  Why was it ever a subscription?  I dunno.

pubsub/websocket.go
```diff
@@ -368,6 +392,30 @@ func (wsh *WebsocketHub) Run() {
...
+		// Signal or unregister the client.
+		timer := time.NewTimer(5 * time.Second)
+		select {
+		case <-client.killed:
+			log.Tracef("Unable to send %s message to client %d: gone (killed)",
+				hubMsg, client.id)
+			wsh.unregisterClient(spoke)
+		case *spoke <- hubMsg:
+			log.Tracef("Sent %s message to client %s.", hubMsg, client.id)
+		case <-timer.C:
```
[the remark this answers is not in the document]
chappjc: Right, that's my assessment too regarding blocking.  Regarding load testing, I've just had it running on alpha for a week or so, but hitting it hard with a lot of clients and address subscriptions would be good too.

chappjc:
> Just a few notes. Also, I was unable to build with 1.13.
> 
> ...`https://sum.golang.org/lookup/github.com/asdine/storm@v2.2.0+incompatible: 410 Gone`
> 
> but I just used 1.12.

Just rebased on master with a recently merged fix for that storm import.

pubsub/websocket.go
chappjc: Just for logging.  I'll switch to a counter.  Although part of the idea is that with the log files that span multiple restarts, grepping for a unique id helps debugging.

PR #1573, opened 2019-10-04
[5.1] pubsub: fix possible dropped messages

chappjc:
When sending a signal to a pubsub client (i.e. `*spoke <- hubMsg`) and
the channel buffer is full, that does not mean the client is dead.
Instead, block the send unless `client.killed` is closed (when its
`WebSocketHandler` returns after both send and receive loops have
returned). There is no longer a `default` case.

Assign each client a random id, and add more logging around address messages.

Remove `SQLT` log subsystem from sample-dcrdata.conf

Add timeout for sends on the `hubSpoke` message channel send to
catch programmer error that could lead to deadlock.


chappjc:
These changes, which will be in 5.1.3-release, require the following new modules tags:
- pubsub/v3.1.0 (from pubsub/v3.0.6)
- pubsub/types/v3.1.0 (from pubsub/types/v3.0.4)

PR #1574, opened 2019-10-05
multi: cleanup whitespace and preallocate a few variables

dajohi:


db/dbtypes/types.go
```diff
@@ -1831,7 +1831,8 @@ func ReduceAddressHistory(addrHist []*AddressRow) (*AddressInfo, float64, float6
 	}
 
 	var received, sent, fromStake, toStake int64
-	var transactions, creditTxns, debitTxns []*AddressTx
+	var creditTxns, debitTxns []*AddressTx
+	transactions := make([]*AddressTx, 0, len(addrHist))
```
chappjc: There's the possibility this will over-allocate because of `!addrOut.ValidMainChain`, but the vast majority of most transactions are on stakeholder-approved main chain blocks, so this is still a good optimization.

gov/agendas/deployments.go
```diff
@@ -176,7 +176,7 @@ func agendasForVoteVersion(ver uint32, client DeploymentSource) ([]AgendaTagged,
 // data contents but chaincfg.Vote does not contain the important vote status
 // field that is found in chainjson.Agenda.
 func (db *AgendaDB) updatedb(activeVersions map[uint32][]chaincfg.ConsensusDeployment) (int, error) {
-	var agendas []AgendaTagged
+	agendas := make([]AgendaTagged, 0, len(activeVersions))
```
chappjc: Doesn't hurt here either, but it might be a bit misleading because each `activeVersions` element can result in more than one new `AgendaTagged` being appended (`agendas = append(agendas, taggedAgendas...)`).

PR #1575, opened 2019-10-05
build: enter the actual module path to run tests

dajohi:


run_tests.sh
```diff
@@ -43,18 +43,19 @@ ROOTPATHPATTERN=$(echo $ROOTPATH | sed 's/\\/\\\\/g' | sed 's/\//\\\//g')
...
-    --enable govet \
-    --enable staticcheck \
-    --enable gosimple \
-    --enable unconvert \
-    --enable ineffassign \
-    --enable structcheck \
-    --enable goimports \
-    --enable misspell \
-    --enable unparam
+  (cd ${module} && \
```
chappjc: A problem with this is that the replace statements in to root module are no longer applied.  The module path being tested would need it's own replaces, but we tend not to do that.  How do we handle this in other repos?

chappjc: @dajohi Any ideas how to address this issue?

db/cache/charts.go
```diff
@@ -1254,13 +1254,13 @@ func dailyHashrate(time, chainwork ChartUints) (ChartUints, ChartUints) {
...
 	for i, t := range time[1:] {
-		tDiff := t - time[i]
+		tDiff := int64(t - time[i])
 		if tDiff <= 0 {
 			tDiff = aDay
 			dupes += 1
 		}
 		workDiff := chainwork[i+1] - chainwork[i]
-		rates = append(rates, (workDiff)*1e6/tDiff)
+		rates = append(rates, (workDiff)*1e6/uint64(tDiff))
```
chappjc: @buck54321 Would you have a look at this?  Any gotchas?  This seems to be a solid bugfix so `tDiff <= 0` can work.

buck54321: Oh yeah. Good catch.

run_tests.sh
chappjc: At least for the tests, https://github.com/decred/dcrd/blob/master/run_tests.sh#L31

I would think the same issue exists for the linter (testing the pinned imports, not the ones in the checked out repo), but I don't know for sure.

db/cache/charts.go
chappjc: Merged separately in https://github.com/decred/dcrdata/pull/1589

PR #1577, opened 2019-10-10
blocks: Redesign Block Page

### LINKED ISSUE #1545: Redesign blockpage
current:
https://explorer.dcrdata.org/blocks
image

redesign:
image
Minimal indent between columns in table - 30px. The table does not fit on the phone screen, so on phone we use horisontal scroll.

specs:
https://xd.adobe.com/spec/029fc9e7-7209-4326-6bab-07544195825f-74ae/

njirap:
Resolves #1545


public/js/controllers/pagenavigation_controller.js
```diff
@@ -4,14 +4,12 @@ import Url from 'url-parse'
 
 export default class extends Controller {
   static get targets () {
-    return ['pagesize', 'votestatus', 'listview']
+    return ['pagesize', 'votestatus', 'listview', 'pageage']
```
dmigwi: ` 'age'` should just work fine here.

views/timelisting.tmpl
```diff
@@ -17,61 +17,50 @@
...
-                          </li>
-                          {{end}}
-                      </ul>
-                  </nav>
+                <span class="fs12 nowrap text-right">
+                    <select
+                        data-target="pagenavigation.pageage"
+                        data-action="change->pagenavigation#setPageAge"
+                        data-rows="{{.Limit}}"
+                        data-offsetkey="offset"
```
dmigwi: This key is clashing with https://github.com/decred/dcrdata/pull/1577/files#diff-410e820c1016f20eeb74ac755a218848R31, please change it.
Actually you should not use `data-offsetkey` here.

views/windows.tmpl
```diff
@@ -17,62 +17,49 @@
...
-                    </li>
-                    {{end}}
-                  </ul>
-                </nav>
+                <span class="fs12 nowrap text-right">
+                    <select
+                        data-target="pagenavigation.pageage"
+                        data-action="change->pagenavigation#setPageAge"
+                        data-rows="{{.Limit}}"
+                        data-offsetkey="offset"
```
dmigwi: Comment about offset also applies here.

public/js/controllers/pagenavigation_controller.js
dmigwi: Actually `datastatus` fits the `Newer` and `older` statuses much more better than `age`.

public/js/controllers/pagenavigation_controller.js
```diff
@@ -21,6 +19,18 @@ export default class extends Controller {
     Turbolinks.visit(url.toString())
   }
 
+  setPageAge () {
+    var url = Url(window.location.href)
+    var q = Url.qs.parse(url.query)
+    q[this.pagesizeTarget.dataset.offsetkey] = this.pageageTarget.selectedOptions[0].value
```
dmigwi: The difference between `setPageAge` and  `setPageSize` is this line.

As it is now, this line is overwriting the offset value set in `setPageSize`, if you come up with a new dataset key and value instead of offset you can easily add this line to `setPageSize` and call it instead.

views/explorer.tmpl
```diff
@@ -16,61 +16,48 @@
...
-                <span class="h2 d-flex pt-4 pb-1 pr-2">Blocks</span>
-
+                <span class="h2 d-flex pt-4 pb-1 pr-2">
+                    Per blocks
+                    <span class="dcricon-info ml-3 mt-1"></span>
+                </span>
                 <div class="pb-1 d-flex justify-content-end align-items-end flex-wrap">
-                  <span class="fs12 nowrap text-secondary px-2 my-2">
-                      {{intComma (add $Offset 1)}} &ndash; {{intComma (add $Offset .Rows)}} of {{intComma (add .BestBlock 1) }} rows
-                  </span>
```
buck54321: I know it's not in the spec,but please leave this row range display.

views/explorer.tmpl
```diff
@@ -106,14 +93,14 @@
                             <span class="d-none d-md-inline">Version</span>
                             <span class="d-md-none position-relative" data-tooltip="block version">Ver</span>
                         </th>
-                        <th class="text-right" class="jsonly">Age</th>
-                        <th class="text-right px-0">Time</th>
+                        <th class="text-center" class="jsonly">Age</th>
+                        <th class="text-center px-0">Time</th>
```
buck54321: So this is an interesting problem. The mockups have this right aligned, but with the title centered over the text, but that's actually pretty hard to do with CSS.

Typically our tables always use right alignment on the last column. It helps to frame the table out. 

If you can figure out how to replicate the mockup, please do. Otherwise, leave this as right-aligned.

views/explorer.tmpl
```diff
@@ -122,8 +109,8 @@
                         <td class="text-center" data-type="value">{{threeSigFigs .Total}}</td>
                         <td class="text-center" data-type="size">{{.FormattedBytes}}</td>
                         <td class="text-center d-none d-sm-table-cell" data-type="version">{{.Version}}</td>
-                        <td class="text-right" data-type="age" class="jsonly" data-target="time.age" data-age="{{.BlockTime.UNIX}}"></td>
-                        <td class="text-right px-0" data-type="time">{{.BlockTime}}</td>
+                        <td class="text-center" data-type="age" class="jsonly" data-target="time.age" data-age="{{.BlockTime.UNIX}}"></td>
+                        <td class="text-center px-0" data-type="time">{{.BlockTime}}</td>
```
buck54321: See note above. This needs to be right-aligned, as per the mockup and general table rules, but it's difficult to get the header centered over the row.

views/explorer.tmpl
```diff
@@ -16,61 +16,48 @@
             {{$Offset := (subtract .BestBlock $topBlock)}}
             {{$pendingBlocks := ((index .Data 0).Height)}}
             <div class="d-flex justify-content-between align-items-end">
-                <span class="h2 d-flex pt-4 pb-1 pr-2">Blocks</span>
-
+                <span class="h2 d-flex pt-4 pb-1 pr-2">
+                    Per blocks
+                    <span class="dcricon-info ml-3 mt-1"></span>
```
buck54321: It's better as just "Blocks", but leave the icon. Maybe make the icon a little bit smaller.

public/scss/navigation.scss
```diff
@@ -12,6 +12,7 @@
   top: 0;
   background: #fff;
   border-bottom: 1px solid #00000017;
+  box-shadow: 0 0 25px #dcdcdc;
```
buck54321: Doesn't work for dark mode. 

image

As a general rule, we've always avoided box shadows. I think it kinda works here though. Thoughts @chappjc? My thinking is partially based on guidance from @gozart1 from https://github.com/decred/dcrdata/pull/921#issuecomment-455096066

views/explorer.tmpl
```diff
@@ -16,61 +16,48 @@
...
-                      {{if eq $blocksCount 20 30 50 100 .WindowSize}}{{else}}<option selected value="{{$blocksCount}}">{{$blocksCount}} per page</option>{{end}}
+                        <option selected disabled>New<option>
+                        {{if lt $topBlock .BestBlock}}
+                            <option value="{{.BestBlock}}">Newest</option>
+                            <option value="{{add $topBlock (int64 $blocksCount)}}">Newer</option>
+                        {{end}}
+                        {{if ge $topBlock .Rows}}
+                            <option value="{{subtract $topBlock (int64 $blocksCount)}}">Older</option>
+                            <option value="{{subtract .Rows 1}}">Oldest</option>
+                        {{end}}
```
buck54321: By making this a dropdown, our no-js users can no longer use this navigation. dcrdata gives high priority to no-js users.

public/scss/navigation.scss
```diff
@@ -126,16 +126,32 @@ ul.pages > li:last-child > a {
...
+select {
+  -webkit-appearance: none;
+  -moz-appearance: none;
   border: none;
+  height: 30px;
   font-size: 14px;
-  height: 24px;
   padding: 0 5px;
+  border-radius: 2px;
+  background-color: #fff !important;
```
buck54321: Doesn't work for dark mode.

image

Also, why the `!important`?

public/scss/icons.scss
```diff
@@ -133,6 +133,11 @@
   content: "\e92c";
 }
 
+.dcricon-info::before {
+  content: "\e916";
+  color: rgba(0, 0, 0, 0.5);
```
buck54321: Drop the color specification here. Icons should default to text color. Also, this breaks dark mode.

public/scss/navigation.scss
chappjc: I certainly don't want my browser eating more CPU than needed, but I've not qualified the assertion that it is more demanding to render.

Regardless, the shadow is OK, although about twice as big as it needs to be.

chappjc:
Please use accompany the PR and any significant changes with a screenshot.

views/explorer.tmpl
njirap: @buck54321 it's a dropdown in the designs 
Screen Shot 2019-10-17 at 3 10 54 PM


https://xd.adobe.com/view/800a9bfa-dcb9-4985-58fb-ae312bfb80d2-f63b/screen/061c3e84-82df-49ce-8a06-72bc360be393/Explore-blockchain

buck54321: I understand that. The design team largely ignored our discussions on this and other aspects of the layout, which is part of the reason these things need to be discussed before you do them. 

Navigation needs to be accessible to no-js users, which is why I left it as a button instead of a select.

public/js/controllers/pagenavigation_controller.js
njirap: this will no longer be needed as it's going back to being buttons

views/timelisting.tmpl
njirap: this will no longer be needed as it's going back to being buttons

views/windows.tmpl
njirap: this will no longer be needed as it's going back to being buttons

views/explorer.tmpl
chappjc: Even if this worked without JS, I honestly don't understand the motivation for putting these in a dropdown.  It forces you to click twice.

njirap:
Screen Shot 2019-10-18 at 12 20 08 AM
Screen Shot 2019-10-18 at 12 20 29 AM

njirap:
The force push does not affect comment history, it was necessary as I have rebased against the latest master due to the merge conflicts that were

njirap:
All requested changes made.

views/extras.tmpl
```diff
@@ -196,23 +196,6 @@
...
-			<option {{if eq . "years"}}selected {{end}}value="years">Years</option>
-			<option {{if eq . "months"}}selected {{end}}value="months">Months</option>
-			<option {{if eq . "weeks"}}selected {{end}}value="weeks">Weeks</option>
-			<option {{if eq . "days"}}selected {{end}}value="days">Days</option>
-			<option {{if eq . "windows"}}selected {{end}}value="ticketpricewindows">Windows</option>
-			<option {{if eq . "blocks"}}selected {{end}}value="blocks">Blocks</option>
-		</select>
-	</div>
-{{end}}
-
```
njirap: this may not be related to the task, just found an unused template and removed it

views/timelisting.tmpl
```diff
@@ -17,27 +17,32 @@
...
-                  <span class="fs12 nowrap text-right">
+                  <span class="fs12 nowrap text-right mr-2">
                       <select
                           data-target="pagenavigation.pagesize"
                           data-action="change->pagenavigation#setPageSize"
                           data-offset="{{$.Offset}}"
                           data-offsetkey="offset"
                           class="dropdown text-secondary my-2 {{if lt $pending 10}}disabled{{end}}"
-                          {{if lt $pending 10}}disabled="disabled"{{end}}
+                          {{if lt $pending 10}}disabled='disabled'{{end}}
```
buck54321: Why the single quotes?

views/timelisting.tmpl
```diff
@@ -17,27 +17,32 @@
...
-                      {{if eq $dropVal $count 20 30 50 100}}{{else}}<option value="{{$dropVal}}">{{$dropVal}} per page</option>{{end}}
-                      {{if eq $count 20 30 50 100 200}}{{else}}<option selected value="{{$count}}">{{$count}} per page</option>{{end}}
+                        <option {{if eq $count 10}}selected{{end}} value="10">10 per page</option>
+                        <option {{if eq $count 20}}selected{{end}} value="20">20 per page</option>
+                        <option {{if eq $count 30}}selected{{end}} value="30">30 per page</option>
+                        <option {{if eq $count 40}}selected{{end}} value="40">40 per page</option>
+                        <option {{if eq $count 50}}selected{{end}} value="50">50 per page</option>
+                        {{if eq $count 10 20 30 40 50}}{{else}}
+                            <option selected value="{{$count}}">{{$count}} per page</option>
+                        {{end}}
```
buck54321: You've changed the behavior here. For instance, right now, on the months view, there is an option to show 50 per page, but Decred only has 45 months of history. What was the reasoning for removing the filters here?

views/extras.tmpl
```diff
@@ -82,7 +82,7 @@
 				<input type="checkbox" data-target="menu.toggle" data-action="change->menu#toggle"/>
 				<span class="patty"></span>
 				<span class="patty"></span>
-				<span class="patty short"></span>
+				<span class="patty"></span>
```
buck54321: Looks better with the short patty.

views/timelisting.tmpl
njirap: @buck54321 I literally copy pasted the designs

njirap: The dropdown has only these options

njirap: reverting back original behavior

views/timelisting.tmpl
njirap: bringing them back...must have been accidental

njirap:
Screen Shot 2019-10-21 at 3 02 01 PM

njirap:
All requested changes made

Screen Shot 2019-10-22 at 8 28 00 PM
Screen Shot 2019-10-22 at 8 28 19 PM

PR #1582, opened 2019-10-15
insight: set a per-connection room address subscription limit

chappjc:


gov/agendas/tracker.go
```diff
@@ -191,7 +191,7 @@ func NewVoteTracker(params *chaincfg.Params, node VoteDataSource, counter voteCo
 func (tracker *VoteTracker) Refresh() {
 	voteInfo, err := tracker.refreshRCI()
 	if err != nil {
-		log.Errorf("VoteTracker.Refresh -> refreshRCI: %v")
+		log.Errorf("VoteTracker.Refresh -> refreshRCI: %v", err)
```
chappjc: @buck54321 I noticed these missing args, so I plugged them in.

I'm seeing this in the dcrdata logs:

```
[ERR] AGDB: VoteTracker.Refresh -> refreshRCI: refreshRCI: Vote information not found: -32603: stake version 8 does not exist
```

api/insight/socket.io.go
```diff
@@ -127,33 +132,49 @@ func NewSocketServer(params *chaincfg.Params, txGetter txhelpers.RawTransactionG
...
 		if _, err := dcrutil.DecodeAddress(room, params); err == nil {
+			// Enforce the maximum address room subscription limit.
+			numAddrSubs, _ := so.Context().(uint32)
+			if numAddrSubs >= maxAddressSubsPerConn {
+				apiLog.Warnf("Client %s failed to subscribe, at the limit.", so.ID())
+				so.Emit("error", `"too many address subscriptions"`)
+				return "too many address subscriptions"
+			}
+			numAddrSubs++
+			so.SetContext(numAddrSubs)
```
chappjc: According to go-socket.io docs, there's no need for locking the context accesses:

https://github.com/googollee/go-socket.io/blob/3f1229f016cfb51926a2ea99838fa869bd6b6246/conn.go#L25-L28

gov/agendas/tracker.go
buck54321: Whoops. I missed this PR. I'll get you a review and look into the error.

chappjc: Thanks.  I was ready to merge this PR (unrelated to this error message though) when I hit a problem in testing on alpha.  For the life of me I cannot understand why it works with my local client, but not other socket.io clients from third parties hitting alpha.

api/insight/socket.io.go
```diff
@@ -100,54 +102,71 @@ func NewSocketServer(params *chaincfg.Params, txGetter txhelpers.RawTransactionG
...
 	// Subscription to a room checks the room name is as expected for an
 	// address, joins the room, and increments the room's subscriber count.
-	server.OnEvent("/", "subscribe", func(so socketio.Conn, room string) {
+	server.OnEvent("/", "subscribe", func(so socketio.Conn, room string) string {
+		return "i got nothin"
 		if len(room) > 64 || !isAlphaNumeric(room) {
-			return
+			return "bad address"
 		}
 		if _, err := dcrutil.DecodeAddress(room, params); err == nil {
```
chappjc: Ah, I was shadowing `err` here, but using it below in `return "bad address: " + err.Error()` where it was unexpectedly nil.

At least all this debugging revealed that other third-party clients are manually subscribing to "inv".  I suspect we should remove `so.Join("inv")` from the `OnConnect` handler.

public/js/test_insight_socket.io.html
```diff
@@ -21,7 +21,18 @@
 
         sio.on('connect', function () {
             update('Connected to dcrdata\'s Insight socket.io server. Updates will be shown below.')
-            sio.emit('subscribe', 'Dcur2mcGjmENx4DhNqDctW5wJCVyT3Qeqkx');
+            so.emit('subscribe', 'inv', (data) => {
+                update('subscribe "inv": ' + data);
+            })
```
chappjc: Test html updated to join "inv" explicitly.

gov/agendas/tracker.go
chappjc: Sorry to change up the PR so much, but I discovered a few issues.

api/insight/socket.io.go
```diff
@@ -100,54 +103,86 @@ func NewSocketServer(params *chaincfg.Params, txGetter txhelpers.RawTransactionG
...
+		txGetter:         txGetter,
 	}
 
+	// OnConnect sets the address room subscription counter to 0. There are no
+	// default subscriptions. The client must subscribe to "inv" if they want
+	// notification of all new transactions.
 	server.OnConnect("/", func(so socketio.Conn) error {
-		// New connections automatically join the inv and sync rooms.
-		so.Join("inv")
-		so.Join("sync")
```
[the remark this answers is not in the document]
chappjc: Pretty certain now that I've spent some time debugging.  Most clients were explicitly joining "inv" with a subscribe event.  I chatted with Exodous about it and they confirmed that it's not assumed to be automatically joined, plus they agreed it shouldn't be by default in order to reduce excessive chatter when it's not needed (the mobile client).
Referencing what little Insight documentation is available, I could find to indication that these rooms were supposed to be auto-subscribe.
This is definitely worth some documentation.

api/insight/socket.io.go
```diff
@@ -100,54 +103,86 @@ func NewSocketServer(params *chaincfg.Params, txGetter txhelpers.RawTransactionG
...
-		apiLog.Debugf("New socket.io connection. %d clients are connected.",
-			server.RoomLen("inv"))
+		so.SetContext(uint32(0))
```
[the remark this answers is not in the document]
chappjc: Yeah, I'll document that since it's not clear until you get to the subscribe event handler.  That's the idea though, it's a general purpose connection-scoped data element.

PR #1584, opened 2019-10-17
try to purge blockdata and gov of invalid storm

chappjc:
The horrible storm message is hard to eliminate:
> verifying github.com/asdine/storm@v2.1.3+incompatible/go.mod: github.com/asdine/storm@v2.1.3+incompatible/go.mod: reading https://sum.golang.org/lookup/github.com/asdine/storm@v2.1.3+incompatible: 410 Gone

Try as we might, we can't get rid of this on our own because `gov` imports `github.com/decred/politeia`, which imports `github.com/decred/dcrdata/pubsub/v2` (old), which imports `blockdata`, which imports `stakedb`, the old version of which imported the broken storm.  Grrrrr.

`stakedb/v3` itself is fixed, no longer importing or referencing the broken storm in it's go.sum.  This PR updates `blockdata` and `gov` so they use the updated `stakedb/v3`.  After this is merged, then `pubsub/v3` can update it's `stakedb/v3` import.  Then `github.com/decred/politeia` will need to be updated from `pubsub/v2` to `pubsub/v3`.  Then then then `gov` can be updated. :nauseated_face:


chappjc:
OK, going through with this and on to the next step of updating `blockdata` so `github.com/decred/politeia` can get off the bad version.

chappjc:
The next step of updating github.com/decred/politeia may proceed when https://github.com/decred/dcrdata/pull/1584 is merged.

PR #1585, opened 2019-10-22
update dompurify to 2.0 and disallow svg and math tags

chappjc:


chappjc:
Funny that doesn't cause a CI error.  Apparently `npm run build` doesn't error when the linter complains.

PR #1595, opened 2019-10-31
rpcutils,gov: require dcrd RPC v6.1.1

chappjc:


chappjc:
Tested on mainnet and testnet3 with current dcrd master (dcrd (JSON-RPC API v6.1.1)).  Required for https://github.com/decred/dcrdata/pull/1596, so merging without delay.

PR #1599, opened 2019-11-01
add cache busting url query to app js links

chappjc:


chappjc:
Already deployed with these query keys, so needs to be tweaked.

chappjc:
Not to be deployed until https://github.com/decred/dcrdata/pull/1606 is merged too.

PR #1600, opened 2019-11-01
set server header

chappjc:
This adds a middleware to set Server in the HTTP response header. This is helpful to deal with reverse proxies like nginx that just set `Server: nginx` (or worse, with the entire version string).


chappjc:
I'm still sitting on this PR because I think it needs to be a config option whether or not to use dcrdata's server string vs. something else.

chappjc:
https://github.com/decred/dcrdata/pull/1600/commits/c8cd92d43212cfb4a7c72b062b31534e28a2bcc5 adds `--server-http-header`

PR #1604, opened 2019-11-04
explorer: Fix a couple mempool bugs. Add js debug logging

buck54321:
I'm not sure that this solves all of the issues we've seen. I do know this solves at least some of the problems we were seeing though.

### DIFF
--- a/public/index.js
+++ b/public/index.js
@@ -37,8 +37,11 @@ async function createWebSocket (loc) {
   ws.connect(uri)
 
   var updateBlockData = function (event) {
-    console.log('Received newblock message', event)
     var newBlock = JSON.parse(event)
+    if (window.loggingDebug) {
+      console.log('Block received')
+      console.log(newBlock)
+    }
     newBlock.block.unixStamp = new Date(newBlock.block.time).getTime() / 1000
     globalEventBus.publish('BLOCK_RECEIVED', newBlock)
   }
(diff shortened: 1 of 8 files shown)


public/index.js
```diff
@@ -48,4 +51,13 @@ async function createWebSocket (loc) {
...
 
+// Debug logging can be enabled by entering logDebug(true) in the console.
+// Your setting will persist across sessions.
+window.loggingDebug = window.localStorage.getItem('loggingDebug') === '1'
+window.logDebug = yes => {
+  window.loggingDebug = yes
+  window.localStorage.setItem('loggingDebug', yes ? '1' : '0')
+  return 'debug logging set to ' + (yes ? 'true' : 'false')
+}
+
```
buck54321: I wanted to reduce the noise in the console so that I could see some temporary prints I was putting in. Turning off "debug logging" disables the websocket messages and pings and new blocks. This verbose output is off by default, though, which has potential consequences. For example, if you see something wrong, it's not always true anymore that you can open your console and see the block history. The flip side is that errors will be obvious, because they won't get buried over time.

public/js/controllers/homepage_controller.js
```diff
@@ -72,7 +72,7 @@ export default class extends Controller {
     })
     ws.registerEvtHandler('getmempooltxsResp', (evt) => {
       var m = JSON.parse(evt)
-      this.mempool.mergeMempool(m)
+      this.mempool.replace(m)
```
buck54321: Looking over it, I think I should have done this from the beginning. If we are getting the full copy of `MempoolInfo`, there is not reason to try to merge anything. Just replace it all.

public/js/helpers/mempool_helper.js
```diff
@@ -54,14 +54,14 @@ export default class Mempool {
...
-    var fauxVal = total / count
+    var fauxVal = count === 0 ? 0 : total / count
     for (var i = 0; i < count; i++) {
       this.mempool.push(makeTx('', txType, fauxVal, null, avgSize))
     }
   }
 
   initVotes (tallyTargets, total, count, avgSize) {
-    var fauxVal = total / count
+    var fauxVal = count === 0 ? 0 : total / count
```
buck54321: This was the cause of the bars going crazy sometimes.

public/js/controllers/homepage_controller.js
chappjc: Hmm, yeah it does seem like the response will be either all of mempool (rather than an increment) or no response at all if the request's mempool ID is current.

Issue #1477, opened 2019-07-22
hashrate units  on hashrate chart


raedah:
Current:
Network Hashrate (terahash/s): 654.453K
Expected:
Network Hashrate: 654 Ph/s
or
Network Hashrate: 654,453 Th/s

RyanBRiley:
I'll give this one a go.

PR #1606, opened 2019-11-05
explorer charts: refactor legend formatting

### LINKED ISSUE #1602: charts: The legend formatter is not hitting any of the label cases
https://github.com/decred/dcrdata/blob/203f793f102726993c3efcf314a9dd2446935a81/public/js/controllers/charts_controller.js#L112-L131

e.g. This prevents `formatHashRate` from setting the proper units on the legend.  `formatHashRate` should be applied to the y-axis labels as well, but that may be a bigger task.

buck54321:
Resolves #1602, #1477

### DIFF
--- a/public/js/controllers/charts_controller.js
+++ b/public/js/controllers/charts_controller.js
@@ -450,6 +450,7 @@ export default class extends Controller {
             color: '#888'
           }
         }
+        yFormatter = customYFormatter(y => `${intComma(y)} tickets &nbsp;&nbsp; (network target ${intComma(ticketPoolSizeTarget)})`)
         break
 
       case 'stake-participation':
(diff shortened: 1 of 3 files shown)


buck54321:
The axis units are fixed. Would you prefer petahash/s? 

The legend units are not fixed, and will pick the best units for whatever point is being displayed. Hovering around a point from 2016 or 2017 would should Th/s, for example. That's easy to do with a single point. If you were interested in y-axis units that can change when the user zooms, it's not impossible, but not a simple task to get right, I'd imagine.

chappjc:
I thought the y-axis units could be chosen similar to the legend, but since it's based on a range of data I see why that's more challenging.  I think issue #1477 was pointing out that "Network Hashrate (terahash/s): 654.453K" should be "Network Hashrate: 654 Ph/s" on the axis as well as the legend, but I appreciate that it's more complicated.  Probably not sinking too much more time into unless there's a straightforward solution that comes to mind.

buck54321:
Petahash/s is probably better, so I'll change it to that. It will only weird if you zoom in to data before 2018 or so, and even then it will still be readable.

PR #1610, opened 2019-11-11
CSPP mix detection

chappjc:
The detection is basic and fast (if ticket price is handy).


txhelpers/cspp.go
```diff
@@ -62,20 +62,49 @@ func IsMixTx(tx *wire.MsgTx) (isMix bool, mixDenom int64, mixCount uint32) {
...
+}
+
+func IsMixedSplitTx(tx *wire.MsgTx, relayFeeRate, ticketPrice dcrutil.Amount) (isMix bool, numTickets, numMixedIns uint32) {
+	ticketTxFee := FeeForSerializeSize(DefaultRelayFeePerKb, soloTicketTxSize())
+	ticketOutAmt := int64(ticketPrice + ticketTxFee)
+
+	var numOtherOut uint32
 	for _, o := range tx.TxOut {
-		if o.Value == ticketPrice {
+		if o.Value == ticketOutAmt {
```
chappjc: This is working, but feels vary fragile.  This needs to agree exactly with how CSPP operates.

chappjc: This is all yanked from dcrwallet code.  A way to keep this in agreement with dcrwallet and the CSPP server would be better.

txhelpers/cspp.go
```diff
@@ -0,0 +1,122 @@
...
+	return EstimateSerializeSizeFromScriptSizes(inSizes, outSizes, 0)
+}
+
+func IsMixedSplitTx(tx *wire.MsgTx, relayFeeRate, ticketPrice dcrutil.Amount) (isMix bool, numTickets, numMixedIns uint32) {
+	ticketTxFee := FeeForSerializeSize(DefaultRelayFeePerKb, soloTicketTxSize())
+	ticketOutAmt := int64(ticketPrice + ticketTxFee)
+
+	var numOtherOut uint32
+	for _, o := range tx.TxOut {
+		if o.Value == ticketOutAmt {
```
chappjc: This is all yanked from dcrwallet code.  A way to keep this in agreement with dcrwallet and the CSPP server would be better.

db/dcrpg/pgblockchain.go
```diff
@@ -5399,16 +5409,44 @@ func (pgb *ChainDB) GetExplorerTx(txid string) *exptypes.TxInfo {
...
+	var mixDenom int64
+	var mixCount uint32
+	if !txhelpers.IsStakeTx(msgTx) {
+		isMix, mixDenom, mixCount = txhelpers.IsMixTx(msgTx)
+		if !isMix {
+			var ticketPrice int64
+			if txraw.Confirmations > 0 {
+				ticketPrice = pgb.GetSBitsByHash(txraw.BlockHash)
+			} else {
+				sdiffRes, err := pgb.Client.GetStakeDifficulty()
```
jrick: Seems racy

db/dcrpg/pgblockchain.go
```diff
@@ -5399,16 +5409,44 @@ func (pgb *ChainDB) GetExplorerTx(txid string) *exptypes.TxInfo {
...
+				if err == nil {
+					sdiff, _ := dcrutil.NewAmount(sdiffRes.CurrentStakeDifficulty) // err means sdiff==0
+					ticketPrice = int64(sdiff)
+				}
+			}
+			log.Infof("ticket price at %s: %d", txraw.BlockHash, ticketPrice)
+			isMix, mixCount, _ = txhelpers.IsMixedSplitTx(msgTx, ticketPrice)
+			if isMix {
+				mixDenom = ticketPrice
```
jrick: The mixed output amounts when ticket buying is not just the current ticket price, but also includes extra DCR in order that the ticket purchase itself may consume just that output and still pay the required transaction fee.

jrick:
Another (better?) way to locally guess whether the transaction is a coinjoin or not would be to do partial sum analysis across the inputs and change outputs, to ensure that each change output corresponds to a unique set of the inputs.  Keep in mind that if the change would be dust, it is omitted, so it's possible that some of the inputs wouldn't correspond to any change at all.

db/dcrpg/pgblockchain.go
chappjc: Thanks for pointing that out.  It needs to be `ticketOutAmt` as computed in `IsMixedSplitTx`: https://github.com/decred/dcrdata/pull/1610/files/fab622a77e4da638b8d88d8923e8486f490ae9ba#diff-74e3b548913f559d292b3e67d1cc1d47R92

db/dcrpg/pgblockchain.go
chappjc: It is.  The tx may be confirmed and the sdiff changed by the time this is called.  I indent to address this.

db/dcrpg/pgblockchain.go
chappjc: BTW, all the stolen code regarding this computation does not sit well with me for a number of reasons.  I'd like a more reliable way to do that.

chappjc:
This now stores mix_count and mix_denom in the transactions table.

```
# select mix_denom, count(*) from transactions where mix_count > 0 group by mix_denom;
  mix_denom  | count 
-------------+-------
      262144 |  2606
     1048576 |  2590
     4194304 |  2633
    16777216 |  2666
    67108864 |  2632
   268435456 |  2558
  1073741824 |  2202
  4294967296 |  1687
 12080771038 |    24
 12101411796 |    21
 12117184413 |    17
 12177824330 |    21
 12192464817 |    19
 12228000099 |    20
 12277628776 |    24
 12367450943 |    12
 12371989479 |    21
... so on for other ticket prices ...
```

I still don't like that the (default) relay fee is used and hard-coded since the relay fee can change over time and there is no indication of that value is current.  However, that does not affect the standard-denomination mixes, just the mixed split txns.

Next steps are to present this data on the web page in the form of charts and/or tables.

chappjc:
Now shows total mixed on block page:

image

chappjc:
And mix data for regular transactions on block page:

image

chappjc:
In addition to the mix data on the transaction page:

image

txhelpers/cspp.go
```diff
@@ -0,0 +1,143 @@
...
+	1 << 34, // 171.79869184
+	1 << 32, // 042.94967296
+	1 << 30, // 010.73741824
+	1 << 28, // 002.68435456
+	1 << 26, // 000.67108864
+	1 << 24, // 000.16777216
+	1 << 22, // 000.04194304
+	1 << 20, // 000.01048576
+	1 << 18, // 000.00262144
+}
```
chappjc: This PR only looks for splits made by the current CSPP server.  Ideally the mix check would look for any denomination in sets of 3 or more plus change outputs that when combined add up to an input amount minus some fees, but that's not so straightforward.
Also, I think that might detect regular old pool ticket split transactions as mixes too since they are structured similarly.

txhelpers/cspp.go
```diff
@@ -0,0 +1,143 @@
...
+// transaction structure and worst-case signature script sizes.
+func calcSoloTicketTxSize() int {
+	inSizes := []int{RedeemP2PKHSigScriptSize}
+	outSizes := []int{P2PKHPkScriptSize + 1, TicketCommitmentScriptSize, P2PKHPkScriptSize + 1}
+	return EstimateSerializeSizeFromScriptSizes(inSizes, outSizes, 0)
+}
+
+var (
+	soloTicketTxSize    = calcSoloTicketTxSize()
+	defaultFeeForTicket = FeeForSerializeSize(dcrutil.Amount(DefaultRelayFeePerKb), soloTicketTxSize)
```
chappjc: This is another thing that only allows `IsMixedSplitTx` to detect mixing ticket split transactions created with the *current* CSPP server, which computes it's output amount for such transactions in this way.

txhelpers/cspp.go
```diff
@@ -0,0 +1,143 @@
...
+		splitPointMap[int64(amt)] = struct{}{}
+	}
+}
+
+// IsMixTx test if a transaction is a CSPP-mixed transaction, which must have 3
+// or more outputs of the same amount, which is one of the pre-defined mix
+// denominations. mixDenom is the largest of such denominations. mixCount is the
+// number of inputs of this denomination.
+func IsMixTx(tx *wire.MsgTx) (isMix bool, mixDenom int64, mixCount uint32) {
+	if len(tx.TxOut) < 3 {
```
chappjc: Should this also require at least this many inputs?

txhelpers/cspp.go
```diff
@@ -0,0 +1,143 @@
...
+		if count < 3 {
+			continue
+		}
+		if val > mixDenom {
+			mixDenom = val
+			mixCount = count
+		}
+	}
+
+	isMix = mixCount >= uint32(len(tx.TxOut)/2)
```
chappjc: Should this also require at least as many *inputs* as mixed outputs?

raedah:
Could also copy the mixed column over to the mempool template which is similar to the block template.

chappjc:
Will do the same on the mempool page in a followup PR.

PR #1625, opened 2019-11-25
when time remaining is 0, show "imminent"

chappjc:
Previously the home page would just say "remaining" when the ticket price window count reached 144 of 144.  This is the number of blocks mined in the current ticket price window.  This could be changed to stop at 143, proceeding to 0 next, but for now this fixes the time remaining when the entire ticket price window has been mined.

image

### DIFF
--- a/explorer/templates.go
+++ b/explorer/templates.go
@@ -400,6 +400,9 @@ func makeTemplateFuncMap(params *chaincfg.Params) template.FuncMap {
 		"threeSigFigs": threeSigFigs,
 		"remaining": func(idx int, max, t int64) string {
 			x := (max - int64(idx)) * t
+			if x == 0 {
+				return "imminent"
+			}
 			allsecs := int(time.Duration(x).Seconds())
 			str := ""
 			if allsecs > 604799 {


chappjc:
This could also say "done", "changed" or "now"...

chappjc:
Rethinking the current/next ticket price business:

When it says 144 of 144, that means then entire window is done.
Instead of that it could start at 0 of 144 (skipping 144 of 144)
Right now it goes ..., 143, 144, 1, ...
Could go ..., 143, 0, 1, ...
(Indicating blocks mined in the window)
"Current Ticket Price" still shows the ticket price of the best block rather than the next since tickets in mempool will be for the next block
So maybe we can change around the semantics of "current".
Current Ticket Price and Next Ticket Price jive with getstakedifficulty, and those refer to current best block and next block, but I might argue that Current Ticket Price should show the ticket price for tickets in mempool (what will be in the next block).

chappjc:
This is better than an incomplete "remaining" message, so merging, but we may wish to rethink this progress meter as described in https://github.com/decred/dcrdata/pull/1625#issuecomment-558242900

PR #1639, opened 2019-12-09
dcrpg: add vouts.mixed and vouts.spend_tx_row_id

chappjc:
This upgrades the database scheme from 5 to 6 (specifically, v1.5.1 -> v1.6.0).
- add `vouts.mixed` (BOOLEAN), true for outputs of mix transactions, and the stake transactions that spend them.
- add `vouts.spend_tx_row_id` (INT8) referencing the row ID (PK) in the transactions table, of the spending transaction, default is NULL

### DIFF
--- a/db/dcrpg/pgblockchain.go
+++ b/db/dcrpg/pgblockchain.go
@@ -11,6 +11,7 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"math"
 	"runtime"
 	"sort"
 	"strconv"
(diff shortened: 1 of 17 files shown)


db/dcrpg/pgblockchain.go
```diff
@@ -6097,3 +6098,73 @@ func (pgb *ChainDB) SignalHeight(height uint32) {
...
 	}
 }
+
+func (pgb *ChainDB) MixedUtxosByHeight() (heights, utxoCountReg, utxoValueReg, utxoCountStk, utxoValueStk []int64, err error) {
+	var rows *sql.Rows
+	rows, err = pgb.db.Query(`SELECT vouts.value, fund_tx.block_height, spend_tx.block_height, vouts.tx_tree
+		FROM vouts
+		JOIN transactions AS fund_tx ON vouts.tx_hash=fund_tx.tx_hash
+		LEFT OUTER JOIN transactions AS spend_tx ON spend_tx_row_id=spend_tx.id
+		WHERE mixed=true and value>0;`)
```
chappjc: This takes a couple seconds an a fast machine, maybe 10-15 on a slow one, so this is only meant for initializing a data set, which should then be kept up to date when new blocks are processed (add and remove outputs).

PR #1640, opened 2019-12-11
Coinjoin chart

ademuanthony:
This PR is a collaboration between me and @raedah 

We added a chart for coinjoin that shows the mixed amount by day block
image

We added a chart for the value of `Anonymity Set` to the `Coin Supply` chart
image


yarn.lock
```diff
@@ -0,0 +1,8525 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
chappjc: git rm pls

chappjc:
Shouldn't we replace "Coinjoin" with "CoinShuffle++" or something generic like "mixed coins" or "privacy" on these pages?  Coinjoin really under sells where Decred has implemented (cspp).

raedah:
CSPP is only part of the server client protocol and does not appear on chain. Coinjoin is the name of the transaction type that is being identified. Mixed coins or privacy are accurate but are more generic, less specific. Perhaps more generic is better since what is being identified is actually several specific transaction cases such as the connected sstx transactions.

chappjc:
I see.  From a marketing perspective, how about "anonymity set"?

chappjc:
Actually, Privacy Participation would jive well with the other charts (e.g. Stake Participation).  How about that?

db/cache/charts.go
```diff
@@ -1001,6 +1064,20 @@ func accumulate(data ChartUints) ChartUints {
...
+	for i, v := range data {
+		accumulator += v
+		if i >= x {
+			accumulator -= data[i - x]
+		}
+		d = append(d, accumulator)
+	}
+	return d
+}
+
```
buck54321: This is not used anywhere.

db/cache/charts.go
```diff
@@ -1155,26 +1232,28 @@ func coinSupplyChart(charts *ChartData, bin binLevel, axis axisType) ([]byte, er
 		switch axis {
 		case HeightAxis:
 			return encode(lengtherMap{
-				supplyKey: accumulate(charts.Blocks.NewAtoms),
+				supplyKey: 				accumulate(charts.Blocks.NewAtoms),
```
buck54321: Formatting here and a few other places is weird too.

db/dcrpg/pgblockchain.go
```diff
@@ -3131,6 +3137,19 @@ func (pgb *ChainDB) blockFees(charts *cache.ChartData) (*sql.Rows, func(), error
 	return rows, cancel, nil
 }
 
+// blockFees sets or updates a series of per-block fees.
+// This is the Fetcher half of a pair that make up a cache.ChartUpdater. The
+// Appender half is appendBlockFees.
```
buck54321: ~~`appendBlockFees`~~ -> `appendBlockCoinJoins`

public/js/controllers/charts_controller.js
```diff
@@ -154,7 +165,7 @@ function zipTvY (times, ys, yMult) {
 
 function zipIvY (ys, yMult, offset) {
   yMult = yMult || 1
-  offset = offset || 1
+  offset = offset || 1 // TODO: check for why offset is set to a default value of 1 when genesis block has a height of 0
```
buck54321: #1643 
@chappjc

db/cache/charts.go
```diff
@@ -507,6 +518,16 @@ func (charts *ChartData) Lengthen() error {
...
+	movingInterval := [][2]int{} // stores the start and end index for the vote reward period(~29.07 days on mainnet) moving sum
+	miStart := start
+	// The actual reward of a ticket needs to also take into consideration the
+	// ticket maturity (time from ticket purchase until its eligible to vote)
+	// and coinbase maturity (time after vote until funds distributed to ticket
+	// holder are available to use).
+	meanVotingBlocks := txhelpers.CalcMeanVotingBlocks(charts.chainParams)
+	avgSSTxToSSGenMaturity := meanVotingBlocks + int64(charts.chainParams.TicketMaturity) + 
+		int64(charts.chainParams.CoinbaseMaturity)
+	rewardPeriod := avgSSTxToSSGenMaturity * int64(charts.chainParams.TargetTimePerBlock.Seconds())
```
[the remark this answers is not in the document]
ademuanthony: This was a way to get approximate value of how long a mixed tx takes before it's spent. It is now removed as a better solution is now provided and there is no need for approximation again.

db/dcrpg/upgrades.go
```diff
@@ -361,7 +361,27 @@ func (u *Upgrader) compatVersion1Upgrades(current, target DatabaseVersion) (bool
...
+			current.maint++
+			if err = updateMaintVersion(u.db, current.maint); err != nil {
+				return false, fmt.Errorf("failed to update maintenance version: %v", err)
+			}
+			fallthrough
+		case 1:
+			// all ready
+		default:
+			return false, fmt.Errorf("unsupported maint version %d", current.maint)
+		}
```
chappjc: Why is this needed again?

ademuanthony:
> Oh the 1.7.1 maint is because of this?
> 
> But I think this change will create far too many false positives. It also allows outputs with different values in the same tx to all be recognized as ticket funding outputs, but they would have to be the same value to be a mix.

I have rolled back these changes for now.
It is now noted and we can look into it later and provide a better solution

chappjc:
Please have a look at the test failure:

```
 --- FAIL: TestChartsCache (0.00s)
##[error]    charts_test.go:53: DeepEqual: expected true, found false for Height after read
    --- FAIL: TestChartsCache/Read_from_an_existing_gob_encoded_file (0.00s)
##[error]        testing.go:864: test executed panic(nil) or runtime.Goexit: subtest may have called FailNow on a parent test
```

db/dcrpg/pgblockchain.go
```diff
@@ -3160,6 +3172,32 @@ func (pgb *ChainDB) blockFees(charts *cache.ChartData) (*sql.Rows, func(), error
...
+	ctx, cancel := context.WithTimeout(pgb.ctx, pgb.queryTimeout)
+
+	rows, err := retrievePrivacyParticipation(ctx, pgb.db, charts)
+	if err != nil {
+		return nil, cancel, fmt.Errorf("chartPrivacyParticipation: %v", pgb.replaceCancelError(err))
+	}
+	return rows, cancel, nil
+}
+
+// anonymitySet sets or updates a series of per-block anonymity set.
```
chappjc: Also wrap this comment please.  I suggest a comment formatter, like the [Rewrap extension](https://marketplace.visualstudio.com/items?itemName=stkb.rewrap) if you happen to be using VSCode.

db/dcrpg/queries.go
```diff
@@ -3619,6 +3620,182 @@ func appendBlockFees(charts *cache.ChartData, rows *sql.Rows) error {
...
+	for rows.Next() {
+		var blockHeight uint64
+		var totalMixed int64
+		if err := rows.Scan(&blockHeight, &totalMixed); err != nil {
+			log.Errorf("Unable to scan for MixedVoutsPerBlock fields: %v", err)
+			return err
+		}
+		if totalMixed < 0 {
+			totalMixed *= -1
+		}
```
chappjc: In what circumstances does this happen?  Surely that indicates a bug in something else.

db/dcrpg/queries.go
```diff
@@ -3619,6 +3620,182 @@ func appendBlockFees(charts *cache.ChartData, rows *sql.Rows) error {
...
+
+// Append the result from retrieveAnonymitySet to the provided ChartData. This
+// is the Appender half of a pair that make up a cache.ChartUpdater.
+func appendAnonymitySet(charts *cache.ChartData, rows *sql.Rows) (err error) {
+	var vals, fundHeights, spendHeights []int64
+
+	var maxHeight int64
+	minHeight := int64(math.MaxInt64)
+
+
```
chappjc: extra newline

db/dcrpg/queries.go
```diff
@@ -3619,6 +3620,182 @@ func appendBlockFees(charts *cache.ChartData, rows *sql.Rows) error {
...
+		}
+
+		vals = append(vals, value)
+		fundHeights = append(fundHeights, fundHeight)
+		spendHeights = append(spendHeights, spendHeight)
+	}
+
+	err = rows.Err()
+	if err != nil {
+		log.Errorf("Unable to scan for BlockCoinJoins fields: %v", err)
```
chappjc: What is `"BlockCoinJoins"`"

db/dcrpg/queries.go
```diff
@@ -3619,6 +3620,182 @@ func appendBlockFees(charts *cache.ChartData, rows *sql.Rows) error {
...
+	}
```
chappjc: Since there is no return on error, justify in a comment the reason for appending anyway.

db/dcrpg/queries.go
```diff
@@ -3619,6 +3620,182 @@ func appendBlockFees(charts *cache.ChartData, rows *sql.Rows) error {
...
+
+	blocks := charts.Blocks
+
+	// block anonymity set
+	for _, h := range blocks.Height[len(blocks.AnonymitySet) :] {
```
chappjc: Is there a guarantee that `blocks.Height` will be at least this long?  This will inevitably panic and crash dcrdata when something goes wrong.

Why not infer the block height like the other appenders, and loop like:

```go
for h := len(blocks.AnonymitySet)+1; h <= len(blocks.Height); h++ {
	...
}
```

And/or first check the length of blocks.Height to catch malfunctions.

db/dcrpg/queries.go
```diff
@@ -3619,6 +3620,182 @@ func appendBlockFees(charts *cache.ChartData, rows *sql.Rows) error {
...
+
+	// Find the index that begins new data.
+	offset := 0
+	for i, t := range blocks.Time {
+		if t > start {
+			offset = i
+			break
+		}
+	}
+
```
chappjc: Why the rest of this function being done here instead of in `(*ChartData).Lengthen`, from which all this days code is taken?  Please do it in `(*ChartData).Lengthen` using  the appropriate ChartUints methods.  If that is not possible, we should create appropriate data types to allow all the day binning to be done in one spot, Lengthen.

chappjc:
Thanks for refactoring to use Length.  Just some trivial test failures related to code formatting (https://github.com/decred/dcrdata/runs/423024195#step:9:102).  I'll do a final review and hopefully we can get this up on alpha today or tomorrow.

chappjc:
For follow-up work, we should investigate triggering a fresh charts data update as part of the upgrade process rather than the usual charts data update on startup.  The reason is that on startup I see"

```
[INF] DATD: Mainchain sync complete.
[INF] DATD: All ready, at height 420483.
[WRN] PSQL: charts.ValidateLengths: dataset at index 9 has mismatched length 0 != 420484
[WRN] PSQL: charts.ValidateLengths: dataset at index 10 has mismatched length 0 != 420484
[WRN] PSQL: ChartData.Lengthen: block data length mismatch detected. Truncating blocks length to 0
[INF] PSQL: Updating charts data...
```
It doesn't say what it is doing, but a look at `top` indicates it is `dcrdata` **not** `postgres`, which suggests it is one of the functions that operates on the retrieved data, probably `appendAnonymitySet`.

chappjc:
After about 5 minutes:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0xc727a8]

goroutine 1 [running]:
github.com/decred/dcrdata/db/dcrpg/v5.appendAnonymitySet(0xc0075f41e0, 0xc00c11e100, 0xc030b6a2b0, 0x0)
	/home/jon/github/decred/dcrdata/db/dcrpg/queries.go:3726 +0x618
github.com/decred/dcrdata/db/cache/v3.(*ChartData).Update(0xc0075f41e0, 0x0, 0x0)
	/home/jon/github/decred/dcrdata/db/cache/charts.go:873 +0x3c1
github.com/decred/dcrdata/db/cache/v3.(*ChartData).Load(0xc0075f41e0, 0xc01f3a63c0, 0x2f, 0x0, 0x0)
	/home/jon/github/decred/dcrdata/db/cache/charts.go:728 +0x1a0
main._main(0x133cc60, 0xc0002a6180, 0x0, 0x0)
	/home/jon/github/decred/dcrdata/main.go:992 +0x3797
main.main()
	/home/jon/github/decred/dcrdata/main.go:58 +0x81
```

Looks like it got through privacy participation and crashed on anonymity set.

chappjc:
Note that to reproduce that panic, you will have to run with master to get a chartscache.gob the has all the charts data except anon set and priv partic.

db/cache/charts.go
```diff
@@ -538,13 +558,27 @@ func (charts *ChartData) Lengthen() error {
...
+			var anonymitySet int64
+			for iu := range charts.MixedVouts.Values {
+				maxHeight := blocks.Height[interval[1]]
+
+				if uint64(charts.MixedVouts.FundingHeight[iu]) < maxHeight &&
+					(uint64(charts.MixedVouts.SpendingHeights[iu]) > maxHeight || charts.MixedVouts.SpendingHeights[iu] == -1) {
+
+					anonymitySet += charts.MixedVouts.Values[iu]
+				}
+			}
```
chappjc: I don't understand what this bit is supposed to be doing.  Unlike the other day data sets, which are a sum or average over an interval, this just seems to be getting the anon set size at the end of the interval.  But this data is already computed and stored in the `blocks.AnonymitySet` slice.

If you really wanted to compute the average in this interval, it should be done similar to the other data sets.

db/dcrpg/queries.go
```diff
@@ -3619,6 +3620,116 @@ func appendBlockFees(charts *cache.ChartData, rows *sql.Rows) error {
...
+	// block anonymity set
+	for h := len(blocks.AnonymitySet) + 1; h <= len(blocks.Height); h++ {
+		var anonymitySet int64
+		for iu := range vals {
+			if int64(h) >= fundHeights[iu] && (int64(h) < spendHeights[iu] || spendHeights[iu] == -1) {
+				anonymitySet += vals[iu]
+			}
+		}
+		blocks.AnonymitySet = append(blocks.AnonymitySet, uint64(anonymitySet))
+	}
```
chappjc: Unfortunately, I don't think this approach to computing the anonymity set *at each height* is scalable.  Even at the current height and mixing status, it takes far too long.  Rather than compute anon set value fresh at each height, please try to do this by adding and removing value from a running anonymity set value depending on what heights are new for the current height.  Feel free to use different data structures as you Scan data from the db (e.g. map by fund height and map by spend height?  but please give it additional though before trying that).

db/cache/charts.go
ademuanthony: I am now taking daily average. Thanks for pointing to the fact that the data already stored in `blocks.AnonymitySet`

db/dcrpg/queries.go
ademuanthony: What I have done here is to use a running total and maintain a map of spending heights for all the values. At every new height, deduct all spent vouts from the running total and append the total as the anonymity set for that height

chappjc: I just saw https://github.com/decred/dcrdata/pull/1640/commits/e3e2d6e3a72e2ec3eca0462409da97ca9e2e9526 addressing this.  Looks to be what I had in mind.  Thanks.

chappjc:
Running well now.  Fresh updating charts data is down to 20 seconds from >5 minutes.  This is still several times longer than before, so we should look to optimizing this in future work.

Last issue, code formatting again: https://github.com/decred/dcrdata/runs/425480283#step:9:97

db/dcrpg/queries.go
chappjc: Please comment on this case.

public/js/controllers/charts_controller.js
```diff
@@ -336,13 +384,19 @@ export default class extends Controller {
...
+    this.settings = TurboQuery.nullTemplate(['chart', 'zoom', 'scale', 'bin', 'axis', 'visibility'])
     this.query.update(this.settings)
     this.settings.chart = this.settings.chart || 'ticket-price'
     this.zoomCallback = this._zoomCallback.bind(this)
     this.drawCallback = this._drawCallback.bind(this)
     this.limits = null
     this.lastZoom = null
+    this.visibility = []
+    if (this.settings.visibility) {
+      this.settings.visibility.split(',', -1).forEach(s => {
```
chappjc: Actually, I do not care for how the URL ends up with `%2C` in it, but I don't have a better suggestion for this.  Since the charts and controls are working, I don't want to delay this PR any longer.  For a future issues.  @buck54321 might also have a suggestion for this.

db/dcrpg/queries.go
ademuanthony: I no longer have this issue, this was based on different implementation where my effort to remove `spent vouts` was leading the negative value. I ought to have removed it when the implementation was changed.

public/js/controllers/charts_controller.js
ademuanthony: The visibility here is an array.
Some chart has multiple lines and checkboxes for toggling the visibility of any of the lines. The encoded value of the array is what lead to the `%2C` from `visibility = [true, false]`.
During the iterations of the PR, I had a recommendation to preserve the visibility across page reload

chappjc: I understand the reason,  I'm just pointing out that in the URL, the query params are ugly as a result:

image

We can resolve this in the future though, but it will have to be very near future and definitely before release otherwise we will get broken urls when we change the scheme.

chappjc:
Letting it run for a while, I get warnings about dataset at index 10 when the charts do their occasional update:

```
[WRN] PSQL: charts.ValidateLengths: dataset at index 10 has mismatched length 420836 != 420837
[WRN] PSQL: ChartData.Lengthen: block data length mismatch detected. Truncating blocks length to 420836
```

Restarting dcrdata seems to repair things, but it looks like there will be a problem while running for long periods.

db/dcrpg/internal/txstmts.go
```diff
@@ -273,6 +273,21 @@ var (
...
+			AND tx.block_height > $1
+		GROUP BY tx.block_height
+		ORDER BY tx.block_height;`
+
+	SelectMixedVouts = `
+		SELECT vouts.value, fund_tx.block_height, spend_tx.block_height, vouts.tx_tree
+		FROM vouts
+		JOIN transactions AS fund_tx ON vouts.tx_hash=fund_tx.tx_hash
+		LEFT OUTER JOIN transactions AS spend_tx ON spend_tx_row_id=spend_tx.id
+		WHERE mixed=true and value>0 ORDER BY fund_tx.block_height;`
```
chappjc: I thought you had a version of this query with a block height constraint like with `SelectMixedTotalPerBlock`.  Related to your comment at https://github.com/decred/dcrdata/pull/1640/files#diff-87f7a577b6af45d0284e4c3fbb98ddf5R3685

Will you follow up on this in a separate PR?  I want to merge this, but this change must be done as this is an expensive query that the server can't afford to have running often.

ademuanthony: I will follow up in a separate PR

chappjc: Excellent, thanks.  And thanks for your persistence and patience in this PR.

chappjc:
> Letting it run for a while, I get warnings about dataset at index 10 when the charts do their occasional update:
> 
> ```
> [WRN] PSQL: charts.ValidateLengths: dataset at index 10 has mismatched length 420836 != 420837
> [WRN] PSQL: ChartData.Lengthen: block data length mismatch detected. Truncating blocks length to 420836
> ```
> 
> Restarting dcrdata seems to repair things, but it looks like there will be a problem while running for long periods.

This does seem to be a consistent problem.  Off by one in this dataset?

[WRN] PSQL: charts.ValidateLengths: dataset at index 10 has mismatched length 421007 != 421008
[WRN] PSQL: ChartData.Lengthen: block data length mismatch detected. Truncating blocks length to 421007

[WRN] PSQL: charts.ValidateLengths: dataset at index 10 has mismatched length 421059 != 421060
[WRN] PSQL: ChartData.Lengthen: block data length mismatch detected. Truncating blocks length to 421059

PR #1645, opened 2019-12-19
pubsub,psclient: fix server->client pings and require them in client

chappjc:
The following changes are made, which effectively require the client to receive pings from the server at the interval the server sends them otherwise the client will shutdown assuming the connection has died.  The clients no longer have to `Ping` at the application level to detect when the connection is lost.  And the server's outgoing pings are used to clean up dead clients.

* Fix the server->client pings that were supposed to be sent automatically to all clients since the ping subscription was removed.
* Export `pubsub.PingInterval`, the server's ping interval, and set it to 30 seconds.
* In the client (`psclient`), set the default read time out to `pubsub.PingInterval * 10 / 9`, although this is still configurable by via `psclient.Opts`.
* **A read timeout in the client's receive loop will now SHUTDOWN THE CLIENT.**

### DIFF
--- a/pubsub/psclient/client.go
+++ b/pubsub/psclient/client.go
@@ -92,7 +92,7 @@ func newPingMsg(reqID int64) []byte {
 	return pingMsg
 }
 
-var defaultTimeout = 10 * time.Second
+const defaultTimeout = pubsub.PingInterval * 10 / 9
 
 // Opts defines the psclient Client options.
 type Opts struct {
@@ -241,9 +241,8 @@ func (c *Client) receiver() {
 
 		resp, err := c.receiveMsg()
 		if err != nil {
-			if pstypes.IsIOTimeoutErr(err) {
-				continue
-			}
+			// Even a timeout should close shutdown the client since that
+			// indicates pings from the server did not arrive in time.
 			log.Errorf("ReceiveMsg failed: %v", err)
 			return
 		}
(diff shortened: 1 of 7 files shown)


pubsub/websocket.go
```diff
@@ -17,7 +17,9 @@ import (
 type hubSpoke chan pstypes.HubMessage
 
 const (
-	pingInterval = 45 * time.Second
+	// PingInterval is how frequently the server will ping all clients. The
+	// clients should set their read deadlines to more than this.
+	PingInterval = 30 * time.Second
```
chappjc: Exported so that the client can set a sensible default read timeout.

pubsub/psclient/client.go
```diff
@@ -92,7 +92,7 @@ func newPingMsg(reqID int64) []byte {
 	return pingMsg
 }
 
-var defaultTimeout = 10 * time.Second
+const defaultTimeout = pubsub.PingInterval * 10 / 9
```
chappjc: Could be additive (e.g. `pubsub.PingInterval + 2 * time.Second`), but the timeout used by the client instance is still configurable via `psclient.Opts`.

chappjc:
@alexlyp This fixes the outgoing server -> client ping signals, but the two other commits change how read timeouts are handles in the client.  Without these changes, it was up to the client application to call `psclient.(*Client).Ping` periodically to verify server connectivity.  I think it should be the way in this PR.  Thoughts about how it should work?

Issue #1624, opened 2019-11-25
Add link to the Tor address


jzbz:
Since dcrdata also runs as a Tor hidden service (http://dcrdata2opeenddl.onion/) there should be a link somewhere (footer possibly) to advertize this fact.

chappjc:
Resolved by https://github.com/decred/dcrdata/pull/1647

PR #1647, opened 2019-12-19
Onion v2 address configuration option and footer link

peterzen:
Display configurable Onion link in the footer.
Solves #1624

### DIFF
--- a/sample-dcrdata.conf
+++ b/sample-dcrdata.conf
@@ -94,3 +94,6 @@
 ; Maximum number of comma-separated addresses allowed in certain Insight API
 ; endpoints, such as /insight/api/addrs/{addr0,..,addrN}
 ;max-api-addrs=3
+
+; TOR hidden service address.  When specified, it will be displayed in the footer.
+;onion-address=
--- a/views/extras.tmpl
+++ b/views/extras.tmpl
@@ -136,6 +136,15 @@
 				target="_blank"
 				rel="noopener noreferrer"
 			>dcrdata v{{.Version}}</a>
+			{{- if ne .Links.OnionURL "" }}
+			<a
+				class="text-nowrap d-none d-md-inline-block pr-3"
+				href="{{.Links.OnionURL}}"
+				title="Access dcrdata over Tor"
+				target="_blank"
+				rel="noopener noreferrer"
+			>Tor site</a>
+			{{ end }}
 			<a
 				class="text-nowrap"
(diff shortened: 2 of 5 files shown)


chappjc:
Thanks for this.  Sorry for the service downtime presently.  Yesterday was chaotic for me, but I'm doing maintenance now.  Will be back shortly.

chappjc:
Note that there is also now a v3 onion service at dcrdata5oppwcotlxkrlsp6afncnxvw54sw6jqftc4bjytm4rn27j3ad.onion

peterzen:
Cool, I wasn't aware of that - I'll add a v3 option.

views/extras.tmpl
```diff
@@ -136,6 +136,15 @@
...
 				rel="noopener noreferrer"
 			>dcrdata v{{.Version}}</a>
+			{{- if ne .Links.OnionV2 "" }}
+			<a
+				class="text-nowrap d-none d-md-inline-block pr-3"
+				href="{{.Links.OnionV2}}"
+				title="Hidden service URL"
+				target="_blank"
+				rel="noopener noreferrer"
+			>Onion v2</a>
```
chappjc: As a Tor user, do you think there is much value in the "v2" part of the link text?  If I were naming it I might go with "Tor host" or something like that.

chappjc: Also, can we just abandon the v2 hostname in favor of the v3 hostname?  I expect most tor versions support it these days, but I don't know for sure.

peterzen:
I don't know either - other block explorers do show both v2 and v3 links.
Will look into it.

views/extras.tmpl
peterzen: Done - there's indeed no point in keeping the v2 hostname.

sample-dcrdata.conf
```diff
@@ -94,3 +94,6 @@
 ; Maximum number of comma-separated addresses allowed in certain Insight API
 ; endpoints, such as /insight/api/addrs/{addr0,..,addrN}
 ;max-api-addrs=3
+
+; TOR hidden service address.  When specified, it will be displayed in the footer.
+;onionaddress=
```
chappjc: This needs to be "onion-address" instead of "onionaddress" to match the config struct tags.  Also, please add a trailing newline at the end of the file.

peterzen: Thanks for spotting, fixed.

Issue #1607, opened 2019-11-07
Blocks: move content higher


xaur:
The "Explore Blockchain" on the new blocks page is not too helpful but consumes precious vertical space. One idea is to just remove it and also make the "Blocks" heading space smaller, in order to have less scrolling to content.

![blocks-1](https://user-images.githubusercontent.com/24484727/68427040-478ce580-01a1-11ea-90c5-f471f89a166e.png)
![blocks-2](https://user-images.githubusercontent.com/24484727/68427041-478ce580-01a1-11ea-8828-c759ca6a34e2.png)

chappjc:
Agreed.  You're not the only one who thinks we have too much dead space.

chappjc:
Because I strongly agree with this, I've manually deployed the blocks page without that text and all the white space it eats.

![image](https://user-images.githubusercontent.com/9373513/68554078-dc9c1280-041d-11ea-92ba-3add65256c98.png)

Is there a middle ground, or should we just shave it down all the way as shown above?

xaur:
That's already a huge improvement that gets us ~70 px back (yay!).

Agreed padding can be further optimized.

And perhaps even table row height. At least for desktop users who have a good old mouse and not fat fingers :D

P.S. Just realized that if we ever need to add some title above the tabs, we can place it to the left of them (kind of unclickable label-type tab).

buck54321:
Much better that way. Nice work, @chappjc. Let's consider your screenshot our working design.

chappjc:
Resolved by https://github.com/decred/dcrdata/pull/1652

PR #1652, opened 2019-12-20
views: remove the Explore Block chain div that wastes space

chappjc:
This was live-demoed for a while, but not committed.  This gets rid of the "Explore Blockchain" div that wastes so much space.

Resolves issue https://github.com/decred/dcrdata/issues/1607


chappjc:
This was all I had done previously.  Any tweaks worth adding here, @buck54321?

PR #1655, opened 2019-12-25
Display Fee rate column in mempool tx/tickets tables

### LINKED ISSUE #1251: mempool: show fee rate
Add a column to show fee rate for tickets and regular tx.

Sometimes when ticket purchasing is very active it helps to see ticket transactions and their fee rate (I assume fee rate and not just fee is what influences the priority of inclusion in blocks).

This can be a useful indicator for regular tx too.

peterzen:
Fixes #1251

### DIFF
--- a/mempool/collector.go
+++ b/mempool/collector.go
@@ -127,9 +127,12 @@ func (t *MempoolDataCollector) mempoolTxns() ([]exptypes.MempoolTx, txhelpers.Me
 		//fee, _ := txhelpers.TxFeeRate(msgTx)
 		// log.Tracef("tx fee: GRM result = %f, msgTx = %f", tx.Fee, fee.ToCoin())
 
+		_, feeRate := txhelpers.TxFeeRate(msgTx)
+
 		txs = append(txs, exptypes.MempoolTx{
 			TxID:      msgTx.TxHash().String(),
 			Fees:      tx.Fee,
+			FeeRate:   feeRate.ToCoin(),
 			VinCount:  len(msgTx.TxIn),
 			VoutCount: len(msgTx.TxOut),
 			Vin:       exptypes.MsgTxMempoolInputs(msgTx),
--- a/views/mempool.tmpl
+++ b/views/mempool.tmpl
@@ -168,6 +168,7 @@
                               <th>Transaction ID</th>
                               <th class="text-right">Total DCR</th>
                               <th class="text-right">Size</th>
+                              <th class="text-right">Fee Rate</th>
                               <th class="text-right">Time in Mempool</th>
                             </tr>
                           </thead>
@@ -183,12 +184,13 @@
                                     {{template "decimalParts" (float64AsDecimalParts .TotalOut 8 false)}}
                                   </td>
                                   <td class="mono fs15 text-right">{{.Size}} B</td>
+                                  <td class="mono fs15 text-right">{{.FeeRate}} DCR/kB</td>
                                   <td class="mono fs15 text-right" data-target="time.age" data-age="{{.Time}}"></td>
                               </tr>
                               {{end}}
                               {{else}}
                               <tr class="no-tx-tr">
-                                  <td colspan="4">No tickets in mempool.</td>
+                                  <td colspan="5">No tickets in mempool.</td>
                               </tr>
                               {{end}}
                           </tbody>
@@ -242,6 +244,7 @@
                               <th>Transaction ID</th>
                               <th class="text-right">Total DCR</th>
                               <th class="text-right">Size</th>
+                              <th class="text-right">Fee Rate</th>
                               <th class="text-right">Time in Mempool</th>
                             </tr>
                           </thead>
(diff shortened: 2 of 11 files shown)


views/mempool.tmpl
```diff
@@ -183,12 +184,13 @@
                                     {{template "decimalParts" (float64AsDecimalParts .TotalOut 8 false)}}
                                   </td>
                                   <td class="mono fs15 text-right">{{.Size}} B</td>
+                                  <td class="mono fs15 text-right">{{.FeeRate}}/kB</td>
```
chappjc: This gives `[ERR] EXPR: Template execute failure: template: mempool.tmpl:187:69: executing "mempool" at <.FeeRate>: can't evaluate field FeeRate in type types.MempoolTx`

peterzen: Fixed in the PR.

peterzen:
As the `MempoolTx` struct was changed in this PR the linter fails ATM - is there a `git` tag needed on `github.com/decred/dcrdata/explorer/types/v2` for the test suite to properly pick it up?

chappjc:
Unfortunately due to the way we have CI running tests (first `cd`ing` into the module paths), each go.mod that requires explorer/types/v2 will need a replace.  Looks like that includes `mempool`, `pubsub`, `pubsub/types`, and `pubsub/democlient`.

mempool/collector.go
```diff
@@ -127,9 +127,12 @@ func (t *MempoolDataCollector) mempoolTxns() ([]exptypes.MempoolTx, txhelpers.Me
 		//fee, _ := txhelpers.TxFeeRate(msgTx)
 		// log.Tracef("tx fee: GRM result = %f, msgTx = %f", tx.Fee, fee.ToCoin())
 
+		_, feeRate := txhelpers.TxFeeRate(msgTx)
+
 		txs = append(txs, exptypes.MempoolTx{
 			TxID:      msgTx.TxHash().String(),
 			Fees:      tx.Fee,
+			FeeRate:   feeRate.ToCoin(),
```
chappjc: Also need something similar in mempool/monitor, (`(*MempoolMonitor).TxHandler`).

chappjc: Looks like `(pgb *ChainDB).GetMempool` needs it too, but there seems to be a number of fields unset in that function too.  I don't recall a reason or see docs explaining why, so perhaps that's a bug.

mempool/collector.go
```diff
@@ -127,9 +127,12 @@ func (t *MempoolDataCollector) mempoolTxns() ([]exptypes.MempoolTx, txhelpers.Me
 		//fee, _ := txhelpers.TxFeeRate(msgTx)
 		// log.Tracef("tx fee: GRM result = %f, msgTx = %f", tx.Fee, fee.ToCoin())
 
+		_, feeRate := txhelpers.TxFeeRate(msgTx)
```
chappjc: Regarding the comment a few lines up, we may need special  handling for coinbase transactions.  I haven't checked.

views/mempool.tmpl
```diff
@@ -168,6 +168,7 @@
                               <th>Transaction ID</th>
                               <th class="text-right">Total DCR</th>
                               <th class="text-right">Size</th>
+                              <th class="text-right">Fee rate</th>
```
chappjc: For consistency with the other columns, best to use title case (i.e. Fee Rate).

mempool/collector.go
peterzen: Can there be a case where a coinbase TX gets added to the mempool?

chappjc:
BTW, dcrdata 6.0 is going to eliminate half of the modules.  It's kinda unmanageable.

db/dcrpg/go.mod
```diff
@@ -25,7 +25,7 @@ require (
 	github.com/decred/dcrdata/blockdata/v5 v5.0.1
 	github.com/decred/dcrdata/db/cache/v3 v3.0.1
 	github.com/decred/dcrdata/db/dbtypes/v2 v2.2.1
-	github.com/decred/dcrdata/explorer/types/v2 v2.1.1
+	github.com/decred/dcrdata/explorer/types/v2 v2.1.2
```
chappjc: There is no v2.1.2 yet, just 2.1.1.  The replace will use the current version though.
git tag -l | grep explorer/types

peterzen: Done, with the replaces in place it should be all good.

mempool/collector.go
chappjc: Nah, but I think stakebase/votes might have the same issue as they're generating coins too.

peterzen: FeeRate is displayed for SSTX and regular TXs but not for votes; weird `feeRate` values coming from coinbase/stakebase txs won't be shown.  If you think it's warranted I can check `feeRate` coming back from `txhelpers.TxFeeRate` and maybe blank it out - what do you think?

chappjc: Yeah, true.  It's good for this.

PR #1661, opened 2019-12-29
gov: fix simnet agenda db and vote tracking

buck54321:
Disables `VoteTracker` and `AgendaDB` when simnet is in use. `VoteTracker` is not initialized, and `AgendaDB` has reasonable return values from exported functions.

### DIFF
--- a/gov/agendas/deployments.go
+++ b/gov/agendas/deployments.go
@@ -21,6 +21,7 @@ import (
 // AgendaDB represents the data for the stored DB.
 type AgendaDB struct {
 	sdb           *storm.DB
+	simnet        bool
 	stakeVersions []uint32
 	deploySource  DeploymentSource
 }
@@ -42,10 +43,6 @@ type AgendaTagged struct {
 }
 
 var (
-	// errDefault defines an error message returned if the agenda db wasn't
-	// properly initialized.
-	errDefault = fmt.Errorf("AgendaDB was not initialized correctly")
-
 	// dbVersion is the current required version of the agendas.db.
 	dbVersion = semver.NewSemver(1, 0, 0)
 )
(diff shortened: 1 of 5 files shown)


gov/agendas/deployments.go
```diff
@@ -252,8 +253,9 @@ func (db *AgendaDB) storeAgenda(agenda *AgendaTagged) error {
 
 // UpdateAgendas updates agenda data for all configured vote versions.
 func (db *AgendaDB) UpdateAgendas() error {
-	if db == nil || db.sdb == nil {
-		return errDefault
```
buck54321: I didn't see a good reason for these. The only time this would be the case is if an error is returned from the constructor, in which case the main loop returns anyway.

gov/agendas/deployments_test.go
```diff
@@ -229,24 +228,9 @@ func TestUpdateAndRetrievals(t *testing.T) {
...
-		t.Run("Test_UpdateAgendas_#"+strconv.Itoa(i), func(t *testing.T) {
-			err := val.db.UpdateAgendas()
-			if err != nil && val.errMsg != err.Error() {
-				t.Fatalf("expect to find error '%s' but found '%v' ", val.errMsg, err)
-			}
-
-			if err == nil && val.errMsg != "" {
-				t.Fatalf("expected to find error '%s' but none was returned ", val.errMsg)
-			}
-		})
```
buck54321: This loop was just testing the removed nil checks.

davecgh:
There will not be any defined agendas at the start of new testnets since they are effectively continuations of the old testnets in terms of rules.  You can see when testnet3 was implemented via commit decred/dcrd@9e1579771b9d97c721a251e31e4584ca80d75df1 that it has no agendas defined and all previously enabled rules are active.

Consequently, there will be no version that could be used with `getvoteinfo`, so it would always return a `VoteVersionError`.

chappjc:
Thanks, @davecgh.  That's exactly what I was wondering and what I suspected.

chappjc:
So we can merge this to get simnet working again, but for testnet4 a more generic solution is needed.  Perhaps at least an iteration limit in `listStakeVersions` for now.  Thoughts about that @buck54321?

buck54321:
Iteration limit had crossed my mind too.

PR #1617, opened 2019-11-20
[WIP] 51% attack cost

ademuanthony:
for issue https://github.com/decred/dcrdata/issues/1022

image

PR #1664, opened 2020-01-03
51% attack cost page

### LINKED ISSUE #1022: calculate 51% attack cost
This can be put anywhere on dcrdata. Its a frequently discussed number, and dcrdata is the best place to be able to reference the equation as well as get the current calculated result.

ademuanthony:
Fix #1022 
This is a replacement for #1617 
It is a continuation of work done by other developers on the 51% attack cost.
This PR adds a page for majority attack cost calculation for both internal and external attack type using network hashrate, ticket price and ticket pool size. 
The user has the ability to adjust parameters like attack time, electricity cost, exchange rate and mining device, and see the attack cost in US dollar for every selected set of options.

image

### DIFF
--- a/public/js/controllers/attackcost_controller.js
+++ b/public/js/controllers/attackcost_controller.js
@@ -98,6 +98,7 @@ export default class extends Controller {
       'actualHashRate', 'attackPercent', 'attackPeriod', 'blockHeight', 'countDevice', 'device',
       'deviceCost', 'deviceDesc', 'deviceName', 'external', 'internal', 'internalHash',
       'kwhRate', 'kwhRateLabel', 'otherCosts', 'priceDCR', 'internalAttackText', 'targetHashRate', 'externalAttackText',
+      'externalAttackPosText', 'additionalTickets', 'newTicketPoolValue', 'internalAttackPosText',
       'additionalHashRate', 'targetPos', 'targetPow',
       'ticketAttackSize', 'ticketPoolAttack', 'ticketPoolSize', 'ticketPoolSizeLabel',
       'ticketPoolValue', 'ticketPrice', 'tickets', 'ticketSizeAttack', 'durationLongDesc',
@@ -294,7 +295,7 @@ export default class extends Controller {
 
     switch (this.settings.attack_type) {
       case 1:
-        this.targetHashRate = hashrate / (1 - parseFloat(this.targetPowTarget.value) / 100)
+        this.targetHashRate = hashrate / (1 - parseFloat(this.targetPowTarget.value) / 100) - hashrate
         this.projectedPriceDivTarget.style.display = 'block'
         this.internalAttackTextTarget.classList.add('d-none')
         this.externalAttackTextTarget.classList.remove('d-none')
@@ -323,6 +324,16 @@ export default class extends Controller {
     this.targetPowTarget.value = Math.floor(randomVal * 100) / 100
     this.internalHashTarget.innerHTML = digitformat((rate * this.targetHashRate), 4) + ' Ph/s '
     this.ticketsTarget.innerHTML = digitformat(val * tpSize) + ' tickets '
+    switch (this.settings.attack_type) {
+      case 1:
+        this.hideAll(this.internalAttackPosTextTarget)
+        this.showAll(this.externalAttackPosTextTargets)
+        break
+      case 0:
+      default:
+        this.hideAll(this.externalAttackPosTextTargets)
+        this.showAll(this.internalAttackPosTextTarget)
+    }
 
     this.calculate(true)
   }
@@ -339,8 +350,16 @@ export default class extends Controller {
     var totalElectricity = totalKwh * parseFloat(this.kwhRateTarget.value)
     var extraCostsRate = 1 + parseFloat(this.otherCostsTarget.value) / 100
     var totalPow = extraCostsRate * totalDeviceCost + totalElectricity
-    var ticketAttackSize = (tpSize * parseFloat(this.targetPosTarget.value)) / 100
-    var DCRNeed = tpValue * (parseFloat(this.targetPosTarget.value) / 100)
+    var ticketAttackSize, DCRNeed
+    if (this.settings.attack_type === 1) {
+      ticketAttackSize = tpSize / (1 - parseFloat(this.targetPosTarget.value) / 100)
+      this.additionalTicketsTarget.innerHTML = digitformat(ticketAttackSize - tpSize, 2)
+      this.newTicketPoolValueTarget.innerHTML = digitformat(ticketAttackSize, 2)
+      DCRNeed = tpValue / (1 - parseFloat(this.targetPosTarget.value) / 100)
+    } else {
+      ticketAttackSize = (tpSize * parseFloat(this.targetPosTarget.value)) / 100
+      DCRNeed = tpValue * (parseFloat(this.targetPosTarget.value) / 100)
+    }
     var projectedTicketPrice = DCRNeed / tpSize
     this.setAllValues(this.ticketPoolAttackTargets, digitformat(DCRNeed))
     this.ticketPoolValueTarget.innerHTML = digitformat(hashrate, 3)
@@ -381,5 +400,13 @@ export default class extends Controller {
 
   setAllValues (targets, data) {
     targets.map((n) => { n.innerHTML = data })
+  }
+
+  hideAll (targets) {
+    targets.map(el => el.classList.add('d-none'))
+  }
+
+  showAll (targets) {
+    targets.map(el => el.classList.remove('d-none'))
   }
 }
--- a/public/js/controllers/attackcost_controller.js
+++ b/public/js/controllers/attackcost_controller.js
@@ -112,7 +112,7 @@ export default class extends Controller {
       'actualHashRate', 'attackPercent', 'attackPeriod', 'blockHeight', 'countDevice', 'device',
       'deviceCost', 'deviceDesc', 'deviceName', 'external', 'internal', 'internalHash',
       'kwhRate', 'kwhRateLabel', 'otherCosts', 'priceDCR', 'internalAttackText', 'targetHashRate', 'externalAttackText',
-      'externalAttackPosText', 'additionalTickets', 'newTicketPoolValue', 'internalAttackPosText',
+      'externalAttackPosText', 'additionalTickets', 'internalAttackPosText',
       'additionalHashRate', 'targetPos', 'targetPow', 'operatorSign',
       'ticketAttackSize', 'ticketPoolAttack', 'ticketPoolSize', 'ticketPoolSizeLabel',
       'ticketPoolValue', 'ticketPrice', 'tickets', 'ticketSizeAttack', 'durationLongDesc',
@@ -376,7 +376,7 @@ export default class extends Controller {
       ticketAttackSize = tpSize / (1 - parseFloat(this.targetPosTarget.value) / 100)
       DCRNeed = tpValue / (1 - parseFloat(this.targetPosTarget.value) / 100)
       this.setAllValues(this.additionalTicketsTargets, digitformat(ticketAttackSize - tpSize, 0))
-      this.setAllValues(this.newTicketPoolValueTargets, digitformat(ticketAttackSize, 0))
+      // this.setAllValues(this.newTicketPoolValueTargets, digitformat(ticketAttackSize, 0))
       this.setAllValues(this.operatorSignTargets, '+')
     } else {
       ticketAttackSize = (tpSize * parseFloat(this.targetPosTarget.value)) / 100
--- a/public/js/controllers/attackcost_controller.js
+++ b/public/js/controllers/attackcost_controller.js
@@ -71,6 +71,9 @@ const deviceList = {
   }
 }
 
+const externalAttackType = 'external'
+const internalAttackType = 'internal'
+
 function legendFormatter (data) {
   var html = ''
   if (data.x == null) {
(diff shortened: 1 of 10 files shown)


zubairzia0:
UI is looking good. At a quick glance, the way the total attack cost is displayed makes it seem like there are two attack vectors (you need PoW + PoS for block production). Imo, dropping the word 'attack' in 'PoS attack cost' and 'PoW attack cost' wherever displayed would be better. 

In addition, would it be possible to expose your local server and checkout the UI? it would allow for some quick UI feedback (I have used https://ngrok.com/ and it works really well).

views/extras.tmpl
```diff
@@ -48,7 +49,7 @@
 	<link rel="preload" href="/images/disconnected.svg?x=55tv3" as="image" type="image/svg+xml" />
 	<link href="/dist/css/style.css?x=9nfyu4m" rel="stylesheet">
 
-	<script src="/js/vendor/turbolinks.min.js?x=55tv3"></script>
+	<script src="/js/vendor/turbolinks.min.js"></script>
```
chappjc: You did not resolve conflicts correctly.  These changes should not be made.

chappjc:
This looks like nice work.  I can give you a full review this afternoon or tomorrow, but there are some issues that snuck in when you rebased.

ademuanthony:
> UI is looking good. At a quick glance, the way the total attack cost is displayed makes it seem like there are two attack vectors (you need PoW + PoS for block production). Imo, dropping the word 'attack' in 'PoS attack cost' and 'PoW attack cost' wherever displayed would be better.
> 
> In addition, would it be possible to expose your local server so that we could check out the UI? it would allow for some quick feedback (I have used https://ngrok.com/ and it works really well).

We have the test code running at https://explorer.planetdecred.org/attack-cost

raedah:
Projected ticket price on external attack looks wrong. Some values show a ticket price lower than the current price. An external attack is always adding additional funds so all cases should show a rise in ticket price.

buck54321:
A couple of things to consider.

- You've plotted hashrate multiplier vs tickets, which is then used as an input the user can click to pick a particular value of tickets, which is then converted to a minimum attack cost. My question is why don't you just plot minimum attack cost vs tickets directly? The existing plot is already 1-dimensional, since all values are restricted to the minimum cost line anyway. At the very least, can the cost of attack value update as the user hovers around the chart?

- There is actually a relationship between ticket fraction, hashrate fraction and minimum attack time that you could leverage to drop the time input. If a user holds 'y' portion of the tickets and has 'x' fraction of the honest hashrate, and with a somewhat naive assumption that the honest hashrate did not change when the attack began, then the attacker would take `t = 5/x` to generate each block, and would not be able to validate (on average) until every `1/P` blocks, with P from 

image


(little `p` is stakeholder participation = 1 in your setup). `(N k)` is the [binomial coefficient](https://en.wikipedia.org/wiki/Binomial_coefficient) and is where all those coefficients in your `rateCalculation` function come from.

So it follows that the minimum average time of attack is `5/xP`. Here it is plotted against some hashrate multipliers.

image

buck54321:
- About that naive hashrate assumption... There are actually two special cases that could potentially be used to establish a **range** of valid attack costs which is more realistic. What you're displaying is a sort of upper minimum attack cost, since you are assuming that the attacker is bringing on new hashpower rather than redirecting existing hashpower from the network. Presenting that info as the "correct" attack cost is misleading at best. A more likely scenario is that the attacker already controls some hashrate mining honestly on the network that is then redirected to their private network. This would appear as a large hashrate dip on the honest network, lowering the attack cost significantly. So if your calculation establishes an "upper minimum", the "lower minimum" is the case where all of the attacker's hashrate up to 100% of the pre-attack hashrate is also subtracted from the honest network.

public/js/controllers/attackcost_controller.js
```diff
@@ -0,0 +1,385 @@
...
+import { nightModeOptions } from '../helpers/chart_helper'
+import dompurify from 'dompurify'
+
+function digitformat (amount, decimalPlaces, noComma) {
+  if (!amount) return 0
+
+  if (noComma) return amount.toFixed(decimalPlaces)
+
+  decimalPlaces = decimalPlaces || 0
+  let result = parseFloat(amount).toLocaleString(undefined, { minimumFractionDigits: decimalPlaces, maximumFractionDigits: decimalPlaces }).replace(/\.0*$/, '')
```
buck54321: Better to define these string formatter objects as constants than to recreate them every time. Consider using sig figs too.

```
const legendFormatter = {
  minimumSignificantDigits: 4,
  maximumSignificantDigits: 5
}
```

Or consider using `humanize.threeSigFigs` which will scale the value and append an appropriate 'k, M, B' unit prefix.

public/js/controllers/attackcost_controller.js
```diff
@@ -0,0 +1,385 @@
...
+    cost: 4199, // $
+    name: 'D1',
+    link: 'https://www.cryptocompare.com/mining/crypto-drilling/microbt-whatsminer-d1-plus-psu-dcr-44ths/'
+  }
+}
+
+function legendFormatter (data) {
+  var html = ''
+  if (data.x == null) {
+    let dashLabels = data.series.reduce((nodes, series) => {
```
buck54321: If `data.x == null`, then there's actually nothing to display, so just `return ''` and the box won't even be drawn, if I'm remembering correctly.

public/js/controllers/attackcost_controller.js
```diff
@@ -0,0 +1,385 @@
...
+    if (this.settings.target_pos || this.settings.target_pow) this.attackPercentTarget.value = (this.targetPowTarget.value || this.targetPosTarget.value) / 100
+
+    this.setDevicesDesc()
+    this.updateSliderData()
+
+    Dygraph = await getDefault(
+      import(/* webpackChunkName: "dygraphs" */ '../vendor/dygraphs.min.js')
+    )
+
+    Dygraph.prototype.doZoomY_ = function (lowY, highY) {}
```
buck54321: Please comment on why this is necessary.

public/js/helpers/chart_helper.js
```diff
@@ -230,3 +230,18 @@ export function synchronize (dygraphs, syncOptions) {
...
+      gridLineColor: '#596D81',
+      colors: ['#2DD8A3', '#2970FF', '#FFC84E']
+    }
+  }
+  return {
+    rangeSelectorAlpha: 0.4,
+    gridLineColor: '#C4CBD2',
+    colors: ['#2970FF', '#006600', '#FF0090']
+  }
+}
```
buck54321: These will likely be very specific to each chart, so until you see a case for reuse, I'd move these to *attackcost_controller.js* so you don't need the import.

public/scss/attackcost.scss
```diff
@@ -0,0 +1,149 @@
+input[type=number] {
+  -moz-appearance: textfield;
+  font-size: 0.95rem;
+  border-radius: 5px;
```
buck54321: Try just `3px`. Can we do something about the borders too?

image

views/attackcost.tmpl
```diff
@@ -0,0 +1,378 @@
+{{define "attackcost"}}
+<!DOCTYPE html>
+<html lang="en">
+{{ template "html-head" "Majority Attack Cost Calculator"}}
+{{template "navbar" . }}
+
```
buck54321: Since there's a lot going on in this file, consider throwing up some section comments

`{{- /* ADJUSTABLE PARAMETERS */ -}}`

The dashes at the beginning and end mean to strip whitespace before and after during template parsing, so not only is the file easier to read, it ends up being smaller.

views/attackcost.tmpl
```diff
@@ -0,0 +1,378 @@
...
+            <div class="justify-content-center align-items-center" data-target="attackcost.graph" data-action="mouseleave->attackcost#updateSliderData" style="width:100%; height:200px;"></div>
+          </div>
+          <div class="col-24 col-sm-12 col-md-24 col-lg-6 bg-grey px-3 position-relative fs13">
+            <div>
+              <div class="pl-1">
+                  <span class="h6 d-inline-block pl-2">Description</span>
+              </div>
+              <div>
+                <input
+                  class="slider"
```
buck54321: I love this slider.

public/js/controllers/attackcost_controller.js
ademuanthony: I can't see how this can be made a constant as it used with varying parameters all through the controller

public/js/controllers/attackcost_controller.js
ademuanthony: The idea is to show `N/A` instead of an empty space for such situations

buck54321: Why, though? To me, when it just randomly says N/A with no context, I assume an error has occurred.

views/extras.tmpl
```diff
@@ -1,6 +1,7 @@
 {{define "html-head"}}
 <head data-turbolinks-eval="false">
 	<meta charset="utf-8">
+	<meta http-equiv="X-UA-Compatible" content="IE=edge">
```
chappjc: Most of the changes to this file still need to be reverted.
This one.

views/extras.tmpl
```diff
@@ -25,15 +26,15 @@
 		<link rel="icon" href="/images/favicon/favicon-16x16.png?v=yHh3NA" type="image/png" sizes="16x16">
 
 		<!-- Android PWA -->
-		<link rel="manifest" href="/images/favicon/site.webmanifest?v=yHh3NA">
+		<link rel="manifest" href="/images/favicon/manifest.json?v=yHh3NA">
```
chappjc: This one.

views/extras.tmpl
```diff
@@ -25,15 +26,15 @@
...
 
 		<!-- Safari -->
 		<link rel="mask-icon" href="/images/favicon/safari-pinned-tab.svg?v=yHh3NA" color="#091440">
 
 		<!-- Windows PWA -->
 		<meta name="msapplication-TileColor" content="#091440">
-		<meta name="msapplication-TileImage" content="/images/favicon/mstile-144x144.png?v=55tv3">
-		<meta name="msapplication-config" content="/images/favicon/browserconfig.xml?v=55tv3">
+		<meta name="msapplication-TileImage" content="/images/favicon/mstile-144x144.png?v=fi5jKKtbwv">
+		<meta name="msapplication-config" content="/images/favicon/browserconfig.xml?v=fi5jKKtbwv">
```
chappjc: These

views/extras.tmpl
```diff
@@ -73,9 +74,8 @@
 				class="top-search mousetrap"
 				placeholder="Search for blocks, addresses, transactions or proposal tokens"
 				spellcheck="false"
-				autocomplete="off"
 			/>
-			<button class="search-bttn" type="submit"><span class="dcricon-search"></span></button>
+			<button class="search-bttn"><span class="dcricon-search"></span></button>
```
chappjc: This

views/extras.tmpl
```diff
@@ -165,12 +166,12 @@
...
 	</div>
 	<script
-		src="/dist/js/vendors~app.bundle.js?x=55tv3"
+		src="/dist/js/vendors~app.bundle.js"
 		data-turbolinks-eval="false"
 		data-turbolinks-suppress-warning
 	></script>
 	<script
-		src="/dist/js/app.bundle.js?x=55tv3"
+		src="/dist/js/app.bundle.js"
```
chappjc: These

chappjc:
Thanks for undoing those extras.tmpl changes, but prefer not to introduce all those whitespace changes.  I suggest to checkout master's version of the file and go from there.

ademuanthony:
> Encountered two issues. Was unable to reproduce the first. It was on a fresh build, fresh start, re-bundled web assets, cleared cache, etc. Keep an eye out for it. The chart was working fine initially, but then after refreshing it wasn't.
> 
> image
> 
> Seems that the ticket pools size is not being set in the controller element.
> 
> image
> 
> After waiting a bit, it eventually started working again, but something else is off.
> 
> image
> 
> We get some boost from hybrid consensus, but 3 trillion dollars is a bit high.

I am unable to reproduce this issue on a fresh build even after clearing my cache and trying from another web browser

ademuanthony:
> > Encountered two issues. Was unable to reproduce the first. It was on a fresh build, fresh start, re-bundled web assets, cleared cache, etc. Keep an eye out for it. The chart was working fine initially, but then after refreshing it wasn't.
> > image
> > Seems that the ticket pools size is not being set in the controller element.
> > image
> > After waiting a bit, it eventually started working again, but something else is off.
> > image
> > We get some boost from hybrid consensus, but 3 trillion dollars is a bit high.
> 
> I am unable to reproduce this issues on a fresh build even after clearing my cache and trying from another web browser

I think this PR can be merged and for the issue pointed out, a follow up PR can be created for the bug when it is isolated.

public/js/controllers/attackcost_controller.js
```diff
@@ -0,0 +1,438 @@
...
+
+    this.targetHashRate *= rateCalculation(this.targetPosTarget.value / 100)
+    this.query.replace(this.settings)
+    var deviceInfo = deviceList[this.selectedDevice()]
+    var deviceCount = Math.ceil((this.targetHashRate * 1000) / deviceInfo.hashrate)
+    var totalDeviceCost = deviceCount * deviceInfo.cost
+    var totalKwh = deviceCount * deviceInfo.power * parseFloat(this.attackPeriodTarget.value) / 1000
+    var totalElectricity = totalKwh * parseFloat(this.kwhRateTarget.value)
+    var extraCost = parseFloat(this.otherCostsTarget.value) / 100 * (totalDeviceCost + totalElectricity)
+    var totalPow = extraCost * totalDeviceCost + totalElectricity
```
buck54321: This seems to be the error. Shouldn't this be `extraCost + totalDeviceCost + totalElectricity`?

If you aren't seeing this error, you should check to make sure your web assets are recompiling.

ademuanthony: 🤦🏻‍♂️
Too bad, I missed this. 
Thank you

raedah:
Im seeing errors in the calc that are causing external attack to show with negative numbers which should not occur.

ademuanthony:
I have two pending issues on this PR

- One is to make the target PoW attack input editable and use the value to calculate the attack cost
- The second one is to get clarity on how to calculate the expected hash rate for an external attack for a giving ticket percentage

zubairzia0:
Regarding the target PoW attack input being editable, the slider already solves this issue and that is why I thought a slider would be a good idea. However, if we want the hashpower to be an input we would solve for equation  `(6 (1-stake)⁵ -15(1-stake)⁴ + 10(1-stake)³) / (6stake⁵-15stake⁴ + 10stake³)=%hashpowe`r and I am not sure how to do that in JS.

raedah:
Ive tested the latest version of this and it looks good to go. What is here works, and it would be good to have it merged to see it get some use and feedback. There are several areas for enhancements that can come in later PRs.

PR #1620, opened 2019-11-22
eat excessive white space in block template

chappjc:
I was hacking on the template while working on a different task and this got a little out of hand, so I'm making a separate PR.

PR #1665, opened 2020-01-08
cleanup blocks html template - whitespace, missing tags, indentation

chappjc:
Replaces PR https://github.com/decred/dcrdata/pull/1620

- There was a missing `</td>` tag that is fixed.
- Switch to tabs to be consistent with extras.html (and use fewer chars and allow IDE set indent depth)
- Eat whitespace in template elements (`{- .adsf -}`)

Space savings is relatively minor.
- Before: block.tmpl 19127 bytes / block.tmpl.gz 2678 bytes
- After: block.tmpl 13013 bytes / block.tmpl.gz 2417

The generated html is however much cleaner and more consistently formatted, with a couple tag fixes.

### DIFF
--- a/views/block.tmpl
+++ b/views/block.tmpl
@@ -3,362 +3,362 @@
 <html lang="en">
 {{template "html-head" printf "Decred Block %d" .Data.Height}}
 {{ template "navbar" . }}
-  <div class="container main" data-controller="time">
-  {{- with .Data -}}
-  {{$Invalidated := and (gt .Confirmations 1) (not .Valid) }}
-    <div class="row mx-2 my-2">
-      <div class="col-24 col-xl-12 bg-white p-3 position-relative">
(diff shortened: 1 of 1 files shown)


views/block.tmpl
```diff
@@ -1,388 +1,365 @@
...
+			</div>
+		</div>
+		<div class="col-24 col-xl-12 secondary-card py-3 px-3 px-xl-4">
+			<div class="h6 d-inline-block my-2 pl-3">Block Details</div>
+			<table class="w-100 fs14 mt-2 details">
+				<tbody>
+					<tr>
+						<td class="text-right font-weight-bold text-nowrap pr-2"
+							><span class="d-none d-sm-inline">Ticket Price</span
+							><span class="d-sm-none">Tkt Price</span>: </td>
```
chappjc: The one actual bug fix.

PR #1666, opened 2020-01-08
Display time ago on block page

### LINKED ISSUE #1632: add time ago to block page
Regression, when looking at a block page it was previously possible to know how old a block was without having to do the math by hand to subtract the current time from the block time stamp. The functionality was lost in one of the page redesigns.

peterzen:
Fixes #1632

### DIFF
--- a/views/block.tmpl
+++ b/views/block.tmpl
@@ -86,7 +86,7 @@
 					<br>
 					<span class="fs18 font-weight-bold lh1rem d-inline-block pt-1">{{.BlockTime.PrettyMDY}}</span>
 					<br>
-					<span class="fs14 text-secondary">{{.BlockTime.HMSTZ}}</span>
+					<span class="fs14 text-secondary">{{.BlockTime.HMSTZ}} <span class="jsonly">(<span data-target="time.age" data-age="{{.BlockTime.UNIX}}"></span> ago)</span></span>
 				</div>
 			</div>
 			<div class="d-flex justify-content-around fs15 text-center text-secondary my-2">


views/block.tmpl
```diff
@@ -84,9 +84,9 @@
                 <div class="col-7 col-sm-8 text-left">
                     <span class="text-secondary fs13">Block Time</span>
                     <br>
-                    <span class="fs18 font-weight-bold lh1rem d-inline-block pt-1">{{.BlockTime.PrettyMDY}}</span>
+                    <span class="fs18 font-weight-bold lh1rem d-inline-block pt-1">{{.BlockTime.PrettyMDY}} <span class="fs13 text-secondary">{{.BlockTime.HMSTZ}}</span></span>
                     <br>
-                    <span class="fs14 text-secondary">{{.BlockTime.HMSTZ}}</span>
+                    <span class="text-secondary fs13"><span data-target="time.age" data-age="{{.BlockTime.UNIX}}"></span> ago</span>
```
chappjc: Please add `jsonly` to the outer/fs13 span class list.

peterzen: Done

peterzen:
Yes, the intention was to free up some space for the relative time this
way.

Forgot to post a screenshot, sorry about that.

buck54321:
This doesn't quite work. This breaks the whole layout pattern and doubles the width of the last element. What we've done elsewhere is to grab the element with javascript and change the hard date to a time since. This has the advantage that it doesn't change the layout, and those with JS disabled still get some data. Take a look at `time_controller.js` and elements with `data-target="time.age"`. 

There are other solutions too, but this doesn't work as is.

peterzen:
@raedah do you have a screenshot of this page in the state before the regression?

chappjc:
It was the major redesign, so it wouldn't be too helpful to see unfortunately.  How about putting the time below the date to keep the width down?

peterzen:
> It was the major redesign, so it wouldn't be too helpful to see unfortunately. How about putting the time below the date to keep the width down?

How about this layout?
image

chappjc:
That sits much better with me.  Lemme check out the code again.

chappjc:
@buck54321 Does the revised layout work for you?

PR #1667, opened 2020-01-08
Timestamp cleanup in table views

### LINKED ISSUE #1614: Move (UTC) from timestamps to column header or label
It would clear up some space if every timestamp didn't repeat the `(UTC)`.

I suggest placing it once in the column header or timestamp's label and removing from all individual timestamps. The users should get used to the general rule that all dates are in UTC.

Pages affected:

- home
- /blocks
- /address
- /agendas
- /proposal
- maybe others

peterzen:
Moved timezone strings from timestamps to column headers in table views in order to clean them up.

Closes #1614

### DIFF
--- a/explorer/templates.go
+++ b/explorer/templates.go
@@ -455,12 +455,12 @@ func makeTemplateFuncMap(params *chaincfg.Params) template.FuncMap {
 			dateTime := time.Unix(int64(a), 0).UTC()
 			return dateTime.Format("2006-01-02 15:04:05 MST")
 		},
-		"timeWithoutDateAndTimeZone": func(a uint64) string {
+		"dateTimeWithoutTimeZone": func(a uint64) string {
 			if a == 0 {
 				return "N/A"
 			}
 			dateTime := time.Unix(int64(a), 0).UTC()
-			return dateTime.Format("2006-01-02")
+			return dateTime.Format("2006-01-02 15:04:05")
 		},
 		"toLowerCase": strings.ToLower,
 		"toTitleCase": strings.Title,
--- a/explorer/types/explorertypes.go
+++ b/explorer/types/explorertypes.go
@@ -35,8 +35,9 @@ type TimeDef struct {
 }
 
 const (
-	timeDefFmtHuman = "2006-01-02 15:04:05 (MST)"
-	timeDefFmtJS    = time.RFC3339
+	timeDefFmtHuman        = "2006-01-02 15:04:05 (MST)"
+	timeDefFmtDateTimeNoTZ = "2006-01-02 15:04:05"
+	timeDefFmtJS           = time.RFC3339
 )
 
 // String formats the time in a human-friendly layout. This ends up on the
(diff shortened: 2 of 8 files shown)


chappjc:
@peterzen Thanks for your fixes, and sorry for the delay reviewing.  This looks good.  I will review today.

explorer/templates.go
```diff
@@ -462,6 +462,13 @@ func makeTemplateFuncMap(params *chaincfg.Params) template.FuncMap {
 			dateTime := time.Unix(int64(a), 0).UTC()
 			return dateTime.Format("2006-01-02")
 		},
+		"dateTimeWithoutTimeZone": func(a uint64) string {
```
chappjc: While you're here, would you please remove `timeWithoutDateAndTimeZone`, which seems to be unused and incorrectly named?

explorer/types/explorertypes.go
```diff
@@ -90,6 +91,12 @@ func (t *TimeDef) HMSTZ() string {
 	return t.T.Format("15:04:05 MST")
 }
 
+// DatetimeWithoutTZ formats the time in a human-friendly layout, without
+// time zone.
+func (t TimeDef) DatetimeWithoutTZ() string {
```
chappjc: Pointer receivers for this and the other one in db/dbtypes please.

public/js/helpers/humanize_helper.js
```diff
@@ -160,8 +160,10 @@ var humanize = {
   },
   date: function (stamp, withTimezone) {
     var d = new Date(stamp)
-    return `${String(d.getUTCFullYear())}-${String(d.getUTCMonth() + 1).padStart(2, '0')}-${String(d.getUTCDate()).padStart(2, '0')} ` +
-          `${String(d.getUTCHours()).padStart(2, '0')}:${String(d.getUTCMinutes()).padStart(2, '0')}:${String(d.getUTCSeconds()).padStart(2, '0')} (UTC)`
+    var dateStr = `${String(d.getUTCFullYear())}-${String(d.getUTCMonth() + 1).padStart(2, '0')}-${String(d.getUTCDate()).padStart(2, '0')} ` +
+          `${String(d.getUTCHours()).padStart(2, '0')}:${String(d.getUTCMinutes()).padStart(2, '0')}:${String(d.getUTCSeconds()).padStart(2, '0')}`
+    if (withTimezone) dateStr += ' (UTC)'
+    return dateStr
```
chappjc: This seems reasonable to me, although since previously `withTimezone` was being dropped, I wonder if there was a latent bug.  I just want to ping @buck54321 about this to see if this rings a bell.

buck54321: Without looking too deeply into it, I would guess it's an artifact of an older implementation, or a feature that was never fully hashed out.

chappjc: Thanks.  I don't see any issues with this change and everything I've checked looks fine.

PR #1669, opened 2020-01-08
Copyright notice update

peterzen:


buck54321:
In 2019, #1025 changed the footer on February 17th, so we're doing much better this year.

PR #1677, opened 2020-01-24
explorer,views: fix charts page target ticket pool size

chappjc:
This fixes the charts page's target ticket pool size, which was showing ticket expiry in blocks. This must have slipped in because on mainnet these are equal values although their units are different and need not be the same value.

In `(*explorerUI).Charts` get target pool size from `exp.pageData.HomeInfo.PoolInfo.Target`, and pass to page template.

`explorer/types`: change `TicketPoolInfo.Target` from `uint16` to `uint32`.


chappjc:
Tests OK:

image

chappjc:
Confirmed OK on mainnet too.  Merging...

PR #1678, opened 2020-01-31
ticket-charts: add smooth mode switch

### LINKED ISSUE #1502: ticket price chart smooth mode
Add switch to change ticket price chart between smooth lines and the current stepped lines. Smooth mode previously existed and would be viewable in older code if needed.

Can also note that this chart is currently also applying the stepped lines to the Tickets Bought line. It generally would not apply to that since tickets are bought per block and not per window.

njirap:
fixes #1502

tkt-smooth-mode

Screen Shot 2020-01-31 at 6 28 05 PM
Screen Shot 2020-01-31 at 6 28 15 PM

### DIFF
--- a/views/charts.tmpl
+++ b/views/charts.tmpl
@@ -198,7 +198,7 @@
                             <li class="nav-item active">
                                 <a
                                     class="nav-link active"
-                                    href="javascript:void(0);"
+                                    href="#"
                                     data-target="charts.modeOption"
                                     data-action="click->charts#setMode"
                                     data-option="smooth"
@@ -207,7 +207,7 @@
                             <li class="nav-item">
                                 <a
                                     class="nav-link"
-                                    href="javascript:void(0);"
+                                    href="#"
                                     data-target="charts.modeOption"
                                     data-action="click->charts#setMode"
                                     data-option="stepped"
--- a/views/charts.tmpl
+++ b/views/charts.tmpl
@@ -199,6 +199,7 @@
                                 <a
                                     class="nav-link active"
                                     href="#"
+                                    onclick="return false;"
                                     data-target="charts.modeOption"
                                     data-action="click->charts#setMode"
                                     data-option="smooth"
--- a/views/charts.tmpl
+++ b/views/charts.tmpl
@@ -198,8 +198,7 @@
                             <li class="nav-item active">
                                 <a
                                     class="nav-link active"
-                                    href="#"
-                                    onclick="return false;"
+                                    href="javascript:void(0);"
                                     data-target="charts.modeOption"
                                     data-action="click->charts#setMode"
                                     data-option="smooth"
--- a/views/charts.tmpl
+++ b/views/charts.tmpl
@@ -207,7 +207,7 @@
                             <li class="nav-item">
                                 <a
                                     class="nav-link"
-                                    href="#"
+                                    href="javascript:void(0);"
                                     data-target="charts.modeOption"
                                     data-action="click->charts#setMode"
                                     data-option="stepped"
--- a/views/charts.tmpl
+++ b/views/charts.tmpl
@@ -189,6 +189,34 @@
                     </div>
                 </div>
 
+                <div class="chart-control-wrapper mr-2 mb-1" data-target="charts.modeSelector">
+                    <div class="chart-control-label">MODE</div>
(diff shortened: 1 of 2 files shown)


njirap:
Live demo also available http://dev.planetdecred.org:7778/charts?chart=ticket-price&zoom=k4ue8xmo-k61uw3eo&bin=window&axis=time&mode=stepped

views/charts.tmpl
```diff
@@ -189,6 +189,34 @@
...
+                <div class="chart-control-wrapper mr-2 mb-1" data-target="charts.modeSelector">
+                    <div class="chart-control-label">MODE</div>
+                    <div class="chart-control">
+                        <ul
+                            class="nav nav-pills"
+                        >
+                            <li class="nav-item active">
+                                <a
+                                    class="nav-link active"
+                                    href="javascript:void(0);"
```
chappjc: PR looks good, thanks.

@buck54321 I've wondered about the use of an anchor with the href javascript hack instead of a button.  Thoughts about changing this?

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#Accessibility

> Anchor tags are often abused as fake buttons by setting their href to "#" or "javascript:void(0)" to prevent the page from refreshing, then listening for their click events .
> 
> These bogus href values cause unexpected behavior when copying/dragging links, opening links in a new tab/window, bookmarking, or when JavaScript is loading, errors, or is disabled. They also convey incorrect semantics to assistive technologies, like screen readers.
> 
> Use a <button> instead. In general, you should only use a hyperlink for navigation to a real URL.

But the inline javascript is what stands out to me as a potential challenge for deploying a good CSP.

buck54321: I don't really understand why these are links at all. We should be applying the styling directly to the `<li>` elements, and making those elements the stimulus targets. This seems to be based on [a pattern from the bootstrap docs](https://getbootstrap.com/docs/4.0/components/navs/). Interestingly, in the docs, they do use the anchor tag for the href. 

@njirap, use `href="#"` for now in this PR. Another job might be to go through and remove all the links.

njirap: this is done

njirap: @buck54321 something's broken with the `href="#"` looking into it

http://dev.planetdecred.org:7778/charts?chart=ticket-price&zoom=k4ue8xmo-k61uw3eo&bin=window&axis=time&mode=stepped

chappjc: Maybe with onclick? https://stackoverflow.com/a/1291944

If that doesn't work, we can stick with the javascript: unit we can entirely nuke these anchors.

gozart1: Using anchor tags is just an old convention since they automatically deal with small things like styling cursor hover and active state in a way that feels native to whatever client the end user is on. 

Also if the application is using url backed state then its kind of semantic to use an anchor tag since the state change of clicking on it ( eg a query param to represent line mode) would get reflected in the url updating as well. 

So I tend to just stick to using empty `<a href="">` tags out of habit and simplicity.

But I think just using a utility class like ` .clickable {
   cursor: pointer;
 }` and using any element that seems most semantic to handle clicks is just fine as well!

Just sharing my 2 cents here since I may have played a role in seeding this pattern in this repo. Don't have any qualms about moving away from the pattern if it feels wrong to folks.

chappjc: > Also if the application is using url backed state then its kind of semantic to use an anchor tag since the state change of clicking on it ( eg a query param to represent line mode) would get reflected in the url updating as well.

That's a good point about updating the URL, but I wouldn't find it odd for a `button` to do the same.

views/charts.tmpl
```diff
@@ -189,6 +189,34 @@
...
+                                    data-target="charts.modeOption"
+                                    data-action="click->charts#setMode"
+                                    data-option="smooth"
+                                >Smooth</a>
+                            </li>
+                            <li class="nav-item">
+                                <a
+                                    class="nav-link"
+                                    href="#"
```
chappjc: Please revert this one back too.

njirap: done

views/charts.tmpl
chappjc: Now tracking this in issue https://github.com/decred/dcrdata/issues/1679

PR #1690, opened 2020-02-11
Mixed chart query optimisation

### LINKED ISSUE #1689: mix chart queries need optimization
After observing this running on alpha.dcrdata.org for a while, this needs to be a high priority to fix.  This query and the processing of the result by dcrdata is bogging down the server.

There needs to be a version `SelectMixedVouts` with a height constraint like with `SelectMixedTotalPerBlock`.  Related to the at https://github.com/decred/dcrdata/pull/1640/files#diff-87f7a577b6af45d0284e4c3fbb98ddf5R3685

This change must be done ASAP as this is an expensive query that the server can't afford to have running often.

Ref: https://github.com/decred/dcrdata/pull/1640#discussion_r375321314

ademuanthony:
Fix #1689 

This PR adds a height constraint to the query that fetches vouts for computing anonymity set.

### DIFF
--- a/db/cache/charts.go
+++ b/db/cache/charts.go
@@ -796,8 +796,8 @@ func (charts *ChartData) TotalMixedTip() int32 {
 	return int32(len(charts.Blocks.TotalMixed)) - 1
 }
 
-// AnonymitySetUpdateOffset is the height offset for update of the anonymity set
-func (charts *ChartData) AnonymitySetUpdateOffset() int32 {
+// AnonymitySetTip is the height of the anonymity set
+func (charts *ChartData) AnonymitySetTip() int32 {
 	charts.mtx.RLock()
 	charts.mtx.RUnlock()
 	return int32(len(charts.Blocks.AnonymitySet)) - 1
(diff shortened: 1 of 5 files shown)


chappjc:
Thanks for the quick response on this @ademuanthony!  I'm glad you see the issue, and I appreciate your follow-through.  Your mix charts work has been very good.

Just a bit more code formatting to make CI happy.

ademuanthony:
> Thanks for the quick response on this @ademuanthony! I'm glad you see the issue, and I appreciate your follow-through. Your mix charts work has been very good.
> 
> Just a bit more code formatting to make CI happy.
> 
> Note that prior to merging the fix, I'll have to instrument a few functions to provide timing data. It would be helpful if you could do the same and provide before and after values.

This is the information from the planentdecred dev server
Before:
retrieveAnonymitySet:  28.039550932s
appendAnonymitySet:  1.154520337s

After:
retrieveAnonymitySet:  25.777803803s
appendAnonymitySet:  67.025679ms

chappjc:
Thanks for the timings.  I'm going to look into the query on my end however.  25 second is still to high, by a lot unfortunately.

ademuanthony:
In my new update on this PR, I have changed the process of the anonymity set computation to only fetch data from the database during first load and whenever there is a length mismatch in the chart data. Otherwise, I used the in coming mixed output and their spending notification to maintain the value of the running anonymity set

PR #1691, opened 2020-02-12
Fix chart loading indicator spinning forever

### LINKED ISSUE #1684: chart loading indicator spinning forever
On an address page that has no transaction history.

example https://explorer.dcrdata.org/address/DsQxuVRvS4eaJ42dhQEsCXauMWjvopWgrVg

ademuanthony:
Closes #1684 
 
This PR resolved the issue of chart loading indicator spinning forever on dummy address pages.
image

### DIFF
--- a/public/js/controllers/address_controller.js
+++ b/public/js/controllers/address_controller.js
@@ -227,7 +227,7 @@ export default class extends Controller {
 
   bindElements () {
     this.flowBoxes = this.flowTarget.querySelectorAll('input')
-    // pagesizeTarget is not available for addresses without txns
+    // pagesizeTarget is not available for dummy addresses
     this.pageSizeOptions = this.data.get('isDummy') === 'false' ? this.pagesizeTarget.querySelectorAll('option') : []
     this.zoomButtons = this.zoomTarget.querySelectorAll('button')
     this.binputs = this.intervalTarget.querySelectorAll('button')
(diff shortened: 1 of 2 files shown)


public/js/controllers/address_controller.js
```diff
@@ -227,7 +227,8 @@ export default class extends Controller {
 
   bindElements () {
     this.flowBoxes = this.flowTarget.querySelectorAll('input')
-    this.pageSizeOptions = this.pagesizeTarget.querySelectorAll('option')
+    // pagesizeTarget is not available for dummy addresses
+    this.pageSizeOptions = this.data.get('isDummy') === 'false' ? this.pagesizeTarget.querySelectorAll('option') : []
```
chappjc: The comment is really what you are after, so I suggest checking for the target directly rather than adding a data element:

```js
this.pageSizeOptions = this.hasPagesizeTarget ? this.pagesizeTarget.querySelectorAll('option') : []
```

https://stimulusjs.org/reference/targets

views/address.tmpl
```diff
@@ -13,6 +13,7 @@
       data-address-dcraddress="{{.Address}}"
       data-address-txn-count="{{$TxnCount}}"
       data-address-balance="{{toFloat64Amount .Balance.TotalUnspent}}"
+      data-address-is-dummy="{{.IsDummyAddress}}"
```
chappjc: If `hashPagesizeTarget` works out, can remove this data attribute.

PR #1692, opened 2020-02-12
Show double-digit numbers for non-leading time units

### LINKED ISSUE #1682: Always show double-digit numbers for non-leading time units
The Age column on many pages has an undesired (imo) dynamic where the label size constantly changes between bigger and smaller as the smaller time units alternate between one and two digits:

```
1m 1s
1m 11s
11m 11s
11m 1s
and so on
```

double-digits-blocks

double-digits-proposals

This breaks vertical uniformity that helps our vision to scan the numbers faster, or in other words is annoying.

To address this I suggest one of two approaches:

- always show two digits for all but the major unit (`4h 01m 01s`)
- always show two digits for all units (`04h 01m 01s`)

ademuanthony:
Closes #1682 

This PR adds 0 to the left of minute and second time units that are less than 10.
image

### DIFF
--- a/public/js/helpers/humanize_helper.js
+++ b/public/js/helpers/humanize_helper.js
@@ -140,7 +140,7 @@ var humanize = {
     }
     var pad = function (x) {
       if (x < 10) {
-        return `0${x}`
+        return `&nbsp${x}`
       }
       return x
     }
--- a/public/js/helpers/humanize_helper.js
+++ b/public/js/helpers/humanize_helper.js
@@ -138,25 +138,31 @@ var humanize = {
       }
       return result
     }
+    var pad = function (x) {
+      if (x < 10) {
+        return `0${x}`
+      }
+      return x
+    }
     interval = Math.floor(seconds / 3600)
     if (interval >= 1) {
       let extra = Math.floor((seconds - interval * 3600) / 60)
       let result = interval + 'h'
       if (extra > 0) {
-        result = result + ' ' + extra + 'm'
+        result = result + ' ' + pad(extra) + 'm'
       }
       return result
     }
     interval = Math.floor(seconds / 60)
     if (interval >= 1) {
       let extra = seconds - interval * 60
-      let result = interval + 'm'
+      let result = pad(interval) + 'm'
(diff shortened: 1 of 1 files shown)


public/js/helpers/humanize_helper.js
```diff
@@ -138,25 +138,31 @@ var humanize = {
       }
       return result
     }
+    var pad = function (x) {
+      if (x < 10) {
+        return `0${x}`
```
chappjc: I suspect padding with a space instead of a zero would look better, at least for the first number that doesn't even need padding since it's on the left.

public/js/helpers/humanize_helper.js
```diff
@@ -138,25 +138,31 @@ var humanize = {
...
-        result = result + ' ' + extra + 'm'
+        result = result + ' ' + pad(extra) + 'm'
       }
       return result
     }
     interval = Math.floor(seconds / 60)
     if (interval >= 1) {
       let extra = seconds - interval * 60
-      let result = interval + 'm'
+      let result = pad(interval) + 'm'
```
chappjc: I don't know that the minutes ever needs padding.  Also zeros don't look good here.  And if there are values >100m then the padding wouldn't be adequate anyway, no?

ademuanthony: There won't be numbers greater than 59m, that will round to 1h.

chappjc: For some reason I thought we had this with minutes as the largest unit.  Thanks for correcting me.

public/js/helpers/humanize_helper.js
```diff
@@ -138,25 +138,31 @@ var humanize = {
       }
       return result
     }
+    var pad = function (x) {
+      if (x < 10) {
+        return `&nbsp${x}`
```
chappjc: This is actually showing "&nbsp" on the page.  Even with a `;`, it still doesn't work.  I don't know what the problem is with that, but why not use `padStart` with a unicode non-breaking space:

```js
x.toString().padStart(2, '\u00a0')
```

PR #1693, opened 2020-02-12
Show date in YYYY-MM-DD format on chart legends

### LINKED ISSUE #1681: proposal: odd date format on chart
The date format on the chart overlay is not the one used everywhere else, i.e. `YYYY/MM/DD` instead of `YYYY-MM-DD`.

odd-date-format

ademuanthony:
Closes #1681 
This PR updated the legend formatter of the chart on all the pages to show date in YYYY-MM-DD format. This include the charts on the main chart page, address, proposal and agenda page.
image

huminzie.date function is updated to accept a third parameter for hiding the H:i:s part of the date string

### DIFF
--- a/public/js/helpers/humanize_helper.js
+++ b/public/js/helpers/humanize_helper.js
@@ -158,10 +158,10 @@ var humanize = {
     }
     return Math.floor(seconds) + 's'
   },
-  date: function (stamp, withTimezone) {
+  date: function (stamp, withTimezone, hideHisForMidnight) {
     var d = new Date(stamp)
     var dateStr = `${String(d.getUTCFullYear())}-${String(d.getUTCMonth() + 1).padStart(2, '0')}-${String(d.getUTCDate()).padStart(2, '0')}`
-    if (d.getUTCHours() > 0 || d.getUTCMinutes() > 0 || d.getUTCSeconds() > 0) {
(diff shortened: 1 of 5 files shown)


public/js/helpers/humanize_helper.js
```diff
@@ -158,10 +158,10 @@ var humanize = {
     }
     return Math.floor(seconds) + 's'
   },
-  date: function (stamp, withTimezone) {
+  date: function (stamp, withTimezone, hideHisForMidnight) {
```
chappjc: I like the idea of hiding the time when the stamp is midnight because that is probably just representing the entire day, but I'm confused by the "i" in "His"/"h:i:s". Isn't that hours:minutes:seconds?

ademuanthony: Yes, it is hours:minutes:seconds
`i` is usually used for minute as `m` has already been used for month in `Y-m-d H:i:s`

chappjc: I don't see that often (php?), but that's OK.  The date formatting I'm used to with nix dates is capital "M" for minutes.  In any case, `hideHisForMidnight` is just a variable name so it might as well have clear meaning.  "His" is a head-scratcher IMO.  The variable could just as well be `hideTimeAtMidnight`.

PR #1697, opened 2020-02-14
supply only

chappjc:
This adds the `/api/supply/circulating` endpoint with the optional `dcr=[true|false]` URL query (default is false, meaning atoms).

Because CMC needs a number not some silly JSON thign.

```
$  curl http://127.0.0.1:7777/api/supply/circulating
1134639036687827
$  curl http://127.0.0.1:7777/api/supply/circulating?dcr=false
1134639036687827
$  curl http://127.0.0.1:7777/api/supply/circulating?dcr=true
11346390.36687827
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -676,12 +676,13 @@ Exchange monitoring is off by default. Server must be started with
 The server will set a default currency code. To use a different code, pass URL
 parameter `?code=[code]`. For example, `/exchanges?code=EUR`.
 
-| Other                           | Path            | Type               |
-| ------------------------------- | --------------- | ------------------ |
-| Status                          | `/status`       | `types.Status`     |
-| Health (HTTP 200 or 503)        | `/status/happy` | `types.Happy`     |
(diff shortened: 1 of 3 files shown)


hypernoob:
The "Reported Supply" on [Messari](https://messari.io/asset/decred/metrics) is wrong too.
It matches CMC's figure so the causes are probably the same.

chappjc:
Apparently this is back on the table, but a parameter or a separate endpoint to return atoms vs. DCR is needed first.

chappjc:
@jzbz @degeri Does the usage in the PR description look like it will work for cmc?

jzbz:
LGTM I don't see how we can do better than that for them.

degeri:
This was the example given. 

http://chainz.cryptoid.info/grs/api.dws?q=totalcoins

Should be fine.

chappjc:
Alrighty then.  That will do it for new 5.3 features.  We'll have to roll a release to get this deployed on the main site, which is overdue anyway considering we have mix charts and attack cost calc in this release.

PR #1699, opened 2020-02-17
Fix ticket window end block and block count.

hypernoob:
A ticket window ends when `(height + 1) % window_size == 0`;
Orphaned blocks should be excluded from ticket window block count in `/ticketpricewindows`.

window

### DIFF
--- a/views/windows.tmpl
+++ b/views/windows.tmpl
@@ -109,15 +109,14 @@
                       <th class="text-center d-none d-sm-table-cell">Difficulty</th>
                       <th class="text-center"><span class="d-none d-sm-inline">Ticket </span>Price<span class="d-none d-sm-inline"> (DCR)</span></th>
                       <th class="text-right pr-0 jsonly">Age</th>
-                      <th class="text-right pr-0 d-none d-sm-table-cell">Start Time</th>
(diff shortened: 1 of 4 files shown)


views/windows.tmpl
```diff
@@ -109,15 +109,14 @@
...
+                      <th class="text-right pr-0 d-none d-sm-table-cell">Start Time (UTC)</th>
                   </tr>
               </thead>
               <tbody>
               {{range .Data}}
                   <tr>
                       <td class="text-left pl-0">
-                          {{$blocksCount := (.BlocksCount)}}
-                          {{if lt .BlocksCount $.WindowSize}}{{$blocksCount = (subtract .BlocksCount 1)}} <span>({{$blocksCount}}/{{$.WindowSize}}) </span>{{end}}
+                          {{if lt .BlocksCount $.WindowSize}} <span>({{.BlocksCount}}/{{$.WindowSize}}) </span>{{end}}
```
chappjc: For these kinds of changes, screenshots of the edge cases need to be shown. First and last block in a window.

hypernoob: Screenshot added.

PR #1700, opened 2020-02-17
Fix block reward reduction progress off by 1.

hypernoob:
reward


chappjc:
Let's be sure the `remaining` and `rewardAdjustmentProgress` html template functions are going to show something sensible at the edge cases.  It's not clear to me that switching from 0-based to 1-based indexing is just going to work.

hypernoob:
Screenshot added.

PR #1701, opened 2020-02-17
charts: limit mode switch to ticket-price only

njirap:


public/js/controllers/charts_controller.js
```diff
@@ -41,6 +42,10 @@ function isScaleDisabled (chart) {
   return lineScales.indexOf(chart) > -1
 }
 
+function isModeEnabled (chart) {
+  return modeScales.indexOf(chart) > -1
```
chappjc: I think `return modeScales includes(chart)` is a bit more expressive here.

njirap: fixing it now

njirap: this is done

Issue #1680, opened 2020-02-05
mempool: always show 8 decimals in fee rate


xaur:
#1655 added fee rate columns to mempool page. I accidentally noticed that small fee rate is displayed in scientific notation instead of the standard 8-decimal notation:

![fee-rate-decimals](https://user-images.githubusercontent.com/24484727/73879099-eb536380-486c-11ea-91f6-8b3ab7ffbd6c.png)

_P.S. it's interesting that someone tries these low fees_

xaur:
Observed on dcrdata v5.3.0-pre+dev.

mkingori:
Can i work on this?

chappjc:
Resolved by https://github.com/decred/dcrdata/pull/1704

PR #1704, opened 2020-02-26
8-decimals notation on fee rate in mempool tickets and transactions

mkingori:
Fix for [#1680](https://github.com/decred/dcrdata/issues/1680)
mempool: small fee rate are displayed in scientific notation instead of the standard 8-decimal notation

_Steps to reproduce_
Go to [mempool](https://explorer.planetdecred.org/mempool) page
_Expected results_
Standard 8-decimal notations for fee rate
_Actual results_
Scientific notations for small fee rates.
_Cause_
The fee rates not being rounded up to 8-decimal notations
_Solution_
Rounded the fee rates to 8-decimal notations

### DIFF
--- a/views/mempool.tmpl
+++ b/views/mempool.tmpl
@@ -184,7 +184,7 @@
                                     {{template "decimalParts" (float64AsDecimalParts .TotalOut 8 false)}}
                                   </td>
                                   <td class="mono fs15 text-right">{{.Size}} B</td>
-                                  <td class="mono fs15 text-right">{{.FeeRate}} DCR/kB</td>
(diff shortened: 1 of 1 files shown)


chappjc:
Thanks for the fix.  I would note that the cause and solution described in your PR description are a bit misleading as rounding is not the issue, but default formatting.  It seems without a formating specifieder, `"%g"` is the effective formatting mode (`%g	%e for large exponents, %f otherwise. Precision is discussed below.`)  By explicitly specifying %f with 8 decimal digits, you the issue is solved.

PR #1707, opened 2020-03-06
explorer: fix default block height and divide by 0 rows

hypernoob:
Correctly handle URLs without rows and height parameters.

### DIFF
--- a/explorer/explorerroutes.go
+++ b/explorer/explorerroutes.go
@@ -387,7 +387,9 @@ func (exp *explorerUI) StakeDiffWindows(w http.ResponseWriter, r *http.Request) 
 	if offsetWindow > bestWindow {
 		offsetWindow = bestWindow
 	}
-	if rows > maxExplorerRows {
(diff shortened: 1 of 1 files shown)


explorer/explorerroutes.go
```diff
@@ -567,6 +569,14 @@ func (exp *explorerUI) timeBasedBlocksListing(val string, w http.ResponseWriter,
 
 // Blocks is the page handler for the "/blocks" path.
 func (exp *explorerUI) Blocks(w http.ResponseWriter, r *http.Request) {
+	bestBlockHeight, err := exp.dataSource.GetHeight()
```
dajohi: This should be moved after parsing query parameters so we don't hit the DB on invalid params.

chappjc: This particular method doesn't hit the DB. `GetHeight()`, which actually just calls `Height()`, does return the cached best block.  The `HeightDB` method on the other hand would hit the DB.

PR #1711, opened 2020-03-15
Allow searching for a transaction outpoint

### LINKED ISSUE #1672: allow searching for an outpoint
The search function should let you search for a tx outpoint of the form "hash:out".  It should be fairly straightforward for the search handler to identify search terms formatted this way, look up the tx hash, and redirect to the tx page with `/out/{out}` appended to the URL to highlight the output from the search term.

leRequinNoir:
Closes #1672

This pull request enables searching for a transaction outpoint formatted as `txid:out` where `txid` is a transaction hash and `out` is a transaction output index. Function `Search` redirects to path `/tx/{txid}/out/{out}` after seeing such search pattern. The transaction page then highlights the corresponding output.

### DIFF
--- a/explorer/explorerroutes.go
+++ b/explorer/explorerroutes.go
@@ -1641,10 +1641,10 @@ func (exp *explorerUI) Search(w http.ResponseWriter, r *http.Request) {
 			"", ExpStatusNotFound)
 		return
 	case len(searchStrSplit) > 1:
-		if _, err := strconv.ParseInt(searchStrSplit[1], 10, 0); err == nil {
+		if _, err := strconv.ParseUint(searchStrSplit[1], 10, 32); err == nil {
 			searchStrRewritten = searchStrRewritten + "/out/" + searchStrSplit[1]
 		} else {
-			exp.StatusPage(w, "search failed", "Transaction output index is not a valid integer: "+searchStrSplit[1],
+			exp.StatusPage(w, "search failed", "Transaction output index is not a valid non-negative integer: "+searchStrSplit[1],
 				"", ExpStatusNotFound)
 			return
 		}
--- a/explorer/explorerroutes.go
+++ b/explorer/explorerroutes.go
@@ -1631,8 +1631,27 @@ func (exp *explorerUI) Search(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


chappjc:
The contribution is much appreciated!  I will get you a review in the next day or two.

explorer/explorerroutes.go
```diff
@@ -1615,8 +1640,18 @@ func (exp *explorerUI) Search(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// Split searchStr to the first part corresponding to transaction hash and
+	// to the second part corresponding to transaction output.
+	searchStrSplit := strings.Split(searchStr, ":")
+	var searchStrRewritten string
+	if len(searchStrSplit) >= 2 {
```
chappjc: A pattern that I tend to prefer is to eliminate `else` branches if the variable initialization can handle the case.  That is,

```go
searchStrRewritten := searchStrSplit[0]
if len(searchStrSplit) > 1 {
	// Validate searchStrSplit[1]
	...
	searchStrRewritten = searchStrSplit[0] + "/out/" + searchStrSplit[1]
}
```

Although I would also want to throw an error if `len(searchStrSplit) > 2`, which would indicate an invalid path, so you might consider a switch on `len(searchStrSplit)` instead.

leRequinNoir:
Thank you for taking the time to review my proposed changes. I am committed to delivering a satisfactory solution. But before that let me ask for some clarification.

The idea behind the issue is to allow searching for a pattern formatted as `{hash}:{out}`. From the issue description I have had a feeling that `{out}` should be *transaction output address*. However your comments indicate that `{out}` should be *transaction output index*. The proposed changes implement the former interpretation of the issue.

Should `{out}` be *transaction output address* or *transaction output index*?

chappjc:
The output index.  Transaction *outpoints* are usually formatted like txid:out, and that is the out to which the endpoint /tx/{hash}/out/{out} refers.  An address does not uniquely specify which output or input.

leRequinNoir:
I hopefully implemented suggested changes. I started from scratch in order to maintain clean history.

leRequinNoir:
Thank you for spotting this edge case.

PR #1713, opened 2020-03-18
api,middlware: add Indent middleware

chappjc:

PR #1714, opened 2020-03-21
api: add address/{address}/exists endpoint

buck54321:
Provides access to the `existsaddresses` RPC call and parses the result to a `[]bool`. Allows multiple comma-separated addresses to be specified in the same way as the `addrs` insight endpoints. Limited to 64 addresses, mostly because it simplifies parsing and use of the RPC result.

The dcrd address existence index is enabled by default and seems to be quite fast. I did a little DB testing with the query

```
SELECT exists(select 1 from addresses where address = a) FROM unnest(ARRAY[...]) AS a;
```

and it seems to be roughly the same speed on my system and network.

### DIFF
--- a/api/apiroutes.go
+++ b/api/apiroutes.go
@@ -37,15 +37,9 @@ import (
 	appver "github.com/decred/dcrdata/v5/version"
 )
 
-const (
-	// maxBlockRangeCount is the maximum number of blocks that can be requested at
-	// once.
-	maxBlockRangeCount = 1000
-
-	// maxExistsAddrs is the highest number of addresses accepted to the
-	// address/{address}/exists request.
-	maxExistAddrs = 64
-)
+// maxBlockRangeCount is the maximum number of blocks that can be requested at
+// once.
+const maxBlockRangeCount = 1000
 
 // DataSource specifies an interface for advanced data collection using the
 // auxiliary DB (e.g. PostgreSQL).
@@ -1719,7 +1713,7 @@ func (c *appContext) addressTotals(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	addresses, err := m.GetAddressCtx(r, c.Params, 1)
+	addresses, err := m.GetAddressCtx(r, c.Params)
 	if err != nil || len(addresses) > 1 {
 		http.Error(w, http.StatusText(422), 422)
 		return
@@ -1751,7 +1745,7 @@ func (c *appContext) addressExists(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	addresses, err := m.GetAddressRawCtx(r, c.Params, maxExistAddrs)
+	addresses, err := m.GetAddressRawCtx(r, c.Params)
 	if err != nil {
 		apiLog.Errorf("addressExists rejecting request: %v", err)
 		http.Error(w, "address parsing error", http.StatusBadRequest)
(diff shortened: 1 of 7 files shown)


chappjc:
I prefer RPC these days anyway, so that's good.

api/apiroutes.go
```diff
@@ -1734,6 +1741,41 @@ func (c *appContext) addressTotals(w http.ResponseWriter, r *http.Request) {
...
+	b, err := hex.DecodeString(strMask)
+	if err != nil {
+		log.Warnf("existsaddress error: %v", err)
+		http.Error(w, http.StatusText(422), 422)
+	}
+	mask := binary.LittleEndian.Uint64(append(b, make([]byte, 8-len(b))...))
+	exists := make([]bool, 0, len(addresses))
+	for n := range addresses {
+		exists = append(exists, (mask&(1<<uint8(n))) != 0)
+	}
```
chappjc: Looks good.  My only concern is that `maxExistAddrs` cannot be more than 64 because of this, and there's no check on `maxExistAddrs`.  I worry about someone bumping up `maxExistAddrs` in the future without understanding it would lead to a panic in `make` here when >64 addresses get passed, especially because it might appear to work OK in many cases.

middleware/apimiddleware.go
```diff
@@ -414,21 +407,61 @@ func GetAddressCtx(r *http.Request, activeNetParams *chaincfg.Params, maxAddrs i
...
+		return nil, fmt.Errorf("address not set")
+	}
+	addressStrs := strings.Split(addressStr, ",")
+	if len(addressStrs) > maxAddrs {
+		return nil, fmt.Errorf("maximum of %d addresses allowed", maxAddrs)
+	}
+
+	return addressStrs, nil
+}
+
```
chappjc: Nice.  Good validation.  Although I'm thinking it could be even better to have this further up in the request handling chain, where `CtxAddress` is set (`middleware.AddressPathCtx`) to preempt the http handlers entirely if any of the addresses are invalid.  If this opens a can of worms or becomes a major overhaul of all the `/addr/...` endpoints, let's skip it though... just a thought.
It's basically the idea applied in the [new indent middleware](https://github.com/decred/dcrdata/pull/1713/files#diff-92898af3f3b4aa2e3e620614c71dbf9aR542-R551) - validate the url bits and call `next.ServerHTTP` otherwise `http.Error` so that no other middlewares, subrouters, or any parts of the routed handler actually have to run if validation fails.

api/apiroutes.go
buck54321: Should I just document `maxExistAddr` better?

chappjc: Yah, that's cool.  I think a note there not to go beyond 64 without modifying addressExists (or whatever) would be sufficient to prevent future breakage.

middleware/apimiddleware_test.go
```diff
@@ -81,13 +86,6 @@ func TestGetAddressCtx(t *testing.T) {
 			wantErr:  true,
 			errMsg:   "invalid address 'DcxxxxcGjmENx4DhNqDctW5wJCVyT3Qeqkx' for this network: checksum mismatch",
 		},
-		{
-			testName: "not_set",
-			args:     args{2, nil},
-			want:     nil,
-			wantErr:  true,
-			errMsg:   "address not set",
-		},
```
buck54321: This was not very useful because the URL reduced to "/address/", which would be a be a different handler so chi was just returning a 404.

PR #1717, opened 2020-03-30
Removed default values from url query string on attack cost page

ademuanthony:
Removed default values from url query string on attack cost page

### DIFF
--- a/public/js/controllers/attackcost_controller.js
+++ b/public/js/controllers/attackcost_controller.js
@@ -3,7 +3,6 @@ import TurboQuery from '../helpers/turbolinks_helper'
 import { getDefault } from '../helpers/module_helper'
 import globalEventBus from '../services/event_bus_service'
 import dompurify from 'dompurify'
-import _ from 'lodash-es'
 
 function digitformat (amount, decimalPlaces, noComma) {
   if (!amount) return 0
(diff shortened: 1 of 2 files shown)


public/js/controllers/attackcost_controller.js
```diff
@@ -3,6 +3,7 @@ import TurboQuery from '../helpers/turbolinks_helper'
 import { getDefault } from '../helpers/module_helper'
 import globalEventBus from '../services/event_bus_service'
 import dompurify from 'dompurify'
+import _ from 'lodash-es'
```
chappjc, line 6: The pattern seems to be to specify what to import (`{ find }`), but I leave this PR to @buck54321 for a proper js review.

public/js/controllers/attackcost_controller.js
```diff
@@ -236,6 +248,20 @@ export default class extends Controller {
...
+    const settingsEntries = Object.entries(this.settings)
+    const defaultSettingsEntries = Object.entries(this.defaultSettings)
+    for (const [key, val] of settingsEntries) {
+      let defaultValue = _.find(defaultSettingsEntries, (entry) => {
+        return entry[0] === key
+      })
+      if (val === null || val.toString() === defaultValue[1].toString()) continue
+      updatedSettings[key] = val
+    }
+    this.query.replace(updatedSettings)
```
buck54321, line 262:
```suggestion
const [query, settings, defaults] = [{}, this.settings, this.defaultSettings]
for (const k in settings) {
  if (!settings[k] || settings[k] === defaults[k]) continue
  query[k] = defaults[k] || null
}
this.query.replace(query)
```

public/js/controllers/attackcost_controller.js
buck54321, line 6: I have yet to find a case where I felt `lodash` was actually an improvement over vanilla js. I'm sure there are some, I just haven't seen one.

public/js/controllers/attackcost_controller.js
buck54321, line 262: How about

```
const [query, settings, defaults] = [{}, this.settings, this.defaultSettings]
for (const k in settings) {
  if (!settings[k] || settings[k] === defaults[k]) continue
  query[k] = settings[k]
}
this.query.replace(query)
```

You may need a string conversion for comparison, but ideally, the `this.settings` would be assigned with the correct type, e.g. using `parseInt` anyway.

PR #1718, opened 2020-04-02
Right-align numeric columns

### LINKED ISSUE #1683: Right-align numeric columns
Not sure if this is a regression, but in any case, right alignment broke on /blocks and Latest Mainnet Blocks on the root page. I didn't check all pages.

align-blocks

align-root-blocks

ademuanthony:
Fix #1683 
This PR right-align numeric columns on all the pages

### DIFF
--- a/views/agendas.tmpl
+++ b/views/agendas.tmpl
@@ -242,7 +242,7 @@
                   <tr>
                     <th>Agenda ID</th>
                     <th class="d-none d-md-table-cell">Description</th>
-                    <th class="text-right">Status</th>
+                    <th class="text-left">Status</th>
                     <th class="text-right">Vote Version</th>
                     <th class="text-right d-none d-sm-table-cell">Start Time (UTC)</th>
                     <th class="text-right d-none d-sm-table-cell">Expire Time (UTC)</th>
@@ -253,7 +253,7 @@
                 <tbody>
                     <td><a href="/agenda/{{.ID}}">{{.ID}}</a></td>
                     <td class="d-none d-md-table-cell truncate">{{.Description}}</td>
-                    <td class="text-right">{{.Status}}</td>
+                    <td class="text-left">{{.Status}}</td>
                     <td class="text-right">{{.VoteVersion}}</td>
                     <td class="text-right d-none d-sm-table-cell">{{dateTimeWithoutTimeZone .StartTime}}</td>
                     <td class="text-right d-none d-sm-table-cell">{{dateTimeWithoutTimeZone .ExpireTime}}</td>
--- a/views/block.tmpl
+++ b/views/block.tmpl
@@ -205,7 +205,7 @@
 				<tr>
 					<th>Transactions ID</th>
 					<th class="text-right">Vote Version</th>
-					<th class="text-right">Last Block</th>
+					<th class="text-left">Last Block</th>
 					<th class="text-right">Total DCR</th>
 					<th class="text-right">Size</th>
 				</tr>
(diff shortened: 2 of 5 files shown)


views/agendas.tmpl
```diff
@@ -253,7 +253,7 @@
                 <tbody>
                     <td><a href="/agenda/{{.ID}}">{{.ID}}</a></td>
                     <td class="d-none d-md-table-cell truncate">{{.Description}}</td>
-                    <td class="text-right">{{.Status}}</td>
```
chappjc, line 256: This one is non-numeric.  Just looks better?

views/block.tmpl
```diff
@@ -205,7 +205,7 @@
 				<tr>
 					<th>Transactions ID</th>
 					<th class="text-right">Vote Version</th>
-					<th class="text-right">Last Block</th>
+					<th class="text-left">Last Block</th>
```
chappjc, line 208: Left align of this column doesn't look right.

Before:

image

This PR:

image

views/explorer.tmpl
```diff
@@ -84,26 +84,26 @@
...
                             <span class="d-md-none position-relative" data-tooltip="revocations">X</span>
                         </th>
-                        <th class="text-center d-table-cell d-sm-none">Txns</th>
-                        <th class="text-center">DCR</th>
-                        <th class="text-center">Size</th>
-                        <th class="text-center d-none d-sm-table-cell">
+                        <th class="text-right d-table-cell d-sm-none">Txns</th>
+                        <th class="text-right">DCR</th>
+                        <th class="text-right">Size</th>
+                        <th class="text-right d-none d-sm-table-cell">
```
chappjc, line 106: I'm not sure this is better either.

Before:

image

This PR:

image

Note the "DCR" and "Size" header positions compared to the data cells below them.  "Version" is similarly off a bit.  And this seems to create a lot of empty space in the "Transactions" column that should be dealt with appropriately.

I might be OK with right aligning the numberic data **if** the column headers remain center aligned.

views/home.tmpl
```diff
@@ -79,10 +79,10 @@
...
                                     <tr data-height="{{.Height}}" data-link-class="fs18">
                                         <td class="text-left pl-1" data-type="height"><a href="/block/{{.Height}}" class="fs18">{{.Height}}</a></td>
-                                        <td class="text-center" data-type="tx">{{.Transactions}}</td>
-                                        <td class="text-center" data-type="votes">{{.Voters}}</td>
-                                        <td class="text-center" data-type="tickets">{{.FreshStake}}</td>
-                                        <td class="text-center" data-type="revocations">{{.Revocations}}</td>
+                                        <td class="text-right" data-type="tx">{{.Transactions}}</td>
+                                        <td class="text-right" data-type="votes">{{.Voters}}</td>
+                                        <td class="text-right" data-type="tickets">{{.FreshStake}}</td>
+                                        <td class="text-right" data-type="revocations">{{.Revocations}}</td>
```
chappjc, line 85: The home page looks OK, however I'm not sure it's clearly better.  Also note the large new white space in the Transactions column here too.

Before:

image


This PR:

image

buck54321: I agree. I decided against right aligning those columns simply because it looks terrible.

buck54321:
I'm not certain where this blanket rule for right-aligning columns came from, but the rule for dcrdata is that we're going to align the columns in whatever way looks better. This does not look better, imo.

views/agendas.tmpl
ademuanthony: Yes, IMO

xaur:
@buck54321 it came from me :) I promote right-aligning numeric columns everywhere because it is easier to read tables with numbers. It allows our visual pattern matching ASICs to quickly spot numbers that have different number of digits.

I cannot imagine too many cases when centrally-aligned numbers look better. Perhaps some very small (few rows) "preview" tables where numbers are big?

> I might be OK with right aligning the numberic data if the column headers remain center aligned.

Agreed, right-aligning should not affect column headers.

xaur:
> Also note the large new white space in the Transactions column here too.

That indeed looks bad, but the question is why the Transactions column is so wide, far beyond its column header text width?

PR #1719, opened 2020-04-03
Address page pagination rows fix and comma formatting

mkingori:
Pagination selector on address page was not changing the number of rows displayed. The param being passed in was not properly being used. The function was hard coded to 20 rows. This has been resolved.

Additionally, changing the transaction type selector caused the total transaction count to lose comma formatting. This has also been fixed.

Also added pagination page number links.

Resolves https://github.com/decred/dcrdata/pull/1719

### DIFF
--- a/explorer/explorerroutes.go
+++ b/explorer/explorerroutes.go
@@ -1270,6 +1270,7 @@ func (exp *explorerUI) AddressPage(w http.ResponseWriter, r *http.Request) {
 		Data         *dbtypes.AddressInfo
 		CRLFDownload bool
 		FiatBalance  *exchanges.Conversion
+		Pages        []pageNumber
 	}
 
 	// Grab the URL query parameters
@@ -1351,12 +1352,19 @@ func (exp *explorerUI) AddressPage(w http.ResponseWriter, r *http.Request) {
 	// endings.
 	UseCRLF := strings.Contains(r.UserAgent(), "Windows")
 
+	linkTemplate := "/address/" + addrData.Address + "?start=%d&n=" + strconv.FormatInt(limitN, 10) + "&txntype=" + fmt.Sprintf("%v", txnType)
+
+	if limitN == 0 {
+		limitN = 20
+	}
+
 	// Execute the HTML template.
 	pageData := AddressPageData{
 		CommonPageData: exp.commonData(r),
 		Data:           addrData,
 		CRLFDownload:   UseCRLF,
 		FiatBalance:    conversion,
+		Pages:          calcPages(int(addrData.TxnCount), int(limitN), int(offsetAddrOuts), linkTemplate),
 	}
 	str, err := exp.templates.exec("address", pageData)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


public/js/controllers/address_controller.js
```diff
@@ -52,6 +52,10 @@ function amountFlowProcessor (d, binSize) {
   }
 }
 
+function numberWithCommas (x) {
+  return x.toString().replace(/\B(?=(\d{3})+(?!\d))/g, ',')
+}
```
chappjc, line 57: I figured there would be something in `public/js/helpers/humanize_helper.js`, but there isn't.  But as much as I respect the regexp you've got here, I think a more robust way would be with [`toLocaleString()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NumberFormat/NumberFormat).

public/js/controllers/address_controller.js
```diff
@@ -380,6 +387,44 @@ export default class extends Controller {
...
+        continue
+      }
+
+      links = links + '\n' + link
+    }
+
+    if ((txCount - offset) > pageSize) {
+      link = '<a href="/address/' + this.dcrAddress + '?start=' + (offset + pageSize) + '&n=' + pageSize + '" class="d-inline-block dcricon-arrow-right m-1 fs20"></a>'
+      links = links + '\n' + link
+    }
```
buck54321, line 424:
```suggestion
var links = tablePagesLink.map(d => {
  if (!d.link) return `<span>${d.str}</span>`
  return `<a href="${d.link}" class="fs18 pager px-1${d.active ? ' active' : ''}">${d.str}</a>`
}).join('\n')
if ((txCount - offset) > pageSize) {
  links += '\n' + `<a href="/address/${this.dcrAddress}?start=${(offset + pageSize)}&n=${pageSize}" class="d-inline-block dcricon-arrow-right m-1 fs20"></a>`
}
```

explorer/explorerroutes.go
```diff
@@ -1351,12 +1352,19 @@ func (exp *explorerUI) AddressPage(w http.ResponseWriter, r *http.Request) {
 	// endings.
 	UseCRLF := strings.Contains(r.UserAgent(), "Windows")
 
+	linkTemplate := "/address/" + addrData.Address + "?start=%d&n=" + strconv.FormatInt(limitN, 10) + "&txntype=" + fmt.Sprintf("%v", txnType)
```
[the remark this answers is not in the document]
mkingori: Cool

public/js/controllers/address_controller.js
mkingori: Fixed.

mkingori:
Done working on the requested changes.

public/js/controllers/address_controller.js
```diff
@@ -380,6 +394,32 @@ export default class extends Controller {
...
+      if (!d.link) return `<span>${d.str}</span>`
+      return `<a href="${d.link}" class="fs18 pager px-1${d.active ? ' active' : ''}" data-action="click->address#pageNumberLink">${d.str}</a>`
+    }).join('\n')
+
+    if ((txCount - offset) > pageSize) {
+      links += '\n' + `<a href="/address/${this.dcrAddress}?start=${(offset + pageSize)}&n=${pageSize}&txntype=${txnType}" ` +
+      `class="d-inline-block dcricon-arrow-right m-1 fs20" data-action="click->address#pageNumberLink"></a>`
+    }
+
+    ctrl.tablePaginationTarget.innerHTML = links
```
buck54321, line 420: Any time we use innerHTML with parameters that might ultimately be set by the user, we should use dompurify. 

```
ctrl.tablePaginationTarget.innerHTML = dompurify.sanitize(links)
```

mkingori: Got it. Its fixed now.

Issue #1583, opened 2019-10-17
Abandoned status incorrect on /proposals


buck54321:
A proposal that is abandoned is marked as "In Discussion" on /proposals.

![image](https://user-images.githubusercontent.com/6109680/67050269-c12f3780-f0fd-11e9-8010-5e56e118ffe9.png)

The proposal is marked appropriately on /proposal

![image](https://user-images.githubusercontent.com/6109680/67050328-df953300-f0fd-11e9-9d3b-2fb0234680f2.png)

xaur:
Still seeing this for https://explorer.dcrdata.org/proposal/amin-rafiee-decred-hackers-congress - the page shows Abandoned correctly but /proposals still shows In Discussion.

xaur:
This affects 3 proposals. On the pic below, only first 2 items are really In Discussion

![proposals-discussion](https://user-images.githubusercontent.com/24484727/73950823-7f1c4280-48f4-11ea-8959-70ad6df03e13.png)

mkingori:
Seems its resolved already.

chappjc:
Confirmed.

xaur:
@chappjc I still experience it on both prod (dcrdata v5.2.2-pre+dev) and alpha (dcrdata v5.3.0-pre+dev).

![polis-prop](https://user-images.githubusercontent.com/24484727/79485472-d598b200-8004-11ea-84ed-282f6738fde3.png)

![polis-props](https://user-images.githubusercontent.com/24484727/79485481-d92c3900-8004-11ea-84da-fa9f808e0fd8.png)

mkingori:
@chappjc I've worked on a fix for proposals showing `In discussion` instead of `Abandoned`

PR #1732, opened 2020-04-19
Fixed vote status for proposals showing in discussion instead of abandoned

mkingori:
Fixes [#1583](https://github.com/decred/dcrdata/issues/1583)

Added a fix for proposals showing `In discussion` instead of `Abandoned`.

### DIFF
--- a/views/proposals.tmpl
+++ b/views/proposals.tmpl
@@ -145,9 +145,15 @@
                         <td class="text-left">
                             {{if .VoteStatus}}
                                 {{if eq .VoteStatus.ShortDesc "Not Authorized"}}
-                                    <span class="text-abandoned position-relative" data-tooltip="Proposal in discussion">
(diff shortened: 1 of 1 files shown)


chappjc:
Thanks for the fix.  In general it is valuable to also give some background on the etiology of the issue.  i.e. Abandon would only show if `.VoteStatus` were nil, meaning...

https://github.com/decred/dcrdata/blob/a1338774619eea0f908389c97f748656ad3a3961/views/proposals.tmpl#L146

mkingori:
An abandoned proposal is expected to have a VoteStatus of 0, but there are actually abandoned proposals with a VoteStatus of 1 which indicates that they are still in discussion. In order to categorize a proposal to either In Discussion or Abandoned, I check if the AbandonedDate is set. If AbandonedDate is not set, it is still in discussion.

Issue #1725, opened 2020-04-14
rework modules and package deps for 6.0


chappjc:
This is not starting yet, but I wanted to make some notes on how we will eliminate several modules and reduce some tight couplings.

eliminate modules:
- gov
- middleware
- db submodules (cache and dcrpg, maybe dbtypes), replace with single db module
- blockdata
- stakedb
- testutil/dbconfig
- dcrrates
- exchanges

remaining:
- psclient (new, remove pubsub dep: ping interval and Version)
- pubsub/types
- explorer/types (replace *exchanges.Conversion with an interface)
- api/types (rework TimeAPI, **remove dbtypes dep**, move TxOutFromDB to caller pkg, etc.)
- txhelpers
- rpcutils

merge dcrrates and exchanges?

PR #1733, opened 2020-04-19
[poc] explorer/api: label swap contract/redemption transactions

itswisdomagain:
Display `Swap Contract` / `Swap Redemption` / `Swap Refund` label for atomic swap transactions on the explorer tx page, with some extra details about the completed swap.

Screenshot 2020-04-19 at 11 01 53 PM
Screenshot 2020-04-19 at 11 02 18 PM

Also added a `/tx/swap/{txid}` API endpoint for fetching data about a swap redemption transaction. This currently only accepts a redemption txid but will be modified to accept a swap contract txid as well.

Sample data returned from the api for a redeemption tx currently:
```
{
   "txid": "4a7948f1eaa589b3e7bb655f44d1ee4daceda7220154ec677e45df708c2e44be",
   "contract_tx_id": "4a14a2d79c1374d286ebd68d2c104343bcf8be44ed54045b5963fbf73667cecc",
   "contract_address": "TchUiUNfb96ji35N767BNesdKTe88KP35nf",
   "contract": "6382012088c020c6de3217594af525fb57eaf1f2aae04c305ddc67d465edd325151685fc5a5e428876a914479eddda81b6ed289515f2dbcc95f05ce80dff466704fe03865eb17576a91498a67ed502adb04173d88fb1ef92d0317711c3816888ac",
   "recipient_address": "TsXYpmRekVsBt8oxS3cSUck4AMEW9aRc2X7",
   "refund_address": "TsewGe2LsyPuX14ECtW9mopW4yS2NqCQ4T6",
   "locktime": "2020-04-02 15:25:50 +0000 UTC",
   "is_refund": false,
   "redeem_address": "TsRgDUySkyq5Miw9axWkCm3SSXdLsCYEKf8"
}
```

**Some Notes**
- One condition for correctly labelling `Swap Contract` transactions is that the contract has to have been redeemed, because from what I gather, it is not possible to definitely identify a swap contract unless it has been redeemed.
- This is a proof-of-concept, would love feedback on if there are other swap-related information that should be displayed on the tx page or returned by the API.
- Immediate next step would be modifying the implementation of the added API endpoint to return data for swap contract txids as well.


db/dcrpg/pgblockchain.go
```diff
@@ -4885,6 +4887,151 @@ func (pgb *ChainDB) GetTrimmedTransaction(txid *chainhash.Hash) *apitypes.Trimme
...
+	return nil, nil
+}
+
+// GetAtomicSwapInfoFromRedemptionTx gets a *dbtypes.AtomicSwapInfo for a given
+// transaction ID, if the transaction exists and represents an atomic swap
+// redemption/refund.
+// Returns nil if the passed txid does not represent a swap redemption/refund.
+// NOTE: both return values may be nil if the txid is a valid transaction but
+// does not represent an atomic swap redemption/refund.
+func (pgb *ChainDB) GetAtomicSwapInfoFromRedemptionTx(txid string) (*dbtypes.AtomicSwapInfo, error) {
```
chappjc, line 4927: The meat of `GetAtomicSwapInfoFromRedemptionTx` really does not need to be a `ChainDB` method since it's only using `getRawAPITransaction` and that just uses `rpcutils.APITransaction`.  This method can still exist, perhaps just wrapping an `rpcutils` function of the same name.  Having it in `rpcutils` would make it more broadly accessible, and more easily testable.

db/dbtypes/types.go
```diff
@@ -1649,6 +1649,22 @@ type Tx struct {
...
+	ContractValue    float64 `json:"contract_value"`
+	FormattedAmount  string  `json:"formatted_amount"`
+	RecipientAddress string  `json:"recipient_address"`
+	RefundAddress    string  `json:"refund_address"`
+	Locktime         string  `json:"locktime"`
+
+	RedemptionTxID      string   `json:"redemption_tx_id"`
+	RedeemedToAddresses []string `json:"redeemed_to_addresses"`
+	IsRefundToInitiator bool     `json:"is_refund_to_initiator"`
+}
```
chappjc, line 1666: I'd prefer this in `api/types` since the data isn't coming from or being used by the DB, but also I want to limit the use of `dbtypes` to other `db/...` packages as much as possible in the future.

api/apiroutes.go
```diff
@@ -743,6 +744,24 @@ func (c *appContext) getDecodedTx(w http.ResponseWriter, r *http.Request) {
...
 }
 
+func (c *appContext) getTxSwapInfo(w http.ResponseWriter, r *http.Request) {
+	txid, err := m.GetTxIDCtx(r)
+	if err != nil {
+		http.Error(w, http.StatusText(422), 422)
+		return
+	}
+
+	tx, err := c.DataSource.GetAtomicSwapInfo(txid.String())
```
chappjc, line 754: `appContext` has it's own node client, so it could call the hypothetical rpcutils function instead.

explorer/explorerroutes.go
```diff
@@ -1101,6 +1101,22 @@ func (exp *explorerUI) TxPage(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
+	// attempt to read swap info assuming this is a swap redemption tx
+	var swapInfo *dbtypes.AtomicSwapInfo
+	swapInfo, err = exp.dataSource.GetAtomicSwapInfoFromRedemptionTx(tx.TxID)
```
chappjc, line 1106: Unlike `appContext`, `explorerUI` doesn't have a node client, so it would have to use the dataSource wrapper.

views/tx.tmpl
```diff
@@ -279,6 +290,59 @@
...
+                    {{template "hashElide" (hashlink $.SwapInfo.ContractTxID (print "/tx/" $.SwapInfo.ContractTxID))}}
+                  </td>
+                </tr>
+                {{end}}
+                <tr>
+                  <td class="text-right medium-sans text-nowrap pr-2 py-2">
+                    Redeemed By:
+                  </td>
+                  <td class="text-left py-1 text-secondary">
+                    {{if $.SwapInfo.IsRefundToInitiator}}Initiator{{else}}Recipient{{end}}
```
chappjc, line 320: Initiator and *Participant* are probably more accepted terms here.

db/dcrpg/pgblockchain.go
chappjc: But I'd eliminate the `Get` prefix (we shouldn't have such names although we have many) and go with `AtomicSwapInfoFromRedemption` or similar.

itswisdomagain: Had similar thoughts re the method name but just wanted to follow the pattern I observed in the existing code.

api/apiroutes.go
chappjc: Oh, well this one is still DataSource I guess since it needs SpendingInfo, or it could follow that logic her.e

db/dbtypes/types.go
itswisdomagain: This is done but it feels wrong having `explorer` reference `apitypes`.

explorer/explorerroutes.go
itswisdomagain: `GetAtomicSwapInfoFromRedemptionTx`, unlike `GetAtomicSwapInfo` (used in `appContext`) does not read from the db at all, so instead of keeping the dataSource method (even though it would only be a wrapper around the rpcutils function), I added a `dataSource.NodeClient()` method similar to the `dataSource.GetChainParams()` method to give `explorerUI` access to the node client.

I believe we can even do away with both `dataSource.GetChainParams()` and `dataSource.NodeClient()` methods as
- neither function performs any db work AND
- `explorerUI` can get those objects directly when initialized from `ExplorerConfig` rather than from `dataSource`; as is similarly done for `appContext` which is currently how `appContext` gets its node client.

db/dbtypes/types.go
chappjc: I understand the feeling, but it's definitely good for dbtypes.  Both `explorer` and `api` need the `AtomicSwapInfo` (now `AtomicSwap`) type.  I'm open to other suggestions, but I believe an effort in 6.0 will be to get down to maybe 1 types package.  ref: https://github.com/decred/dcrdata/issues/1725

chappjc: oops, meant to say **not** good for dbtypes

api/apiroutes.go
```diff
@@ -743,6 +744,73 @@ func (c *appContext) getDecodedTx(w http.ResponseWriter, r *http.Request) {
...
+		done(err)
+		return
+	}
+
+	// It is theoretically possible for a contract tx to have multiple contract
+	// outputs, so check all txs that spend from this tx's outputs and store all
+	// completed swaps returned.
+	allCompletedSwaps := make([]*apitypes.AtomicSwap, 0)
+	for _, spendingTxID := range spendingTxIDs {
+		swaps, err = rpcutils.AtomicSwapsRedeemedInTx(c.nodeClient, spendingTxID, c.Params)
```
buck54321, line 793: You're requesting a `TxRawResult` for every spent output over RPC, even if it's not a p2sh script. Since you're already requesting the raw transaction data for `txid` in `AtomicSwapsRedeemedInTx`, maybe keep a hold of that and check that the output pubkey script is p2sh first.

explorer/explorerroutes.go
```diff
@@ -1101,6 +1103,41 @@ func (exp *explorerUI) TxPage(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
+	// Check if tx is an atomic swap redemption tx or swap contract tx.
+	var swapType string
+	var atomicSwaps []*apitypes.AtomicSwap
+
+	// First check if this tx redeems 1 or more atomic swap contracts.
+	atomicSwaps, err = rpcutils.AtomicSwapsRedeemedInTx(exp.nodeClient, hash, exp.ChainParams)
```
buck54321, line 1111: We could potentially skip this if 1) this is a stake tree transaction, or 2a) there are no p2sh outputs and 2b) there are no inputs with the correct sigscript type. If not now, then add a `TODO` here.

views/tx.tmpl
```diff
@@ -41,6 +41,9 @@
               <div class="d-inline-block mix-box mx-2 fs14">Mix: {{.MixCount}} outputs of {{template "decimalParts" (amountAsDecimalParts .MixDenom false)}} DCR
               </div>
               {{- end}}
+              {{if gt (len $.Swaps) 0}}
+              <div class="d-inline-block swap-box mx-2 fs14">Swap {{$.AtomicSwapTxType}}</div>
+              {{end}}
```
buck54321, line 46: How about separate labels: **Swap**, **Redemption**, and **Refund**. More than one could possibly be displayed, though it wouldn't be common. The shorter words = smaller labels wouldn't wrap there either, which would look better. Here it is at a common mobile width. 

image

views/tx.tmpl
```diff
@@ -279,6 +282,59 @@
...
+                  </td>
+                  <td class="text-right medium-sans text-nowrap pr-2 py-2">
+                    Lock Time:
+                  </td>
+                  <td class="text-left py-1 text-secondary clipboard">
+                    {{.Locktime}}
+                  </td>
+                </tr>
+              {{end}}
             </tbody>
```
buck54321, line 338: Since there can and will be multiple redemptions or swaps in a single transaction, the best place to display this info is in the input and output tables. 

In the output table, maybe use the `Type` column to say **swap** when the output is a swap output.

For the inputs, it's a little harder, because we are already tight on space in that table. I would be ok with not displaying the `txid:vout` directly for redemptions and refunds, and instead just displaying a link with the text **redemption** or **refund**. The txid and vout detail would still be visible on hover, and could also be put in a tooltip. Thoughts @chappjc?

buck54321: Tables something like

image
 
Though now that I'm looking at it, "Swap Refund" doesn't really work there. Just **Swap** would be better.

api/apiroutes.go
```diff
@@ -743,6 +744,73 @@ func (c *appContext) getDecodedTx(w http.ResponseWriter, r *http.Request) {
...
+
+	swapTx := &apitypes.AtomicSwapTx{
+		TxID: txid.String(),
+	}
+
+	done := func(err error) {
+		if err != nil {
+			var errString = fmt.Sprintf("Unable to get atomic swap info for transaction %v: %v", txid, err)
+			apiLog.Errorf(errString)
+			http.Error(w, errString, 422)
```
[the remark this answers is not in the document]
chappjc: I agree the error shown to the user should be more generic here.

views/tx.tmpl
itswisdomagain: Re the inputs table:
Since the **Previous Outpoint** column links to the output that was spent, shouldn't this be `Contract` or `Swap Contract`? The spent output for a redeem/refund tx input would be a contract output, not a refund or redemption...

api/types/apitypes.go
```diff
@@ -185,28 +185,29 @@ type TxInputID struct {
...
-type AtomicSwapTx struct {
-	TxID  string        `json:"tx_id"`
-	Type  string        `json:"type"`
-	Swaps []*AtomicSwap `json:"swaps"`
+// TxAtomicSwaps defines information about completed atomic swaps that are
+// related to a transaction.
+type TxAtomicSwaps struct {
+	TxID        string                 `json:"tx_id"`
+	Found       string                 `json:"found"`
+	Contracts   map[uint32]*AtomicSwap `json:"contracts,omitempty"`
```
itswisdomagain, line 193: Named this `Contracts` instead of `Swaps` because this property contains references to txouts that define atomic swap contracts, just as `Redemptions` and `Refunds` link txins to contract redeems/refunds.

explorer/explorerroutes.go
```diff
@@ -1238,6 +1202,23 @@ func (exp *explorerUI) TxPage(w http.ResponseWriter, r *http.Request) {
...
+	if tx.Type == txhelpers.TxTypeRegular && !tx.Coinbase {
+		// use following spendingTxsFn to return the already fetched data for
+		// spent outputs instead of re-hitting the database.
+		spendingTxsFn := func(txid string) ([]string, []uint32, []uint32, error) {
+			if txid != tx.TxID {
+				return nil, nil, nil, fmt.Errorf("unexpected txid")
+			}
+			return spendingTxHashes, spendingTxVinInds, voutInds, nil
+		}
+		swapsInfo, err = rpcutils.TxAtomicSwapsInfo(exp.nodeClient, exp.ChainParams, tx.TxID, spendingTxsFn)
```
itswisdomagain, line 1216: An alternative to passing a `spendingTxsFn` is to pass the already fetched spending info with the call to `rpcutils.TxAtomicSwapsInfo` but that would require modifying the api `getTxSwaps` method to also fetch spending info before calling the rpcutils function. As mentioned earlier, I hope to prevent calls to the db to fetch spending info when there isn't a need to do so.

txhelpers/swap.go
```diff
@@ -0,0 +1,125 @@
...
+	AtomicSwapInitiator     string = "Initiator"
+	AtomicSwapParticipant   string = "Participant"
+)
+
+// ParseInputScriptForSwapContract checks if a tx input redeems a swap contract
+// and returns the contract that was redeemed, along with a string describing the
+// identity of the redeemer.
+// Returns an empty contract script and nil error if the provided script does not
+// redeem a contract. Returns a non-nil error if the script could not be parsed.
+func ParseInputScriptForSwapContract(inputScriptHex string, params *chaincfg.Params) ([]byte, string, error) {
```
itswisdomagain, line 37: Defined `ParseInputScriptForSwapContract` and `ParseAtomicSwapContract` here to make them usable outside dcrdata.

views/tx.tmpl
```diff
@@ -248,6 +248,16 @@
...
+              {{if $.SwapsFound}}
+                <tr>
+                  <td class="text-right medium-sans text-nowrap pr-2 py-2"
+                    >Swaps Info:
+                  </td>
+                  <td class="text-left py-1 text-secondary"><a
+                    href="/api/tx/swaps/{{.TxID}}?indent=true" data-turbolinks="false">swaps</a>
+                  </td>
+                </tr>
+              {{end}}
```
itswisdomagain, line 260: Since we are no longer able to show details of swaps related to a tx, I added this link to the api endpoint that can display that info.

main.go
```diff
@@ -494,6 +494,7 @@ func _main(ctx context.Context) error {
 
 	// Create the explorer system.
 	explore := explorer.New(&explorer.ExplorerConfig{
+		Client:          dcrdClient,
```
buck54321, line 497: I think this is a clue that `TxAtomicSwapsInfo` function should be routed through a method on the `ChainDB`, which already has a copy of the `rpcclient.Client` and the `chaincfg.Params`. It also fits the prevailing model of `Explorer` and the `explorerDataSource` interface.

itswisdomagain: `TxAtomicSwapsInfo` was a method on the `ChainDB`, but since the method doesn't _really_ perform a db query, @chappjc and I felt it was better to make it a `rpcutils` function, as the method _really_ only need to make rpc calls. Related conversation: https://github.com/decred/dcrdata/pull/1733#discussion_r411545218 and https://github.com/decred/dcrdata/pull/1733#discussion_r411548496.

I italicised _really_ because there is one area where a db call is necessary, when calling `TxAtomicSwapsInfo` from the api route handler. The db call is not necessary when accessing the tx explorer page, so explorer really does not need to make db calls from the `TxAtomicSwapsInfo` method.

Re the current `Explorer` / `explorerDataSource` model, we've discussed making some changes there in [chat](https://matrix.to/#/!LmKzrmxJIXNHNiEmIh:decred.org/$158740395316769DahOV:decred.org?via=decred.org&via=matrix.org&via=decred.org), to not mix rpc calls and db calls in one db pkg.

rpcutils/rpcclient.go
```diff
@@ -631,3 +633,185 @@ func APITransaction(client txhelpers.VerboseTransactionGetter, txid *chainhash.H
...
+		}
+	}
+	return contracts, nil
+}
+
+// checkTxInputForSwapInfo parses the scriptsig of the provided transaction input
+// for information about a completed atomic swap.
+// Returns (nil, nil) if the scriptsig of the provided txin does not redeem a
+// swap contract.
+func checkTxInputForSwapInfo(txid string, inputIndex uint32, input chainjson.Vin, params *chaincfg.Params) (*apitypes.AtomicSwap, error) {
```
chappjc, line 790: This would be a better fit for `txhelpers`.  One of the functions I'm alluding to in my above review comment.

rpcutils/rpcclient.go
```diff
@@ -631,3 +633,185 @@ func APITransaction(client txhelpers.VerboseTransactionGetter, txid *chainhash.H
...
+
+	tx := &apitypes.TxAtomicSwaps{
+		TxID:        txid,
+		Contracts:   make(map[uint32]*apitypes.AtomicSwap),
+		Redemptions: make(map[uint32]*apitypes.AtomicSwap),
+		Refunds:     make(map[uint32]*apitypes.AtomicSwap),
+	}
+
+	// Check if tx is stake tree tx or coinbase tx and return empty swap info.
+	for _, input := range txraw.Vin {
```
dnldd, line 684: Moving this check before `tx`'s initialization would save some allocations since you'd return nil if the tx happens to be a coinbase or a stakebase.

txhelpers/swap.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+	if err := tokenizer.Err(); err != nil {
+		return nil, "", fmt.Errorf("error parsing input script: %v", err)
+	}
+
+	if contract == nil || !tokenizer.Done() {
+		// script should contain contract as the last data
+		// if contract has been extracted, tokenizer.Done() should be true
+		return nil, "", nil
+	}
```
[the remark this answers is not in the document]
itswisdomagain: Validating the contract here would require parsing it, which is done later, separately. I'll modify this function to also parse the script to ensure that it is actually a swap contract.

itswisdomagain:
@buck54321, want to confirm that the following is the UI changes you recommended before updating the PR description with the new screenshots:

#### Contract tx
Screenshot 2020-12-30 at 13 35 19

#### Redemption tx
Screenshot 2020-12-30 at 13 36 30

#### Refund tx
Screenshot 2020-12-30 at 13 42 10

chappjc:
I rebased and squashed this after the mega code reorganization was merged in https://github.com/decred/dcrdata/pull/1796, which needed to get in to begin preparing for treasury changes.

I'm going to merge this shortly so we can start evaluating it in the wild on alpha.  It is working fine for me on my dev machine.

chappjc (approved):

PR #1735, opened 2020-04-22
explorer: display alt. block hashes on /block page

itswisdomagain:
Display (link to) other block hashes if two or more blocks are found for a given height. Helps to provide visual feedback when there are multiple blocks for a given height.

Screenshot 2020-04-22 at 3 26 05 AM
Screenshot 2020-04-22 at 3 26 14 AM

### DIFF
--- a/db/dcrpg/queries.go
+++ b/db/dcrpg/queries.go
@@ -4133,11 +4133,11 @@ func RetrieveBlockStatuses(ctx context.Context, db *sql.DB, idx int64) (blocks [
 
 	for rows.Next() {
 		var bs dbtypes.BlockStatus
-		err = rows.Scan(&bs.IsValid, &bs.IsMainchain, &bs.Height, &bs.Hash)
+		err = rows.Scan(&bs.IsValid, &bs.IsMainchain, &bs.Hash)
 		if err != nil {
 			return
 		}
-
+		bs.Height = uint32(idx)
 		blocks = append(blocks, &bs)
 	}
 	err = rows.Err()
--- a/db/dcrpg/queries.go
+++ b/db/dcrpg/queries.go
@@ -4121,6 +4121,30 @@ func RetrieveBlockStatus(ctx context.Context, db *sql.DB, hash string) (bs dbtyp
 	return
 }
 
+// RetrieveBlockStatuses retrieves the block chain statuses of all blocks at
+// the given height.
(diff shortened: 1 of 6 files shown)


db/dcrpg/queries.go
```diff
@@ -4121,6 +4121,30 @@ func RetrieveBlockStatus(ctx context.Context, db *sql.DB, hash string) (bs dbtyp
...
+	var rows *sql.Rows
+	rows, err = db.QueryContext(ctx, internal.SelectBlockStatuses, idx)
+	if err != nil {
+		return
+	}
+	defer closeRows(rows)
+
+	for rows.Next() {
+		var bs dbtypes.BlockStatus
+		err = rows.Scan(&bs.IsValid, &bs.IsMainchain, &bs.Height, &bs.Hash)
```
chappjc, line 4136: Since height from `SelectBlockStatuses` is necessarily equal to `idx`, I'd leave that out of the query and just set `bs.Height = idx` after a error-free `Scan`.

views/block.tmpl
```diff
@@ -55,6 +55,18 @@
...
+				<div class="fs13 text-secondary pb-1">
+					Alt. Block (on {{if .IsMainchain}}main chain{{else}}side chain{{end}})
+				</div>
+				<div class="d-inline-block fs14 break-word rounded font-weight-bold">
+					<a href="/block/{{.Hash}}" title="Valid: {{.IsValid}}">
+						{{.Hash}}
+					</a>
+				</div>
+			</div>
+			{{end}}
```
chappjc, line 69: This obviously makes the "Block #" parent div taller, but I'm fine with that since this is rare.  What would be bad is if there are several alt blocks, which happens occasionally on testnet.

Maybe we could make it like this to be more compact in the case of more than one alt block:

image

views/block.tmpl
```diff
@@ -55,6 +55,19 @@
 				<div class="fs13 text-secondary pb-1">Block Hash</div>
 				<div class="d-inline-block fs14 break-word rounded font-weight-bold">{{.Hash}}</div>
 			</div>
+			{{if gt (len $.AltBlocks) 0}}
+			<div class="text-left lh1rem py-1">
+				<div class="fs13 text-secondary pb-1">
+					Alternative Block{{if gt (len $.AltBlocks) 1}}s{{end}}
+				</div>
+				{{range $.AltBlocks}}
```
dnldd, line 63: This could make the view grow in height quite a bit since `AltBlocks` is unbounded when there are a lot of valid side chains. Perhaps the range should be populating a drop down or a collapsible view which by default shows one entry?

PR #1738, opened 2020-05-01
Exchange rate prediction

ademuanthony:
This PR predicts the price increase in DCR that will result from any attempt to attack the network.
image

The predicted price is obtained by using the size and value (total ask price) of DCR offered on all exchange to get the increase in price per unit increase in demand which is in turn multiplied by the total amount of DCR needed for the attack.
The increase rate is calculated using the cumulative growth rate formula
image

The total cost of the DCR acquired for the attack is gotten by accumulating the cost of getting each DCR which increases through out the period of the purchase based on the increase rate.
A user can choose between the predicted rate (the default) and the current exchange rate for the cost calculation.
If the user is using the predicted rate, the accumulated cost of acquiring the needed DCR is taken as the PoS attack cost.

A warning is shown to the user if the DCR is greater than the amount offered on all exchange.


raedah:
Tests correctly. Verified results are in the expected range. Amount of DCR available on orderbooks will be effected by how many OB's are showing on /markets

chappjc:
Thanks for this PR.  I acknowledge that assuming market price for purchasing any huge amount of DCR is very very conservative and certainly underestimates the practical cost of attack. However, we need some pretty solid justification for the approach you guys have taken. Could you give a plain language description of how CAGR is applied and an explicit justification for using it that way.  From what I see online, CAGR is used for a different purpose and the analogy here is not clear to me: https://www.investopedia.com/terms/c/cagr.asp

raedah:
The cumulative growth rate formula is used to recreate the projected orderbook price based on the increase rate we have from the actual orderbook data. Since its a projection, I do not expect an exact science here. These numbers are an interpretation that are on the conservative side. Compared to the actual orderbook, this estimate undershoots comparisons to the orderbook data from which it is based. This was intentional. Better to underestimate rather then overestimate. Due to this being simply a best projection, the red warning is displayed to indicate to the user that even with the funds available they would still likely have other complications in attempting to acquire the required DCR amount.

If someone was going to criticize this work currently, it would more likely come from not including additional market liquidity out there that is available but not showing on /markets, and thats something to be worked on in another PR.

For the final cost we arrive at, in any case it is going to be a very large number because that is the truth.

chappjc:
> The cumulative growth rate formula is used to recreate the projected orderbook price **based on the increase rate we have from the actual orderbook data**

(emphasis added by me)  This helps explain the use of CAGR in this work.  Thanks.  But the details of how it's used are still only in the code.  We'll review, but the average user needs a clear explanation, or a link one perhaps as a big comment block in our code.

> If someone was going to criticize this work currently, it would more likely come from not including additional market liquidity out there that is available but not showing on /markets, and thats something to be worked on in another PR.

This actually came up on matrix chat.  Realistically the market will rapidly reshape itself in response to large trading volume and swings in market price.  Orders will retreat, book depth will replenish as people move funds on to exchanges, etc.  It gets more and more speculative as we try to make it more realistic.

It may be more accurate, but since it is based on a number of assumptions I just think we need to spell it out and explain the intuition.

buck54321:
This isn't just some little nudge towards more accurate numbers. Before we had concrete figures. We had a defensible definition for attack cost. "This is the value in coin and equipment that an attacker would need to control **right now** in order to initiate an attack". 

With these changes, we move from the static, objective case to something wholly different. The definition would be something like "This is how much it might cost for the attacker to acquire the coin and equipment needed to initiate an attack, based on assumptions A, B, and C"

My biggest question is this. **Are you able to link to some published work that shows how CAGR fits market data under varying market conditions?** If not, it might as well be vodoo. 

To be clear, I think this work is really important. I've often wanted to play with a model there too. But it does create some complications in data presentation. If you do add market modeling, I would ask for the following.

1. A separate dialogue on the page that contains all of the modeling info, including the main equation, and a link to a description of the model and some justification for its use. 
2. If there are variable parameters in the model, the user should have access to adjustments for those parameters over some reasonable range. 
3. The limitations of the model must be made clear.
4. Off by default, though I'm flexible here if the UI is just right.

I'll also echo @chappjc's observation here that CAGR does not seem made for this. If the model does fit market data, it's likely just a mathematical coincidence. It's a pretty simple model though, so if that is the case, you might consider renaming it to avoid confusion.

raedah:
@buck54321 The original mode you mentioned above is still there when in 'Current' mode. This work currently adds additional functionality, a 'Predicted' mode. Nothing is lost. If the attack type is 'External' to the community, which in most cases it would be, the 'Current' market price is the one that is _not accurate_. Its a completely low ball figure that lacks accuracy because its not at all possible to acquire the large amount of DCR needed at that rate. This predicted mode moves us closer to reality by giving an estimation of what the actual figure would be based on the best available data we have. Not saying its perfect and its certainly open to feedback if anyone has a contribution that moves this forward. This is much more accurate to reality for the cost of the attack.

Regarding CAGR, I agree that perhaps the use of the word definition in the PR description above is misleading. It was only used in the respect that the nature of a markets order book is compounding and based on available data we have a measure of its growth rate. So we have a compounding growth rate model we built and it fits nicely, again still on the safe side and intentionally undershooting the available data. Since a model will have a margin or error to either side, I rather the model used goes under this and takes the lower bound rather then trying to fit precisely along the middle and give higher values.

Regarding your question of how it fits under varying market conditions, thats easy, the model is based on the current orderbook data and is derived from the current condition, so it updates itself accordingly when refreshed. As @chappjc mentioned above, "the market will rapidly reshape itself", "orders will retreat, book depth will replenish". So the model updates to the current actual data. Rather than using just the current market order books, I suppose it would be possible to integrate historic order book data as well over time so that the model is not based simply on the present conditions, but even that case wont make much of a difference because its still not able to integrate data about the _future_ demand...for the obvious reason that it does not yet exist. That being said, we can measure the variance over time of the orderbook and see how much it changes. It does have a large degree of constancy, particularly in market conditions that are non-volatile. Thats why the model used in this PR is delivering fairly constant measurements.

So if we really want to improve this, what is important is that we are using the right data sources to account for available market liquidity depth. dcrdex will need to get included on the dcrdata /markets page. Its important when looking at this that we can say that the sources from /markets are a realistic measure of actual market liquidity available out there for DCR. We can also continue to investigate as a team that the model used for the prediction is the best model to be using. I would love to participate in discussions of how the model can be improved. I will work with @ademuanthony to see that an improved description of the model is added as well as additional code comments.

raedah:
Just tested this and noticed that the 'Current' button is broken and is continuing to display 'Predicted' data. This is a bug that needs to be fixed.

buck54321:
Thanks @raedah. That clarifies a lot. 

I wonder if "prediction" should be disabled for "internal" attack.

I agree that we need more data. DEX will certainly be on there when it's ready. 

With regards to the model, if it fits, it fits. But we should be able to show that if we're claiming it. It seems like it is "accurate" by definition at the point `totalObUnits = totalObUnits`, at least in the sense that it accurately predicts the price for the last unit if you buy all the current aggregated asks and nothing changes while you're doing it. Is that correct? I do have some questions about the implementation, but I'll save those for the code review. Could you just verify that your use of CAGR is this.
image

If that is correct, I might do a couple fittings just to see how it fits the data  in the `totalObUnits < totalObUnits` range, and I'll share those here.

raedah:
Will have to confirm some of this with ademuanthony and he is dealing with a hardware failure at the moment but regarding "in the sense that it accurately predicts the price for the last unit" would be true if the OB was completely uniform in slope but its noisy. Also the ends of OB's tend to not be part of the general trend. So we only look at the first 95%, and instead of targeting the last unit we are targeting the avg price which is found by summing the total cost of all units being considered on the book and dividing it by the total number of units considered bought. This gives us the avg price, and that is used as the target for the generated slope. I did not check the code for ademuanthony's application of cagr to get the slope, but I did verify that it does indeed match the avg price derived in the first step. The price of the last unit is also around 5x-6x higher than the avg price, so avg is the lower measure, but as well I consider the avg price to be the more realistic measure. More realistic because its targeting the bulk weight of whats available rather than the end shape.

That sounds correct that it should get disabled for internal. 

The only other thing to consider would be time delay factors such as acquiring over a targeted period of time into the future and how that would play out, but thats a can of worms I dont want to open up. It would just create many more unknown variables about the future. While deriving from the present moment we have hard data available. So it would be fair to say that this is the attack cost in this moment based on actual market data. Showing this cost to the user I believe gives a better idea of the difficulty in executing the attack, rather then just saying 'not possible' even though there is a rather expensive and highly unlikely scenario in which it could actually be played out. Im all for anything else that can be added on the front end to provide the best possible explanation to the user.

ademuanthony:
Yes @buck54321 , this is how the model is used
image

ademuanthony:
Updates.
- The current DCR price is used by default for attack cost calculation.
- All prediction related information are hidden for current DCR price.
- A tooltip that shows more information about the price prediction is added.
- The text has been rewritten to make it clearer

Issue #1741, opened 2020-05-07
Chart mode parameter gets stripped from the URL and is not applied


xaur:
I open:

https://explorer.dcrdata.org/charts?chart=ticket-price&zoom=k8f3rvwm-k9oqhibz&bin=window&axis=time&visibility=true-false&mode=stepped

it redirects to

https://explorer.dcrdata.org/charts?chart=ticket-price&zoom=k8f3rvwm-k9oqhibz&bin=window&axis=time&visibility=true-false

thus stripping the `mode=stepped` and not applying it on the UI.

explorer.dcrdata.org dcrdata v5.3.0-pre+dev

raedah:
looks like it may be resolved or partially resolved in https://github.com/decred/dcrdata/pull/1765

Ekeh:
This issue has been resolved in #1765

ukane-philemon:
Not fixed. I'll look into this soon.

PR #1755, opened 2020-05-27
Market url and param string improvements

Ekeh:
This PR fixed the url parameters of /market page so they are updated automatically when a setting is changed. That way a user who shared a url will be sharing the specific view they are looking at. Also to simplify the url, only params are shown that are set other than the default value. Also added support for the url to be /markets so that it matches the plurality of the rest of the dcrdata urls, while still maintaining backwards-compatible support for /market.

### DIFF
--- a/explorer/explorerroutes.go
+++ b/explorer/explorerroutes.go
@@ -2184,7 +2184,12 @@ func (exp *explorerUI) StatsPage(w http.ResponseWriter, r *http.Request) {
 	io.WriteString(w, str)
 }
 
-// MarketPage is the page handler for the "/agendas" path.
+// MarketPageRedirect is the page handler that redirects "/market" url permanently to "/markets".
+func (exp *explorerUI) MarketPageRedirect(w http.ResponseWriter, r *http.Request) {
+	http.Redirect(w, r, "/markets", http.StatusMovedPermanently)
+}
+
+// MarketPage is the page handler for the "/markets" path.
 func (exp *explorerUI) MarketPage(w http.ResponseWriter, r *http.Request) {
 	str, err := exp.templates.exec("market", struct {
 		*CommonPageData
--- a/main.go
+++ b/main.go
@@ -771,7 +771,8 @@ func _main(ctx context.Context) error {
 		r.Get("/charts", explore.Charts)
 		r.Get("/ticketpool", explore.Ticketpool)
 		r.Get("/stats", explore.StatsPage)
-		r.Get("/market", explore.MarketPage)
+		r.Get("/market", explore.MarketPageRedirect) // Moved to markets but still keeping for backward compatibility.
+		r.Get("/markets", explore.MarketPage)
 		r.Get("/statistics", func(w http.ResponseWriter, r *http.Request) {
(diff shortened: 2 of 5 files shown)


xaur:
#1741 is different but may have some common URL processing code with this one.

main.go
```diff
@@ -771,7 +771,8 @@ func _main(ctx context.Context) error {
 		r.Get("/charts", explore.Charts)
 		r.Get("/ticketpool", explore.Ticketpool)
 		r.Get("/stats", explore.StatsPage)
-		r.Get("/market", explore.MarketPage)
+		r.Get("/market", explore.MarketPage) // Moved to markets but still keeping for backward compatibility.
```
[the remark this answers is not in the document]
Ekeh: Done.

public/js/controllers/market_controller.js
```diff
@@ -1128,32 +1150,42 @@ export default class extends Controller {
...
     if (btn.nodeName !== 'BUTTON' || !this.graph) return
     this.setZoomPct(parseInt(btn.name))
     var stats = this.graph.getOption('stats')
-    var spread = stats.midGap * parseFloat(btn.name) / 100
-    this.graph.updateOptions({ dateWindow: [stats.midGap - spread, stats.midGap + spread] })
+    if (stats) {
+      var spread = stats.midGap * parseFloat(btn.name) / 100
+      this.graph.updateOptions({ dateWindow: [stats.midGap - spread, stats.midGap + spread] })
+    }
+    settings.zoom = parseInt(btn.name)
```
ademuanthony, line 1166: Can we set a `data-percentage` on the buttons rather than using the `name` attribute for this purpose?

Ekeh: Done.

explorer/explorerroutes.go
```diff
@@ -2185,6 +2185,9 @@ func (exp *explorerUI) StatsPage(w http.ResponseWriter, r *http.Request) {
 }
 
 // MarketPage is the page handler for the "/agendas" path.
```
dnldd, line 2187: s **/** "/agendas" **/** "/markets"

explorer/explorerroutes.go
```diff
@@ -2185,6 +2185,9 @@ func (exp *explorerUI) StatsPage(w http.ResponseWriter, r *http.Request) {
...
+func (exp *explorerUI) MarketPageMoved(w http.ResponseWriter, r *http.Request) {
```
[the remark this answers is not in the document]
Ekeh: Resolved.

explorer/explorerroutes.go
```diff
@@ -2185,6 +2185,9 @@ func (exp *explorerUI) StatsPage(w http.ResponseWriter, r *http.Request) {
...
+	http.Redirect(w, r, "/markets", http.StatusMovedPermanently)
+}
```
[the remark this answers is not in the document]
Ekeh: Resolved.

explorer/explorerroutes.go
Ekeh: Resolved.

dnldd (approved):

PR #1769, opened 2020-06-29
build: update deps

dajohi:


chappjc:
Thanks.  Looks straightforward.  Just a note that the `websocket-extensions` vuln does not impact production builds, being only used for the webpack dev server (the think that is used in `npm run watch` to constantly update the pack while a dev is hacking).

Issue #1734, opened 2020-04-21
Chart showing 


jholdstock:
https://testnet.dcrdata.org/address/TsZkKizNMtX1tuc4B12HfSbjqTnmzHuXabY?chart=balance&zoom=k7uu1gg0-k92aom80&bin=month

chappjc:
Seems that when the page is loaded from just the address, the URL gets written as `https://testnet.dcrdata.org/address/TsZkKizNMtX1tuc4B12HfSbjqTnmzHuXabY?chart=balance&zoom=k7uu1gg0-k92aom80&bin=month` (notice `bin=month`), which is incorrect.  If you click on the Group By options, only Day and Block remain, so I believe Week and Month are not intended to be Group By options (deserves investigation though).  @mkingori Interested in this one?

mkingori:
Yes i am.

PR #1774, opened 2020-07-28
Fixed Group By bug on address data chart

mkingori:
Solves #1734

Added a grouping logic for the graph as per the oldest address transaction block time as follows:
- `< 10 days` ago to group by `All`
- `< 70 days` ago to group by `Day`
- `< 12 months` ago to group by `Week`
- `< 5 years` ago to group by `Month`
- `> 5 years` ago to group by `Year`

### DIFF
--- a/db/cache/addresscache.go
+++ b/db/cache/addresscache.go
@@ -577,6 +577,8 @@ func (d *AddressCacheItem) OldestAddressTransaction() (*dbtypes.AddressRowCompac
 	var oldestTxBlockTime int64
 	oldestTransaction := &dbtypes.AddressRowCompact{}
 
+	d.mtx.Lock()
+	defer d.mtx.Unlock()
 	if d.rows == nil {
 		return nil, fmt.Errorf("Not transactions found")
 	}
--- a/db/cache/addresscache.go
+++ b/db/cache/addresscache.go
@@ -571,6 +571,31 @@ func (d *AddressCacheItem) SetBalance(block BlockID, balance *dbtypes.AddressBal
 	d.balance = balance
 }
 
+// OldestAddressTransaction returns data of the oldest address transaction
+func (d *AddressCacheItem) OldestAddressTransaction() (*dbtypes.AddressRowCompact, error) {
+
+	var oldestTxBlockTime int64
+	oldestTransaction := &dbtypes.AddressRowCompact{}
+
+	if d.rows == nil {
+		return nil, fmt.Errorf("Not transactions found")
+	}
+
+	for k, v := range d.rows {
+		if k == 0 {
+			oldestTxBlockTime = v.TxBlockTime
+			oldestTransaction = v
+			continue
+		}
+
+		if oldestTxBlockTime > v.TxBlockTime {
+			oldestTxBlockTime = v.TxBlockTime
+			oldestTransaction = v
+		}
+	}
+	return oldestTransaction, nil
+}
+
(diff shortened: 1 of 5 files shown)


db/cache/addresscache.go
```diff
@@ -571,6 +571,31 @@ func (d *AddressCacheItem) SetBalance(block BlockID, balance *dbtypes.AddressBal
 	d.balance = balance
 }
 
+// OldestAddressTransaction returns data of the oldest address transaction
+func (d *AddressCacheItem) OldestAddressTransaction() (*dbtypes.AddressRowCompact, error) {
+
+	var oldestTxBlockTime int64
+	oldestTransaction := &dbtypes.AddressRowCompact{}
+
+	if d.rows == nil {
```
chappjc, line 580: Not thread-safe

db/dcrpg/pgblockchain.go
```diff
@@ -6324,3 +6324,8 @@ func (pgb *ChainDB) MixedUtxosByHeight() (heights, utxoCountReg, utxoValueReg, u
 	return
 
 }
+
+// OldestTransaction queries the database for the data of the oldest address transaction
```
chappjc, line 6328: Does not query the database though. Is that what you intend?

views/address.tmpl
```diff
@@ -152,11 +152,11 @@
...
-                  <button name="year">Year</button>
-                  <button class="btn-selected" name="month">Month</button>
-                  <button name="week">Week</button>
-                  <button name="day" data-fixed="1">Day</button>
-                  <button name="all" data-fixed="1">Block</button>
+                  <button {{if eq $.GroupBy "year"}} class="btn-selected" {{end}} name="year">Year</button>
+                  <button {{if eq $.GroupBy "month"}} class="btn-selected" {{end}} name="month">Month</button>
+                  <button {{if eq $.GroupBy "week"}} class="btn-selected" {{end}} name="week">Week</button>
+                  <button {{if eq $.GroupBy "day"}} class="btn-selected" {{end}} name="day" data-fixed="1">Day</button>
+				          <button{{if eq $.GroupBy "all"}} class="btn-selected" {{end}} name="all" data-fixed="1">Block</button>
```
chappjc, line 159: Mixing tabs and spaces.

db/dcrpg/pgblockchain.go
mkingori: I am just getting the oldest transaction of an address from the cache.

mkingori:
I've made the changes.

dnldd:
I'm grouping by block in the image below but nothing on the chart shows I am, shouldn't the x-axis be the block height instead of the time there?

group-by-block

mkingori:
I believe its correct as it is. The objective was to show when the transactions of a particular address occurred. The block height won't be much relevant here.

dnldd:
Labelling that `group by` option as `block` seems misleading to me. Personally I'd expect to see the block heights if we are showing when the transactions of a particular address occurred relative to blocks. Perhaps the label should be `exact` which would denote the exact time the transaction for the address occurred. That'd fit well with the current graph I think.

raedah:
The way we have dealt with this on other /charts is by having a separate widget to toggle the x axis display. That would probably work here. Also on the other charts we call these BINS instead of GROUP BY, so perhaps better to standardize on that.

dnldd (approved):

PR #1780, opened 2020-09-02
Bad request on proposals page

mkingori:
Closes [#1784](https://github.com/decred/dcrdata/issues/1784)

Fixes the `Bad request` error that occur on the `/proposals` page when one changes the vote status on the drop down from `All` to any other option, then back to `All`.

### DIFF
--- a/explorer/explorerroutes.go
+++ b/explorer/explorerroutes.go
@@ -530,6 +530,15 @@ func (exp *explorerUI) timeBasedBlocksListing(val string, w http.ResponseWriter,
 		return
 	}
 
+	lastOffsetRows := uint64(maxOffset) % rows
+	var lastOffset uint64
+
+	if lastOffsetRows == 0 && uint64(maxOffset) > rows {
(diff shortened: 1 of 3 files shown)


dnldd:
Confirmed the error and the code looks good as well I just haven't been able to test it out on simnet (need to figure that out), is there a test url you could share?

mkingori:
No.

chappjc (approved):
OK this fixes the byvotestatus issue (string instead of int for all), but this took a long time because the diff included an extraneous change that was not even mentioned anywhere afaict.  I'm referring to the LastOffset change.  Merging as is, but this should have been split into 2 commits, had appropriate docs in both the PR and in the code where `lastOffset` is computed, and ideally done without adding a page data field and instead using a template function to do the mod and subtract as appropriate.

chappjc:
I can't resolve the conflicts on this PR. Either fix or we'll merge https://github.com/decred/dcrdata/pull/1838 instead.

PR #1787, opened 2020-11-04
add dex.decred.org exchange

buck54321:
Add an `exchanges.Exchange`-implementing type for a Decred DEX. On by default.

Aggregated depth chart stacks deepest books first. Exchange colors are now static.


dcrrates/rateserver/go.sum
```diff
@@ -1,72 +1,177 @@
 cloud.google.com/go v0.26.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
+decred.org/cspp v0.2.0/go.mod h1:KVnB49sueBFCldRa/ivZCaWZbrPNEiXWwxHCf1jTYKI=
```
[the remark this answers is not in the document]
chappjc: Really no big deal.  If it were dcrdex's go.sum exploding that might be different.

exchanges/exchanges.go
```diff
@@ -2851,3 +2887,285 @@ func (poloniex *PoloniexExchange) Refresh() {
...
+		return
+	}
+	side := dcr.asks
+	if ord.Side == msgjson.BuyOrderNum {
+		side = dcr.buys
+	}
+	rateKey := int64(ord.Rate)
+	bucket := side.order(rateKey, float64(ord.Rate)/1e8)
+	bucket.volume -= float64(ord.Quantity) / 1e8
+	if bucket.volume < 1e-8 {
```
chappjc, line 3144: Would you put a comment about this?

exchanges/exchanges.go
```diff
@@ -2851,3 +2887,285 @@ func (poloniex *PoloniexExchange) Refresh() {
...
+		delete(side, rateKey)
+	}
+}
+
+// updateRemaining processes the 'update_remaining' notification.
+// updateRemaining should only be called with the orderMtx write-locked.
+func (dcr *DecredDEX) updateRemaining(update *msgjson.UpdateRemainingNote) {
+	oid := update.OrderID.String()
```
chappjc, line 3152: I think OrderID could be nil if the server made it so with null json.  Just want to be defensive.  Although I think (Bytes).String might actually not be a problem even with nil slice

exchanges/go.mod
```diff
@@ -3,12 +3,11 @@ module github.com/decred/dcrdata/exchanges/v2
 go 1.12
 
 require (
+	decred.org/dcrdex v0.1.1-beta1
```
chappjc, line 6: I stupidly made a tag that qualified as semver and we're stuck with it.  Will want to make this go for whatever we have on either the master or release-0.1 branches.  Seeing as I made that dumb tag I think we have no choice but to tag v0.1.1 at the same place as release-v0.1.1, so I'll probably do that.

chappjc (changes requested):
Looks great.  I think we can get this in today.  Just need to deal with book purge and a few other little things.

exchanges/exchanges.go
chappjc: And I'm wondering if we really want to check abs(bucket.volume) < 1e-8

exchanges/exchanges.go
```diff
@@ -2851,3 +2887,322 @@ func (poloniex *PoloniexExchange) Refresh() {
...
+			note := new(msgjson.TradeSuspension)
+			err := msg.Unmarshal(note)
+			if err != nil {
+				dcr.setWsFail(fmt.Errorf("suspension Unmarshal error: %v", err))
+				return
+			}
+			if !dcr.checkSeq(note.Seq) {
+				return
+			}
+			if note.SuspendTime > 0 || note.Persist {
```
chappjc, line 3074: If `note.Persist == true`, `Seq` will be zero/unset, so I think you want to checkSeq after checking note.Persist

exchanges/exchanges.go
chappjc: Still seems like this could go slightly negative as easily as positive with successive floating point subtractions.

buck54321: Still works though

chappjc: Oh right!  :)

chappjc:
Two thoughts about the "Decred Markets" table.

image

1. Would be nice to retain some defined capitalization so that instead of "Dcrdex" it would be "DCRDEX"
2. Instead of 0 for DCR Vol., would be better to put "-" or "N/A"

chappjc:
When selecting just the dcrdex market, something is wrong with the depth chart

image

vs.

image

buck54321:
@chappjc It appears you're on an older commit. I've updated the name in the "Decred Markets" table. See what you think now.

chappjc:
So I was. I see "dex.decred.org" now, which makes sense since there can be any number of DCRDEXs.

image

Something is still up with the depth chart though.

buck54321:
This is actually a pre-existing Dygraphs bug. For some reason, the left-most section is often unshaded if just one of its points is out of on the edge of the viewport. I see the same thing if I carefully zoom in on say Bittrex. It's just more apparent on dcrdex because the book only has a few price levels instead of the hundreds on other books. Once the book has more price bins, the chart will look normal.

chappjc (approved):

PR #1791, opened 2020-11-09
market: fix new agg chart bug

buck54321:
Whoops. Forgot that empty arrays are truthy.


chappjc:
Does this resolve the "perputal loading state" bug observed by @davecgh?

buck54321:
Yes

chappjc (approved):

PR #1807, opened 2021-02-21
fix broken links

degeri:
Fixed a couple of links in readme.

Cant seem to find https://github.com/decred/dcrdata/blob/master/cmd/scanblocks/README.md though .

### DIFF
--- a/README.md
+++ b/README.md
@@ -709,13 +709,6 @@ dcrdata daemon, but may be called alone with rebuilddb2. See the
 [README.md](./cmd/rebuilddb2/README.md) for `rebuilddb2` for important usage
 information.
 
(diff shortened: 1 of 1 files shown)


chappjc:
Thanks.  scanblocks was removed in https://github.com/decred/dcrdata/commit/e2619e47f9519184a8afa45dba12b41eb10df91a#diff-3d3476bbe7a80fe6ea1efef4d0c856747873e2abe36a445bea9eec31236d51c5, so if you want to nuke those bits too that would be nice.

degeri:
Done

chappjc (approved):
Thanks!

PR #1812, opened 2021-04-01
house cleaning and dep updates

chappjc:
Lots of housecleaning, plus some dep updates:

- dcrdex latest
- dcrwallet latest on release-v1.6 branch (v1.7.0 tag)
- chi v5
- tollbooth v6 (for insight api)
- lib/pq 1.10
- grpc 1.32 -> 1.36
- env v6.4 -> 6.5 (I want to remove env var support for config though)
- patch: gops, cors, protobuf


chappjc:
Been going for 4 days on a mainnet backend without trouble.

PR #1820, opened 2021-04-29
resolve ssri dep, remove webpack-dev-server, add 404 svg

chappjc:


chappjc:
Already deployed on a staging node. Merging to deploy to testnet.dcrdata.org

Issue #1819, opened 2021-04-27
db support for treasury data


chappjc:
Next step to support a new treasury page is new PG queries or even a new `treasury` table to fill out a treasury page with the three types of txns:

- `treasuryadd` burns coins and credits the treasury account.
- `treasuryspend` creates coins and debits the treasury account.
- `treasurybase` is analogous to a coinbase in that it credits the treasury without spending/burning any coins, instead creating them out of thin air like coinbases

collect and store todo here: https://github.com/decred/dcrdata/pull/1817/files#diff-1256fb31f67b81d11af39cef8f5f93723821756f546f2b939b77d96280d5615aR3942

Also, the votes with a 4th output are used to authorize treasury spends so we should consider how to track these types of votes and/or their tspend decisions (decoded via `stake. GetSSGenTreasuryVotes` or `stake.CheckSSGenVotes`).  Related is the [gettreasuryspendvotes](https://github.com/decred/dcrd/blob/master/docs/json_rpc_api.mediawiki#gettreasuryspendvotes) RPC that can tally the yes/no votes on a particular tspend.

_Originally posted by @chappjc in https://github.com/decred/dcrdata/issues/1817#issuecomment-826906927_

PR #1821, opened 2021-05-02
add treasury db table, begin treasury page

chappjc:
Resolves https://github.com/decred/dcrdata/issues/1819

### DIFF
--- a/cmd/dcrdata/explorer/explorer.go
+++ b/cmd/dcrdata/explorer/explorer.go
@@ -356,7 +356,7 @@ func New(cfg *ExplorerConfig) *explorerUI {
 		"rawtx", "status", "parameters", "agenda", "agendas", "charts",
 		"sidechains", "disapproved", "ticketpool", "visualblocks", "statistics",
 		"windows", "timelisting", "addresstable", "proposals", "proposal",
-		"market", "insight_root", "attackcost"}
+		"market", "insight_root", "attackcost", "treasury", "treasurytable"}
 
 	for _, name := range tmpls {
 		if err := exp.templates.addTemplate(name); err != nil {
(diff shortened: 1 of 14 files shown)


buck54321:
I've got eyes on this today.

chappjc:
> I've got eyes on this today.

Oh, the page?  I'm getting ready to push an update with the select queries for the basic stuff to go on the page.

chappjc:
Ok, the DB bits and a start on the treasury page are there now.  I can pass this off to you @buck54321

chappjc:
The broken /treasury page is there just to ensure the basic data is retrieved. Charts and tx pagination totally broken. You can totally scrap it.

Next steps:
- Fixing up the treasury page tables and maybe charts. Possibly just eliminate the chart from the treasury page for now
- Add the **unconfirmed** tadd and tspends to the page.  Since tspends can sit in mempool for a very long time, I believe unconfirmed treasury txns need their own little table, which will only every include a couple txns.
- Starting with the unconfirmed txns (not in the DB tables yet) would be the next easy step. `getrawmempool` works to get unconfirmed tspends, along with `gettreasuryspendvotes` to tally the votes. So we don't urgently have to parse the 4th output of votes yet to show this.
- Eventually parse the 4th vout of votes and decode those ssgen treasury choices.  I've not decided if we should have a treasury_votes table, but I do think we should not have to modify or query the existing votes table for sake of efficiency.  The idea being we'll want to show unconfirmed tspends, and integrate the approve/disapprove tallies from those new version votes that bothered to set a preference in vout 4.

chappjc:
image

Also the home page's treasury balance and usd value are from the sum of legacy and decentralized treasuries.

chappjc:
Included in https://github.com/decred/dcrdata/pull/1824

PR #1825, opened 2021-05-10
explorer: add mixed percent to homepage

buck54321:
Is "mixed" the right word there?


chappjc (approved):
LGTM. Going up on the HS rebased on master with the latest changes before going to the main site.

Issue #1823, opened 2021-05-06
Proposal discussion link broken


xaur:
https://explorer.dcrdata.org/proposal/politeia-pi-2021-q3-proposal-system-

has a `Discussion` link pointing to

https://proposals.decred.org/proposals/91cfcc8c62af2022

but it doesn't open. The correct link is

https://proposals.decred.org/record/91cfcc8

Possibly part of #1816 but this issue is about a specific bug reproducible on the main site.

chappjc:
The main site is in 6.0 release now.  Resolved (aside from legacy proposals being absent)?

xaur:
Yeah links work now. I'm also seeing that each proposal row is duplicated but that's a separate issue. Closing this, thanks.

chappjc:
Seeing that too.  That's new, yay.

PR #1829, opened 2021-06-02
[gov] Migrate to new politeia tlog API

### LINKED ISSUE #1816: Migrate to new politeia API
These are the politeiawww enpoints that dcrdata currenlty uses:
- RouteAllVetted
- RouteProposalDetails
- RouteVoteStatus
- RouteAllVoteStatus

These endpoints have been deprecated in the recent politeia [release](https://github.com/decred/politeia/releases/tag/v1.0.0).

Listed below are the routes that should be used instead.  dcrdata currently pulls the politeia git repos manually in order to build the vote graphs.  This is not required anymore.  All data that dcrdata needs should be available through the politeia API.  

The release also includes a politeiawww client to make interacting with the API easier.  This will replace the dcrdata `piclient` package.
[politeiawww client](https://github.com/decred/politeia/blob/master/politeiawww/client/client.go#L164)

The new API uses generic record routes.  A record is application agnostic.  It can be a proposal, a cms invoice, a cms DCC, or any other type of record.  
[Record structure](https://github.com/decred/politeia/blob/master/politeiawww/api/records/v1/v1.go#L219-L230)

**RouteAllVetted**

Replaced by inventory ordered route.

[req/reply payloads](https://github.com/decred/politeia/blob/master/politeiawww/api/records/v1/v1.go#L429-L441)
[client method](https://github.com/decred/politeia/blob/master/politeiawww/client/records.go#L149)

**RouteProposalDetails**

Replaced by record details route. 

[req/reply payloads](https://github.com/decred/politeia/blob/master/politeiawww/api/records/v1/v1.go#L306-L316)
[client method](https://github.com/decred/politeia/blob/master/politeiawww/client/records.go#L80)
[pictl reference cmd](https://github.com/decred/politeia/blob/master/politeiawww/cmd/pictl/cmdproposaldetails.go)

Here's a function that converts a record to a proposal from the legacy `www/v1` API.  This shows how proposal data is formatted in the generic record structure.
https://github.com/decred/politeia/blob/master/politeiawww/proposals.go#L863

**RouteVoteStatus and RouteAllVoteStatus**

Replaced by ticket vote summaries route.

[req/reply payloads](https://github.com/decred/politeia/blob/master/politeiawww/api/ticketvote/v1/v1.go#L515-L534)
[client method](https://github.com/decred/politeia/blob/master/politeiawww/client/ticketvote.go#L126)
[pictl reference cmd](https://github.com/decred/politeia/blob/master/politeiawww/cmd/pictl/cmdvotesummaries.go)

**Cast vote data**

The cast vote replies now contain the vote timestamp so that the dcrdata vote graphs can be built using the API data.

[req/reply payloads](https://github.com/decred/politeia/blob/master/politeiawww/api/ticketvote/v1/v1.go#L453-L476)
[client method](https://github.com/decred/politeia/blob/master/politeiawww/client/ticketvote.go#L109)
[pictl reference cmd](https://github.com/decred/politeia/blob/master/politeiawww/cmd/pictl/cmdvoteresults.go)

Feel free to ping me if you have any questions.

thi4go:
This diff implements the changes necessary to migrate to the new politeia tlog APIs. All data is synced through politeia APIs and stored on stormdb for now. Talked to @chappjc and we might want to drop the stormdb dependency in the near future, so this data might go to a pg table instead of being stored on storm. Major changes on this PR were: 

- Drops `go-piparser` that dcrdata needed as part of the logic of fetching the git repos and parsing votes data.
- The pg tables `proposals` and `proposal_votes` are no longer used anywhere in the code.
- Chart data is now pre-parsed from the api and stored on stormdb with the proposal object
- Code cleanups and refactoring of the gov module for improved readability and reasoning.

I think there isn't a need for the gov package to be a standalone module anymore. I left it be, but if you agree on this, I can make it part of this PR.

About supporting the legacy proposals, politeiagui is using [legacyproposals.json](https://github.com/decred/politeiagui/blob/master/src/legacyproposals.json) and [legacyvotestatuses.json](https://github.com/decred/politeiagui/blob/master/src/legacyvotestatuses.json) to display part of the data. These would be enough for the dcrdata UI if we import them, except for the tickets vote data for the charts. So I think we have two options:

- Show charts on legacy proposals, and keep `proposals` and `proposal_votes` tables. 
- Decide to not show charts on legacy proposals, and nuke the pg tables completely.

With this migration we lose the ability to reconstruct the pg tables if we ever need it, since we dropped the `go-piparser` dep and all the legacy code behind it, so I'm not sure how I feel about the first option. The data will always be availabe on git if someone ever wants the parse it themselves and verify it, so I tend to think it's okay not showing charts for the legacy props.

Adding support for legacy proposals will be done next in a fresh PR in order to facilitate review.

closes #1816 and #1823


gov/politeia/tlog/proposals.go
```diff
@@ -0,0 +1,596 @@
...
+	errDef = fmt.Errorf("ProposalDB was not initialized correctly")
+
+	// dbVersion is the current requirxed version of the proposals.db.
+	dbVersion = semver.NewSemver(2, 0, 0)
+)
+
+// dbinfo defines the property that holds the db version.
+const dbinfo = "_proposals.db_"
+
+// TODO: remove tlog from namings (?)
```
chappjc, line 41: Yeah, the tlog aspect seems irrelevant since afaict it's just hitting /api endpoints and not dealing with anything tlog-related, correct?

thi4go: yep correct, have to clean a lot of stuff up :)

chappjc: Cool.  Take your time.  I like the way this is progressing.  Very eager to see the go-piparser (the Pi git repo package) go away and to pull vote data even for charts from the new (tlog) Pi API.

thi4go:
This is ready for a review @chappjc and @buck54321 if also interested mate. Need to delete `proposals.db` from dcrdata data folder before testing it out. Working on rewriting `proposals_test.go` in the meantime, will add it soon.

chappjc:
Thanks, @thi4go .  I got it.

chappjc:
Going to review this today after I push a couple fixes.  Don't worry about the conflicts for now @thi4go

thi4go:
Sounds good mate @chappjc 👍

chappjc:
I'm started reviewing, but the merge commit from master makes squashing it a little trickier than usual. Please prefer rebase to merge in the future.

But even with numerous commits, it's often easier to squash first without changing the base commit and then rebase rather than doing it in one step with `git rebase -i`.

What I'm reviewing now is this PR squashed, then rebased on master.  I chose to do it this time like the following (after pulling latest master):

```
git checkout tlog
git branch tlog-pre-squash
git reset --hard master             # start tlog branch over at master
git merge --squash tlog-pre-squash  # replay tlog branch commits in one commit
# then resolve conflicts
```

but could also have done:

```
git checkout tlog
git reset --soft ec6b0e6  # the initial branch point, with tlog's working tree and index kept
git commit                # single commit for entire branch
git rebase master         # replay the commit on master
# then resolve conflicts and git rebase --continue
```

This is the single rebased commit I came up with: https://github.com/chappjc/dcrdata/commit/6e5d94077a84a6b9c1b89c5099ec81a2bbb73066
Please see if that looks right and see if you can squash and rebase this PR down similarly, thanks.

chappjc (changes requested):
I had about 30 other line comments that github just discarded.  Here's one and I'll write my review again later.

gov/politeia/proposals.go
```diff
@@ -35,32 +35,29 @@ var (
...
+	client   *piclient.Client
+	APIPath  string
 }
 
-// NewProposalsDB opens an exiting database or creates a new DB instance with
-// the provided file name. Returns an initialized instance of proposals DB, http
-// client and the formatted politeia API URL path to be used. It also checks the
-// db version, Reindexes the db if need be and sets the required db version.
-func NewProposalsDB(politeiaURL, dbPath string) (*ProposalDB, error) {
+func NewProposalsDB(politeiaURL, dbPath string) (*ProposalsDB, error) {
```
chappjc, line 47: Need to re-doc

gov/politeia/proposals.go
```diff
@@ -35,32 +35,29 @@ var (
...
+	// Validate arguments
 	if politeiaURL == "" {
-		return nil, fmt.Errorf("missing politeia API URL")
+		return nil, fmt.Errorf("missing Politeia URL")
```
chappjc, line 50: errors.New here pls

gov/politeia/proposals.go
```diff
@@ -90,371 +87,509 @@ func NewProposalsDB(politeiaURL, dbPath string) (*ProposalDB, error) {
...
-// ProposalByRefID returns the single proposal identified by the provided refID.
-// RefID is generated from the proposal name and used as the descriptive part of
-// the URL to proposal details page on the /proposal page.
-func (db *ProposalDB) ProposalByRefID(RefID string) (*pitypes.ProposalInfo, error) {
-	if db == nil || db.dbP == nil {
-		return nil, errDef
+// fetchVettedTokens fetches all vetted tokens ordered by the timestamp of
+// their last status change.
+func (db *ProposalsDB) fetchVettedTokensInventory() ([]string, error) {
+	page := 0
```
chappjc, line 209: Do you mean to have a `page++` at the end of this loop?  It looks like it would iterate forever on the same page if the number of tokens were `recordsv1.InventoryPageSize`.
Otherwise, `page` is not a useful var.

gov/politeia/proposals.go
```diff
@@ -90,371 +87,509 @@ func NewProposalsDB(politeiaURL, dbPath string) (*ProposalDB, error) {
...
+	vettedTokens := []string{}
```
chappjc, line 210: To be consistent with "dcrdata style", please use var declaration zero vals are adequate.  i.e. `var vettedTokens []string`

gov/politeia/proposals.go
```diff
@@ -90,371 +87,509 @@ func NewProposalsDB(politeiaURL, dbPath string) (*ProposalDB, error) {
...
+		if len(reply.Tokens) < int(recordsv1.InventoryPageSize) {
+			break
+		}
+	}
+	return vettedTokens, nil
+}
+
+// fetchRecordDetails fetches the record details of the given proposal tokens.
+func (db *ProposalsDB) fetchRecordDetails(tokens []string) (map[string]recordsv1.Record, error) {
+	records := make(map[string]recordsv1.Record, len(tokens))
```
chappjc, line 234: Why not go with `map[string]*ecordsv1.Record` since `Record` is not a small struct and you already have a pointer from `RecordDetails`.

gov/politeia/proposals.go
```diff
@@ -90,371 +87,509 @@ func NewProposalsDB(politeiaURL, dbPath string) (*ProposalDB, error) {
...
+		return nil, fmt.Errorf("Pi client TicketVoteSummaries err: %v", err)
+	}
+	voteSummaries := sr.Summaries
 
-// CheckProposalsUpdates updates the proposal changes if they exist and updates
-// them to the proposal db.
-func (db *ProposalDB) CheckProposalsUpdates() error {
-	if db == nil || db.dbP == nil {
-		return errDef
+	var proposals []*pitypes.ProposalRecord
```
chappjc, line 279: Looks safe to pre-allocate to `len(recordDetails)`.

gov/politeia/proposals.go
```diff
@@ -90,371 +87,509 @@ func NewProposalsDB(politeiaURL, dbPath string) (*ProposalDB, error) {
...
+		}
+		proposal.UserID = um.UserID
+
+		// Comments count
+		proposal.CommentsCount =
+			int32(commentsCounts[record.CensorshipRecord.Token])
+
+		// Vote summary data
+		summary := voteSummaries[proposal.Token]
+		proposal.VoteStatus = summary.Status
```
chappjc, line 313: Let's assume the API can serve bad data.  Check like

```go
summary, found := voteSummaries[proposal.Token]
if !found {
    <log and continue>
```

I suspect zero values for all the fields would be problematic.

gov/politeia/proposals.go
```diff
@@ -90,371 +87,509 @@ func NewProposalsDB(politeiaURL, dbPath string) (*ProposalDB, error) {
...
-// CheckProposalsUpdates updates the proposal changes if they exist and updates
-// them to the proposal db.
-func (db *ProposalDB) CheckProposalsUpdates() error {
-	if db == nil || db.dbP == nil {
-		return errDef
+	var proposals []*pitypes.ProposalRecord
+
+	// Iterate through every record and feed data used by dcrdata
+	for _, record := range recordDetails {
+		proposal := pitypes.ProposalRecord{}
```
chappjc, line 283: `var proposal pitypes.ProposalRecord` usually.

Although in this case,

```go
proposal := &pitypes.ProposalRecord{
    State: record.State
...
```

gov/politeia/proposals.go
```diff
@@ -90,371 +87,509 @@ func NewProposalsDB(politeiaURL, dbPath string) (*ProposalDB, error) {
...
+		proposal.TotalVotes = totalVotes
+
+		// Status change metadata
+		ts, changeMsg, err := statusChangeMetadataDecode(record.Metadata)
+		if err != nil {
+			return nil, fmt.Errorf("statusChangeMetadataDecode err: %v", err)
+		}
+		proposal.PublishedAt = ts[0]
+		proposal.CensoredAt = ts[1]
+		proposal.AbandonedAt = ts[2]
```
chappjc, line 334: It would be clearer and less error prone for `statusChangeMetadataDecode` to return something like


```go
type statusTime struct {
    published int64
    censored int64
    abandoned int64
}
```

and then 

```go
proposal.PublishedAt = times.published
proposal.CensoredAt = times.censored
proposal.AbandonedAt = times.abandoned
```

gov/politeia/proposals.go
```diff
@@ -90,371 +87,509 @@ func NewProposalsDB(politeiaURL, dbPath string) (*ProposalDB, error) {
...
-		return err
+	// Parse reply to proposal chart data
+	var chart pitypes.ProposalChartData
+	var timestamps []int64
+	for _, v := range tvr.Votes {
+		bit, err := strconv.ParseUint(v.VoteBit, 16, 64)
+		if err != nil {
+			return nil, err
+		}
+		switch voteOptsMap[bit] {
```
chappjc, line 374: I think we should be doing a bitwise `&` with the bit options otherwise extraneous bits would break the switch.

gov/politeia/proposals.go
```diff
@@ -90,371 +87,509 @@ func NewProposalsDB(politeiaURL, dbPath string) (*ProposalDB, error) {
...
 	}
 
-	n, err := db.saveProposals(publicProposals)
-	if err != nil {
-		return err
+	// Parse reply to proposal chart data
+	var chart pitypes.ProposalChartData
+	var timestamps []int64
+	for _, v := range tvr.Votes {
+		bit, err := strconv.ParseUint(v.VoteBit, 16, 64)
```
chappjc, line 370: Docs for `CastVoteDetails` say

```go
VoteBit   string `json:"votebit"`   // Selected vote bit, hex encoded
```

**hex encoded**

Are the docs incorrect?

gov/politeia/proposals.go
```diff
@@ -90,371 +87,509 @@ func NewProposalsDB(politeiaURL, dbPath string) (*ProposalDB, error) {
...
-	if len(lastProposal) > 0 && lastProposal[0].TokenVal != "" {
-		queryParam = fmt.Sprintf("?before=%s", lastProposal[0].TokenVal)
-	}
-	publicProposals, err := db.fetchAPIData(queryParam)
+	tvr, err := db.client.TicketVoteResults(ticketvotev1.Results{
+		Token: token,
+	})
 	if err != nil {
-		return err
+		return nil, fmt.Errorf("Pi client TicketVoteResults err: %v", err)
```
chappjc, line 363: `%w` for error args with `fmt.Errorf` throughout please

gov/politeia/proposals.go
```diff
@@ -90,371 +87,509 @@ func NewProposalsDB(politeiaURL, dbPath string) (*ProposalDB, error) {
...
+		case "no":
+			chart.No = append(chart.No, 1)
+			chart.Yes = append(chart.Yes, 0)
+		}
+		timestamps = append(timestamps, v.Timestamp)
 	}
+	sort.Slice(timestamps, func(i, j int) bool {
+		return timestamps[i] < timestamps[j]
+	})
+	chart.Time = timestamps
```
chappjc, line 387: This sort is reordering just the `timestamps` slice but not the `Yes` and `No` slices.  I think you mean to sort a slice of structs pair times and counts like


```go
[]*struct {
    Yes, No uint64
    Time int64
}
```

gov/politeia/proposals.go
```diff
@@ -90,371 +87,509 @@ func NewProposalsDB(politeiaURL, dbPath string) (*ProposalDB, error) {
...
-	// Add the sum of the newly added proposals.
-	numRecords += n
+	return &chart, nil
+}
 
-	log.Infof("%d politeia proposal DB records were updated", numRecords)
+// proposalsSave saves the proposals data to the db.
+func (db *ProposalsDB) proposalsSave(proposals []*pitypes.ProposalRecord) error {
+	for _, proposal := range proposals {
+		proposal.Synced = false
```
chappjc, line 395: Pls doc that `proposalsSave` also sets `Synced=false` if this is an intended side-effect.

gov/politeia/util.go
```diff
@@ -0,0 +1,118 @@
...
+import (
+	"encoding/base64"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"strings"
+
+	"github.com/decred/politeia/politeiad/plugins/usermd"
+	umplugin "github.com/decred/politeia/politeiad/plugins/usermd"
```
chappjc, line 12: Importing `github.com/decred/politeia/politeiad/plugins/usermd` twice just with different names.

gov/politeia/util.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+	return pmp, nil
+}
+
+// statusChangeMetadataDecode returns the published, censored and abandoned
+// dates from status change metadata streams.
+func statusChangeMetadataDecode(md []recordsv1.MetadataStream) ([]uint64, string, error) {
+	var (
+		statuses = make([]umplugin.StatusChangeMetadata, 0, 16)
+	)
```
chappjc, line 69: Let's avoid var blocks unless there are many lines.  Just `statuses := make([]umplugin.StatusChangeMetadata, 0, 16)` please

gov/politeia/util.go
```diff
@@ -0,0 +1,118 @@
...
+	for _, v := range md {
+		if v.PluginID != umplugin.PluginID {
+			continue
+		}
+
+		// Search for status change metadata
+		switch v.StreamID {
+		case umplugin.StreamIDStatusChanges:
```
chappjc, line 77: `userMetadataDecode` combined these conditions in a single else.  Doesn't seem like this `switch` is needed.

gov/politeia/util.go
```diff
@@ -0,0 +1,118 @@
...
+			d := json.NewDecoder(strings.NewReader(v.Payload))
```
chappjc, line 78: Pls doc that the payload is a stream of encoded json objects.

gov/politeia/util.go
```diff
@@ -0,0 +1,118 @@
...
+			changeMsgTimestamp = v.Timestamp
+		}
+		switch recordsv1.RecordStatusT(v.Status) {
+		case recordsv1.RecordStatusPublic:
+			publishedAt = v.Timestamp
+		case recordsv1.RecordStatusCensored:
+			censoredAt = v.Timestamp
+		case recordsv1.RecordStatusArchived:
+			abandonedAt = v.Timestamp
+		}
```
chappjc, line 110: The `*At` vars are overwritten on each iteration regardless of `v.Timestamp`.  But `changeMsgTimestamp` and `changeMsg` are only updated for the latest `Timestamp`.  This seems incorrect.

gov/politeia/proposals.go
```diff
@@ -90,371 +87,509 @@ func NewProposalsDB(politeiaURL, dbPath string) (*ProposalDB, error) {
...
+		return nil, fmt.Errorf("Pi client TicketVoteResults err: %v", err)
 	}
 
-	n, err := db.saveProposals(publicProposals)
-	if err != nil {
-		return err
+	// Parse reply to proposal chart data
+	var chart pitypes.ProposalChartData
+	var timestamps []int64
+	for _, v := range tvr.Votes {
```
chappjc, line 369: `CastVoteDetails` is a large struct and I hate to iterate like this over a slice of them (not pointers).  I'd be tempted to do `for iv := range tvr.Votes` and access like `tvr.Votes[i].VoteBit`, but it's not a big deal.

chappjc (changes requested):

gov/politeia/proposals.go
thi4go: omg yes, sorry about that, my bad

gov/politeia/proposals.go
thi4go: hmmm yes seems like it. Will warn luke

chappjc: Do you have an example URL where I can pull the response from?

thi4go: https://proposals.decred.org/record/e2d7b7d

You can click on the magnifier that is besides the total votes, which is used to search for a specific ticket vote. It will then load a pop-up which starts to load the vote results request. then you check the response on the browser's network console logs 👍 

will be pushing some changes to address the comments soon mate

gov/politeia/util.go
thi4go: We only need to get the `changeMsgTimestamp` and `changeMsg` from the newest status change metadata, which will correspond to the current status of that proposal. About the `*At` timestamps, not sure what you mean about overwritten. We key-off the timestamps by the record status from that status change metadata. an example of this in pi: [gh link](https://github.com/decred/politeia/blob/84a52cfc0975cb1ca2df676d15e7ad0637a8ce42/politeiawww/proposals.go#L973-L991)

gov/politeia/proposals.go
thi4go: not sure if useful but: https://github.com/decred/politeia/blob/84a52cfc0975cb1ca2df676d15e7ad0637a8ce42/politeiad/backendv2/tstorebe/plugins/ticketvote/cmds.go#L1467-L1468

chappjc: OK, so:

```
curl -X POST -d '{"token":"e2d7b7d6b837431f"}' https://proposals.decred.org/api/ticketvote/v1/results
```

gives vote records like:

```json
{
  "votes": [
    {
      "token": "e2d7b7d6b837431f",
      "ticket": "0002dcdc9f7bb52be9f5ee09549a2c4a630d844bf8c9c8e36e6b3e89ad45c56b",
      "votebit": "2",
      "address": "DsVYYG1nJKGm7ozrvh4gDnXCN1g6DexJ2KT",
      "signature": "1f76c0bb04cbdffb2bd248e79b8fbaab70228951a28d60b65bb121a76a98a90b5e11663a77ea3ff894fcfd858f9a0ff6380d70ef857370636b500c4849baeabbd2",
      "receipt": "f2f7b4a9a69882ee14b166ed1afa300c36099bbe3e31f78a70a8a811a0569d4fd46de7eacddfcec749fe778f01b97373ad58e6f2c2e824c33ca5a235bf15460f",
      "timestamp": 1626540343
    },
...
```

So the `votebit` field really is a text representation of a number, rather than just a number.  Any idea why that is?

thi4go: hmmm actually no, I'm not sure why that is.

chappjc: > I'll be adding a similar bitwise `&` with the bit options for validation like you asked above

Nice.  I see that being done in `plugins/ticketvote.voteBitVerify` with `voteDetails.Params.Options` and `voteDetails.Params.Mask`, so I guess that's the way it's supposed to be done.

https://github.com/decred/politeia/blob/84a52cfc0975cb1ca2df676d15e7ad0637a8ce42/politeiad/backendv2/tstorebe/plugins/ticketvote/cmds.go#L184

gov/politeia/util.go
chappjc: Do you mean there is only supposed to be one of any any kind of `RecordStatusT` in the `status` slices?  That's not clear.

gov/politeia/proposals.go
chappjc: OK so I'm guessing hex encoded means it's a hexadecimal number like "20" instead of 32.
Ah that makes sense as base 16 is passed to strconv.ParseUint.  OK so the docs looks right then.

thi4go: oh yes that's it, got confused there for a moment also 😅

gov/politeia/util.go
thi4go: Yes, since it represents status change transitions from proposals, there can only be one status change metadata for each of the statuses. There is also no way for a proposal to have passed through a status twice, so there will only be one status change metadata for each status.

chappjc: That makes perfect sense now.  Please put that in comments here

thi4go:
Hey mate @chappjc thanks for the review. This is ready for another look, comments addressed.

chappjc:
Great.  Apologies for failing to alert you to the axios elimination.

gov/go.mod
```diff
@@ -1,4 +1,4 @@
-module github.com/decred/dcrdata/gov/v4
+module github.com/decred/dcrdata/gov/v5
```
chappjc, line 1: Oh dang actually we never reached a gov/v4 release (no gov/v4.x.y tag exists).  Sorry, can you go back to gov/v4 since we never closed dev cycle that would have stabilized it?  My bad.

thi4go: no problem at all mate, all good 👍 rewinded the package version changes

gov/politeia/util.go
```diff
@@ -0,0 +1,154 @@
...
+		if err != nil {
+			return nil, err
+		}
+		pmp = &pm
+		break
+	}
+	if pmp == nil {
+		return nil, fmt.Errorf("proposal metadata not found")
+	}
+	return pmp, nil
```
chappjc, line 61: I suggest tightening this up:

```suggestion
		return &pm, nil
	}
	return nil, fmt.Errorf("proposal metadata not found")
```

And get rid of the `pmp` delc at the top of the function.

Can take the same approach in `userMetadataDecode`.  (Is not not an error in that function to not find it?)

gov/politeia/util.go
```diff
@@ -0,0 +1,154 @@
...
+			}
+		}
+	}
+
+	// Status change metadata represents the data associated with each
+	// status change a proposal undergoes. A proposal can only ever be
+	// on one status once. Therefore, walk the statuses metadatas and
+	// parse the data we need from the public, censored and abandoned
+	// status, as well as the status change message from the latest
+	// one.
```
chappjc, line 105: Rewrap this block pls.

gov/politeia/util.go
```diff
@@ -0,0 +1,154 @@
...
+
+// voteBitVerify verifies that the vote bit corresponds to a valid vote option.
+// This verification matches the one used on politeia's code base to verify vote
+// bits.
+func voteBitVerify(options []ticketvotev1.VoteOption, mask, bit uint64) error {
+	if len(options) == 0 {
+		return fmt.Errorf("no vote options found")
+	}
+	if bit == 0 {
+		return fmt.Errorf("invalid bit 0x%x", bit)
```
chappjc, line 137:
```suggestion
		return fmt.Errorf("invalid bit %#x", bit)
```

gov/politeia/proposals.go
```diff
@@ -91,373 +90,607 @@ func NewProposalsDB(politeiaURL, dbPath string) (*ProposalDB, error) {
...
+
+		var totalVotes uint64
+		for _, v := range summary.Results {
+			totalVotes += v.Votes
+		}
+		proposal.TotalVotes = totalVotes
+
+		// Status change metadata
+		statusTimestamps, changeMsg, err :=
+			statusChangeMetadataDecode(record.Metadata)
```
chappjc, line 292: Only comments need to wrap strictly to 80.  This can go on 1 line as can a number of other fmt.Errorf statements with err on the next line.

The guidline for dcrdata style is that when it comes to code, columns are less important than readibility of the code, just keep is to less than a 100 or so columns, but no fixed limit.  Comments on the other hand I really like formatted consistently.  Setup your editor to wrap at 80 with some key combo.

gov/politeia/proposals.go
```diff
@@ -91,373 +90,607 @@ func NewProposalsDB(politeiaURL, dbPath string) (*ProposalDB, error) {
...
 }
 
-// LastProposalsSync returns the last time a sync to update the proposals was run
-// but not necessarily the last time updates were synced in proposals.db.
-func (db *ProposalDB) LastProposalsSync() int64 {
-	return atomic.LoadInt64(&db.lastSync)
+// fetchVettedTokens fetches all vetted tokens ordered by the timestamp of
+// their last status change.
+func (db *ProposalsDB) fetchVettedTokensInventory() ([]string, error) {
+	page := 1
```
chappjc, line 311: Just declare with the needed type like `page := uint32(1)` an remove the casts below.

gov/politeia/proposals.go
```diff
@@ -91,373 +90,607 @@ func NewProposalsDB(politeiaURL, dbPath string) (*ProposalDB, error) {
...
+	var vettedTokens []string
+	for {
+		inventoryReq := recordsv1.InventoryOrdered{
+			State: recordsv1.RecordStateVetted,
+			Page:  uint32(page),
+		}
+		reply, err := db.client.RecordInventoryOrdered(inventoryReq)
```
chappjc, line 318: What would be returned for a non-existent/empty page?  This is an important question since the condition below does not guarantee it is the last page since the last page can also have `len(reply.Tokens) == int(recordsv1.InventoryPageSize)`.

gov/politeia/proposals.go
```diff
@@ -91,373 +90,607 @@ func NewProposalsDB(politeiaURL, dbPath string) (*ProposalDB, error) {
...
-	// Retrieve and update any new proposals created since the previous
-	// proposals were stored in the db.
-	lastProposal, err := db.lastSavedProposal()
-	if err != nil && err != storm.ErrNotFound {
-		return fmt.Errorf("lastSavedProposal failed: %v", err)
+// fetchTicketVoteSummaries fetches the vote summaries for the given proposal tokens.
+func (db *ProposalsDB) fetchTicketVoteSummaries(tokens []string) (map[string]ticketvotev1.Summary, error) {
+	voteSummaries := make(map[string]ticketvotev1.Summary, len(tokens))
+	paginatedTokens := paginateTokens(tokens, ticketvotev1.SummariesPageSize)
+	for i := 0; i < len(paginatedTokens); i++ {
```
chappjc, line 376: Go with range pls.  `for i := range paginatedTokens {`

gov/politeia/proposals.go
```diff
@@ -91,373 +90,607 @@ func NewProposalsDB(politeiaURL, dbPath string) (*ProposalDB, error) {
...
+// page size.
+func paginateTokens(tokens []string, pageSize uint32) [][]string {
+	n := len(tokens)/int(pageSize) + 1 // number of pages needed
+	ts := make([][]string, n)
+	page := 0
+	for i := 0; i < len(tokens); i++ {
+		if len(ts[page]) >= int(pageSize) {
+			page++
+		}
+		ts[page] = append(ts[page], tokens[i])
```
chappjc, line 400: Not handling edge cases well here.  Namely when `len(tokens)` is a multiple of `pageSize`.

```suggestion
// paginateTokens paginates tokens in a matrix according to the provided page
// size. pageSize must not be zero.
func paginateTokens(tokens []string, pageSize int) [][]string {
	n := len(tokens) / pageSize
	if len(tokens)%pageSize != 0 {
		n++
	}
	ts := make([][]string, n)
	var page int
	for i := range tokens {
		ts[page] = append(ts[page], tokens[i])
		if (i+1)%pageSize == 0 {
			page++
		}
```

gov/politeia/proposals.go
```diff
@@ -91,373 +90,607 @@ func NewProposalsDB(politeiaURL, dbPath string) (*ProposalDB, error) {
...
 	}
+	return voteSummaries, nil
+}
 
-	var queryParam string
-	if len(lastProposal) > 0 && lastProposal[0].TokenVal != "" {
-		queryParam = fmt.Sprintf("?before=%s", lastProposal[0].TokenVal)
+// paginateTokens paginates tokens in a matrix according to the provided
+// page size.
+func paginateTokens(tokens []string, pageSize uint32) [][]string {
```
chappjc, line 392: Given this function is non-trivial, pls add test coverage. e.g.

```go
func Test_paginateTokens(t *testing.T) {
	tests := []struct {
		name     string
		tokens   []string
		pageSize int
		want     [][]string
	}{
		{
			name:     "2 / 4",
			tokens:   []string{"x", "y"},
			pageSize: 4,
			want:     [][]string{{"x", "y"}},
		},
		{
			name:     "2 / 3",
			tokens:   []string{"x", "y"},
			pageSize: 3,
			want:     [][]string{{"x", "y"}},
		},
		{
			name:     "2 / 2",
			tokens:   []string{"x", "y"},
			pageSize: 2,
			want:     [][]string{{"x", "y"}},
		},
		{
			name:     "1 / 2",
			tokens:   []string{"x"},
			pageSize: 2,
			want:     [][]string{{"x"}},
		},
		{
			name:     "0 / 2",
			tokens:   []string{},
			pageSize: 2,
			want:     [][]string{},
		},
		{
			name:     "1 / 1",
			tokens:   []string{"x"},
			pageSize: 1,
			want:     [][]string{{"x"}},
		},
		{
			name:     "0 / 1",
			tokens:   []string{},
			pageSize: 1,
			want:     [][]string{},
		},
	}
	for _, tt := range tests {
		t.Run(tt.name, func(t *testing.T) {
			if got := paginateTokens(tt.tokens, tt.pageSize); !reflect.DeepEqual(got, tt.want) {
				t.Errorf("paginateTokens() = %v, want %v", got, tt.want)
			}
		})
	}
}

```

gov/politeia/proposals.go
```diff
@@ -91,373 +90,607 @@ func NewProposalsDB(politeiaURL, dbPath string) (*ProposalDB, error) {
...
+			return nil, err
+		}
+
+		// Parse relevant data.
+		var vd voteData
+		switch voteOptsMap[bit] {
+		case "yes":
+			vd.yes = 1
+			vd.no = 0
+		case "no":
```
chappjc, line 457: Pls use `ticketvotev1.VoteOptionIDApprove` and `ticketvotev1.VoteOptionIDReject` in these cases

gov/politeia/proposals.go
```diff
@@ -91,373 +90,607 @@ func NewProposalsDB(politeiaURL, dbPath string) (*ProposalDB, error) {
...
+			vd.no = 1
+			vd.yes = 0
+		}
```
chappjc, line 460:
```go
default:
	log.Warnf("Unknown vot option ID %v", voteOptsMap[bit])
	continue
```

gov/politeia/proposals.go
```diff
@@ -91,373 +90,607 @@ func NewProposalsDB(politeiaURL, dbPath string) (*ProposalDB, error) {
...
-func (db *ProposalDB) lastSavedProposal() (lastP []*pitypes.ProposalInfo, err error) {
-	err = db.dbP.Select().Limit(1).OrderBy("Timestamp").Reverse().Find(&lastP)
-	return
+// proposal is used to retrieve proposals from stormdb given the search
+// arguments passed in.
+func (db *ProposalsDB) proposal(searchBy, searchTerm string) (*pitypes.ProposalRecord, error) {
+	var proposal pitypes.ProposalRecord
+	err := db.dbP.Select(q.Eq(searchBy, searchTerm)).Limit(1).First(&proposal)
+	if err != nil {
+		log.Errorf("Failed to fetch data from Proposals DB: %w", err)
```
chappjc, line 522: Check for all these `%w` pls.

chappjc (changes requested):

gov/politeia/proposals.go
```diff
@@ -91,373 +90,607 @@ func NewProposalsDB(politeiaURL, dbPath string) (*ProposalDB, error) {
...
 	// Return the proposals listing starting with the newest.
+	var proposals []*pitypes.ProposalRecord
 	err = query.Skip(offset).Limit(rowsCount).Reverse().OrderBy("Timestamp").
 		Find(&proposals)
-	if err != nil && err != storm.ErrNotFound {
-		log.Errorf("Failed to fetch data from Proposals DB: %v", err)
-	} else {
-		err = nil
+	if err != nil && !errors.Is(err, storm.ErrNotFound) {
+		log.Errorf("Failed to fetch data from Proposals DB: %w", err)
```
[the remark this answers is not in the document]
thi4go: hrmmm right, this function didn't change much actually, I should have reviewed this more carefully

gov/politeia/util.go
```diff
@@ -0,0 +1,154 @@
...
+	published int64
+	censored  int64
+	abandoned int64
+}
+
+// statusChangeMetadataDecode returns the published, censored and abandoned
+// dates from status change metadata streams in the statusTimestamps struct.
+// It also returns the status change message for the latest metadata stream.
+func statusChangeMetadataDecode(md []recordsv1.MetadataStream) (*statusTimestamps, string, error) {
+	statuses := make([]usermd.StatusChangeMetadata, 0, 16)
```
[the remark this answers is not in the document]
thi4go: That's just a random number politeia uses that's big enough for the status changes. As of now, a record can only undergo a status change two times at max, but this might change in the future.

reference: [link](https://github.com/decred/politeia/blob/84a52cfc0975cb1ca2df676d15e7ad0637a8ce42/politeiad/backendv2/tstorebe/plugins/usermd/hooks.go#L267)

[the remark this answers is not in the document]
chappjc: Huh, alright.  Normally I'd request a comment and/or const for a magic number like this, esp when it's not even necessary to prealloc at all.  For just a handful of status changes, where the number is unknown, I'd go with `var statuses []usermd.StatusChangeMetadata`.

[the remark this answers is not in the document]
thi4go: Makes sense mate, sounds good

gov/politeia/util.go
thi4go: sweet

gov/politeia/proposals.go
thi4go: Yep important question. The request goes through fine for an empty page, no errors returned. The tokens slice of the reply will return empty though, so in the case of the last page being full, the next request would fetch an empty token slice and we'd know that it was indeed the last page. Should I doc this?

chappjc: Can't hurt to doc intent and assumptions whenever it's not obvious.

gov/politeia/proposals.go
thi4go: Hrmmmmm of course, sorry I missed this mate, my bad. Fixing this up and including the test cases

chappjc: No biggie.  Lots of loc always means saying bugs, nothing unexpected.

BTW, the test cases should probably include multiple page results.  The ones I suggested were limited

thi4go:
Hey @chappjc thx again for the review. Second round of comments addressed 👍

chappjc:
> I think there isn't a need for the gov package to be a standalone module anymore. I left it be, but if you agree on this, I can make it part of this PR.

This was to isolate the storm dependency too.

chappjc:
On the Daylight Robery prop that's voting right now, I'm getting the following at http://127.0.0.1:7777/proposal/ae609f183a031c34: 

image

Seeing the following on the Dubai prop that's not voting yet (http://127.0.0.1:7777/proposal/51c412822e5e4b59), but also on the Storyteller prop:

image

Seeing this on an accepted prop like Automatic Ticket Revocations (http://127.0.0.1:7777/proposal/e2d7b7d6b837431f), same as the ones that are currently voting: 

image


I've rebuilt everything a couple times.

thi4go:
The charts broke after I ditched axios, should have changed the way we get the response, my bad. Pushed a fix for it. The `Error: Missing target element "proposal.token"` we are also getting on master, still investigating what it might be.

chappjc:
Ahhh, yes, having removed a ton of `.data`s when axios was dropped, I should have noticed that.  Thanks for fixing.

chappjc:
I'm just hacking wildly around Stimulus now, but I got around the proposal.token error with the following:

```diff
diff --git a/cmd/dcrdata/public/js/controllers/proposal_controller.js b/cmd/dcrdata/public/js/controllers/proposal_controller.js
index b27fb022..a24cc85f 100644
--- a/cmd/dcrdata/public/js/controllers/proposal_controller.js
+++ b/cmd/dcrdata/public/js/controllers/proposal_controller.js
@@ -118,20 +118,20 @@ export default class extends Controller {
   }
 
   async connect () {
-    if (this.hasApprovalMeterTarget) {
-      const d = this.approvalMeterTarget.dataset
-      const opts = {
-        darkMode: darkEnabled(),
-        segments: [
-          { end: d.threshold, color: '#ed6d47' },
-          { end: 1, color: '#2dd8a3' }
-        ]
-      }
-      this.approvalMeter = new MiniMeter(this.approvalMeterTarget, opts)
+    if (!this.hasApprovalMeterTarget) return // there will be no meter or charts
+
+    const d = this.approvalMeterTarget.dataset
+    const opts = {
+      darkMode: darkEnabled(),
+      segments: [
+        { end: d.threshold, color: '#ed6d47' },
+        { end: 1, color: '#2dd8a3' }
+      ]
     }
+    this.approvalMeter = new MiniMeter(this.approvalMeterTarget, opts)
 
     chartData = await requestJSON('/api/proposal/' + this.tokenTarget.dataset.hash)
-
+    if (!chartData) return
     Dygraph = await getDefault(
       import(/* webpackChunkName: "dygraphs" */ '../vendor/dygraphs.min.js')
     )
```

chappjc (approved):
Looks good now, and working well.

Some follow-up for me:
- the stimulus fix described in my last comment for when there are no votes and thus no approval meter or charts
-  DB upgrade to drop the orphaned proposals table
- <s>Somehow snapshotting the old props (vote data extracted from git repo), and importing all that into the new `ProposalsDB`</s> @thi4go is working on loading the legacy json files from the politeiagui repo *into the actual tlog* so we will eventually get the legacy prop info for free

xaur:
> closes #1816 and #1823

Not sure if this PR is deployed yet, but just in case I can still reproduce #1823 at explorer.dcrdata.org (that is, `Discussion` links still use wrong paths).

xaur:
Nevermind, just learned about `tip.dcrdata.org` and `Discussion` links [are correct](https://tip.dcrdata.org/proposal/91cfcc8c62af2022) there so I'll just wait for the deployment.

Issue #1679, opened 2020-02-03
charts buttons using bogus href and inline javascript


chappjc:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#Accessibility

> Anchor tags are often abused as fake buttons by setting their href to "#" or "javascript:void(0)" to prevent the page from refreshing, then listening for their click events .
> 
> These bogus href values cause unexpected behavior when copying/dragging links, opening links in a new tab/window, bookmarking, or when JavaScript is loading, errors, or is disabled. They also convey incorrect semantics to assistive technologies, like screen readers.
> 
> Use a `<button>` instead. In general, you should only use a hyperlink for navigation to a real URL.

But the inline javascript is what stands out to me as a potential challenge for deploying a good CSP.

_Originally posted by @chappjc in https://github.com/decred/dcrdata/pull/1678#discussion_r373548994

Discussion follows at the above link, but it is not being addressed in that PR, hence this issue.

PR #1832, opened 2021-06-08
remove inline js from charts page

chappjc:
Resolves https://github.com/decred/dcrdata/issues/1679

### DIFF
--- a/cmd/dcrdata/public/scss/navigation.scss
+++ b/cmd/dcrdata/public/scss/navigation.scss
@@ -38,6 +38,10 @@
   color: inherit;
 }
 
+.nav-link {
+  cursor: pointer;
+}
+
 ul.pages > li > a {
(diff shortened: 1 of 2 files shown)


cmd/dcrdata/public/scss/navigation.scss
```diff
@@ -38,6 +38,10 @@
   color: inherit;
 }
 
+.nav-link {
+  cursor: pointer;
+}
```
chappjc, line 43: Bad?  I tried `$nav-link-cursor: pointer;` in _variables.scss as suggested by the comment in bootstrap_overrides.scss, but that didn't work.

buck54321: Good.

buck54321 (approved):

PR #1833, opened 2021-06-08
replace axios with fetch

chappjc:
This drops the axios dependency, using `fetch` instead.

This reduces the size of the vendors bundle (4.bundle.js) from 104 KiB to 90.3 KiB, and grows app.bundle.j by 1 KiB.

### DIFF
--- a/cmd/dcrdata/public/js/helpers/http.js
+++ b/cmd/dcrdata/public/js/helpers/http.js
@@ -1,0 +1,18 @@
+
+export async function requestJSON (url) {
+  const conf = {
+    headers: {
+      accept: 'application/json'
+    },
+    method: 'GET'
+  }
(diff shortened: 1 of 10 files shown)


cmd/dcrdata/public/js/helpers/http.js
```diff
@@ -0,0 +1,26 @@
+
+export async function requestJSON (url, reqBody) {
```
buck54321, line 2: Noticing that we're never actually using the `reqBody` arg. Should we simplify?

chappjc: Ah right, there's no proper POST version of this.  Might as well simplify

buck54321 (approved):

PR #1834, opened 2021-06-08
migrate from turbolinks to turbo drive

chappjc:
This starts the migration from Turbolinks to Turbo Drive.  Help needed.

We are using some wacky Turbolinks hacks to deal with url query and patge anchor navigation issues.

https://github.com/turbolinks/turbolinks/issues/75#issuecomment-445325162
The above is apparently no longer needed with Turbo Drive, but I don't follow how TurboQuery works or how to remove it.  See https://github.com/hotwired/turbo/pull/125

https://github.com/decred/dcrdata/pull/845
https://github.com/decred/dcrdata/pull/870

I'm pretty sure we need to nuke TurboQuery now, but someone else will need to pick this up.


buck54321:
Yuck.

Thanks for doing the footwork, @chappjc. I'll get into it.

chappjc:
Turns out a CSP with `script-src` set to `self` does **not** break with turbolinks after all.  The change in https://github.com/decred/dcrdata/pull/1832 was sufficient it seems.  So I guess that makes this work super low priority.  Maybe even could wait until the hotwired devs get turbo 7.0 out of beta.

chappjc:
Just saw that Turbo 7 was released a few days ago.
https://world.hey.com/hotwired/turbo-7-0dd7a27f
https://github.com/hotwired/turbo/releases

PR #1836, opened 2021-06-23
rigorous html escaping

chappjc:
Both the insight and dcrdata http APIs need to be more cautious with their responses that might include html.
For the most part this involves using `html.EscapeString` with any string that might contain user input, such as an error message.


buck54321 (approved):
lgtm

degeri:
Tested . LGTM

PR #1837, opened 2021-06-23
prevent double Close'ing of the signalr and websocket connections

buck54321:
Also some basic Refresh tests, but they didn't catch the bug.

### DIFF
--- a/exchanges/exchanges_test.go
+++ b/exchanges/exchanges_test.go
@@ -10,6 +10,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"net/http"
 	"os"
 	"strconv"
 	"sync"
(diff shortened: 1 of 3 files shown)


exchanges/exchanges_test.go
```diff
@@ -231,6 +232,31 @@ func TestPoloniexWebsocket(t *testing.T) {
...
+	resp := &http.Response{
+		Body: io.NopCloser(bytes.NewReader(b)),
+	}
+	d.responses = append(d.responses, resp)
+	return resp
+}
+
 type testBittrexConnection struct {
 	xc           *BittrexExchange
 	subscription hubs.ClientMsg
```
chappjc, line 262: current linters on my machine have started fussing about the `conn.subscription = subscription` in Send being ineffective/unused, and it really comes back to this field being unused.  Can we just remove this?

exchanges/exchanges_test.go
```diff
@@ -231,6 +232,31 @@ func TestPoloniexWebsocket(t *testing.T) {
...
+	return resp, nil
+}
+
+func (d *tDoer) queue(body interface{}) *http.Response {
+	b, err := json.Marshal(body)
+	if err != nil {
+		panic("tDoer.queue error:" + err.Error())
+	}
+	resp := &http.Response{
+		Body: io.NopCloser(bytes.NewReader(b)),
```
chappjc, line 254: oh yeah go 1.15 still needs ioutil

chappjc (approved):
bittrex connection is now able to reestablish on the main site.   We can look into signalr replacements without the Close usability issue in the future.

xaur:
```
type Doer interface {
	Do(...) (...)
}
```

Deep stuff.

PR #1847, opened 2021-08-06
js: fix proposal contoller for props with no progress or votes

chappjc:
This resolves the JS error described in https://github.com/decred/dcrdata/pull/1829#issuecomment-893836773

image

The above happens when a proposal has no votes yet and thus cannot have charts or a progress meter.

### DIFF
--- a/cmd/dcrdata/public/js/controllers/proposal_controller.js
+++ b/cmd/dcrdata/public/js/controllers/proposal_controller.js
@@ -118,20 +118,20 @@ export default class extends Controller {
   }
 
   async connect () {
-    if (this.hasApprovalMeterTarget) {
-      const d = this.approvalMeterTarget.dataset
-      const opts = {
-        darkMode: darkEnabled(),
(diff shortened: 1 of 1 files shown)


cmd/dcrdata/public/js/controllers/proposal_controller.js
```diff
@@ -118,20 +118,20 @@ export default class extends Controller {
...
+      segments: [
+        { end: d.threshold, color: '#ed6d47' },
+        { end: 1, color: '#2dd8a3' }
+      ]
     }
+    this.approvalMeter = new MiniMeter(this.approvalMeterTarget, opts)
 
     chartData = await requestJSON('/api/proposal/' + this.tokenTarget.dataset.hash)
-
+    if (!chartData) return
```
chappjc, line 134: This seems redundant, and may be, but if `/api/proposal/{token}` requests a prop with no votes, the json response is `null`

PR #1850, opened 2021-08-27
gov/politeia: refactor proposalsNewUpdate

chappjc:
This changes (*ProposalsDB).proposalsNewUpdate to retrieve all new
proposals, not just one page worth since there may be more than
one page worth of new props depending on how fast they are created
or if dcrdata was stopped for a long period.

Also more logging.


chappjc:
Has been deployed on tip.dcrdata.org for some time.  Have also discussed the `proposalsNewUpdate` change with @thi4go, and it does appear needed to cover the cases described in the PR desc.

PR #1861, opened 2021-09-07
gov/politeia: prevent multiple concurrent updates

chappjc:
The explorer package is responsible for periodically (every 5th block) triggering updates to Politeia proposals, however cmd/dcrdata does an initial update on startup too.  This modifies `(*ProposalsDB).ProposalsSync` so that it will not start multiple concurrent updaters.  Since the initial fetch can take a couple minutes, fast blocks could even cause this even without cmd/dcrdata starting a startup sync.


thi4go (approved):
ACK

PR #1862, opened 2021-09-14
show /nextreduction

Ekeh:
This creates a page similar to the many bitcoin halving pages out there. This page shows a countdown till the next block reward reduction for decred. It also shows basic information about the block reward itself and links to related documentation. People from our community can link to this page to provide an entryway for new people to be curious about how decred works and learn more.

### DIFF
--- a/explorer/explorerroutes.go
+++ b/explorer/explorerroutes.go
@@ -140,6 +140,11 @@ type homeConversions struct {
 	TreasurySplit   *exchanges.Conversion
 	TreasuryBalance *exchanges.Conversion
 }
+type nextReductionConversions struct {
+	PowSplit        *exchanges.Conversion
+	TreasurySplit   *exchanges.Conversion
+	TreasuryBalance *exchanges.Conversion
+}
 
 // Home is the page handler for the "/" path.
 func (exp *explorerUI) Home(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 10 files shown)


chappjc:
Most of these files are in cmd/dcrdata now

chappjc:
Happy to review, but you need to rebase so these diffs are applied to these files that are now under cmd/dcrdata instead of the repository root.

explorer/templates.go
```diff
@@ -600,5 +600,15 @@ func makeTemplateFuncMap(params *chaincfg.Params) template.FuncMap {
 			return intStr
 		},
 		"floor": math.Floor,
+		"getCurrentRewardWindowRemainingTime": func(idx int, max, t int64) int {
```
chappjc, line 603: `"getRewardWindowRemainingTime"` to match function call in nextreduction.tmpl

chappjc (changes requested):

explorer/explorerroutes.go
```diff
@@ -158,19 +163,64 @@ func (exp *explorerUI) Home(w http.ResponseWriter, r *http.Request) {
...
+		}
+	}
 
-	// Safely retrieve the current inventory pointer.
-	inv := exp.MempoolInventory()
-
-	// Lock the shared inventory struct from change (e.g. in MempoolMonitor).
-	inv.RLock()
-	exp.pageData.RLock()
+	str, err := exp.templates.exec("home", struct {
```
chappjc, line 178: The `Home` method got messed up. It shouldn't have changed.  I think something went wrong since your initial diff because `NextReduction` is using `homeConversions` instead of `nextReductionConversions`, and the `Home` handler is changed a lot.

PR #1863, opened 2021-09-16
ui: show tx & script version on tx details

### LINKED ISSUE #1851: Feature Request: Show Transaction and Script Versions
With the upcoming explicit version upgrades consensus change, the transaction and script versions are going to play a more prominent role moving forward, so I think it would be nice to show the the versions in the GUI details section.

Currently, you can see this information by hitting the "decoded" option for the raw tx.

thi4go:
This diff adds the tx and script versions to the tx details areabox.

Edit:

Before:

image

After: 

image

Let me know if any UI changes are needed for a better UX 👍 

closes #1851

### DIFF
--- a/explorer/types/explorertypes.go
+++ b/explorer/types/explorertypes.go
@@ -173,6 +173,7 @@ type TrimmedTxInfo struct {
 // TxInfo models data needed for display on the tx page
 type TxInfo struct {
 	*TxBasic
+	TxVersion        int32
 	SpendingTxns     []TxInID
 	Vin              []Vin
 	Vout             []Vout
@@ -427,6 +428,7 @@ type Vout struct {
 	OP_RETURN       string
 	OP_TADD         bool
 	Index           uint32
+	Version         uint16
 }
 
 // TrimmedBlockInfo models data needed to display block info on the new home page
--- a/cmd/dcrdata/views/tx.tmpl
+++ b/cmd/dcrdata/views/tx.tmpl
@@ -228,25 +228,23 @@
                   {{end}}
               {{end}}
               <tr>
-                <td class="text-right medium-sans text-nowrap pr-2 py-2"
-                  >Raw Tx:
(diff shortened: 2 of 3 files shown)


davecgh:
Every output script has an independent version.  From the screenshot, it appears this is assuming there is only one version for the entire transaction, which isn't the case.

davecgh:
I just hacked up the HTML of the first transaction I came across on the explorer, but I think this would work:

image

thi4go:
Hm understood @davecgh was mistakenly assuming all output scripts from a tx had to be the same version. Sweet, will add it to the outputs table 👍

davecgh:
To be clear, the tx version field looks good and is correct.  It's just the script version that is unique per output.

chappjc:
> I just hacked up the HTML of the first transaction I came across on the explorer, but I think this would work:
> 
> image

That looks good to me too.  We'll just have to play with the responsive design a bit to see if anything gets wonky (e.g. resizing narrow)

thi4go:
@davecgh @chappjc hey mates, fixed the diff with the suggestions 👍

davecgh (approved):
Looks proper to me.

explorer/types/explorertypes.go
```diff
@@ -173,6 +173,7 @@ type TrimmedTxInfo struct {
 // TxInfo models data needed for display on the tx page
 type TxInfo struct {
 	*TxBasic
+	TxVersion        int32
```
chappjc, line 176: This works.  I'm cooking up a PR to do some mempool and DB work, and in that I'm moving `TxVersion` to `TxBasic` and renaming it to just `Version`.  Just a heads up.  This is good like this for now.

chappjc (changes requested):

chappjc (approved):
This works, and I have the minor requested changes in a subsequent diff that adjust the mempool structures.

PR #1865, opened 2021-09-17
fix missing mixedPctTarget breaking home page updates

chappjc:
The following line in homepage_controller.js would error when a new block arrived:

```js
this.mixedPctTarget.innerHTML = ex.mixed_percent.toFixed(0)
```

`this.mixedPctTarget` was undefined because the `'this.mixedPct'` target was not listed in output of the homepage controller's `targets` method.


chappjc:
Page updates not causing errors on tip.dcrdata.org.  Merging.

Issue #1822, opened 2021-05-04
OpenGraph tags


jholdstock:
There are no OpenGraph tags in the dcrdata HTML, which means that links shared on social media do not have any image, logo or description.

The following tools can be used to preview how shared links will appear on Twitter and Facebook.

https://cards-dev.twitter.com/validator
https://developers.facebook.com/tools/debug/?q=https%3A%2F%2Fexplorer.dcrdata.org%2F

Compare with the [preview of dex.decred.org](https://developers.facebook.com/tools/debug/?q=https%3A%2F%2Fdex.decred.org) to see the potential improvement.

I haven't opened a PR for this myself because the image tags require absolute URLs, and I'm unsure how to access the current site URL in `extras.tmpl` - all of the resource URLs in there are relative.

chappjc:
Resolved by https://github.com/decred/dcrdata/pull/1870

PR #1870, opened 2021-09-23
Base url html head and twitter:image meta tag

chappjc:
This adds a `ProxyHeaders` middleware to get the actual request scheme and host when there is a (trusted) proxy between the actual client and the dcrdata backend.  This middleware updates the `http.Request.Host` and `http.Request.URL.Scheme` fields if the proxy headers are set.

`CommonPageData` is supplemented with `BaseURL`, which is something like "https://dcrdata.decred.org" (`scheme + "://" + host`).  The scheme and host are taken from `http.Request.Host` and `http.Request.URL.Scheme`, and the `ProxyHeaders` middleware is required if dcrdata is behind a proxy.

Update the `html-head` template to accept a new anonymous struct that includes `CommonPageData` and a `Title string`, instead of just a title as before.  Use the `BaseURL` and `Title` fields of this struct to set the twitter tags as well as the actual page title.


cmd/dcrdata/views/extras.tmpl
```diff
@@ -2,11 +2,17 @@
...
-	<!-- The above 3 meta tags *must* come first in the head; any other head content must come *after* these tags -->
+	<!-- The above meta tags *must* come first in the head -->
 	<meta name="description" content="dcrdata, an original Decred block explorer powered by Go">
 	<meta name="author" content="The Decred developers">
-	<meta name="application-name"           content="Decred - Autonomous Digital Currency">
-	<meta name="apple-mobile-web-app-title" content="Decred - Autonomous Digital Currency">
+	<meta name="application-name"           content="dcrdata - The Decred Block Explorer">
+	<meta name="apple-mobile-web-app-title" content="dcrdata - The Decred Block Explorer">
+
+	<meta name="twitter:card" content="summary"/>
```
jholdstock, line 11: OpenGraph tags are also needed to enable preview cards on non-Twitter websites (eg. Facebook). 

```
<meta property="og:title"        content="" >
<meta property="og:type"         content="" />
<meta property="og:url"          content="" />
<meta property="og:image"        content="" />
<meta property="og:description"  content="" />
```

chappjc: Thanks @jholdstock.  Was going to look into that next.  

One thing that I noticed on https://developer.twitter.com/en/docs/twitter-for-websites/cards/overview/abouts-cards, is that they recommend pretty huge sizes for the `summary_large_image` card type, so I went with summary and just a reasonably sized symbol image.  The docs there also say not to use static images, so presumably we'd want to move toward a slightly more dynamic image depending on the page. e.g. Supplement the logo image with a little "block" drawing for a /block page, some arrows for the /tx page, maybe a postage stamp or for /address.  Not planning to do this though.

jholdstock: I wouldn't invest too much time into getting this "perfect", I only opened #1822 to make sure we have **something** instead of absolutely nothing.

Making sure we have at least an image and a bit of text is the key thing. If that can be dynamic depending on the page type, thats great, but optimising for every possible resolution/pixel-density isn't so important IMO.

jholdstock (approved):

PR #1875, opened 2021-10-05
gov: use dcrjson/v4 to catch unrecognized vote version

chappjc:
This updates the decred modules again to fix a couple problems:
- Unable to get RPC version:  `*types.VersionCmd is not registered`
- `gov/agendas.listStakeVersions` was not recognizing `dcrjson.RPCError`
  because it was using v3 instead of v4.


thi4go (approved):
tACK

PR #1878, opened 2021-11-21
update for dex 0.4

chappjc:
Update from dcrdex v0.2.2 to v0.4 (master).  This is primarily the rework of the candles types and CertStore now having an extra level keyed first by network (mainnet/testnet/simnet)


chappjc:
Has to go in before txscript PR #1879

PR #1879, opened 2021-11-21
txscript/v4/stdscript migration

chappjc:
This migrates to the new txscript/v4/stdscript package.

New script class types in dbtypes and api/types are created to maintain the same script class strings.

This uncovered a pg DB bug that has existed since the birth of the dcrpg package. Tickets have always had the `is_multisig BOOL` column set to `false`. A DB upgrade would be needed to identify p2sh tickets, or just remove the `is_multisig` column entirely.  I think the idea of this column was to distinguish solo (P2PKH) and legacy stakepool (P2SH) tickets.  For some reason the code was checking the subtype of the stake submission but only setting `is_multisig` was multisig, which was never the case since it is either pkh or sh.  I've confirmed that there are no instances of `is_multisig` being `true` in a mainnet `tickets` table.  I think I will remove this since clearly no one looks at the column, and the corresponding output type can be determined via the `vouts` table.


db/dcrpg/queries.go
```diff
@@ -610,8 +610,10 @@ func InsertTickets(db *sql.DB, dbTxns []*dbtypes.Tx, txDbIDs []uint64, checked,
 			if len(tx.Vouts[0].ScriptPubKeyData.Addresses) > 0 {
 				stakesubmissionAddress = tx.Vouts[0].ScriptPubKeyData.Addresses[0]
 			}
-			scriptSubClass, _ := txscript.GetStakeOutSubclass(tx.Vouts[0].ScriptPubKey, false) // only looking for multisig, so treasury types are irrelevant
-			isMultisig = scriptSubClass == txscript.MultiSigTy                                 // err return nonstandard, so false
+			isMultisig = stdscript.IsStakeSubmissionScriptHashScript(tx.Vouts[0].Version, tx.Vouts[0].ScriptPubKey)
+			// This was historically broken, always setting isMultisig false! A
```
davecgh, line 614: While it's not an entirely unreasonable guess given actual usage patterns, it's perhaps worth point out that this is still not really accurate, since the only you could know is to retroactively set it when the ticket is redeemed to first determine that this output is a p2sh as you're doing here and then also look to see if the signature script redeeming the output is `stdscript.IsMultiSigSigScriptV0`.

txhelpers/swap.go
```diff
@@ -149,7 +150,10 @@ func ExtractSwapDataFromInputScript(inputScript []byte, params *chaincfg.Params)
 // contact and returns the data pushes of the contract.
 func ParseAtomicSwapContract(scriptVersion uint16, script []byte, params *chaincfg.Params) (*AtomicSwapContractPushes, error) {
 	// validate the contract by calling txscript.ExtractAtomicSwapDataPushes
-	contractDataPushes, _ := txscript.ExtractAtomicSwapDataPushes(scriptVersion, script)
+	if scriptVersion != 0 {
```
davecgh, line 153: 🥇

db/dcrpg/pgblockchain.go
```diff
@@ -5838,13 +5840,18 @@ func (pgb *ChainDB) GetExplorerTx(txid string) *exptypes.TxInfo {
 		} else {
 			opTAdd = strings.HasPrefix(vout.ScriptPubKey.Asm, "OP_TADD")
 		}
+		// Get a consistent script class string from dbtypes.ScriptClass. Alt.
+		// to using NewScriptClassFromString, decode Hex or use
+		// msgTx.TxOut[i].PkScript, get stdscript.ScriptType, convert with
+		// dbtypes.NewScriptClass
+		dbScriptClass := dbtypes.NewScriptClassFromString(vout.ScriptPubKey.Type)
```
chappjc, line 5847: One of many things in this changeset that's still not right.  Need to do what's described in the comment and decode the pkScript in the msgTx

txhelpers/swap.go
```diff
@@ -262,7 +266,7 @@ func TxAtomicSwapsInfo(tx *chainjson.TxRawResult, outputSpenders map[uint32]*Out
 	// Check if any of this tx's outputs are contracts. Requires the output to
 	// be spent AND the spending input to have the correct sigscript type.
 	for _, vout := range tx.Vout {
-		if vout.ScriptPubKey.Type != txscript.ScriptHashTy.String() {
+		if vout.ScriptPubKey.Type != stdscript.STScriptHash.String() {
```
chappjc, line 269: Another pervasive problem, using the TxRawResult's strings instead of decoding the tx bytes and determining script types.  Much yet to do in this PR.  I likely won't wrap this up and move this out of draft until the dcrwallet PR is merged, and blockchain/v4, but I will be slowly resolving these legacy issues.

db/dcrpg/queries.go
chappjc: Indeed, it was ill-conceived with the assumption that the scripthash script was itself multi-sig.  Considering the historical 1-of-2 stake pool tickets, that made sense, but it's generally a bad inference.  The bool and the flag in the database would have been better as isP2SH or something like that.  But as I say in the comment, it has never been flagging anything right, and clearly it's not used, so I'm inclined to drop this column of the tickets table.

chappjc:
Script classes are now being kept unchanged despite the more fine-grained stdscript types on account of the new api and db types in the following key places:

1. Display on explorer pages
2. Encoded in JSON API endpoints
3. Stored in the postgresql DB

In addition to keeping the different stake subtypes combined (e.g. `sstxchange-scripthash` || `sstxchange-pubkeyhash` == `sstxchange`), both (1) and (2) also translate the special odd numbered `nulldata` outputs of tickets/stakesubmissions as `sstxcommitment`.

Running as expected on tip.dcrdata.org with dcrd 1.7/master.

chappjc:
3 days on tip.dcrdata.org.  script_type summary in postgres:

```
# select script_type, count(*) from vouts group by script_type;
   script_type   |  count   
-----------------+----------
 nonstandard     |        1
 nulldata        | 11039755
 pubkey          |      951
 pubkeyalt       |        1
 pubkeyhash      | 24744771
 scripthash      |   684404
 sstxchange      |  4278104
 stakegen        |  4150118
 stakerevoke     |    83265
 stakesubmission |  3123067
 treasuryadd     |    63718
 treasurygen     |        1
(12 rows)
```

This indicates the stored script types are stable with `dbtypes.ScriptClass` translating from the `stdscript` type to the internal DB type.

The explorer pages are also showing the same script types, including the "sstxcommitment" pseudo-type, along with the script version:  https://tip.dcrdata.org/tx/78f070e11fbccdfc2bb13e0f7b221c873d1656997ac47b91b90c3617d367170c

image

API:

```json
   "vout": [
    {
      "value": 196.73779166,
      "n": 0,
      "version": 0,
      "scriptPubKey": {
        "asm": "OP_SSTX OP_DUP OP_HASH160 064f47471cf7378c93b2330c9ac954cf69692b5d OP_EQUALVERIFY OP_CHECKSIG",
        "hex": "ba76a914064f47471cf7378c93b2330c9ac954cf69692b5d88ac",
        "version": 0,
        "reqSigs": 1,
        "type": "stakesubmission",
        "addresses": [
          "DsRYGWwfSz3pnGgWmSzAzLMVb9NrZWDVa19"
        ]
      }
    },
    {
      "value": 0,
      "n": 1,
      "version": 0,
      "scriptPubKey": {
        "asm": "OP_RETURN 2a372905e48a34601d842ff88bbed0d117932afa8217a694040000000058",
        "hex": "6a1e2a372905e48a34601d842ff88bbed0d117932afa8217a694040000000058",
        "version": 0,
        "type": "sstxcommitment",
        "addresses": [
          "DsUp7x65r9eBnrzoeF6Y6ncZvuaLfXWdoSh"
        ],
        "commitamt": 196.73782146
      }
    },
    {
      "value": 0,
      "n": 2,
      "version": 0,
      "scriptPubKey": {
        "asm": "OP_SSTXCHANGE OP_DUP OP_HASH160 0000000000000000000000000000000000000000 OP_EQUALVERIFY OP_CHECKSIG",
        "hex": "bd76a914000000000000000000000000000000000000000088ac",
        "version": 0,
        "reqSigs": 1,
        "type": "sstxchange",
        "addresses": [
          "DsQxuVRvS4eaJ42dhQEsCXauMWjvopWgrVg"
        ]
      }
    }
  ]
```

PR #1880, opened 2021-12-04
db/dcrpg: enforce a min required postgres version

chappjc:
This adds a PostgreSQL version check to `NewChainDB`, which will return an error if the postgres service in use indicates a version less than the required minimum version defined by the new `pgVerNumMin` `const`.

This enforces the requirement already documented in README.md.


chappjc:
@victorgcramos With this on master, when you get a chance, see if the one machine trips this or if the `INCLUDE` syntax error is due something else.

victorgcramos:
@chappjc okay 👌🏼

PR #1882, opened 2021-12-05
Purge eslint plugins

chappjc:
eslint was not configured as per .eslintrc.json to use any of these three plugins, which were causing issues upgrading eslint itself.  Remove them.


chappjc:
npm is such horses**t

PR #1886, opened 2021-12-15
trylock: Use sync.Mutex in go1.18+, update CI for beta1

chappjc:
Go 1.18 supplements `Mutex` with a `TryLock` method.  For the on-demand concurrent client accesses to the cache-guarded DB, use this directly instead of our dumb homebrew trylock mutex when Go 1.18 is in use.

Also update CI to use go 1.18-beta1.


.github/workflows/build.yml
```diff
@@ -24,7 +24,7 @@ jobs:
       working-directory: ./cmd/dcrdata
 
     - name: Install Linters
-      run: "curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.48.0"
+      run: "curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.50.1"
```
ukane-philemon, line 27: current version master  is v1.55.2, consider updating to v1.58.0

trylock/trylock.go
```diff
@@ -1,6 +1,9 @@
-// Copyright (c) 2021, The Decred developers
+// Copyright (c) 2021-2022, The Decred developers
```
ukane-philemon, line 1: Can update this the next time you commit.

trylock/trylock_go1.18.go
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) 2022, The Decred developers
```
ukane-philemon, line 1: This too.

ukane-philemon (changes requested):
Looks straight-forward.

PR #1890, opened 2022-01-08
deps: Upgrade to stimulus 3.0

### LINKED ISSUE #1859: upgrade to stimulus 3.0
UPDATE: Stimulus 3 was released a few days ago (https://world.hey.com/hotwired/stimulus-3-c438d432) with Turbo 7.

We should upgrade stimulus from 1.1.1 to 2.0.

See https://onrails.blog/2020/12/04/updating-to-stimulus-js-2-0/

I had also started a branch with changes to this effect: https://github.com/chappjc/dcrdata/commit/39c3a51cf9cbdbcedadb8c85f4e9ab918784cb9c
It branches off an April commit, so it does not rebase cleanly on today's master.
I'm not sure what remained to be completed with that.

thi4go:
This diff updates the stimulus package to the latest version (3.0.1). It
updates data targets and uses the new API for data mapping on the
controllers and templates.

closes #1859


chappjc:
I will get you a proper review soon, but one thing I noticed is the new Data Mappings https://onrails.blog/2020/12/04/updating-to-stimulus-js-2-0/
 
so changing 

```
<div data-controller="meta-state" data-meta-state-key="account-id">
```
to

```
<div data-controller="meta-state" data-meta-state-key-value="account-id">
```

you then change `this.data.get("key")` -> `this.keyValue`

Any value to doing that?  I can't tell if the old `get` way is deprecated or not.

thi4go:
Hey mate @chappjc please take your time, no rush at all to review this 👍 

Upon reading more on the new data value mapping API, we should definitely update this as well, since it provides some cool additional features.

- The previous data mapping API only works with strings, so we had to manually convert it when needed:

   `height = parseInt(this.data.get('height'))`

   The new API handles this type conversion automatically. It introduces a static `values` object to the controllers, where 
   we can specify the data key and its type.

   ```
   static values = {
     url: String,
     height: Number
   }
   ```

   It also generates three properties for each object entry.

   ```
   this.urlValue     // getter
   this.urlValue =   // setter
   this.hasUrlValue  // exists
   ```

Besides the out of the box type conversions and properties, imo it adds documentation value to the controller, since we can check the types being handled on a glance, while being a cleaner syntax.

I think they will not drop support for the old API, but we should also update this to reap the improvements benefits and be up-to-date with the package. I'll get to it and commit these changes 👍 


Reference: https://github.com/hotwired/stimulus/pull/202

chappjc:
On a related note, I'd really love if you could follow up on a few more bootstrap 5 quirks I noticed.
If you open one tab with https://tip.dcrdata.org/ and another at https://dcrdata.decred.org/ you'll see a few minor formatting glitches.
- Mempool section header on home page shifting left
- Voting section header on the home page shifting up
- A few text size deviations like "more blocks..." and "votes approved" and "more transactions..."

It may well be that we'll decide that some of these things are "better", but we're going to be upgrading the mainnet site pretty soon and we'll loose the reference and what we have on tip will become the new formatting whether it's right or not.

Awesome work on everything @thi4go !

thi4go:
My pleasure @chappjc had some trouble with my pgsql env, so took a while to test everything out. This is ready for a review mate. Will open a fresh PR for the bootstrap 5 fixes soon as well 👍

chappjc (approved):
Ready for this as-is.  Will take this in stages and get to the data mappings discussed in https://github.com/decred/dcrdata/pull/1890#issuecomment-1011231352 in a subsequent PR

PR #1891, opened 2022-01-09
upgrade exchange api versions

### LINKED ISSUE #1810: upgrade exchange API versions
```
https://api.binance.com/api/v1 -> v3
https://bittrex.com/api/v1.1 and v2.0 -> https://api.bittrex.com/v3
```

ukane-philemon:
https://api.coindesk.com/v1  -> v2
https://api.binance.com/api/v1 -> v3

Closes #1810

### DIFF
--- a/exchanges/exchanges.go
+++ b/exchanges/exchanges.go
@@ -98,17 +98,17 @@ var (
 		Price: "https://api.coinbase.com/v2/exchange-rates?currency=BTC",
 	}
 	CoindeskURLs = URLs{
-		Price: "https://api.coindesk.com/v1/bpi/currentprice.json",
+		Price: "https://api.coindesk.com/v2/bpi/currentprice.json",
 	}
 	BinanceURLs = URLs{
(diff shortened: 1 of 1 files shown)


chappjc:
Thanks!  No difference with these endpoints?
Can you comment on the 1000->5000 data point change?

ukane-philemon:
Thanks for reviewing, there are no difference to these endpoint. Changes in binance v3 include new endpoints and removal of old endpoints not currently used by decred.
Change in limit
1000- previous max limit
5000- new max limit
Increasing  max limit allow endpoint to get ALL  current DCRBTC bids in binance order book.

chappjc:
Sounds good.  Will test it out.  Interestingly, the v1 endpoint already seems happy to return more than a thousand data points (presently about 1700 asks), but that's probably because they upgraded a common component at some point and it shouldn't be allowing it.  Anyway, thanks again.

ukane-philemon:
@chappjc Thanks for the feedback.

chappjc (approved):

PR #1893, opened 2022-01-18
ui: Bootstrap 5 leftover design fixes

thi4go:
This diff fixes some leftover design issues caused by the bootstrap 5
package upgrade.

### DIFF
--- a/cmd/dcrdata/public/scss/bootstrap_overrides.scss
+++ b/cmd/dcrdata/public/scss/bootstrap_overrides.scss
@@ -16,7 +16,3 @@
     overflow: visible;
   }
 }
-
-.small {
-  font-size: 80%;
-}
--- a/cmd/dcrdata/public/scss/bootstrap_overrides.scss
+++ b/cmd/dcrdata/public/scss/bootstrap_overrides.scss
@@ -16,3 +16,7 @@
     overflow: visible;
   }
 }
+
+.small {
+  font-size: 80%;
+}
--- a/cmd/dcrdata/public/scss/utils.scss
+++ b/cmd/dcrdata/public/scss/utils.scss
@@ -142,6 +142,10 @@
   width: 142px;
 }
 
+.mw50 {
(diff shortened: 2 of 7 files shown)


cmd/dcrdata/public/scss/utils.scss
```diff
@@ -158,6 +162,11 @@
   padding: 0.3rem 0;
 }
 
+.px7-5 {
```
thi4go, line 165: It seems like previous bootstrap (< 5) applied a left/right 7.5 padding to the `col` class. This helper is used to tilt back the `Mempool` homepage section to where it originally were.

cmd/dcrdata/public/scss/bootstrap_overrides.scss
```diff
@@ -16,3 +16,7 @@
     overflow: visible;
   }
 }
+
+.small {
+  font-size: 80%;
+}
```
thi4go, line 22: The `small` bootstrap class for texts also changed. The new version 5 one applies a `font-size: .875em;`. This overrides it to the previous setting that dcrdata used.

cmd/dcrdata/views/ticketpool.tmpl
```diff
@@ -37,7 +37,7 @@
         <br>
         <div class="justify-content-between">
           <div class="dygraph-label dygraph-title mb-1">Distribution of Tickets by Reward Outputs</div>
-          <div class="row col-lg-12 col-sm-8 d-flex text-center m-auto">
+          <div class="col d-flex text-center mw50 w-100 m-auto ">
```
thi4go, line 40: `col-lg-12` used to apply a `max-width: 50%` setting, which changed to `width: 50%` with the update. This sets it back to how dcrdata used it with the `mw50` helper

chappjc (approved):
Formatting matches perfectly.
TY for using _variables.scss instead of overrides.

PR #1894, opened 2022-01-21
Treasury balance updates

chappjc:
**db**: fix treasury spend tally amount

This fixes the treasury spend sum that uses a postgresql query.

This also renames the TGenCount and TGen fields of the
dbtypes.TreasuryBalance struct to TBaseCount and TBase.
The bins of the balance correspond to the treasury types: treasury
adds, spends, bases.  The "tgen" wording confuses this categorization
with output types, where TGEN is the opcode in the outputs of a
treasury spend, while the opcodes in the outputs of a both treasury add
and treasurybase txns are TADD.
Note that when rows are inserted into the treasury table, they are are
currently the change to the treasury and the txn type.

This adds Height and MaturityHeight to the TreasuryBalance struct.

**api:** add /api/treasury/balance


chappjc:
On tip.dcrdata.org

PR #1895, opened 2022-01-23
agendas page: show all agenda meters

chappjc:
image

Room for improvement but will at least show all the voting meters for multiple agendas at the same stake version.

Was showing like the following:

image

### DIFF
--- a/cmd/dcrdata/views/agendas.tmpl
+++ b/cmd/dcrdata/views/agendas.tmpl
@@ -22,12 +22,7 @@
     {{template "html-head" headData .CommonPageData "Consensus Deployment Agendas"}}
         {{template "navbar" .}}
         <div class="container main pb-5" data-controller="agendas">
-
-
-
-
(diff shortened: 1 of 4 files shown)


cmd/dcrdata/views/agendas.tmpl
```diff
@@ -104,137 +99,136 @@
...
+                      {{else}}
+                        <div class="bg-white d-inline-block p-2 fs16 text-center mx-2 border">Voting on <span class="fw-bold">{{.ID}}</span></div>
+                      {{end}}
+                    </div>
+                    <div class="d-flex justify-content-around align-items-start flex-wrap mt-2">
+                      <div class="d-inline-block text-center p-2">
+                        <span class="fs16 fw-bold">Quorum</span><br>
+                        <div
+                          class="quorum meter text-center js-only large-gap"
+                          data-target="agendas.quorumMeter"
```
chappjc, line 165: The problem was that these targets are not usable since this is in a gohtml template **loop** and it makes several identically-named targets.  I may as well remove this line.  The controller that initializes the page now just uses `querySelectorAll` and does a `forEach` through all of these meters, which now have an added class to identify them.

PR #1897, opened 2022-01-26
stylelint v14 and new standard scss

chappjc:


chappjc:
No more changes planned for v6.1, which should get pushed this week.

PR #1898, opened 2022-01-28
ui: Fix customcheck width on treasury page

thi4go:
Bootstrap 5 updated the `row` class to apply a `width: 100%` to its childrens. Apparently this was done to prevent the columns from becoming too narrow when dealing with smaller grids. This made the custom input checks from the treasury page to be overflown from its parent div. Since we do want the columns to be narrow in this specific case,
this diff applies a `width: unset` to the column children's class `.customcheck`.

Before:

image

After:
image

### DIFF
--- a/cmd/dcrdata/public/scss/charts.scss
+++ b/cmd/dcrdata/public/scss/charts.scss
@@ -418,6 +418,7 @@ body.darkBG .chartview .dygraph-y2label {
   margin: 0.3rem auto;
   border: black;
   cursor: pointer;
+  width: unset;
   -webkit-user-select: none;
   -moz-user-select: none;
   -ms-user-select: none;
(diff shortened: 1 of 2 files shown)


chappjc (approved):
Change as discussed in matrix. Tested prior.

cmd/dcrdata/public/scss/charts.scss
```diff
@@ -418,6 +418,7 @@ body.darkBG .chartview .dygraph-y2label {
   margin: 0.3rem auto;
   border: black;
   cursor: pointer;
+  width: unset;
```
chappjc, line 421: Oh, one thing I'd like to request is a bump to the app.bundle.js cache buster at https://github.com/decred/dcrdata/blob/5a46be532ea72ff7fac403bd091c350b55e3dd25/cmd/dcrdata/views/extras.tmpl#L189

thi4go:
As per discussion on matrix, this diff also sneaks in an inline webpack plugin to update `extras.tmpl` with cache busts whenever it is needed.

chappjc:
I meant for the webpack work to be split into a separate PR actually.  Let's just fix the css the old way for now.

But looking at the webpack changes, how come the hash is the same for the css and js bundles.  That implies we'll invalidate the cache for one when we only update the other.  Is there a different hash variable to apply?

PR #1899, opened 2022-02-01
webpack: Add plugin to replace bundle hashes

thi4go:
This diff adds automatic cache bust by adding an inline webpack
plugin, which parses the bundles hashed filenames through the app
entrypoints and replaces them on the `extras.tmpl` template.

### DIFF
--- a/cmd/dcrdata/views/extras.tmpl
+++ b/cmd/dcrdata/views/extras.tmpl
@@ -61,7 +61,7 @@
 	<link rel="preload" href="/images/connected.svg?x=65tv3" as="image" type="image/svg+xml" />
 	<link rel="preload" href="/images/disconnected.svg?x=65tv3" as="image" type="image/svg+xml" />
 
-	<link href="/dist/css/style.css?v=PdgRb7t" rel="stylesheet">
+	<link href="/dist/css/style.ab1819205fd75f31.css" rel="stylesheet">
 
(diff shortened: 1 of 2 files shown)


cmd/dcrdata/views/extras.tmpl
```diff
@@ -61,7 +61,7 @@
 	<link rel="preload" href="/images/connected.svg?x=65tv3" as="image" type="image/svg+xml" />
 	<link rel="preload" href="/images/disconnected.svg?x=65tv3" as="image" type="image/svg+xml" />
 
-	<link href="/dist/css/style.css?v=PdgRb7t" rel="stylesheet">
+	<link href="/dist/css/style.ab1819205fd75f31.css" rel="stylesheet">
```
chappjc, line 64: I think we can go with the new file name, at least on a trial basis on tip.dcrdata.org.  I may need to think a bit more about practical consequences in deployment with the move from a query token to actually changing the file name.

My deploy scripts don't look for the specific file names, just paths, so I think that's OK, but they do nuke the existing "public" folder before copying the new one, so I'm wondering if clients with stale html are going to get 404s on the style and js if their browsers request the old ones them.  Perhaps that's better than getting a style or js that doesn't necessarily match the html they have loaded.

chappjc (approved):

PR #1902, opened 2022-02-06
Rremove support for cockroachDB

ukane-philemon:
This diff removes support for cockroachDB.


db/dcrpg/tables.go
```diff
@@ -196,69 +196,9 @@ func CheckColumnDataType(db *sql.DB, table, column string) (dataType string, err
...
-	if barLoad != nil {
-		barLoad <- &dbtypes.ProgressBarLoad{BarID: dbtypes.InitialDBLoad, Subtitle: " "}
-	}
-	return nil
-}
-
-// DeleteDuplicatesRecovery attempts to delete "duplicate" rows in all tables
-// where unique indexes are to be created.  This is like DeleteDuplicates, but
-// it also includes transactions, tickets, votes, and misses.
-func (pgb *ChainDB) DeleteDuplicatesRecovery(barLoad chan *dbtypes.ProgressBarLoad) error {
```
chappjc, line 261: No need to have the regular `DeleteDuplicates` process the tickets, votes, or misses tables.  Duplicates during IBD (before the unique indexes prevent any) only occur as a result of transactions that were invalidated due to stake disapproval of the regular tx tree in a block ending up in a subsequent block.  This does not apply to tickets, votes, or misses.

chappjc (changes requested):

PR #1904, opened 2022-02-21
views: stimulus targets loose ends

chappjc:
Resolves a failure going between pages of the address table:
image


chappjc:
1 wk deployed w/o trouble

PR #1906, opened 2022-03-24
replace deprecated io/ioutil

dajohi:


chappjc (approved):
Thanks!  My mind is blown that I hadn't done this already.

PR #1918, opened 2022-06-21
multi: Expose more treasury mechanics on TSpend tx page

### LINKED ISSUE #1835: Expose more treasury mechanics on tspend tx page
Currently the [tspend tx page](https://explorer.dcrdata.org/tx/c322f3f93c8aa6d75846ea8ca1d7c9228d23bdef6c21413f8101e7922f265735) shows voting block interval and Yes/No vote counts. These numbers alone do not directly say if the tspend is likely to be approved or rejected. To help with that, as well as the general education, let's expose more stuff from the [DCP-0006](https://github.com/decred/dcps/blob/master/dcp-0006/dcp-0006.mediawiki):

- max possible votes for the interval
- min required votes
- approx. calendar dates of start and end blocks (because we love humans)
- quorum pass progress
- Yes threshold pass progress
- show which Pi key signed the tspend
- show if tspend amount passes or violates the policy (TEW and TEP params)
- anything else?

A side bonus is that when people will screenshot and post this info on social media, it will educate everyone and expose the non-trivial and cool bits like the limits on the burn rate.

ukane-philemon:
- show max possible votes for the voting interval.
- show min required votes(quorum).
- show approx. calendar dates of start and end blocks.
- show quorum pass progress.
- show required yes votes pass progress.
- show which Pi key signed the TSpend.
closes #1835

### DIFF
--- a/db/dcrpg/pgblockchain.go
+++ b/db/dcrpg/pgblockchain.go
@@ -5924,12 +5924,17 @@ func (pgb *ChainDB) GetExplorerTx(txid string) *exptypes.TxInfo {
 			tx.TSpendMeta.QuorumAchieved = totalVotes >= tx.TSpendMeta.QuorumCount
 			tx.TSpendMeta.TotalVotes = totalVotes
 
-			var maxRemainingVotes int64
+			var maxRemainingBlocks int64
 			if !voteStarted {
-				maxRemainingVotes = (tx.TSpendMeta.VoteEnd - tx.TSpendMeta.VoteStart) * int64(pgb.chainParams.TicketsPerBlock)
+				maxRemainingBlocks = tx.TSpendMeta.VoteEnd - tx.TSpendMeta.VoteStart
+			} else if tx.BlockHeight != 0 && tx.TSpendMeta.VoteEnd > tx.BlockHeight {
+				// tspend was short-circuited but we still account for the max
+				// remaining votes at the block it the short-circuited.
+				maxRemainingBlocks = tx.TSpendMeta.VoteEnd - tx.BlockHeight
 			} else if tx.TSpendMeta.VoteEnd > tipHeight {
-				maxRemainingVotes = (tx.TSpendMeta.VoteEnd - tipHeight) * int64(pgb.chainParams.TicketsPerBlock)
+				maxRemainingBlocks = tx.TSpendMeta.VoteEnd - tipHeight
 			}
+			maxRemainingVotes := maxRemainingBlocks * int64(pgb.chainParams.TicketsPerBlock)
 
 			requiredYesVotes := (totalVotes + maxRemainingVotes) * int64(pgb.chainParams.TreasuryVoteRequiredMultiplier) / int64(pgb.chainParams.TreasuryVoteRequiredDivisor)
 			tx.TSpendMeta.RequiredYesVotes = requiredYesVotes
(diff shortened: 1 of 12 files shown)


ukane-philemon:
Before the second commit, the TSspend vote meter looks like the one we have on the proposal page:
newtspend

db/dcrpg/pgblockchain.go
```diff
@@ -5891,10 +5892,82 @@ func (pgb *ChainDB) GetExplorerTx(txid string) *exptypes.TxInfo {
...
+				tx.TSpendMeta.VoteEndDate = time.Unix(voteEndTimeStamp, 0).UTC()
+			}
+
+			// Retrieve Public Key a.k.a Politieia key.
+			var pubKey []byte
+			txIn, err := hex.DecodeString(tx.Vin[0].TreasurySpend)
+			if err != nil {
+				log.Errorf("failed to retrieve pikey: %v", err)
+			} else {
+				pubKey = txIn[66 : 66+secp256k1.PubKeyBytesLenCompressed]
```
chappjc, line 5968: Should check length before slicing.

chappjc (commented):
Meant to deploy while a tspend was in mempool but missed the window, sorry.

ukane-philemon:
> Meant to deploy while a tspend was in mempool but missed the window, sorry.

Thanks for the review. I'm looking re-test this with a live tspend which might come by in the next few days.

chappjc:
> The second commit modified the vote meter to display the actual percentage of yes votes of possible votes (total votes + Maxremaining Votes), leaving the table to show the percentage of yes votes of total votes.

This makes it look like it's on track to fail.  If it has no "no" votes, it's approval is 100%.  Quorum is a separate concept and trying to consider either quorum (3456) or max possible votes (13,534?) makes it more confusing. "X% approval of N _possible_" has little meaning because the max possible is rarely achieved, so it will look to be failing until the moment the voting interval is over and it actually passes (assuming quorum met).

chappjc:
PR running at https://tip.dcrdata.org

ukane-philemon:
>  "X% approval of N possible" has little meaning because the max possible is rarely achieved, so it will look to be failing until the moment the voting interval is over and it actually passes 

So you mean, we can ignore "possible votes" while calculating the percentage for the text display just as it is on the vote meter? i.e (x% of total votes).

The text display tries to provide overall percentage (including possible votes) but maybe a correct behavior is to display yes% of total votes, right?

ukane-philemon:
> PR running at https://tip.dcrdata.org

Cool.

xaur:
"Min Required Yes Votes" was discussed at lenght in chat. I found a simple way to explain why I found it confusing, and more generally what a "normal" user may expect from tspend info section.

As a user I'd like to see if things are "looking good" for the current tspend. TSpend has two milestones to achieve - quorum and threshold.

- First I want to see the quorum thing, so I need the votes cast and votes for quorum (e.g. 1,883 cast, and 3,456 quorum). Personally I love seeing the 3,456 number because I keep forgetting it and deriving it from DCP-6 every time, which is not productive.
- Second is I need the Yes-vs-No breakdown *as of now* - e.g. 1,883 Yes, 0 No, 100% approval.
- Now that I have the *facts*, next I need some *projection* of what to expect. There can be many algos to simulate different scenarios (iirc matheusd had some script for that), but maybe a simple one would do for now. For example: "If vote continues at current approval (100%) and turnout (41%) rates, the spend will be (approved|rejected) in N days".

I guess some of this is already taken care of in the last commits but just in case. Thanks.

ukane-philemon:
> Now that I have the facts, next I need some projection of what to expect. There can be many algos to simulate different scenarios (iirc matheusd had some script for that), but maybe a simple one would do for now. For example: "If vote continues at current approval (100%) and turnout (41%) rates, the spend will be (approved|rejected) in N days".

Yeah, thanks for the suggestions. Concerning `projection of what to expect`, there is a `Yes Votes Still Needed` value displayed on the table and also I think we'd like not to give a projection that we know won't be even 90% correct in practice.

This is because the supposed `projection` will be affected by the number of votes and will always be incorrect.

chappjc (approved):

PR #1922, opened 2022-06-22
begin 6.2 cycle

chappjc:
This begins the v6.2 (application version) dev cycle, which will target the Decred/dcrd 1.8 release.

This begins with several major module version bumps that will see breaking changes in https://github.com/decred/dcrdata/pull/1876

- dcrdata's main module v7 => v8 (apitypes in the dcrdata modules has breaking change)
- db/dcrpg/v7 => v8 (methods being removed)
- gov/v5 => v6 (gov has dcrdata/v8/db/dbtypes types in it's exported API)

This also moves the `api`, `explorer`, `middleware`, and `notification` packages in the `cmd/dcrdata` folder to the `cmd/dcrdata/internal` folder.  These packages are only intended for use by the `cmd/dcrdata` explorer _application_ and there are no other current consumers.


chappjc:
So begins the dcrdata v8 module (dcrdata explorer v6.2)

PR #1929, opened 2022-08-06
views,explorer: Fix treasury pagination

### LINKED ISSUE #1925: Pagination problems on treasury page
I did notice that the pagination seems to have a few bugs though.  For example, if you refresh the page with a transaction type selected (spend in this case), it'll show the correct number of transactions (e.g. `showing 1 -- 4 of 4 transactions`), but if you select `All` from the type dropdown and then go back to `Spend`, it doesn't update properly (e.g. `showing 1 – 20 of 123,607 transactions`).

I also noticed the next link is not disabled when when there are no more results as would be expected.

For example, if you start at https://explorer.dcrdata.org/treasury?txntype=tspend, the next link is active and if you click it, the pagination changes to `showing 21 – 40 of 123,608 transactions` although there are not any actual results displayed.

_Originally posted by @davecgh in https://github.com/decred/dcrdata/issues/1924#issuecomment-1182880672_

martonp:
The treasury page was returning all of the transactions for the tx count regardless of the selected tx type. Also, the pagination section is now hidden if there are no transactions returned.

Closes #1925

### DIFF
--- a/cmd/dcrdata/views/address.tmpl
+++ b/cmd/dcrdata/views/address.tmpl
@@ -196,7 +196,7 @@
     <div class="position-relative" data-address-target="listbox">
         <div class="row align-items-center">
           <div class="me-auto mb-0 h4 col-24 col-sm-6">Transactions</div>
-          <div class="d-flex flex-wrap-reverse align-items-center justify-content-end py-1 col-24 col-sm-18">
+          <div data-address-target="paginationheader" class="d-flex flex-wrap-reverse align-items-center justify-content-end py-1 col-24 col-sm-18">
               <span class="fs12 nowrap text-end" data-address-target="range">
                   showing {{intComma (add .Offset 1)}} &mdash; {{intComma (add .Offset .NumTransactions)}} of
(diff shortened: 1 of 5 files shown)


cmd/dcrdata/views/address.tmpl
```diff
@@ -196,7 +196,7 @@
     <div class="position-relative" data-address-target="listbox">
         <div class="row align-items-center">
           <div class="me-auto mb-0 h4 col-24 col-sm-6">Transactions</div>
-          <div class="d-flex flex-wrap-reverse align-items-center justify-content-end py-1 col-24 col-sm-18">
+          <div address-target="paginationheader" class="d-flex flex-wrap-reverse align-items-center justify-content-end py-1 col-24 col-sm-18">
```
chappjc, line 199: `address-target` -> `data-address-target`
It's right on treasury.tmpl

chappjc (changes requested):

chappjc (approved):

PR #1938, opened 2022-11-24
db/dcrpg: deleted oldest dups, not newest

chappjc:
Duplicate deletion queries were keeping the vouts/vins that were from the stake-invalided transactions instead of the newer/replaced ones.

Before

image

After (the "unspent" is now a link)

image

@dajohi 

Requires full resync/ibd to resolve


buck54321 (commented):
The SQL changes here are basically hieroglyphics to me, but this seems to require a full re-sync to have an accurate db, no?

chappjc:
> The SQL changes here are basically hieroglyphics to me, but this seems to require a full re-sync to have an accurate db, no?

Yeah, the correct records were deleted, and I don't see a way to discern what they were.  Fortunately there's less than 20 affected txns.

PR #1941, opened 2022-12-05
ui: add margin between last element display and footer on `/charts` page

ukane-philemon:
Before:
Screenshot 2022-12-05 at 8 32 28 PM
After:
Screenshot 2022-12-05 at 8 32 14 PM

### DIFF
--- a/cmd/dcrdata/views/charts.tmpl
+++ b/cmd/dcrdata/views/charts.tmpl
@@ -189,7 +189,7 @@
         </div>
 
 
-        <div data-charts-target="chartWrapper" class="chart-wrapper ps-2 pe-2 mb-5">
+        <div data-charts-target="chartWrapper" class="chart-wrapper px-2 pb-5">
             <div
                 class="chartview"
                 data-charts-target="chartsView">


chappjc:
Only seems to affect Chromium in my testing.  On firefox I can always scroll down to see the whole thing.  On chromium, scrolling down still does not show the entire white box (the mb-3 you're adding to that div doesn't really address the bigger issue).

Chromium, **with** the mb-3:

image

Above, the mb-3 at least gets the url on the visible page, but the div containing everything is still going below the footer.

Firefox, **without** the mb-3:

image

Above, unlike Chromium, you can just scroll down.

Firefox, **with** the mb-3:

image

image

So the mb-3 makes the outer white box div weird looking even though it forces Chromium to at least keep the url part visible. 

I don't think this fixes the issue.

I think the question is why the `chartview` inside the `chart-wrapper` div is eating too much vertical space, not leaving enough for the two flex boxes below.

ukane-philemon:
I didn't investigate with firefox but will propose a better fix so it looks good regard less of what browser.

chappjc:
Thanks.  Note that it's still not right even with Chromium.  The div is still behind the footer even though you've pushed the visible text up just enough.

cmd/dcrdata/views/charts.tmpl
```diff
@@ -189,7 +189,7 @@
         </div>
 
 
-        <div data-charts-target="chartWrapper" class="chart-wrapper ps-2 pe-2 mb-5">
+        <div data-charts-target="chartWrapper" class="chart-wrapper px-2 pb-5">
```
chappjc, line 192: OK!  Looks consistent with `padding-bottom`

chappjc (approved):

cmd/dcrdata/views/charts.tmpl
ukane-philemon: Yeah, just like we do for the `main` div in other pages.

PR #1942, opened 2022-12-06
UI: fix dcrdata UI quirks

ukane-philemon:
Before:
p before
After:
p after
Before:
p header before
After:
p header after
Before:
markets before
After:
Screenshot 2022-12-06 at 10 46 12 PM

Before:
blocks before
After:
blocks after

### DIFF
--- a/cmd/dcrdata/views/market.tmpl
+++ b/cmd/dcrdata/views/market.tmpl
@@ -18,11 +18,11 @@
 
         {{- /* PRICE */ -}}
         <div class="ms-4 me-2 my-4 p-2 p-lg-4 bg-white text-center">
-            <div class="text-center fs18 text-secondary">1 DCR =</div>
+            <span class="text-center fs18 text-secondary">1 DCR =</span>
             {{if eq $botState.BtcIndex "USD"}}
-                <span class="fs18 fw-bold d-inline-block pt-1" style="vertical-align:top;">$</span>
+                <span class="fs22 fw-bold d-inline-block">$</span>
             {{end}}
-            <span class="fs28" data-market-target="price">{{printf "%.2f" $botState.Price}}</span> <span class="fs16 text-secondary">{{$botState.BtcIndex}}</span>
+            <span class="fs24" data-market-target="price">{{printf "%.2f" $botState.Price}}</span> <span class="fs16 text-secondary">{{$botState.BtcIndex}}</span>
         </div>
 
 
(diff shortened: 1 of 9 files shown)


cmd/dcrdata/views/market.tmpl
```diff
@@ -18,11 +18,11 @@
 
         {{- /* PRICE */ -}}
         <div class="ms-4 me-2 my-4 p-2 p-lg-4 bg-white text-center">
-            <div class="text-center fs18 text-secondary">1 DCR =</div>
+            <span class="text-center fs18 text-secondary">1 DCR =</span>
             {{if eq $botState.BtcIndex "USD"}}
-                <span class="fs18 fw-bold d-inline-block pt-1" style="vertical-align:top;">$</span>
```
chappjc, line 23: Not sure about this.  I certainly wouldn't call the dollar sign formatting a "quirk", since it's an intentional stylization, which both allows the dollar sign to be smaller (it looks huge with this PR), as well as a bit fancy.

image

But again, if others think this style is more odd than pleasant, we can change it.  Others with opinions on these?

chappjc (commented):
Thanks for the touch-ups.  I'm not sure that a couple of the changes are really fixing "quirks" but changing intentional style, so let's see what others think.

cmd/dcrdata/views/market.tmpl
ukane-philemon: While it looks intentional, I think it is more popular to this display the dollar sign right beside the figure using "almost" the same size?

cmd/dcrdata/views/market.tmpl
```diff
@@ -18,11 +18,11 @@
 
         {{- /* PRICE */ -}}
         <div class="ms-4 me-2 my-4 p-2 p-lg-4 bg-white text-center">
-            <div class="text-center fs18 text-secondary">1 DCR =</div>
+            <span class="text-center fs18 text-secondary">1 DCR =</span>
```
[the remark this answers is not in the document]
ukane-philemon: > so that this whole box centers on and draws your attention to just the number

Okay.

cmd/dcrdata/views/market.tmpl
chappjc: Fine with me either way.  Opinions @buck54321 @xaur @jzbz @davecgh 
Big dollar sign or funky stylization we have presently?

xaur: I'm okay with both as long as I can select `$ 16,966.68` and copy it into clipboard (i.e. it is plain ASCII and not a fancy icon or an exotic Unicode char). I slightly prefer `$` of the "almost" same size/font because my eyes got used to it.

While at it, adding a thousand separator like `1,000,000` to any number that can go above 1,000 would help readability.

davecgh: I'm fine with either.  If I absolutely had to choose, I find the currency symbol being the same size as the digits to more aesthetically pleasing.  It's probably just because that's how I'm more used to seeing it though and it really doesn't bother me stylized either.

chappjc: > While at it, adding a thousand separator like `1,000,000` to any number that can go above 1,000 would help readability.

Good suggestion.  Can probably use `humanize.CommafWithDigits` in a new entry in the `template.FuncMap`.

chappjc: Oh right.  toLocaleString with 2 for min and max decimals should work.

chappjc:
Please rebase and regenerate app js file name.  Otherwise looks g2g.

PR #1943, opened 2022-12-06
UI: rework visual blocks page

ukane-philemon:
Before:
Screenshot 2022-12-06 at 12 57 48 PM

After:
Screenshot 2022-12-06 at 12 57 59 PM

This diff modifies the `/visualblocks` page to match other pages that display block information e.g(`/blocks`, `/months`, `/windows` etc) and includes a link to the `/visualblocks` page in all relevant pages.

TODO: @chappjc, I'm thinking it would make sense to display for older blocks too(i.e not limiting to only recent/30 blocks max). Each page could display up to 10 blocks(including mempool block) with the option for viewers to visit pages with older blocks.


ukane-philemon:
If the new look is okay, I'm suggesting removing this line since it will look redundant? https://github.com/decred/dcrdata/blob/36ec5089c7a3243193403edf19eb0bb5422c3437/cmd/dcrdata/views/visualblocks.tmpl#L244

PR #1949, opened 2022-12-22
multi: implement stake reward calculator

### LINKED ISSUE #1772: Staking rewards calculator
User would input amount of DCR to start, a starting date, and an end date. It would tells the user how much they would have earned by staking that amount in tickets over the given time period. Code from the overview page can be reused. Most of the necessary logic is in the code that calculates the annual reward return rate.

ukane-philemon:
This implements a stake reward calculator and reactors an existing method to work with the stake reward calculator. Closes #1772.

-- Add a new `GET /stake-reward` endpoint.
-- Add a new `POST /stake-reward` endpoint.
-- Add stake reward calculator page.

While I could shutdown anytime to test other stuffs you can see the calculator here:
http://134.209.25.156:7777/stake-reward 

UI view:
Screenshot 2022-12-21 at 8 40 01 PM
Screenshot 2022-12-22 at 6 02 01 AM
Screenshot 2022-12-22 at 6 02 43 AM

### DIFF
--- a/cmd/dcrdata/views/extras.tmpl
+++ b/cmd/dcrdata/views/extras.tmpl
@@ -112,6 +112,7 @@
 					<a class="menu-item" data-keynav-skip href="/treasury" title="Decred Treasury">Treasury</a>
 					<a class="menu-item" data-keynav-skip href="/decodetx" title="Decode or send a raw transaction">Decode/Broadcast Tx</a>
 					<a class="menu-item" data-keynav-skip href="/verify-message" title="Verify Message">Verify Message</a>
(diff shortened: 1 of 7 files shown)


cmd/dcrdata/views/extras.tmpl
```diff
@@ -98,7 +98,7 @@
 				<span class="patty"></span>
 				<span class="patty"></span>
 				<span class="patty short"></span>
-				<div id="menu">
+				<div id="menu" style="max-height: 90vh; overflow-x: hidden;">
```
ukane-philemon, line 101: I have noticed the menu cuts a lot of the menu items out of view when the current view height is less than its full height. This will ensure dcrdata users can scroll the menu because we "could" add more items(maybe) but will not be limited anymore.

PR #1950, opened 2022-12-22
UI: add margin to connection icon

ukane-philemon:
Before:
Screenshot 2022-12-22 at 6 21 06 AM

After:
Screenshot 2022-12-22 at 6 20 49 AM

### DIFF
--- a/cmd/dcrdata/views/extras.tmpl
+++ b/cmd/dcrdata/views/extras.tmpl
@@ -172,7 +172,7 @@
 		</span>
 		<span
 			id="connection"
-			class="text-nowrap align-items-center clickable d-inline-block"
+			class="text-nowrap me-3 align-items-center clickable d-inline-block"
 			data-turbolinks-permanent
 			data-controller="connection"
 			data-connection-target="indicator"


chappjc:
I've never seen it look like that "before" shot.  Something isn't right here.
Also, the margin in the "after" is excessive.
Can you please test with some other browsers, ideally on different platforms (os)?

ukane-philemon:
The first images were from chrome, and this one is from firefox.
Screenshot 2022-12-22 at 6 55 39 AM

chappjc:
Something is wrong on your end unless there is a regression that's not yet deployed.

It looks nothing like that on Chromium or Firefox on Linux, Chrome or Firefox on Windows (7), or both on Android.  There are screenshots all over the web showing it's fine.

Why does the screen appear curved and cropped with a black boundary?  Looks like a problem with your window manager.

ukane-philemon:
> Something is wrong on your end unless there is a regression that's not yet deployed.

I guess so. I will just close this then. But it didn't look right on my end.

chappjc:
Well I'm curious what OS this is and why the window appears unnaturally cropped like that.  Is it a mobile device or a special desktop window theme that makes big (and rounded) window borders that run into the content?

ukane-philemon:
> Well I'm curious what OS

I'm running macOS. Oh, darn. I think that's why the window borders are rounded and causing a different view 😆.

PR #1951, opened 2022-12-24
cmd/dcrdata: version dcrdata APIs

### LINKED ISSUE #48: Version the API

ukane-philemon:
This adds versioning to dcrdata's REST APIs. Closes #48 

http://134.209.25.156:7777/api/v1

While this is just a fundamental approach to the whole versioning system, It is a footwork for future API versioning-related work ([Comment](https://github.com/decred/dcrdata/issues/48#issuecomment-1364567356)).

### DIFF
--- a/cmd/dcrdata/main.go
+++ b/cmd/dcrdata/main.go
@@ -723,15 +723,24 @@ func _main(ctx context.Context) error {
 	// SyncStatusAPIIntercept returns a json response if the sync status page is
 	// enabled (no the full explorer while syncing).
 	webMux.With(explore.SyncStatusAPIIntercept).Group(func(r chi.Router) {
-		// Mount the dcrdata's REST API.
+		// Mount the default dcrdata REST API.
 		r.Mount("/api", apiMux.Mux)
-		// Setup and mount the Insight API.
+		// Mount the versioned dcrdata REST API.
+		versionPrefix := fmt.Sprintf("/api/v%d", apiMux.Version())
+		r.Mount(versionPrefix, apiMux.Mux)
+
+		// Setup the Insight API.
 		insightApp := insight.NewInsightAPI(dcrdClient, chainDB,
 			activeChain, mpm, cfg.IndentJSON, app.Status)
 		insightApp.SetReqRateLimit(cfg.InsightReqRateLimit)
 		insightMux := insight.NewInsightAPIRouter(insightApp, cfg.UseRealIP,
 			cfg.CompressAPI, cfg.MaxCSVAddrs)
+
+		// Mount the default dcrdata insight REST API.
 		r.Mount("/insight/api", insightMux.Mux)
+		// Mount the versioned dcrdata insight REST API.
+		insightVersionPrefix := fmt.Sprintf("/insight/api/v%d", insightMux.Version())
+		r.Mount(insightVersionPrefix, insightMux.Mux)
 
 		if insightSocketServer != nil {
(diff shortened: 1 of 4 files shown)


cmd/dcrdata/main.go
```diff
@@ -723,15 +723,24 @@ func _main(ctx context.Context) error {
...
 			activeChain, mpm, cfg.IndentJSON, app.Status)
 		insightApp.SetReqRateLimit(cfg.InsightReqRateLimit)
 		insightMux := insight.NewInsightAPIRouter(insightApp, cfg.UseRealIP,
 			cfg.CompressAPI, cfg.MaxCSVAddrs)
+
+		// Mount the default dcrdata insight REST API.
 		r.Mount("/insight/api", insightMux.Mux)
+		// Mount the versioned dcrdata insight REST API.
+		insightVersionPrefix := fmt.Sprintf("/insight/api/v%d", insightMux.Version())
+		r.Mount(insightVersionPrefix, insightMux.Mux)
```
ukane-philemon, line 743: @chappjc, I'm contemplating accepting the API version as a URL param in another implementation e.g `/api/v{verison}`. But I am concerned about what behaviour we want for consumers providing versions above the current ver. 

Middleware adds the version to context and:
1.  request handlers reject the request if version > current version. OR
2. request honour the request using the current version.

chappjc: > 1. request handlers reject the request if version > current version

This.  If we can't serve the requested version, fail.  Probably one of the 4xx response codes.

chappjc: Why do you want it as a url param though?  Seems like having the urls prefixes with explicit mounts and version middlewares works.

ukane-philemon: > Why do you want it as a url param though? Seems like having the urls prefixes with explicit mounts and version middlewares works.

url params will allow handling multiple versions better without having to mount several times to support multiple versions. Plus makes the API versoning look cleaner?

chappjc: > url params will allow handling multiple versions better without having to mount several times to support multiple versions. Plus it makes the REST API versioning look cleaner?

Meh.  If we get to v4 any time soon, we're moving too fast.

ukane-philemon: > Seems like having the urls prefixes with explicit mounts and version middlewares works.

Took this approach in #1953.

ukane-philemon: > url params will allow handling multiple versions better without having to mount several times to support multiple versions. Plus it makes the REST API versioning look cleaner?

However, a major issue with this approach is that we won't be able to reject a request to an invalid version path(e.g`/api/v1000`) until it gets to the handler. Plus the versions in the request context is string based so verifying them in the request handler will take extra work.

chappjc: Right.  The _router_, not the handler, should be able to 404 for an unknown path like /api/v1000

chappjc:
Going with https://github.com/decred/dcrdata/pull/1953

PR #1952, opened 2022-12-26
cmd/dcrdata: update node dep

ukane-philemon:
This updates some node dependencies to their current version. Part of  # 1737

Deps like `webpack-cli`  had major changes but they don't affect dcrdata.

`npm out` printed some version diffs but these are the only ones with a major bump.
```
babel-loader                      8.2.5    8.3.0    9.1.0  node_modules/babel-loader                    dcrdata
stylelint-config-standard        27.0.0   27.0.0   29.0.0  node_modules/stylelint-config-standard       dcrdata
stylelint-config-standard-scss    5.0.0    5.0.0    6.1.0  node_modules/stylelint-config-standard-scss  dcrdata
webpack-cli                      4.10.0   4.10.0    5.0.1  node_modules/webpack-cli                     dcrdata
```

Links to change logs:
[stylelint-config-standard-scss](https://github.com/stylelint-scss/stylelint-config-standard-scss/blob/HEAD/CHANGELOG.md) 
[babel-loader  ](https://github.com/babel/babel-loader/releases)
[stylelint-config-standard  ](https://github.com/stylelint/stylelint-config-standard/blob/HEAD/CHANGELOG.md)  
[webpack-cli](https://github.com/webpack/webpack-cli/releases)


cmd/dcrdata/public/scss/connection.scss
```diff
@@ -6,12 +6,8 @@
     box-shadow: rgb(0 255 0 / 0%) 0 0 0;
   }
 
-  50% {
-    box-shadow: rgb(0 255 0 / 100%) 0 0 10px;
-  }
-
   to {
-    box-shadow: rgb(0 255 0 / 0%) 0 0 0;
+    box-shadow: rgb(0 255 0 / 100%) 0 0 10px;
```
ukane-philemon, line 10: update to `keyframe-selector-notation rule` require `from`: 0% and `to`: 100%(or anything but not 0%)

chappjc:
Two ways I would suggest to approach this process:
- **`ncu --target latest -u`** (https://www.npmjs.com/package/npm-check-updates) followed by `npm install`
- `rm -rf node_modules && npm ci` followed by **`npm-check -u`** (https://www.npmjs.com/package/npm-check), it seems to need `node_modules` deleted first.

Then sometimes even with the above you need to do `npm up` if some packages require `latest` for various transitive deps.

Finally `npm dedup`.

ukane-philemon:
I choose the second option. I had intended to only update modules with major version bump.

chappjc (changes requested):
Thank you very much for putting the scss change in a separate commit.  Please just squash the two commits that update node deps.  We will merge the two remaining commits without squashing them.

ukane-philemon:
> Please just squash the two commits that update node deps. We will merge the two remaining commits without squashing them.

Done.

chappjc (approved):

PR #1953, opened 2022-12-27
cmd/dcrdata: support multiple REST API versions

### LINKED ISSUE #48: Version the API

ukane-philemon:
This should replace #1951. It adds support for multiple dcrdata REST API versions, versioned the `apitypes` package and sets a path for future REST API versions. Closes #48.


cmd/dcrdata/internal/middleware/apimiddleware.go
```diff
@@ -1101,3 +1102,24 @@ func RetrieveStickWidthCtx(r *http.Request) string {
 	}
 	return bin
 }
+
+// APIVersionCtx adds supported API version to a request context.
+func APIVersionCtx(version string) func(next http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			r = r.WithContext(context.WithValue(r.Context(), ctxAPIVersion, version))
```
chappjc, line 1110: Why would this not be an integer?

ukane-philemon: I think it will read better as `v1, v2` when used by request handlers, but it can be an integer.

chappjc: hander logic:

```go
switch verCtx {
case 0:
...
case 1:
...
default:
...
}
```

better than string matching.

PR #1954, opened 2023-02-10
UI: show tx block on address page

### LINKED ISSUE #1748: Show tx block height on address page.
It'd be great to see the block the tx belongs to on the address page, currently you have to click on the txid to view the "included in block", adding the block height to the list entries would make it easier to scan through txs for an address.

ukane-philemon:
Closes #1748.
Screenshot 2023-02-10 at 9 57 20 AM

### DIFF
--- a/cmd/dcrdata/views/extras.tmpl
+++ b/cmd/dcrdata/views/extras.tmpl
@@ -264,7 +264,8 @@
 	{{- end}}
 		<th class="d-none d-sm-table-cell text-end">Time (UTC)</th>
 		<th class="text-end">Age</th>
-		<th class="text-end"><span class="d-sm-none position-relative" data-tooltip="Confirmations">Cons</span><span class="d-none d-sm-inline">Confirms</span></th>
+		<th class="d-none d-sm-table-cell text-end">Block</th>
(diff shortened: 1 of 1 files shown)


chappjc:
OK, let's try this.  It might cramp the table too much.  We'll find out how it really feels on tip.dcrdata.org.
Alternatively, we could just have made the confirmations number clickable, taking you to the block.

chappjc:
image

Block for unconfirmed tx makes no sense.  I think we should redo this to just have the confirm count be a clickable link to the block if it is mined.

ukane-philemon:
one moment.

chappjc:
> have the confirm count be a clickable link to the block if it is mined

That looks to get kinda messy because of how `setConfirmationText` works in JS

chappjc:
Since I say that, this Block table entry isn't updated when mined either.

ukane-philemon:
Uhmm, can hide the block column if unconfirmed.

chappjc:
There are multiple rows, some may be confirmed others not.  Besides, adding/removing columns on a table isn't great, too jarring.

We can revert and rework this from scratch with more though about the newblock controller in mind.  I didn't really love an extra column anyway.

ukane-philemon:
Modifying `setConfirmationText` to set the confirmation text as links to the tx block will affect all displays that use the function, are we cool with that?

PR #1958, opened 2023-03-02
cmd/dcrdata: allowed hosts

chappjc:
Since reverse proxy config is often not adequate or correct, this adds an allowed hosts middleware to dcrdata to ensure that any `Host` header value set in the request header is on a whitelist of hosts.

However, the reverse proxy should be configured with:
- a `default_server` black hole or redirect-only `server`, as a catch-all, preventing one of the other `server`s from becoming the default
- all other `server`s should specify `server_name` values including fqns, IPs, and `""` for when the `Host` header is not set.  These configuration changes prevent invalid or unexpected `Host` values.

However, to use without the above settings (or no reverse proxy at all), this change adds the `AllowedHosts` middleware, which ensures the `http.Request.Host` matches one of a number of hosts specified in dcrdata.conf.  If it does not, the value is simply cleared and the request continues to process.  In particular, this causes the social media tags to be omitted from the rendered html.

As a side note, the default `proxy_cache_valid` nginx setting does NOT include 400 (bad request) responses, for good reason.  The caching of bad requests is not a dcrdata bug, but the operators responsibility to avoid when using a caching proxy.

### DIFF
--- a/cmd/dcrdata/internal/explorer/explorermiddleware.go
+++ b/cmd/dcrdata/internal/explorer/explorermiddleware.go
@@ -37,6 +37,28 @@ var (
 	xForwardedScheme = http.CanonicalHeaderKey("X-Forwarded-Scheme")
 )
 
+// AllowedHosts will clear the Host request header entry unless it matches one
+// of the provided hosts names.
+func AllowedHosts(hosts []string) func(http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		hf := func(w http.ResponseWriter, r *http.Request) {
+			var ok bool
+			for _, host := range hosts {
+				if strings.EqualFold(r.Host, host) {
+					r.Host = host
+					ok = true
+					break
+				}
+			}
+			if !ok {
+				r.Host = ""
+			}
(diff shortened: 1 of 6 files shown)


cmd/dcrdata/internal/explorer/explorermiddleware.go
```diff
@@ -37,6 +37,29 @@ var (
...
+			var ok bool
+			for _, host := range hosts {
+				if strings.EqualFold(r.Host, host) {
+					r.Host = host
+					ok = true
+					break
+				}
+			}
+			if !ok {
+				// http.Error(w, http.StatusText(http.StatusBadRequest), http.StatusBadRequest)
```
jholdstock, line 54: Did you forget to remove this comment?

jholdstock (commented):
`cmd/dcrdata/sample-nginx.conf` should be updated to include your comments.

https://github.com/decred/dcrdata/blob/6d2bea99ac8afda060fd7a3cf5dd3e6c746fa7a2/cmd/dcrdata/sample-nginx.conf#L168

jholdstock (approved):

Issue #1961, opened 2023-04-04
Incorrect Daily Address Flow Enpoint Filtering


davecgh:
It appears that the daily address flow endpoint likely isn't filtering out sidechain/invalid outputs.

As a concrete example, If you add up the data from https://explorer.dcrdata.org/api/address/DshMNsvETDWpVoCe1re9NTAChiJagzsFV7J/amountflow/day, you get `~950,296.17` instead of the correct value of `~950,200.86`.

```
$ curl -ks https://explorer.dcrdata.org/api/address/DshMNsvETDWpVoCe1re9NTAChiJagzsFV7J/amountflow/day | jq '.received | add'
950296.167252989
```
This exact incorrect value can be reproduced by querying the database directly and just not filtering the sidechain/invalid outputs. Queries in particular for verification:

Incorrect (no filtering of sidechains and invalid outputs):
`SELECT SUM(vo.value)/1e8 FROM vouts vo WHERE 'DshMNsvETDWpVoCe1re9NTAChiJagzsFV7J'=ANY(vo.script_addresses);`
Result: `950296.167252990000`

Correct:
`SELECT SUM(vo.value)/1e8 FROM vouts vo INNER JOIN transactions t ON vo.tx_hash=t.tx_hash WHERE t.is_valid=true AND t.is_mainchain=true AND 'DshMNsvETDWpVoCe1re9NTAChiJagzsFV7J'=ANY(vo.script_addresses);`
Result: `950200.864535380000`

Finally, you can also see that the net flows aren't adding up properly either (almost positively for the same reason):

```
$ curl -ks https://explorer.dcrdata.org/api/address/DshMNsvETDWpVoCe1re9NTAChiJagzsFV7J/amountflow/day | jq '.net | add'
-32.410757609998996
```

PR #1962, opened 2023-04-04
dcrpg: query for only valid mainchain data in address chart queries

chappjc:
Resolves https://github.com/decred/dcrdata/issues/1961

```
$  curl -ks http://127.0.0.1:7777/api/address/DshMNsvETDWpVoCe1re9NTAChiJagzsFV7J/amountflow/day | jq '.net | add'
-3.637978807091713e-12
$  curl -ks http://127.0.0.1:7777/api/address/DshMNsvETDWpVoCe1re9NTAChiJagzsFV7J/amountflow/day | jq '.received | add'
950200.8645353797
```

jq is doing lossy floating point addition there and I didn't feel like multiplying by 1e8 and rounding before adding.

### DIFF
--- a/cmd/dcrdata/sample-dcrdata.conf
+++ b/cmd/dcrdata/sample-dcrdata.conf
@@ -49,6 +49,7 @@
 ; with no value for Host. Multiple values may be specified. (no default)
 ; For example:
 ;allowedhost=127.0.0.1
+;allowedhost=127.0.0.1:7777
 ;allowedhost=explorer.dcrdata.org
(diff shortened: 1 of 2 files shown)


davecgh:
Looks right.  If you want to confirm the .net comes out at 0, I believe it should be `| jq '[.net[]*1e8] | add | round / 1e8'.
'

chappjc:
That came out to zero, thanks.

davecgh (approved):

PR #1968, opened 2023-06-16
multi: update to dcr 1.8 mods

chappjc:
image

The txhelpers/subsidy.go functions need work though. They are using deprecated functions and won't work as expected I think


dajohi (approved):

chappjc:
Continuing the conversation in https://github.com/decred/dcrdex/pull/2403, this PR will need a follow-up to use the V3 subsidy cache calc methods.  After that there's the relatively minor issue of difficulty charts adjusting every block.

I don't know about handling the blake3 block hashes.

PR #1970, opened 2023-07-18
multi: Show PowHash on blocks view page for DCP0011

### LINKED ISSUE #1965: Feature Request: Show Separate Proof-of-Work Hash
With the upcoming blake3 consensus change, assuming it passes, the block hash and the proof-of-work hash will no longer be the same.

In light of this, I think the explorer should add a separate field that displays the PoW hash.  I suppose it's a matter of taste of whether you prefer the consistency of showing it for all blocks even though it will be the same for all of them prior to the activation of the agenda, or the compactness of only showing the new field when the associated agenda is active.

The relevant code to calculate the new hash was added in decred/dcrd#3115.  In short, there are new `PoWHashV1` and `PoWHashV2` methods on the `wire.MsgBlock` (and `wire.BlockHeader`) that return the respective PoW hash variant that the relevant consensus validation logic uses.  For `PoWHashV1`, it will obviously match `BlockHash`, but having the separate methods makes it clear the PoW hash is no longer tied to the block hash.

Here are a couple of mockups I hacked together for potential placement:

1) Under the block hash on the left:

image

2) Above the Merkle Root on the right:

image

ukane-philemon:
Closes #1965.

### DIFF
--- a/db/dcrpg/pgblockchain.go
+++ b/db/dcrpg/pgblockchain.go
@@ -5811,6 +5811,7 @@ func (pgb *ChainDB) GetExplorerBlock(hash string) *exptypes.BlockInfo {
 	block := &exptypes.BlockInfo{
 		BlockBasic:            b,
 		Confirmations:         data.Confirmations,
+		PoWHash:               b.Hash,
 		StakeRoot:             data.StakeRoot,
 		MerkleRoot:            data.MerkleRoot,
 		Nonce:                 data.Nonce,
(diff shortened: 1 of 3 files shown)


db/dcrpg/pgblockchain.go
```diff
@@ -5828,6 +5829,23 @@ func (pgb *ChainDB) GetExplorerBlock(hash string) *exptypes.BlockInfo {
...
+			log.Errorf("error parsing hash for block %d (hash=%s): %v", data.Height, data.Hash, err)
+			return nil
+		}
+
+		// Get the block header
+		header, err := pgb.Client.GetBlockHeader(pgb.ctx, blockChainHash)
+		if err != nil {
+			log.Errorf("failed to fetch header for block hash %s: %v", data.Hash, err)
+			return nil
+		}
```
ukane-philemon, line 5844: This looks a bit expensive?

chappjc: Ideally `PowHash` would be added to `chainjson.GetBlockVerboseResult`.  We get this just above, at the start of this method.  AFAIK, we can only get this at present by hashing the block header.

chappjc: Would require: https://github.com/decred/dcrd/pull/3154

chappjc (approved):
Verified working.  Will follow up if https://github.com/decred/dcrd/pull/3154 becomes available so we can skip the request for the raw header.

PR #1971, opened 2023-07-19
dcrpg: use PoW hash from block result when avail

davecgh:
This modifies the recently-added logic to include the proof of work hash in the explorer block info to make use of the new PoWHash field of the verbose block output when available.

### DIFF
--- a/db/dcrpg/pgblockchain.go
+++ b/db/dcrpg/pgblockchain.go
@@ -5829,7 +5829,9 @@ func (pgb *ChainDB) GetExplorerBlock(hash string) *exptypes.BlockInfo {
 		Subsidy:               pgb.BlockSubsidy(b.Height, b.Voters),
 	}
 
-	if pgb.IsDCP0011Active(b.Height) {
+	if data.PoWHash != "" {
+		block.PoWHash = data.PoWHash
+	} else if pgb.IsDCP0011Active(b.Height) {
 		blockChainHash, err := chainhash.NewHashFromStr(data.Hash)
 		if err != nil {
 			log.Errorf("error parsing hash for block %d (hash=%s): %v", data.Height, data.Hash, err)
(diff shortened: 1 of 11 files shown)


db/dcrpg/pgblockchain.go
```diff
@@ -5829,7 +5829,9 @@ func (pgb *ChainDB) GetExplorerBlock(hash string) *exptypes.BlockInfo {
 		Subsidy:               pgb.BlockSubsidy(b.Height, b.Voters),
 	}
 
-	if pgb.IsDCP0011Active(b.Height) {
+	if data.PoWHash != "" {
```
davecgh, line 5832: I decided to use this approach over just directly setting the field above to `data.PoWHash` because the release version of dcrd with rpc server version 8.0.0 will not have the field populated in the verbose output.

I believe this approach will allow the faster path when it's backed by a dcrd with rpc server version 8.1.0 without having to explicitly require it.

chappjc: I think that's best too.

chappjc:
Looks like CI wants a `go mod tidy` when you have a chance.

davecgh:
> Looks like CI wants a `go mod tidy` when you have a chance.

Aye, I was figuring out where it wanted it.  I had already done it in the main module, but realized there were the submodules that needed to be bumped too.  Should be good to go now.

davecgh:
Ugh, looks like this is still using Go1.20.5 which has the broken `go.mod`.  I'll update it to use that variant.

chappjc:
> I had already done it in the main module, but realized there were the submodules that needed to be bumped too.

Ah, yeah, that's pretty annoying, sorry.  I never made a decent resolution because I tend to just run `./lint.sh` to do the tidy for me until it's happy.  Kinda dumb.

davecgh:
> Ah, yeah, that's pretty annoying, sorry. I never made a decent resolution because I tend to just run `./lint.sh` to do the tidy for me until it's happy. Kinda dumb.

No worries.  It's happy now.  Sorry for the churn.

chappjc (approved):
Still working with dcrd 1.8.0+release.
Also verified with 1.9.0-pre+e3b4363ff, where it would seem to be using the PoWHash from `getblock` (not making the extra `getblockheader` RPC).  Good stuff!

PR #1972, opened 2023-07-19
CI: use latest Go with check-latest

chappjc:
TEST: let's see if this makes it fetch Go 1.20.6

https://github.com/actions/setup-go#check-latest-version

> If check-latest is set to true, the action first checks if the cached version is the latest one. If the locally cached version is not the most up-to-date, a Go version will then be downloaded. Set check-latest to true if you want the most up-to-date Go version to always be used.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,6 +17,7 @@ jobs:
       uses: actions/setup-go@fac708d6674e30b6ba41289acaab6d4b75aa0753 #v4.0.1
       with:
         go-version: ${{ matrix.go }}
+        check-latest: true
 
     - name: Go Build dcrdata
       run: go build -v ./...


chappjc:
Does not seem to use the cache as documented.  For re-runs, I would it expect it to find the go toolchain in cache and use it, but it fetches it again (and claims to cache it).
Most recent commit on `master`, for reference: https://github.com/decred/dcrdata/actions/runs/5604443746/jobs/10252422249#step:3:9

chappjc:
Oh, when a hosted running finishes, it resets to the initial state.  The cache referred to in setup-go with reference to the `check-latest` flag appears to be the cached toolchains in runners initial state.  Fetching and extracting the latest toolchain takes about 6 seconds, and that will only happen until the hosted runner gets updated so it's initial state includes the latest.

PR #1974, opened 2023-09-25
docs: Removed an old sketchy URL that appeared to be pwnd

matthawkins90:
I was reading `contributing.md`, and I noticed there was a link to coffee. I clicked it and was immediately warned by Google that it was dangerous due to some security breach and phishing attacks. I didn't investigate much further, but I figured I should remove the link.

While there, I just did a quick update to the suggested Go line. Figured 1.8.3 was a bit old.

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -57,11 +57,11 @@ Only submit a PR once the intended edits are either done or nearing completion. 
 
 ### Suggested Toolkit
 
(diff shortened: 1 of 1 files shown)


davecgh (approved):

xaur:
The link was introduced in #192 in 2017. There are minor [TLS issues](https://www.ssllabs.com/ssltest/analyze.html?d=haiticoffeeacademy.com) but other than that I can't quickly find a confirmation that it is a malicious or hacked site. I wonder what Google complained about.

Issue #1973, opened 2023-09-04
Feature Request: Dynamic Scaling of Displayed Values on Home Page


davecgh:
Notice in the screenshot how the Mining difficulty says "0 Mil".  I suggest this should be a dynamically scaled humanized value instead.  Similarly, the hashrate is being reproted in Ph/s.  It should scale dynamically too.

![image](https://github.com/decred/dcrdata/assets/2115102/8e2e7c87-6be8-46d7-b0b7-920690f6ece9)

Example code:

```Go
func humanizeHPS(b uint64) string {
	const unit = 1000
	if b < unit {
		return fmt.Sprintf("%d h/s", b)
	}
	div, exp := uint64(unit), 0
	for n := b / unit; n >= unit; n /= unit {
		div *= unit
		exp++
	}
	return fmt.Sprintf("%.2f %ch/s", float64(b)/float64(div), "kMGTPE"[exp])
}
```

```Go
func main() {
	const maxUint64 = 1<<64 - 1
	testVals := []uint64{0}
	for hps := uint64(1); hps < 10e18; hps *= 10 {
		testVals = append(testVals, hps)
	}
	testVals = append(testVals, maxUint64)

	for _, hps := range testVals {
		fmt.Printf("%v (input: %d)\n", humanizeHPS(max(hps, 1)), hps)
	}
}
```

Output:

```
1 h/s (input: 0)
1 h/s (input: 1)
10 h/s (input: 10)
100 h/s (input: 100)
1.00 kh/s (input: 1000)
10.00 kh/s (input: 10000)
100.00 kh/s (input: 100000)
1.00 Mh/s (input: 1000000)
10.00 Mh/s (input: 10000000)
100.00 Mh/s (input: 100000000)
1.00 Gh/s (input: 1000000000)
10.00 Gh/s (input: 10000000000)
100.00 Gh/s (input: 100000000000)
1.00 Th/s (input: 1000000000000)
10.00 Th/s (input: 10000000000000)
100.00 Th/s (input: 100000000000000)
1.00 Ph/s (input: 1000000000000000)
10.00 Ph/s (input: 10000000000000000)
100.00 Ph/s (input: 100000000000000000)
1.00 Eh/s (input: 1000000000000000000)
18.45 Eh/s (input: 18446744073709551615)
```

xaur:
Yes please! Big zero in Difficulty and Hashrate look like something is broken.

[...]

PR #1975, opened 2023-09-25
views/home: Use ThreeSigFigs to display difficulty on homepage

matthawkins90:
This is a quick fix for displaying the difficulty on the homepage. I just changed it to show the raw value instead of being divided by 1e6 and showing a "Mil" value.

I thought about trying to do something complicated and make a scaling value that switched formats if it became longer than 1e6, but I don't think it's worth the time investment it would take me.

### DIFF
--- a/cmd/dcrdata/public/js/controllers/homepage_controller.js
+++ b/cmd/dcrdata/public/js/controllers/homepage_controller.js
@@ -167,7 +167,7 @@ export default class extends Controller {
 
   _processBlock (blockData) {
     const ex = blockData.extra
-    this.difficultyTarget.innerHTML = humanize.decimalParts(ex.difficulty / 1000000, true, 0)
+    this.difficultyTarget.innerHTML = humanize.decimalParts(ex.difficulty, true, 0)
     this.bsubsidyPowTarget.innerHTML = humanize.decimalParts(ex.subsidy.pow / 100000000, false, 8, 2)
     this.bsubsidyPosTarget.innerHTML = humanize.decimalParts((ex.subsidy.pos / 500000000), false, 8, 2) // 5 votes per block (usually)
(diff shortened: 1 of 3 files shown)


davecgh:
I'll leave this one up to the dcrdata maintainers since it might be a controversial change without dynamic scaling.  That said, I provided code in https://github.com/decred/dcrdata/issues/1973 for dynamic scaling.

matthawkins90:
If dynamic scaling is important, I think I would like to get some feedback about how it should be displayed. Difficulty is unitless, so it just needs some kind of easy-to-interpret "scaling" suffix.

1234 = 1.23 k? 1.23e3? 1,234?
10000 = 10 k? 1e4? 10,000?
100000 = 100 k? 1e5? 100,000?
1000000 = 1 Mil? 1 M? 1.00M? 1e6? 1,000,000?

I don't think it's too hard, but I actually kinda like how it looks as `356,765`, without decimals or scaling factors, so I didn't want to change it to something weird like `3.57e5` or `357 k`

xaur:
`1eX` style is hard to read. If there are more than 3 digits it needs a thousand separator `356,765`. `Mil` looks odd. IMO.

I vote for `999.00`, `1.23 K`, `10.00 K`, `100.00 K`, `1.00 M`. Spelling out `thousand`, `million`, `billion` is also fine.

buck54321:
What about `humanize.threeSigFigs`?

matthawkins90:
Yeah, threeSigFigs is the easiest solution here. It already converts Floats into 3 digits with a possible "k", "M", or "B". The only downside is that it's formatted like `123k` instead of `123 k`, but all the other places using threeSigFigs suffer from the same problem. I could update it everywhere, but I think it was designed to look that way intentionally.

I'm not quite sure if updating `homepage_controller.js` actually did anything. I tried it multiple different ways, but everything looked fine regardless of what I did to `homepage_controller.js`. Please let me know if I need to revert that file.

PR #1988, opened 2024-06-04
Use `-n` to test for bash script argument

ukane-philemon:
`./run_tests.sh` fails on my Mac. 

Error message: `./run_tests.sh: line 16: conditional binary operator expected.` 

```
% bash --version
GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin22)
Copyright (C) 2007 Free Software Foundation, Inc.
```

This PR uses a stable conditional expression `-n` to test for bash script argument in `./run_tests.sh`.

### DIFF
--- a/run_tests.sh
+++ b/run_tests.sh
@@ -13,7 +13,7 @@ set -ex
 GV=$(go version | sed "s/^.*go\([0-9.]*\).*/\1/")
 echo "Go version: $GV"
 
-if [[ -v TESTTAGS ]]; then
+if [[ -n "${TESTTAGS}" ]]; then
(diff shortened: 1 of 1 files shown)


chappjc (approved):
Ok, but... 2007 warez?

jrick:
> Ok, but... 2007 warez?

it's stuck on the last gpl2 release. apple won't upgrade it past that to a gpl3 version.

they also changed the default shell to zsh (which is MIT licensed fwiw).

chappjc:
Get your brew on then

ukane-philemon:
> Get your brew on then

I've got brew, but I only use it for third-party software that doesn't come packaged 🤷

PR #1992, opened 2024-06-05
{cmd/dcrdata/views}: Fix orphan block `api` link on the block view

### LINKED ISSUE #1745: Wrong api link for orphan blocks.
I noticed the api link for orphan blocks link to the block on the main chain instead. [orphan block 448716](https://explorer.dcrdata.org/block/0000000000000000006028125bf9695bd8fc00e1bbcda8ab3e034d7f25d90dbd) api link opens up the details of the [block 448716 on the mainchain](https://explorer.dcrdata.org/api/block/448716/verbose?indent=true).

ukane-philemon:
Closes #1745

Issue: Visit the https://dcrdata.decred.org/side and click on any block, then click on the `api` link on the displayed block view. It'll display data for the mainchain block.

### DIFF
--- a/cmd/dcrdata/views/block.tmpl
+++ b/cmd/dcrdata/views/block.tmpl
@@ -46,7 +46,11 @@
 						{{- else }}
 						<a class="fs13" href="/mempool">mempool </a>|
 						{{- end}}
+						{{if .MainChain}}
 						<a class="fs13" href="/api/block/{{.Height}}/verbose?indent=true" data-turbolinks="false">api</a>
+						{{- else}}
+						<a class="fs13" href="/api/block/hash/{{.Hash}}/verbose?indent=true" data-turbolinks="false">api</a>
+						{{- end}}
 					</div>
 				</div>
 			</div>


cmd/dcrdata/views/block.tmpl
```diff
@@ -46,7 +46,11 @@
 						{{- else }}
 						<a class="fs13" href="/mempool">mempool </a>|
 						{{- end}}
+						{{if .MainChain}}
 						<a class="fs13" href="/api/block/{{.Height}}/verbose?indent=true" data-turbolinks="false">api</a>
+						{{- else}}
+						<a class="fs13" href="/api/block/hash/{{.Hash}}/verbose?indent=true" data-turbolinks="false">api</a>
```
chappjc, line 52: How about hash in both cases?

chappjc (commented):
Thanks for this

cmd/dcrdata/views/block.tmpl
ukane-philemon: The user would proly look at the URL to note the block they are viewing, I think we'd want to make this exception just for side blocks unless you want us to use hash for both.

PR #1997, opened 2024-06-06
{cmd/dcrdata}: Use template func for footer year

ukane-philemon:
This PR adds a `currentYear` template function to display the correct current year so we don't need to update this every year.

### DIFF
--- a/cmd/dcrdata/internal/explorer/templates.go
+++ b/cmd/dcrdata/internal/explorer/templates.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2018-2022, The Decred developers
+// Copyright (c) 2018-2024, The Decred developers
 // Copyright (c) 2017, The dcrdata developers
 // See LICENSE for details.
 
@@ -368,6 +368,9 @@ func makeTemplateFuncMap(params *chaincfg.Params) template.FuncMap {
 		},
 		"intMultiply": func(a, b int) int {
 			return a * b
+		},
+		"currentYear": func() string {
+			return strconv.Itoa(time.Now().Year())
 		},
 		"timezone": func() string {
 			t, _ := time.Now().Zone()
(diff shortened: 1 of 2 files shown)


cmd/dcrdata/internal/explorer/templates.go
```diff
@@ -369,6 +369,9 @@ func makeTemplateFuncMap(params *chaincfg.Params) template.FuncMap {
 		"intMultiply": func(a, b int) int {
 			return a * b
 		},
+		"currentYear": func() string {
+			return fmt.Sprint(time.Now().Year())
```
davecgh, line 373: `fmt.Sprint` should basically never be used for trivial numeric conversions because it is significantly more expensive (at least 2x in the best case) than `strconv.Itoa`, both in terms of CPU usage.  It also needlessly forces an additional allocation.

```Go
return strconv.Itoa(time.Now().Year())
```

This is fairly trivial to show via a benchmark with 10 iterations on both variants:


```
name   old time/op    new time/op    delta
---------------------------------------------------------------
Year   92.3ns ± 3%    40.4ns ± 2%    -56.23%  (p=0.000 n=10+10)

name   old alloc/op   new alloc/op   delta
---------------------------------------------------------------
Year   16.0B ± 0%     4.0B ± 0%      -75.00%  (p=0.000 n=10+10)
```

davecgh (changes requested):

cmd/dcrdata/internal/explorer/templates.go
ukane-philemon: Thanks, I've updated the PR.

davecgh (approved):
Code looks fine now, so approved.

That said, it's arguable whether or not this should be automatic.  After all, if nobody from the Decred project touched the code in the current year, the copyright shouldn't really be updated to claim that they have.

On the other hand, in practice, at least one of the files somewhere is very likely to have been modified and it's exceedingly easy to forgot updating this off in some file nested down in a `cmd` subdir, so it is nice from that perspective.

PR #1999, opened 2024-06-18
build: update deps and linter

dajohi:


davecgh:
Why are all of the modules being updated to require go 1.21 when they don't actually require it?

PR #2003, opened 2024-08-16
Update copyright year

peterzen:


davecgh (approved):
Approving.  It looks like #1997 will make this automatic.

PR #2008, opened 2024-09-19
multi: remove `/statistics`, `/stats` and redirect `/stats` to homepage

### LINKED ISSUE #2000: Remove `/stats` page
https://dcrdata.org/stats is still in the Stone Age. I want to redesign and implement the new design for the /stats page but I want to be clear about if that's what we want for the project or if we want to retire the page completely considering that some of those stats are on the homepage.

Screenshot 2024-07-23 at 12 56 04 AM

cc: @chappjc

ukane-philemon:
closes #2000


cmd/dcrdata/main.go
```diff
@@ -779,10 +779,9 @@ func _main(ctx context.Context) error {
 		r.Get("/search", explore.Search)
 		r.Get("/charts", explore.Charts)
 		r.Get("/ticketpool", explore.Ticketpool)
-		r.Get("/stats", explore.StatsPage)
 		r.Get("/market", explore.MarketPage)
-		r.Get("/statistics", func(w http.ResponseWriter, r *http.Request) {
-			http.Redirect(w, r, "/stats", http.StatusPermanentRedirect)
+		r.Get("/stats", func(w http.ResponseWriter, r *http.Request) {
```
davecgh, line 783: Should it keep statistics as a redirect too so the old endpoint doesn't error?

ukane-philemon: Errm, thought about it, but it has been there for over 6 years, enough time to keep a legacy redirect, no?

davecgh: I'm fine with removing it too.  I just thought I'd point it out since redirects are basically free and who knows if there are old links to it, etc.

No objections to removing it.

davecgh (approved):

ukane-philemon:
Will merge tomorrow, if no other review on this. @davecgh you can still do the honours.

davecgh:
> Will merge tomorrow, if no other review on this. @davecgh you can still do the honours.

I've already approved it.

PR #2029, opened 2025-10-31
multi: add fee and feerate fields to TrimmedTx and Tx API types 

### LINKED ISSUE #470: Show fee in decoded and unconfirmed txs
It would be nice to see the fee and fee rate of decoded (eg: prior to manually broadcasting a tx) and unconfirmed transactions.

Though to get the decoded fee you'd have to look at the utxos, so it might be tricky to do on the current flow of the decoding call.

ukane-philemon:
This PR adds fee and feerate fields (in atoms) to **apitypes.TrimmedTx** and **apitypes.Tx**. These fields
are now available across the api endpoints listed below:

```txt
- GET /tx/{txid}
- GET /tx/{txid}/trimmed
- GET /tx/decoded/{txid}
- POST /txs
- POST /txs/trimmed
```

The new fields provide precise transaction fee information in atoms for both mempool
and mined transactions.

Closes #470

### DIFF
--- a/cmd/dcrdata/views/mempool.tmpl
+++ b/cmd/dcrdata/views/mempool.tmpl
@@ -231,8 +231,8 @@
                             <tr>
                               <th>Transaction ID</th>
                               <th class="text-end">Total DCR</th>
-                              <th class="text-end">Size</th>
                               <th class="text-end">Fee Rate</th>
+                              <th class="text-end">Size</th>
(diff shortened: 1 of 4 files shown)


dajohi (changes requested):
needs /mempool transactions columns to match /block transactions columns

hilljessica8787-dot (approved):

